PR #57, opened 2024-10-23
remove exposing ports for redis

LegendWorrier:
• Removed exposing ports for redis at validator 
• Removed exposing ports for redis at miner


LegendWorrier:
• Removed exposing ports for redis

PR #61, opened 2024-10-24
remove the unnecessary  logs on validator

LegendWorrier:
• Removed unnecessary logs on validator(miner_client.py, miner_service.py, task_service.py)

### DIFF
--- a/neurons/validators/src/clients/miner_client.py
+++ b/neurons/validators/src/clients/miner_client.py
@@ -137,6 +137,10 @@ class MinerClient(abc.ABC):
                     )
                 if self.debounce_counter:
                     sleep_time = self.sleep_time()
+                    logger.info(
+                        f"Retrying connection to miner in {sleep_time:0.2f}",
+                        extra=get_extra_info(self.logging_extra),
+                    )
                     await asyncio.sleep(sleep_time)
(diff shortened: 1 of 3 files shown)


neurons/validators/src/clients/miner_client.py
```diff
@@ -137,10 +137,6 @@ async def await_connect(self):
                     )
                 if self.debounce_counter:
                     sleep_time = self.sleep_time()
-                    logger.info(
```
pyon12, line 140: @LegendWorrier I think we need to keep this log

neurons/validators/src/services/task_service.py
```diff
@@ -137,13 +125,6 @@ async def create_task(
                         )
                         return machine_spec, executor_info, 0
 
-                    logger.info(
```
pyon12, line 140: @LegendWorrier I think we need to keep this log

pyon12, line 140: @LegendWorrier @talentwebdev I think we need to keep this log

pyon12:
@LegendWorrier I don't think we need to keep these 2 logs. please remove them.
image

LegendWorrier:
• Fixed task_service.py of validator.

PR #63, opened 2024-10-28
feat(validator): DAH-523 - update data that sends to backend

LegendWorrier:
• Add the score per miner
• gather the logs per miner
• send logs to App BE


LegendWorrier:
• Add job id per miner
• Update the logs(even though failed)

PR #82, opened 2024-11-13
Feature/dah 637 check docker digest

LegendWorrier:
- check if there is duplicate  image digests.
- check if there is any docker image digest which doesn't exist in the list of digests from docker hub.

### DIFF
--- a/neurons/validators/src/miner_jobs/machine_scrape.py
+++ b/neurons/validators/src/miner_jobs/machine_scrape.py
@@ -9,6 +9,7 @@ import threading
 import psutil
 from functools import wraps
 import hashlib
+import docker
 from base64 import b64encode
 from cryptography.fernet import Fernet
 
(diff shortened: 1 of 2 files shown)


neurons/validators/src/miner_jobs/machine_scrape.py
```diff
@@ -516,6 +517,53 @@ def get_network_speed():
...
+    """Retrieve the Docker container ID if running inside a container."""
+    try:
+        with open('/proc/self/cgroup', 'r') as f:
+            for line in f:
+                if 'docker' in line:
+                    return line.strip().split('/')[-1]
+    except FileNotFoundError:
+        return None
+    
+def get_container_digest(container_id):
```
pyon12, line 530: @LegendWorrier I don't think we need this function

neurons/validators/src/services/miner_service.py
```diff
@@ -44,6 +45,7 @@
 
 
 JOB_LENGTH = 300
+REPOSITORY = ["daturaai/compute-subnet-executor", "daturaai/compute-subnet-executor-runner", "containrrr/watchtower"]
```
pyon12, line 48: There are more repositories
daturaai/pytorch:1.13.0-py3.10-cuda11.7.1-devel-ubuntu22.04
daturaai/pytorch:1.9.1-py3.9-cuda11.1.1-devel-ubuntu20.04

neurons/validators/src/services/miner_service.py
```diff
@@ -240,6 +251,88 @@ async def request_job_to_miner(self, payload: MinerJobRequestPayload):
...
+
+                # Update the all_digests dictionary with the current repository's tag-digest pairs
+                all_digests.update(tag_digests)
+
+            except requests.exceptions.RequestException as e:
+                print(f"Error retrieving data for {repository}: {e}")
+
+        return all_digests 
+
+    def check_digests(self, results, list_digests):
```
pyon12, line 309: these functions check_digests, check_duplicated_digests, validate_digests should be implemented on TaskService, not MinerService

pyon12: You can get the DockerHub Digests, before run the job and pass the result to TaskService

Fetching the DockerHub Digests could be done on MinerService or core/validator.py for better performance

neurons/validators/src/miner_jobs/machine_scrape.py
```diff
@@ -516,6 +517,53 @@ def get_network_speed():
         data["network_speed_error"] = repr(exc)
     return data
 
+def get_container_id():
+    """Retrieve the Docker container ID if running inside a container."""
+    try:
+        with open('/proc/self/cgroup', 'r') as f:
```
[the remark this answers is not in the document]
LegendWorrier: Screenshot_7

neurons/validators/src/miner_jobs/machine_scrape.py
LegendWorrier: okay, will remove it.

neurons/validators/src/services/miner_service.py
LegendWorrier: okay.

PR #87, opened 2024-11-15
feat(validator): DAH-661-subtensor-logs-on-grafana

LegendWorrier:
- valdiator-Subtensor set_weight, etc logs on grafana

### DIFF
--- a/neurons/validators/src/clients/compute_client.py
+++ b/neurons/validators/src/clients/compute_client.py
@@ -19,13 +19,14 @@ from protocol.vc_protocol.validator_requests import (
     AuthenticateRequest,
     ExecutorSpecRequest,
     RentedMachineRequest,
+    LogValidatorRequest
 )
 from pydantic import BaseModel
 
 from clients.metagraph_client import create_metagraph_refresh_task, get_miner_axon_info
 from core.utils import _m, get_extra_info
 from services.miner_service import MinerService
-from services.redis_service import MACHINE_SPEC_CHANNEL_NAME, RENTED_MACHINE_SET
+from services.redis_service import MACHINE_SPEC_CHANNEL_NAME, LOG_ERROR_VALIDATOR_CHANNEL_NAME, RENTED_MACHINE_SET
 
 logger = logging.getLogger(__name__)
 
@@ -100,9 +101,13 @@ class ComputeClient:
         try:
             # subscribe to channel to get machine specs
             pubsub = await self.miner_service.redis_service.subscribe(MACHINE_SPEC_CHANNEL_NAME)
+            # subscribe to channel to get logs of validator
+            pubsub_logs = await self.miner_service.redis_service.subscribe(LOG_ERROR_VALIDATOR_CHANNEL_NAME)
 
             # send machine specs to facilitator
             self.specs_task = asyncio.create_task(self.wait_for_specs(pubsub))
+            # send logs of validator to facilitator
+            self.logs_task = asyncio.create_task(self.wait_for_logs_of_validator(pubsub_logs))
         except Exception as exc:
             logger.error(
                 _m("redis connection error", extra={**self.logging_extra, "error": str(exc)})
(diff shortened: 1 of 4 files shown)


neurons/validators/src/core/validator.py
```diff
@@ -156,7 +157,7 @@ def fetch_miners(self, subtensor: bittensor.subtensor):
         bittensor.logging.info("Found %d miners", "fetch_miners", "fetch_miners", len(miners))
         return miners
 
-    async def set_weights(self, miners, subtensor: bittensor.subtensor):
+    async def set_weights(self, miners, subtensor: bittensor.subtensor) -> bool:
         bittensor.logging.info(f"[set_weights] scores: {self.miner_scores}")
 
         if not self.miner_scores:
```
pyon12, line 163: @LegendWorrier we need to log this as well.

neurons/validators/src/core/validator.py
```diff
@@ -208,10 +209,35 @@ async def set_weights(self, miners, subtensor: bittensor.subtensor):
             wait_for_finalization=False,
             wait_for_inclusion=False,
         )
+
+        current_block = self.get_current_block(subtensor)
         if result is True:
-            bittensor.logging.info("set_weights on chain successfully!")
+            log_text = "set_weights on chain successfully!"
```
pyon12, line 215: we need to have detailed data for score, weight here.

pyon12:
@talentwebdev @LegendWorrier I think it's better to have logs for miner score data updates.
image

neurons/validators/src/core/validator.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, debug_miner=None):
         self.should_exit = False
         self.is_running = False
         self.last_job_run_blocks = 0
+        self.set_weights_success = True
```
pyon12, line 43: @LegendWorrier it should be False for the initial data, but  I don't understand why we need this...

neurons/validators/src/core/validator.py
```diff
@@ -298,7 +324,7 @@ async def sync(self):
             # fetch miners
             miners = self.fetch_miners(subtensor)
 
-            if await self.should_set_weights(subtensor):
+            if await self.should_set_weights(subtensor) or not self.set_weights_success:
```
pyon12, line 327: @LegendWorrier why we are checking the success status False here?

LegendWorrier: image
so, I did try include retry when setting weights is failed.
for that, I define self.set_weights_success  member variable.
if setting weight in fail, self.set_weights_success will set as false or if not,  true.

pyon12: @LegendWorrier so if the success flag is True, then you are not setting weights?

LegendWorrier: if self.should_set_weights(subtensor) is true, it will set the weights.

neurons/validators/src/core/validator.py
LegendWorrier: f"[set_weights] scores: {self.miner_scores}" will be added in log_text?

neurons/validators/src/protocol/vc_protocol/validator_requests.py
```diff
@@ -57,6 +57,11 @@ class ExecutorSpecRequest(BaseValidatorRequest):
     log_status: str | None
     job_batch_id: str
 
+class LogValidatorRequest(BaseValidatorRequest):
```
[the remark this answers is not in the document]
LegendWorrier: image
I did like this. Is it okay?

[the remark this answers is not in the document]
LegendWorrier: @pyon12

neurons/validators/src/clients/compute_client.py
```diff
@@ -272,6 +277,82 @@ async def wait_for_specs(self, channel: aioredis.client.PubSub):
...
+                            extra=self.logging_extra,
+                        )
+                    )
+                    continue
+
+                msg = json.loads(msg["data"])
+                specs = None
+                try:
+                    specs = LogValidatorRequest(
+                        current_block=msg["current_block"],
```
pyon12, line 307: @LegendWorrier you need to add validator_hotkey and timestamp, etc here.

neurons/validators/src/clients/compute_client.py
```diff
@@ -272,6 +277,82 @@ async def wait_for_specs(self, channel: aioredis.client.PubSub):
...
+                        )
+                    )
+                    continue
+
+                logger.info(
+                    "Sending logs of validator to compute app",
+                    extra={**self.logging_extra, "specs": str(specs)},
+                )
+
+                specs_queue.append(specs)
```
pyon12, line 329: @LegendWorrier why are you using specs_queue here?

neurons/validators/src/protocol/vc_protocol/validator_requests.py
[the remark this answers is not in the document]
pyon12: @LegendWorrier you need to discuss the detailed data format with @talentwebdev , but we need to have the validator_hotkey, timestamp and also score/weights results here.

AND it would be good to get the `timestamp` from the bittensor api when setting weight like what we did for job_batch_id.

@talentwebdev ??

PR #94, opened 2024-11-22
hotfix: fix-log_text-of-task_service

LegendWorrier:
- fix log_text at task_service.py


neurons/validators/src/services/task_service.py
```diff
@@ -215,7 +215,7 @@ async def create_task(
                         "Docker digests are not valid",
                         extra=get_extra_info({
                             **default_extra,
-                            "docker_digests": result.get('all_container_digests', [])
+                            "docker_digests": machine_specs.get('all_container_digests', [])
```
pyon12, line 218: @talentwebdev it should be machine_spec, not machine_specs

PR #97, opened 2024-11-22
Feat(docker_service): DAH-729 ports mapping issue renting flow

LegendWorrier:
- select port among the available ports returned by miner


neurons/validators/src/services/docker_service.py
```diff
@@ -41,15 +41,33 @@ def __init__(
...
         internal_ports = [22, 22140, 22141, 22142, 22143]
 
         mappings = []
         used_external_ports = set()
+        
+        # Parse the range_external_ports
+        if range_external_ports:
+            if '-' in range_external_ports:
+                # Handle range like "100-1500"
+                start, end = map(int, range_external_ports.split('-'))
```
pyon12, line 54: @LegendWorrier  could you please update this that allow space between dash?

pyon12: image

LegendWorrier: @pyon12 when space between dash, it return value correctly.
image

PR #107, opened 2024-11-28
feat: DAH-758-validator-check-ssh-connection

LegendWorrier:
- create the docker container with the docker image that is provided by customer
- after created the container, generate pub/private key and add public key into the container with docker exec command
- check the ssh connection against to the docker container


neurons/validators/src/services/docker_service.py
```diff
@@ -394,3 +408,74 @@ async def get_docker_hub_digests(self, repositories) -> dict[str, str]:
...
+        result = await ssh_client.run(command)
+        if result.exit_status != 0:
+            log_text = "Error creating docker connection"
+            log_status = "error"
+            logger.error(log_text, exc_info=True)
+
+            return False, log_text, log_status
+
+        # Run the docker inspect command to get container details
+        result = await ssh_client.run(f'docker inspect {container_name}', check=True)
```
pyon12, line 443: you don't have to get external port here. you will have the portmappings

neurons/validators/src/services/docker_service.py
```diff
@@ -403,3 +417,78 @@ async def get_docker_hub_digests(self, repositories) -> dict[str, str]:
...
 
         return all_digests
+    
+    async def setup_ssh_access(
+            self,
+            ssh_client: asyncssh.SSHClientConnection,
+            container_name: str,
+            ip_address: str,
+            username: str = "root", 
+            port_flags: str = ""
```
pyon12, line 427: @talentwebdev please use port_mappings, not port_flags

PR #131, opened 2024-12-12
Update const.py to add 550.127.08

BENgin33r:
added 550.127.08 drivers


neurons/validators/src/services/const.py
```diff
@@ -304,4 +304,5 @@
     "560.35.03": "93a3f8ef77af86b79314c00b0788aeed",
     "560.35.05": "1eec299b50e33a6cfa5155ded53495ab",
     "565.57.01": "c801dd3fc4660f3a8ddf977cfdffe113",
+    "570.127.08": "ac925f2cd192ad971c5466d55945a243",
```
pyon12, line 307: @BENgin33r could you provide us ssh connection into your machine, so we can verify this?

BENgin33r: @pyon12 send me a ssh key on here or discord.

BENgin33r: @pyon12 confirmed to be the correct hash by another user. can you merge? https://discord.com/channels/799672011265015819/1291754566957928469/1317901068276990032

PR #139, opened 2024-12-17
fix: DAH-827 - runtime error for default templates

talentwebdev:
## Describe your changes

Runtime error happened for default template, which miss volumes by default. This is caused by last commit in this branch for sanitizing volumes.

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -296,7 +296,9 @@ class DockerService:
                 )
 
                 # Prepare extra options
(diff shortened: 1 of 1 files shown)


neurons/validators/src/services/docker_service.py
```diff
@@ -296,7 +296,9 @@ async def create_container(
                 )
 
                 # Prepare extra options
-                sanitized_volumes = [volume for volume in custom_options.volumes if volume.strip()]
+                sanitized_volumes = [
```
pyon12, line 299: @talentwebdev I had some issues when the custom_options is null

PR #156, opened 2024-12-24
feat: DAH-848 - real time gpu monitoring

LegendWorrier:
- generate and inject signatures to monitoring script.


neurons/executor/src/gpus_utility.py
```diff
@@ -0,0 +1,168 @@
...
+                            "gpu_utilization": gpu_utilization,
+                            "cpu_ram_utilization": cpu_ram_utilization,
+                            "disk_utilization": disk_utilization,
+                            "timestamp": timestamp,
+                            "program_id": program_id,
+                            "signature": signature,
+                            "executor_id": executor_id,
+                        }
+                        # Send HTTP POST request
+                        async with session.post(http_url, json=payload) as response:
```
pyon12, line 125: @LegendWorrier not sure how to implement the authentication against compute-app endpoint

neurons/executor/src/gpus_utility.py
```diff
@@ -0,0 +1,168 @@
...
+                        "ram_utilization_in_percent": ram.percent
+                    }
+
+                    disk_utilization = {
+                        "disk_utilization_in_bytes": disk.used,
+                        "disk_utilization_in_percent": disk.percent
+                    }
+                    
+                    # Only send if there's a significant change in any GPU
+                    if True:
```
pyon12, line 114: @LegendWorrier I don't understand why this if statement needs here and why you are checking the threshold_percent & significant_change

neurons/validators/src/services/task_service.py
```diff
@@ -79,13 +79,70 @@ async def upload_directory(
                 # Await all upload tasks for the current directory
                 await asyncio.gather(*upload_tasks)
 
+    async def is_script_running(
```
pyon12, line 82: @LegendWorrier @talentwebdev do we need to check this? every tasks running every 10 mins and I think the gpu spec report will be finished in 10 mins?

neurons/executor/src/gpus_utility.py
LegendWorrier: sorry, I didn't update it if statement is "if should_send". And If there is no significant change, we don't need to save this info in db.

neurons/validators/src/services/task_service.py
LegendWorrier: @talentwebdev is going to answer.

PR #184, opened 2025-01-13
Update to use speedtest

BENgin33r:
Revised to remove speedtest-cli and references
Added using ookla speedtest


neurons/executor/Dockerfile
```diff
@@ -29,7 +29,7 @@ LABEL builder=false
 
 RUN apt-get update -y \
   && apt-get upgrade -y \
-  && apt-get install -y speedtest-cli hashcat hashcat-data nvidia-cuda-toolkit \
+  && apt-get install -y speedtest hashcat hashcat-data nvidia-cuda-toolkit \
```
pyon12, line 32: I tested this on my local and seems like speedtest does require license for the first setup.
image

neurons/miners/docker-compose.app.yml
```diff
@@ -17,7 +17,7 @@ services:
     image: daturaai/compute-subnet-miner:latest
     env_file: ./.env
     environment:
-      - SQLALCHEMY_DATABASE_URI=postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@db:5432/${POSTGRES_DB}
+      - SQLALCHEMY_DATABASE_URI=postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@db:${POSTGRES_PORT}/${POSTGRES_DB}
```
pyon12, line 20: 5432 should be static as it's the internal port of postgres docker container.

BENgin33r: does POST_GRES port need to be removed from the .env template then?

BENgin33r:
removed speedtest-cli and replaced with ookla speedtest.
added flags to accept speedtest user agreement when ran. 
output file to json-pretty
updated machine_scrape.py to use speedtest and capture output
updated Dockerfile to install speedtest
reverted Postgres commits

neurons/executor/Dockerfile
BENgin33r: revised speedtest cmd

neurons/executor/Dockerfile
```diff
@@ -29,7 +29,10 @@ LABEL builder=false
 
 RUN apt-get update -y \
   && apt-get upgrade -y \
-  && apt-get install -y speedtest-cli hashcat hashcat-data nvidia-cuda-toolkit \
+  && sudo apt-get remove speedtest-cli \
```
pyon12, line 32: @BENgin33r I don't think we need to have this line. and "sudo" is not available inside docker

neurons/validators/src/miner_jobs/machine_scrape.py
```diff
@@ -533,10 +533,22 @@ def get_network_speed():
     """Get upload and download speed of the machine."""
     data = {"upload_speed": None, "download_speed": None}
     try:
-        speedtest_cmd = run_cmd("speedtest-cli --json")
+        speedtest_cmd = run_cmd("speedtest --accept-license --accept-gdpr --format=json-pretty")
```
pyon12, line 536: @BENgin33r is this accepting the licence without user interact? The licence does require it's non-commercial use case. is it possible we can use the cli long-term without any issue across? or does it work only short-term and will be  broken after a while?

neurons/validators/src/services/task_service.py
```diff
@@ -835,20 +835,24 @@ async def create_task(
...
                     upload_speed = machine_spec.get("network", {}).get("upload_speed", 0)
                     download_speed = machine_spec.get("network", {}).get("download_speed", 0)
+                    download_time = machine_spec.get("network", {}).get("download_time", 0)
+                    upload_time = machine_spec.get("network", {}).get("upload_time", 0)
 
                     # Ensure upload_speed and download_speed are not None
                     upload_speed = upload_speed if upload_speed is not None else 0
                     download_speed = download_speed if download_speed is not None else 0
 
+                    job_taken_time = hashcat_taken_time - download_time - upload_time
```
pyon12, line 854: @BENgin33r we don't have to deduce the download/upload time here. hashcat is running in separated task from speedtest.

neurons/executor/Dockerfile
BENgin33r: fixed the sudo typo. this is meant to remove any conflicts prior to install of new speedtest version

BENgin33r: @pyon12 done

neurons/validators/src/miner_jobs/machine_scrape.py
BENgin33r: Yes. this accepts it to eliminate the user prompt. I've never had an issue with it but I see your point. other options that offer docker integration:

https://github.com/librespeed/speedtest/blob/master/doc_docker.md
https://github.com/openspeedtest/Speed-Test

neurons/validators/src/services/task_service.py
BENgin33r: @pyon12 I found that hashcat runs in series and after speedtest. This is meant to eliminate the duration of the speedtest from the job_taken_time for hashcat scoring

PR #185, opened 2025-01-14
correct postgres port

BENgin33r:
corrected postgres port to .env file


neurons/miners/docker-compose.local.yml
```diff
@@ -13,7 +13,7 @@ services:
     volumes:
       - db_data:/var/lib/postgresql/data
     ports:
-      - ${POSTGRES_PORT}:5432
+      - POSTGRES_PORT=${POSTGRES_PORT}
```
pyon12, line 16: incorrect port map

PR #186, opened 2025-01-14
Update docker-compose.app.yml

BENgin33r:
Update POSTGRES_PORT

### DIFF
--- a/neurons/miners/docker-compose.app.yml
+++ b/neurons/miners/docker-compose.app.yml
@@ -17,7 +17,7 @@ services:
     image: daturaai/compute-subnet-miner:latest
     env_file: ./.env
     environment:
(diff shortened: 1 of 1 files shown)


neurons/miners/docker-compose.app.yml
```diff
@@ -17,7 +17,7 @@ services:
     image: daturaai/compute-subnet-miner:latest
     env_file: ./.env
     environment:
-      - SQLALCHEMY_DATABASE_URI=postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@db:5432/${POSTGRES_DB}
```
pyon12, line 20: 5432 is static as it's the internal port of postgres container

PR #273, opened 2025-02-20
feat(validator): DAH-1007 Integrate CUDA Program into GPU Validity Check

Royalone94:

[Task Title](https://www.notion.so/Compute-SN-c27d35dd084e4c4d92374f55cdd293f2?p=f9b26856f1a6406892b5db46446260da&pm=s)

- [ ] I wrote tests.


neurons/validators/Dockerfile
```diff
@@ -25,6 +25,11 @@ ENV PYTHONPATH=/opt/pypackages/lib:/root/app
 
 COPY . .
 
+FROM nvidia/cuda:12.8.0-devel-ubuntu24.04 AS cuda-build
```
pyon12, line 28: if we are going to use just binary file, why we need to install cuda lib here?

neurons/validators/src/services/matrix_validation_service.py
```diff
@@ -0,0 +1,226 @@
...
+
+            # Read the matrix values
+            self.matrix = [
+                float(value)
+                for line in contents[2 : 2 + self.dim_n]
+                for value in line.split()
+            ]
+
+            # Read the bandwidth
+            bandwidth_line = contents[2 + self.dim_n].strip()
```
pyon12, line 133: @Royalone94 is it possible to set the bandwidth on vaildator side rather than from the job result?

neurons/validators/src/services/task_service.py
```diff
@@ -862,6 +865,39 @@ async def create_task(
                         clear_verified_job_info=True,
                     )
 
+                is_high_end = self.validation_service.is_high_end_gpu(machine_spec)
+                if is_high_end:
```
pyon12, line 869: matrix job should be run for the unrented machines, not all machines.

neurons/validators/src/services/matrix_validation_service.py
```diff
@@ -0,0 +1,226 @@
...
+                "dim_k": self.dim_k,
+                "bandwidth": bandwidth
+            }
+
+            logger.info(_m("Read successfully the validation result file", extra=get_extra_info(default_extra)))
+        except Exception as e:
+            logger.error(_m(f"Error reading the file: {e}", extra={}))
+
+
+class ValidationService:
```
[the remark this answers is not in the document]
Royalone94: @pyon12 , task_service.py has more than 1500 lines of codes and hard to read and maintain this service.
So instead of adding matrix validation into task_service.py, I have added new service file.

neurons/validators/src/services/task_service.py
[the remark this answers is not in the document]
Royalone94: I follow up, @talentwebdev . I will add some comments.

Royalone94: ok, @pyon12

neurons/validators/src/services/matrix_validation_service.py
Royalone94: It is impossible to set the bandwidth on validator side since bandwidth is coming from executor side which has GPU.

PR #330, opened 2025-04-01
Feat/dah 1101 subnet protect cuda program

Royalone94:
## Describe your changes
- [ ] Embed matrix multiplication validation into so file
- [ ] Instead of returning result csv file, Executor return decrypted uuid from executor sent by cipher text.
- [ ] Validator generate random seed, dim_n, dim_k, generate cipher text based on machine info

https://www.notion.so/Subnet-Protect-CUDA-Program-1adb8bfdbde98002bc46de4a84a7846b?pvs=4)

[Task Title](https://www.notion.so/Compute-SN-c27d35dd084e4c4d92374f55cdd293f2?p=f9b26856f1a6406892b5db46446260da&pm=s)

- [ ] I wrote tests.


neurons/validators/src/services/matrix_validation_service.py
```diff
@@ -169,55 +167,58 @@ def get_max_matrix_dimensions(self, gpu_memory, dim_n):
...
-        remote_result_validation_file_path = f"{remote_dir}/validate_result.txt"
         verifier_params = VerifierParams.generate()
         gpu_memory = self.get_gpu_memory(machine_spec)
         verifier_params.dim_k = int(self.get_max_matrix_dimensions(gpu_memory, verifier_params.dim_n))
-        verifier_params.result_path = remote_result_validation_file_path
+
+        verifier_params.cipher_text = encrypt_challenge(verifier_params.dim_n, verifier_params.dim_k, verifier_params.seed, machine_info, verifier_params.uuid)
+
+        command = (
+            f"{executor_info.python_path} {script_path} {verifier_params}"
```
pyon12, line 196: @Royalone94 seems like you are sending the uuid as well here. Please make sure it's not being sent to miners.

Royalone94: def __str__(self) -> str:
        return f"--dim_n {self.dim_n} --dim_k {self.dim_k} --seed {self.seed} --cipher_text {self.cipher_text}"
        
        
@pyon12 , we're only sending cipher_text and don't expose uuid to miners.

pyon12: okay, cool

neurons/validators/src/services/const.py
```diff
@@ -369,12 +369,4 @@
 }
 
 DOCKER_DIGEST = "52d8fcc2c4370bf324cdf17cbc586784:90ca53cb65a70096c5767941c4518f26b1e350f903d134c0a0667c1477532ee9"
```
pyon12, line 371: @Royalone94 did you check the Docker and Python digest is correct in the new Executor docker image?

PR #361, opened 2025-04-10
Fix: DAH-1101-Subnet-Protect-Cuda-Program

Royalone94:
## Describe your changes
- [ ] modify VerifierParams generate method to instance method
- [ ] Fix getCipherText_str method which is generating same cipher text and implement without memory leaks.
- [ ] Add try catch exception block in gpu verification module.

[Task Title](https://www.notion.so/Subnet-Protect-CUDA-Program-1adb8bfdbde98002bc46de4a84a7846b?pvs=4)

- [ ] I wrote tests.


Royalone94:
Screenshot 2025-04-11 at 9 05 37 AM
This is bug screenshot that generate same cipher text even though we've used different seed, iv, and aes key.

PR #406, opened 2025-04-23
feat: give 20% of the max score as plus

LegendWorrier:

-  give 20% of the max score as plus if executor is compatible with DinD

- [ ] I wrote tests.

### DIFF
--- a/neurons/validators/src/services/task_service.py
+++ b/neurons/validators/src/services/task_service.py
@@ -443,7 +443,7 @@ class TaskService:
         # if uptime in the subnet is exceed 1 month, then it'll get max score 
         # if uptime is less than 1 month, then it'll get score based on the uptime
         # give 50% of max still to avoid 0 score all miners at deployment
-        #if sysbox_runtime is true, then it'll get 20% of max score as plus
+        # If sysbox_runtime is true, then the score will be increased by PORTION_FOR_SYSBOX percent.
         one_month_in_minutes = 60 * 24 * 30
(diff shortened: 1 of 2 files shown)


neurons/validators/src/services/task_service.py
```diff
@@ -606,7 +614,8 @@ async def create_task(
                 storage = machine_spec.get("hard_disk", {}).get("free", 0)
 
                 gpu_processes = machine_spec.get("gpu_processes", [])
-
+                
+                sysbox_runtime = machine_spec.get("sysbox_runtime", None)
```
Royalone94, line 615: `sysbox_runtime = machine_spec.get("sysbox_runtime", False)`
We can simply improve code like this.
In that case, no need to set  `sysbox_runtime=bool(sysbox_runtime)`.
Instead can use `sysbox_runtime=sysbox_runtime`

Royalone94: ok, then, how you could convert to bool if it is string here?
sysbox_runtime=bool(sysbox_runtime)

sysbox value will be coming from machine scrape code?

 sysbox_runtime_cmd = 'docker info | grep sysbox'
    try:
        data["data_sysbox_runtime"] = "sysbox" in run_cmd(sysbox_runtime_cmd).strip()
    except Exception as exc:
        data["data_sysbox_runtime_scrape_error"] = repr(exc)
        
In that case it will be boolean value.

PR #425, opened 2025-05-09
feat: DAH-1311 - enhance logging and monitoring

LegendWorrier:
## Describe your changes

- Need to log normalized score per setting weights in validator 

[Subnet / Backend - Enhance Logging and Monitoring](https://www.notion.so/Subnet-Backend-Enhance-Logging-and-Monitoring-1edb8bfdbde9803fa27ef3f175afc6db)

- [ ] I wrote tests.


neurons/validators/src/core/validator.py
```diff
@@ -294,6 +294,15 @@ async def set_weights(self, miners):
                 extra=get_extra_info(self.default_extra),
             ),
         )
+        miner_hotkeys = [miners[i].hotkey for i in range(len(miners))]
+        await self.redis_service.publish(
+            NORMALIZED_SCORE_CHANNEL,
+            {
```
pyon12, line 300: @LegendWorrier I think we need to send list of map like
```
[
{"uid": 0, "weight": 12, "miner_hotkey": "miner_hotkey"
]
```

Otherwise, we don't know which miner gets what score.

LegendWorrier: sure, I will update

pyon12 (approved):

PR #535, opened 2025-07-23
feat: DAH-1383 - new incentive algorithm

LegendWorrier:

[Subnet/Backend - New Incentive Algorithm](https://www.notion.so/Subnet-Backend-New-Incentive-Algorithm-238b8bfdbde980eeb6bdd217a4e07a9b)

- [ ] I wrote tests.


neurons/validators/src/services/ioc.py
```diff
@@ -33,7 +33,8 @@ async def initiate_services():
     ioc["MinerService"] = MinerService(
         ssh_service=ioc["SSHService"],
         task_service=ioc["TaskService"],
-        redis_service=ioc["RedisService"]
+        redis_service=ioc["RedisService"],
+        collateral_contract_service=ioc["CollateralContractService"]
```
pyon12, line 37: @LegendWorrier we don't need to have collateral_contract_service here.

neurons/validators/src/services/task_service.py
```diff
@@ -50,6 +50,7 @@ class JobResult(BaseModel):
     executor_info: ExecutorSSHInfo
     score: float
     job_score: float
+    collateral_deposited: bool
```
pyon12, line 53: I think it's better to make this field optional (default: False). so this way, we don't have to pass this field in every line of job result if it's failed.

PR #714, opened 2025-10-15
Customize executor container to force liumuser ssh to bash in pod container

slickroot:
## Describe your changes

A few changes in Dockerfile:
- Added a new user to the executor container `liumuser` 
- Changes into `sshd_config` to force `docker exec bash` into a container name set up by the caller.
- AcceptEnv allows the caller to specify the name of the container which is very convenient for us.

About `run.sh`:
- That change makes sure liumuser has access to docker by adding her to docker group by GID.

```bash
AcceptEnv CONTAINER_NAME
Match User liumuser
    ForceCommand bash -c '[ -n "$CONTAINER_NAME" ] && exec docker exec -it "$CONTAINER_NAME" bash || { echo "CONTAINER_NAME not set"; exit 1; }'
```

You can easily test this by building an image using the command in the Notes bellow.

[Backend / Frontend / Subnet - Web Terminal Feature without Agent SSH Key ](https://www.notion.so/Backend-Frontend-Subnet-Web-Terminal-Feature-without-Agent-SSH-Key-285b8bfdbde98098a1d8f8406074f596?v=91f1cf39ae1840cab5239391e6b51ec1&source=copy_link)

## Notes

We should start including a new environment variable during docker builds of the executor docker image

```bash
SSH_PUBLIC_KEY="$(cat ~/.ssh/id_ed25519.pub)" TAG=dev ./docker_build.sh
```

Since this is enforced in the executor container I don't see why we would include any validation about it, as Waris says miners won't change something that doesn't affect their score.

### DIFF
--- a/.github/workflows/executor_cd_dev.yml
+++ b/.github/workflows/executor_cd_dev.yml
@@ -14,6 +14,7 @@ jobs:
       TARGET_FILE: "docker-compose.app.dev.yml"
       DOCKERHUB_PAT: ${{ secrets.DOCKERHUB_PAT }} 
       DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+      SSH_PUBLIC_KEY: ${{ secrets.SSH_PUBLIC_KEY }}
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


talentwebdev (approved):
looks good

talentwebdev (approved):
looks good — updated repository secrets as well.

PR #800, opened 2025-11-14
Feat: DAH-1589 gpu splitting

pyon12:
## Describe your changes
- update rent flow and restrict gpu/cpu/memory
  - restrict gpu by gpu uuid
- tracking port mapping by pod and executor id
- update port connectivity check
  - make port connectivity optional (fatal=False)
  - run port connectivity check before rental check
  - in the clean up containers process, cleaning only the containers that were used in the port connectivity check
 - update renting-in-progress and rented pods logic to support multiple pods and multiple concurrent rental request

DAH-1590
[Subnet (gpu splitting) - pod utilization](https://www.notion.so/Subnet-gpu-splitting-pod-utilization-2a8b8bfdbde98095a294fce6f6c32efb?pvs=23)

- [ ] I wrote tests.


neurons/validators/src/services/task/pipeline_factory.py
```diff
@@ -224,4 +224,5 @@ def build_pipeline(self, checks: list[Check]) -> Pipeline:
         Returns:
             Configured Pipeline ready to run
         """
+        checks = self.build_checks()
```
slickroot: Why make the pipeline coupled with only the checks we have built-in? Can we let them nicely be injected instead? This probably broke all unit tests.

talentwebdev (approved):

PR #801, opened 2025-11-14
validator: use saved port mapping validation and test updates

arhangel66:
## Summary

  Implement persistent port mapping to ensure stable SSH ports across container
  reboots and enable future GPU splitting functionality.

##  Problem

  Unstable SSH ports after container reboot:
  - Each container restart generates new random ports from available pool
  - SSH port changes break existing connections and automation
  - Users must manually discover new port after each reboot

  No support for GPU splitting:
  - Cannot run multiple independent containers on single executor
  - No mechanism to track port allocation between different pods
  - Missing port resource isolation

###  Database Schema

  Added to PortMapping model:
  - rented_for_pod_id: UUID | None - links port to specific pod
  - docker_port: int | None - container's internal port
  - Index on rented_for_pod_id for fast lookups

###  DAO Layer (port_mapping_dao.py)

  New methods:
  - get_ports_for_pod(pod_id) - retrieve reserved ports for pod
  - reserve_ports_for_pod(executor_id, mappings, pod_id) - reserve ports and cleanup
  unused
  - release_ports_for_pod(pod_id) - free all ports on pod deletion
  - get_available_ports_excluding_rented() - get only unreserved ports
  - get_busy_external_ports() - get rented ports for filtering

### Updated cleanup:
  - clean_ports() now preserves ports with rented_for_pod_id IS NOT NULL

###   Docker Service

  Updated generate_portMappings():
  1. Check existing mapping for pod_id
  2. Reuse saved ports if available
  3. Allocate new ports only for missing ones
  4. Call reserve_ports_for_pod() to persist mapping

###   Removed Redis fallback:
  - Deleted generate_port_mapping_from_redis()
  - Database is now single source of truth
  - Simplified logic

###   Benefits

  1. Stable connections: SSH port persists across container reboots
  2. GPU splitting ready: Port tracking infrastructure for multiple pods per executor
  3. Simplified architecture: Single source of truth (removed Redis dependency)
  4. Resource isolation: Each pod has dedicated persistent port set
  5. Data freshness: Continuous port verification regardless of rental status

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -42,7 +42,6 @@ from core.utils import _m, get_extra_info, retry_ssh_command
 from daos.port_mapping_dao import PortMappingDao
 from services.const import PREFERRED_POD_PORTS, MIN_PORT_COUNT
 from services.redis_service import (
-    AVAILABLE_PORT_MAPS_PREFIX,
     STREAMING_LOG_CHANNEL,
     RedisService,
 )
@@ -614,6 +613,9 @@ class DockerService:
                     extra=get_extra_info(default_extra),
                 )
                 logger.error(log_text)
+                
+                # Release ports reserved for this pod
+                await self.port_mapping_dao.release_ports_for_pod(payload.pod_id)
 
                 return FailedContainerRequest(
                     miner_hotkey=payload.miner_hotkey,
@@ -630,6 +632,9 @@ class DockerService:
                     extra=get_extra_info(default_extra),
                 )
                 logger.error(log_text)
+                
+                # Release ports reserved for this pod
+                await self.port_mapping_dao.release_ports_for_pod(payload.pod_id)
 
                 return FailedContainerRequest(
                     miner_hotkey=payload.miner_hotkey,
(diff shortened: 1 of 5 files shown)


neurons/validators/migrations/versions/20251114_5a2cd580f6ca_add_docker_port_and_index_on_rented_for_.py
```diff
@@ -0,0 +1,38 @@
...
+down_revision: Union[str, None] = 'ba52e4725227'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('portmapping', sa.Column('rented_for_pod_id', sa.Uuid(), nullable=True))
+    op.add_column('portmapping', sa.Column('docker_port', sa.Integer(), nullable=True))
+    op.drop_constraint(op.f('uq_executor_external_port'), 'portmapping', type_='unique')
```
pyon12: @arhangel66 this port can't be unique

neurons/validators/src/services/docker_service.py
```diff
@@ -87,13 +87,22 @@ async def generate_portMappings(
         internal_ports: list[int] | None = None,
         initial_port_count: int | None = None,
         enable_jupyter: bool | None = False,
+        pod_id: UUID | None = None,
```
pyon12: pod_id is not None all the time

PR #930, opened 2026-02-01
Fix async task cancellation resource leak

OpenSourceSoul:
## Problem

When validator jobs timeout, the code cancels pending tasks using `task.cancel()` but does NOT wait for the tasks to actually complete their cancellation. This is a classic async/await anti-pattern that causes resource leaks.

## Impact

- **SSH connections** to miners not being properly closed (via asyncssh)
- **File descriptors** remaining open (the code already monitors FD count at line 401)
- **Network connections** remaining in CLOSE_WAIT state
- **Redis locks** not being released properly
- Eventually leads to "too many open files" errors requiring validator restart

## Changes

Added proper awaiting of task cancellation with exception handling for `CancelledError` in:
- `neurons/validators/src/core/validator.py` (line 399)
- `neurons/validators/src/cli.py` (line 244)

This is the standard asyncio pattern that ensures resources are properly released when tasks are cancelled.

## Testing

This fix follows Python asyncio best practices and prevents resource accumulation during job timeouts.


arhangel66:
@OpenSourceSoul 
Thanks for the PR! The issue you've identified is real - calling task.cancel() without awaiting is indeed a classic asyncio anti-pattern that can cause resource leaks.

However, I have a couple of suggestions to improve the implementation:

1. Use asyncio.gather() instead of sequential awaits

The current approach awaits each cancelled task one by one, which can be slow when there are many pending tasks:
```python
for task in pending:
    task.cancel()
    await task  # sequential - slow if many tasks
```
Better approach - cancel all first, then await together:

for task in pending:
    task.cancel()
await asyncio.gather(*pending, return_exceptions=True)

2. Add a timeout for cancellation

If a task is blocked in a synchronous operation (e.g., a blocking DNS lookup), it won't respond to cancellation and await task will hang indefinitely. Consider adding a timeout:

```python
for task in pending:
    task.cancel()
try:
    await asyncio.wait_for(
        asyncio.gather(*pending, return_exceptions=True),
        timeout=5.0
    )
except asyncio.TimeoutError:
    logger.warning("Some cancelled tasks did not complete in time")
```

Could you update the PR with these changes?

slickroot (approved):

OpenSourceSoul:
Hi! @arhangel66 @slickroot please check

PR #936, opened 2026-02-06
fix: await cancelled async tasks to prevent resource leaks

thecaptain789:
Fixes #931

task.cancel() only schedules cancellation, it doesn't wait for cleanup. This leaves SSH connections and file descriptors hanging until the task eventually finishes, causing 'too many open files' errors over time.

Added asyncio.gather with return_exceptions=True after cancelling pending tasks in both validator.py and cli.py.


pixel29913:
Thanks for the PR. We made a lot of improvements to async tasks handling in past few months so closing this. Please send a new PR, in case you still run into it.

PR #1003, opened 2026-04-24
DAH-1991: fix port 9101 race between health_check_* and user rentals

fortunelucky777:
## Summary

Production incident 2026-04-24 14:26:13 UTC: user rental of pod `599e3ed0-...` on executor `7704d95d-...` failed with `Bind for 0.0.0.0:9101 failed: port is already allocated`. Root cause: a backend-created `health_check_1777040756` container (from `RentalVerificationCheck` → compute-app) held port 9101 for ~27s, and `docker run -p 9101:9101` landed inside that window.

Three validator-side guards, one atomic PR:

1. **`services/const.py`** — new `RENTAL_CONTAINER_PREFIXES = ("pod_", "container_", "health_check_")` constant as single source of truth for short-lived rental-related prefixes.
2. **`services/container_cleanup.py`** — `_get_all_rental_containers` iterates the constant (was hardcoded `pod_*` + `container_*` only). Stale `health_check_*` orphans from a crashed backend are now swept on every 15-min cycle.
3. **`services/docker_service.py`** — `wait_for_port_check_containers` OR-filters `^container_{hotkey}_` (preserved, hotkey-scoped) AND `^health_check_` (new, hotkey-agnostic — the backend creates these without a hotkey segment). `create_container` wraps `docker run` in a narrow retry: on stderr matching `Bind for 0\.0\.0\.0:\d+ failed: port is already allocated`, parse colliding port, drop from `available_ports_raw` (prevents `generate_portMappings` re-picking the same port via `max()/min()` determinism), regenerate mappings, retry. 2 retries × 3s backoff. Non-port-allocated errors surface as today.

Absorbs DAH-1957 (stale `health_check_*` cleanup). **Archive DAH-1957 after this merges.**

Architect ask 1 (port reservation drift across retries) was investigated and dismissed: backend writes `pod.ports_mapping` only post-success (`lium-io-backend/apps/server/src/services/executor.py:212`); retry re-pops from the in-memory payload under the same lock — no backend-state drift.

Scale: prod `rental_history` shows 5–9 `rent_failed`/day across many distinct executors (Apr 18–24). This is a class of bug, not a single broken host.

Design doc (consensus: Planner → Architect × 2 → Critic × 2 → APPROVE): [`.omc/plans/port-9101-race-fix.md`](.omc/plans/port-9101-race-fix.md)

## Why three layers?

The original-incident timing recon established that the collision window cannot be closed by a single guard:

- Validator-DinD / backend probe created 5s **after** the pre-flight wait check returned "no port check containers found", and 11s **before** the user's `docker run`.
- A pre-flight check alone cannot close that window → **retry** is load-bearing.
- A retry alone cannot fix orphans left by a crashed backend → **sweep** catches those.
- A sweep alone cannot block a rent landing on a live, healthy probe → **pre-flight wait** handles the common case.

## Test plan (15 new, all green locally)

```bash
cd neurons/validators && set -a && source .env.test && set +a && cd tests/
pdm run pytest test_container_cleanup.py test_docker_service.py -v
```
**Result:** 59 passed, 0 failed. No regressions in pre-existing tests.

**New — `tests/test_container_cleanup.py` (5):**
- [x] `test_cleanup_removes_stale_running_health_check` — port-freeing case, 15-min threshold
- [x] `test_cleanup_removes_stale_exited_health_check` — cosmetic case (port already released on exit)
- [x] `test_cleanup_preserves_young_health_check` — normal ~30s probe lifecycle
- [x] `test_cleanup_preserves_rented_health_check` — `rented_data` filter precondition
- [x] `test_cleanup_filter_includes_all_rental_prefixes` — regression guard for unified constant

**New — `tests/test_docker_service.py` (10):**
- [x] `TestParseAllocatedPort` (5 cases) — full docker error, unrelated error, empty, partial-match false-positive guard, multiple ports
- [x] `TestDropExternalPort` (3 cases) — drop matching, `None` input, absent port
- [x] `test_wait_for_port_check_filter_includes_health_check` — asserts docker ps command includes both filters
- [x] `test_wait_for_port_check_does_not_block_other_miner` — Scenario 3 regression guard (hotkey-scoped `container_*` isolation preserved)

## Pre-deploy gates — deployer-owned (staging)

These require a live staging environment and cannot be verified in the agent pipeline:

- [ ] **PRE-3:** on a staging executor, `docker run -d --name health_check_$(date +%s) alpine sleep 3600`. At +5 min — container still present. At +17 min — gone. Verifies 15-min sweep lands on the new prefix.
- [ ] **PRE-4:** end-to-end: start a long `health_check_*` holding port 9101, trigger a user rent. Verify: one `PORT_ALREADY_ALLOCATED_RETRY` log, then `Container created`, pod transitions to `RUNNING`.
- [ ] **PRE-6 — LOAD-BEARING:** Force a staging `create_container` failure (e.g., pre-occupy a port). Within 5 min, run
  ```
  {namespace="subnet", container="validator"} |= "Failed create_container"
  ```
  in Loki. Must return ≥1 hit. **If zero, Loki label/logger config has drifted — POST-3 and POST-4 rollout metrics are unverifiable. Escalate before merging.**

## Post-deploy metrics

- **POST-1 (+1 h):** `{namespace="subnet", container="validator"} |= "PORT_ALREADY_ALLOCATED_RETRY"` — any matches; each paired with a later `Container creation step finished` or clean fail.
- **POST-2 (+1 d):** `{namespace="subnet", container="validator"} |= "Removed stale container health_check_"` — ≥1 event (they exist in the wild per recon).
- **POST-3 (+1 d):** `SELECT COUNT(*) FROM rental_history WHERE close_reason = 'rent_failed' AND rental_end_time > now() - interval '1 day'` — below 7-day baseline (median 8/day for Apr 18–24).
- **POST-4 (+7 d):** Same SQL over 7-day window — ≥ 50% reduction vs Apr 18–24 baseline.
- **POST-5 (+7 d):** Spot-check 5 random executors — zero orphan `health_check_*` older than 15 min.

## Notes

- **Target branch:** this PR targets `main` because `staging` does not exist on origin today. Retarget to `staging` if/when that branch is restored per `contrib/DEVELOPMENT_WORKFLOW.md`.
- **Backend untouched.** `lium-io-backend/executor_health_check.py` not modified — the validator is the safety net.
- **Rollback:** single `git revert` of this commit cleanly undoes all three code surfaces + tests (no schema, no data shape, no new infra, additive code paths only).

### DIFF
--- a/neurons/validators/src/services/const.py
+++ b/neurons/validators/src/services/const.py
@@ -214,6 +214,15 @@ PREFERRED_POD_PORTS = [20000, 20001, 20002, 20003, 20004, 20005, 20006, 20007, 2
 
 POD_CONTAINER_PREFIX = "pod_"
 
+# Container name prefixes that count as "rental-related" on an executor.
+# All producers of short-lived containers competing for the 9100-9130 port range
+# MUST be listed here so that container_cleanup and wait_for_port_check_containers
+# both see them. Adding a new prefix is a one-line edit that both guards inherit.
+#   pod_*          — long-lived user rentals (validator-owned)
+#   container_*    — validator DinD/port-check probes (hotkey-scoped)
+#   health_check_* — backend executor_health_check probes (hotkey-agnostic, epoch-suffixed)
+RENTAL_CONTAINER_PREFIXES = ("pod_", "container_", "health_check_")
+
 # For simplicity, store whitelist in code. Can be updated to use DB if needed. 
 TDX_WHITELIST = {
     "OS_IMAGE_HASH": set(
--- a/neurons/validators/src/services/container_cleanup.py
+++ b/neurons/validators/src/services/container_cleanup.py
@@ -4,7 +4,7 @@ from typing import Optional
 from protocol.vc_protocol.compute_requests import RentedExecutorsResponse
 from core.docker_utils import DockerCommand
 from core.utils import _m
-from services.const import POD_CONTAINER_PREFIX
+from services.const import POD_CONTAINER_PREFIX, RENTAL_CONTAINER_PREFIXES
 
 logger = logging.getLogger(__name__)
 
@@ -103,12 +103,16 @@ class ContainerCleanup:
         return len(removed_names), removed_names
 
     async def _get_all_rental_containers(self, ssh_client) -> list[str]:
-        """Get all containers with rental-related prefixes."""
+        """Get all containers with rental-related prefixes.
+
+        Iterates RENTAL_CONTAINER_PREFIXES (services/const.py) so any short-lived
+        container that competes for the rental port range is caught by cleanup.
+        A stale running `health_check_*` from a crashed backend would otherwise
+        hold ports 9100-9130 indefinitely.
+        """
         try:
-            # Get both pod_ and container_ prefixed containers in one command
-            result = await ssh_client.run(
-                DockerCommand.ps_filter(f"{POD_CONTAINER_PREFIX}*", "container_*")
-            )
+            patterns = [f"{prefix}*" for prefix in RENTAL_CONTAINER_PREFIXES]
+            result = await ssh_client.run(DockerCommand.ps_filter(*patterns))
             if result.stdout and result.stdout.strip():
                 return result.stdout.strip().split('\n')
         except Exception as e:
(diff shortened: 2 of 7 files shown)


fortunelucky777:
Verified that it's working on executor. Create health container and proceed rent flow before deploying this branch. Result is that it failed to create. 
error-case

After deploying this branch, it worked as expected!

arhangel66 (approved):
looks good

PR #1010, opened 2026-04-29
DAH-2013: floor positive weights so they never collapse to u16=0 on emit

fortunelucky777:
## Summary
- bittensor's `convert_weights_and_uids_for_emit` max-upscales floats to u16 via `round(w / max(w) * 65535)` and **silently drops** any uid that rounds to 0. With our burn-heavy weighting (10 uids at the cap), any organic miner with score below ~7.6e-6 of the burn target disappears from chain — indistinguishable from miners that earned nothing.
- Replace the SDK call in `SubtensorClient.set_weights` with a tiny local helper that does the same max-upscale but lifts `u16=0 → 1` whenever the input float was strictly positive. SDK-zeroed entries (quantile / permit / max-limit exclusions) remain dropped.
- Add `floored_from_zero_count` to the structured log payload so we can track how often this fires in production.

## Test plan
- [ ] `pdm run pytest neurons/validators/tests/test_incentive_flow.py -v -k "convert_weights_with_positive_floor or set_weights_floors"` — four new unit/flow tests pass
- [ ] `pdm run pytest neurons/validators/tests/test_incentive_flow.py -v` — existing scenarios still pass
- [ ] `pdm run ruff check neurons/validators/src/clients/subtensor_client.py neurons/validators/tests/test_incentive_flow.py`
- [ ] After deploy: grep validator logs for `floored_from_zero_count` to confirm the floor fires on every cycle with a long tail (cross-check against `latestsetweights` row size in the DB).

## Notes / out of scope
- Burn-allocation cap (currently u16::MAX, compresses dynamic range available to the tail) is **not** changed — separate conversation.
- Rank-stable tie-break nudge to differentiate miners with near-equal scores is **not** added — separate change.
- Chain side: `vec_u16_max_upscale_to_u16` (`pallets/subtensor/src/subnets/weights.rs:833`) is an identity for inputs already at u16-cap, so a `1` we send survives the chain's defensive re-upscale.

### DIFF
--- a/neurons/validators/src/clients/subtensor_client.py
+++ b/neurons/validators/src/clients/subtensor_client.py
@@ -52,24 +52,24 @@ def _log_sync_block(name: str, *, extra: dict | None = None):
 def _convert_weights_with_positive_floor(
     uids,
     weights,
-) -> tuple[list[int], list[int], int]:
+) -> tuple[list[int], list[int], list[int]]:
     """Max-upscale floats to u16 like bittensor's `convert_weights_and_uids_for_emit`,
     but bump u16=0 → u16=1 whenever the source float was strictly positive. SDK-zeroed
     entries (quantile/permit/max-limit exclusions) stay dropped.
 
-    Returns (uids, weights, floored_count) where floored_count is the number of
-    entries that would have rounded to 0 but were lifted to 1.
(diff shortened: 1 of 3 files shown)


arhangel66 (approved):
looks good

PR #1014, opened 2026-04-30
Close stale subtensor websockets - DAH-1964

pyon777:
## Describe your changes

- Add safe subtensor cleanup that attempts to close the subtensor, substrate, and underlying websocket independently.
- Close the previous validator subtensor instance after a successful replacement.
- Close a newly constructed subtensor if initialization fails before it is installed.

Related DAH-1964 PRs:
- Datura-ai/lium-io-backend#592
- Datura-ai/lium-miner-portal#119

[DAH-1964](https://www.notion.so/DAH-1964)

- [ ] I wrote tests.

### DIFF
--- a/neurons/validators/src/clients/subtensor_client.py
+++ b/neurons/validators/src/clients/subtensor_client.py
@@ -26,6 +26,33 @@ SUBTENSOR_BACKOFF_INITIAL = 12
 SUBTENSOR_BACKOFF_MAX = 300
 
 
+def _close_subtensor_safely(subtensor) -> None:
+    """Close a bittensor.subtensor and its substrate/WS independently.
+
+    Why: the leaking-mid-rate-limit case can leave the WS in a half-broken
+    state where one close path raises while a sibling still works; we want
+    to drive the TCP FIN through whichever path still functions.
+    """
+    if subtensor is None:
+        return
+    try:
+        subtensor.close()
+    except Exception:
+        pass
+    substrate = getattr(subtensor, "substrate", None)
+    if substrate is not None:
+        try:
+            substrate.close()
+        except Exception:
+            pass
+        ws = getattr(substrate, "ws", None)
+        if ws is not None:
+            try:
+                ws.close()
+            except Exception:
+                pass
+
+
 @contextlib.contextmanager
 def _log_sync_block(name: str, *, extra: dict | None = None):
     """Time a synchronous call that runs inside the asyncio event loop. Emits a single
@@ -143,6 +170,7 @@ class SubtensorClient:
         return SubtensorClient._subtensor
 
     def initialize_subtensor(self):
+        new_subtensor = None
         try:
             logger.info(
                 _m(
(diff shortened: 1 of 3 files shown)


arhangel66 (changes requested):
please solve merge conflicts

surcyf123:
Triage: a real leak fix (stale subtensor websockets, DAH-1964) that Mikhail asked to be rebased 124 days ago; the author no longer contributes and the branch conflicts. @arhangel66 / @taiberium: adopt (small diff, worth carrying) or close? Left open.

neurons/validators/src/clients/subtensor_client.py
```diff
@@ -143,19 +170,26 @@ def subtensor(self):
...
-            SubtensorClient._subtensor = subtensor
+            self.check_registered(new_subtensor)
+
+            old_subtensor = SubtensorClient._subtensor
+            SubtensorClient._subtensor = new_subtensor
+            # Hand-off complete; clear local ref so the finally block does not
+            # close the now-installed instance.
+            new_subtensor = None
+            if old_subtensor is not None:
+                _close_subtensor_safely(old_subtensor)
```
surcyf123, line 192: Closing the old instance immediately after the swap breaks any caller that grabbed `_subtensor` before the swap and is still mid-call. Are all subtensor calls on the loop thread? If not, defer this close.

neurons/validators/src/clients/subtensor_client.py
```diff
@@ -26,6 +26,33 @@
...
+
+    Why: the leaking-mid-rate-limit case can leave the WS in a half-broken
+    state where one close path raises while a sibling still works; we want
+    to drive the TCP FIN through whichever path still functions.
+    """
+    if subtensor is None:
+        return
+    try:
+        subtensor.close()
+    except Exception:
```
surcyf123, line 40: Three bare `pass` in a row hide which close path failed. A debug log with the path name makes the next source-IP-cap incident diagnosable.

neurons/validators/tests/fixtures/incentive_fixtures.py
```diff
@@ -41,6 +41,7 @@ def mock_settings(monkeypatch):
     monkeypatch.setattr(settings, "NEW_BURNERS", [100, 101])
     monkeypatch.setattr(settings, "ENABLE_NEW_BURN_LOGIC", True)
     monkeypatch.setattr(settings, "DRY_RUN", False)
+    monkeypatch.setattr(settings, "SKIP_COLLATERAL_PENALTY", False)
```
surcyf123, line 44: Unrelated to the websocket cleanup. Split into its own commit, or drop after the rebase if main already has it.

surcyf123 (commented):
Comment only: this needs a rebase on main before a verdict. Mikhail asked for that on 7 May, the branch has been conflicting since, and the author no longer contributes. If nobody adopts it within 14 days (by 3 Oct 2026) the loop will propose closing it; the fix itself is small and worth carrying.

On the content at head `6821f3ef`:

1. **[medium]** `neurons/validators/src/clients/subtensor_client.py:192`: the class attribute is swapped and the old instance is closed right away. Any caller that read `SubtensorClient._subtensor` before the swap and is still inside a call now gets a closed websocket instead of a working one. Before this PR the old object leaked but kept working. Question for whoever adopts: are all subtensor calls on the event loop thread, so nothing can be mid-call at the swap? If not, defer the close (after the next block, or with a short delay) rather than closing immediately.
2. **[low]** `neurons/validators/src/clients/subtensor_client.py:40`: three bare `except Exception: pass`. Log at debug with the path that failed so a half-closed socket is visible the next time the source-IP cap is hit.
3. **[low]** `neurons/validators/tests/fixtures/incentive_fixtures.py:44` and `tests/prod_snapshot/test_prod_snapshot_incentive.py:134`: the `SKIP_COLLATERAL_PENALTY` monkeypatches have nothing to do with websocket cleanup. If they were needed to make the suite pass at the time, they belong in their own commit, or are already on main after a rebase.

Owner for the decision: Mikhail (arhangel66), with Rustam for the validator side.

PR #1017, opened 2026-05-03
Prevent rental port collisions - DAH-2035

pyon777:
## Describe your changes

- Recheck validator/backend port-check containers on the existing SSH session immediately before the rental docker run.
- Remove stale Created-state rental containers between port-allocation retries so the next retry does not collide with a reserved pod name.
- Add validator Docker service unit coverage for stale-name cleanup, cleanup failures, non-port errors, reused SSH sessions, and stale health_check force-cleaning.

Tests were added but not run during this push/PR workflow.

[DAH-2035](https://www.notion.so/DAH-2035)

- [x] I wrote tests.


arhangel66 (approved):
looks ok

PR #1015, opened 2026-05-01
DAH-1959: replace Docker Hub pull with local docker info probe in DinD sysbox check

arhangel66:
## Describe your changes

Validator's DinD `Sysbox check` previously ran `docker pull hello-world` inside the DinD container and treated **any** non-zero exit as a sysbox failure. When Docker Hub rate-limited that pull (`toomanyrequests`, HTTP 403), executors were marked sysbox=False even though that response itself proves Docker daemon and network are healthy. Downstream `incentive/rental_price.py:221` cuts the executor's `sysbox_multiplier`, so honest miners lost rewards.

This change removes the external registry dependency entirely. Inside the DinD container the probe now runs:

```
docker info --format '{{json .Runtimes}}'
```

and sets `sysbox_runtime=True` iff `sysbox-runc` is present in the inner Docker daemon's runtime list. sysbox-runc registers itself in the inner daemon at DinD startup (when the outer container is launched with `--runtime=sysbox-runc`), so this is a strictly correct, registry-free check.

## Production data motivating the fix

Single validator (`5F7X5UpKSr26KU3jKfpLmT8kuKtBNyHhEnfS8xtxPCqCb13p`), 24h:

| Metric | Value |
| --- | --- |
| Total `Sysbox check failed` events | 577 |
| Caused by Docker Hub ratelimit (false negatives) | **465 (~81%)** |
| Real "docker daemon down" failures | 96 |
| Other network/TLS errors | 11 |
| Distinct executors hit by ratelimit | 24 |
| Distinct IPs hit by ratelimit | **3** (executors share anonymous Docker Hub quota of 100 pulls / 6h) |

After this change, those 465 false negatives stop entirely.

## Why not just whitelist `toomanyrequests` in stderr

Earlier commit on this branch tried that. It only handled the *known* error string and still coupled the sysbox signal to an unrelated external service (Docker Hub). The current commit supersedes the whitelist approach: switching to a local probe makes the check immune to Docker Hub rate-limits, outages, TLS issues, DNS issues, and any other registry-side problem. The whitelist diff is reverted in the final state.

## Issue ticket number and link

[DAH-1959 — Sysbox check incorrectly fails when Docker Hub rate limit is hit](https://www.notion.so/Sysbox-check-incorrectly-fails-when-Docker-Hub-rate-limit-is-hit-345b8bfdbde981959740e87d38f5d274)

## Checklist before requesting a review

- [x] Self-review performed
[...]

PR #1018, opened 2026-05-04
DAH-2016: fix incentive tests after collateral-penalty default flip

arhangel66:
## Why

[PR #1011](https://github.com/Datura-ai/lium-io/pull/1011) (DAH-2016) flipped `SKIP_COLLATERAL_PENALTY` default to `True`, disabling the uptime/collateral penalty in production until slashing ships. Tests were not updated in that PR and have been red on `main` ever since: 12 failures across `tests/test_incentive_flow.py`, `tests/test_rental_price_incentive_flow.py`, and `tests/prod_snapshot/`.

The penalty branch in [incentive/default.py:141](neurons/validators/src/incentive/default.py#L141) is still in the code (just behind the flag), and it remains exercised in production whenever `SKIP_COLLATERAL_PENALTY=false` is set on the validator pod.

## What

1. **`tests/fixtures/incentive_fixtures.py`** — `mock_settings` now patches `SKIP_COLLATERAL_PENALTY=False`. Restores the historical math that the `expected_score()` helper in `test_incentive_flow.py` already models, so the penalty branch stays under test instead of silently going dark.

2. **`tests/prod_snapshot/fixtures/`** — regenerated all 3 snapshot fixtures via `pytest --update-snapshot`. These snapshots intentionally track the *current* production default (penalty disabled) and would otherwise lock to an obsolete pre-DAH-2016 state.

## Verification

```
pdm run pytest tests/prod_snapshot/ tests/test_incentive_flow.py tests/test_rental_price_incentive_flow.py
# 76 passed, 4 skipped
```

All 12 originally failing tests pass. Local `test_attestation_service` and `test_port_mapping_dao` errors are pre-existing env issues (TDX env vars, test DB) unrelated to this diff and pass on CI.

## Coverage note

Penalty math is now exclusively covered by the unit-test path (`SKIP_COLLATERAL_PENALTY=False`), while the prod snapshots cover the live default (`True`). Both branches stay under test.


arhangel66:
Superseded — folded test fixes into PR #1015 to unblock its CI.

PR #1029, opened 2026-05-08
DAH-2068 - Fix duplicate external port allocation in generate_portMappings

fortunelucky777:
## Describe your changes

On `edit_pod=true` (pod reboot/edit), `generate_portMappings` reused existing pod_mapping entries without removing their `external_port` from `available_ports`. A subsequent user-defined container port that was not in the pod_mapping fell into the `random.choice` / `min` branch and could pick the same `external_port`, producing two `-p host:container` flags with identical host ports. Docker then fails with:

```
Bind for 0.0.0.0:<port> failed: port is already allocated
```

**Fix:** one line in the pod_mapping reuse branch — pop the reused `external_port` from `available_ports` before `continue` so it cannot be picked again.

```python
# Before
reused_count += 1
continue

# After
reused_count += 1
available_ports.pop(port_mapping["external_port"], None)
continue
```

Confirmed in production logs across multiple executors (executor `97aac548`, `68654ef8`, `483c010a`, `7704d95d`) — all `edit_pod=true`, all failing with duplicate host port in the `docker run` command.

[DAH-2068](https://www.notion.so/Compute-SN-c27d35dd084e4c4d92374f55cdd293f2?p=f9b26856f1a6406892b5db46446260da&pm=s)

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -247,6 +247,7 @@ class DockerService:
                         port_mapping = pod_mapping[port]
                         mappings.append((port, port_mapping["internal_port"], port_mapping["external_port"]))
                         reused_count += 1
+                        available_ports.pop(port_mapping["external_port"], None)
                         continue
 
                     if not len(available_ports):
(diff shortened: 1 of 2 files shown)


arhangel66 (approved):
Looks good

PR #1043, opened 2026-05-21
DAH-2146 - feat(validator): add GPU model↔VRAM pre-check before libdmcompverify

fortunelucky777:
## Summary

Adds a Python-side fail-fast check that rejects obviously inconsistent GPU specs **before** invoking `libdmcompverify.so`'s `generateChallenge`. Catches:

- **Unknown model** — claimed GPU isn't in our supported set.
- **Missing field** — `gpu_model` empty or `gpu_capacity_mb <= 0`.
- **VRAM out of range** — declared VRAM doesn't match the claimed model (e.g., miner claims H100 80GB but reports 24064 MB).

The matmul-based `.so` check remains authoritative; this is a UX + cost layer on top so the validator stops wasting SSH + native-allocator cycles on obviously bad executor reports and produces typed log errors instead of opaque downstream failures.

Scope is intentionally **model ↔ VRAM only**. Heterogeneous-rig policy and same-VRAM model substitution (e.g., H100 80GB vs A100 80GB) are explicitly out of scope.

## Changes

**New modules:**
- `services/gpu_spec_table.py` — 79-entry `GPU_VRAM_RANGES` table at ±5% of NVIDIA nominal specs, `KNOWN_UNRANGED` allowlist, lookup-table normalization (`NORMALIZATION_MAP`) for CUDA-reported names that differ from `GPU_MODEL_RATES` keys (e.g., \`Tesla V100-SXM2-32GB\` → \`NVIDIA Tesla V100 Tensor Core GPU\`).
- `services/gpu_precheck.py` — `precheck_gpu_spec(gpu_model, gpu_capacity_mb)` raises one of `MissingGpuFieldError` / `UnsupportedGpuModelError` / `VramRangeMismatchError` on rejection; returns \`None\` on pass.

**Wired in two places:**
- `services/matrix_validation_service.py::encrypt_challenge()` — bad spec short-circuits to return \`\"\"\` before \`setDimension\` / \`generateChallenge\`. Safe JSON parse so malformed \`machine_info\` doesn't crash.
- `preflight/checks/matrix_check.py::_generate_cipher_text()` — bad spec raises \`CipherGenError\` before any native allocation.

**Tests** (34 total, all pass):
- \`test_gpu_precheck.py\` — 21 unit tests: all exception types, VRAM boundary parametrize, normalization fixtures, `KNOWN_UNRANGED` passthrough, CI parity test asserting every active `const.GPU_MODEL_RATES` key is covered.
- \`test_matrix_validation_service_precheck.py\` + \`test_matrix_check_precheck.py\` — 13 integration tests asserting short-circuit-before-native-call for each rejection class, happy path proceeds, malformed JSON returns safely.

Full validators suite green; no regressions introduced.

## Test plan

- [ ] CI green
- [ ] Reviewer eyeballs `gpu_spec_table.GPU_VRAM_RANGES` for any obvious VRAM mistakes (top-rented models verified: H100 80GB, H100 NVL, H200, A100-SXM4-80GB, RTX 4090)
- [ ] Reviewer confirms the multi-variant ranges (e.g., RTX 3060 8/12GB, V100 16/32GB) are acceptable trade-offs
- [ ] Reviewer confirms scope cuts (heterogeneous rigs, same-VRAM substitution) are deferred deliberately
- [ ] Staging spot-check: send manipulated `machine_spec` with VRAM out of range for the claimed model; observe pre-check rejection log and that no SSH round-trip to executor happens

### DIFF
--- a/neurons/validators/src/preflight/checks/matrix_check.py
+++ b/neurons/validators/src/preflight/checks/matrix_check.py
@@ -10,6 +10,7 @@ from typing import Optional, Tuple
 
 from preflight.base import PreflightCheck, CheckResult, CheckStatus
 from preflight.utils import suppress_library_output, get_gpu_info
+from services.gpu_precheck import GpuPrecheckError, precheck_gpu_spec
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


arhangel66 (approved):
looks good

PR #1047, opened 2026-05-26
DAH-2160 - Refactor GPU VRAM table to store only VRAM size

fortunelucky777:
## Describe your changes

Refactor the validator-side GPU VRAM table so it stores only the nominal VRAM size per GPU model; the min/max acceptance windows are now computed dynamically by a shared utility instead of being hand-maintained. **Format-only change — no pre-check value or range changes.**

- `gpu_spec_table.py`: replaced `GPU_VRAM_RANGES` (`Dict[str, List[Tuple[int, int]]]`, hand-stored `(min, max)` windows) with `GPU_VRAM_SIZES_MB` (`Dict[str, List[int]]`, nominal MB size per real SKU). Multi-variant cards keep one size per variant (e.g. RTX 4060 Ti → `[8192, 16384]`).
- Added `vram_window_for_size(nominal_mb)` — the single place min/max is derived — plus named calibration constants `VRAM_FLOOR_RATIO = 0.90` / `VRAM_CEIL_RATIO = 1.05`.
- `get_expected_vram_windows()` now computes windows from the sizes table but keeps its signature/return type, so `gpu_precheck.py` (docstring-only change), `task/checks/gpu_vram_precheck.py`, and `preflight/checks/matrix_check.py` are functionally untouched.
- Rewrote the module docstring for the new shape while preserving the calibration rationale (0.90 floor / 1.05 ceiling, discrete-windows-reject-impostors, B200-only-192GB note) and the `libdmcompverify.so` SYNC note.

Verified up front that all 79 entries' windows reproduce exactly from `round(nominal_mb × 0.90)` / `round(nominal_mb × 1.05)` (no off-formula entries, no `.5` rounding ties). The new table and golden snapshot were generated mechanically from the previous table — no hand transcription.

Performance: windows are computed per call (~1 per single-SKU model), which is negligible; no caching needed.

[DAH-2160 — Refactor GPU VRAM mapping table to store only VRAM size](https://www.notion.so/Refactor-GPU-VRAM-mapping-table-to-store-only-VRAM-size-36cb8bfdbde98157844ed339783f76d8)

- [x] I wrote tests.

### Tests
- New `test_computed_windows_match_pre_refactor_snapshot` freezes the pre-refactor windows for all 79 models (copied verbatim from the old `GPU_VRAM_RANGES`) and asserts the computed windows match exactly + the model key set is unchanged — proving no range changed.
- Added unit tests for `vram_window_for_size`, the ratio constants, and `get_expected_vram_windows` (single/multi-variant/unknown); updated the parity test and split the well-formed test into source-shape + computed-window invariants.
- `test_gpu_precheck.py` + `test_gpu_vram_precheck_check.py`: 61 passed. Full validator suite: 593 passed, 4 skipped (pre-existing).

### DIFF
--- a/neurons/validators/src/services/gpu_precheck.py
+++ b/neurons/validators/src/services/gpu_precheck.py
@@ -24,7 +24,7 @@ class GpuPrecheckError(Exception):
 
 
 class UnsupportedGpuModelError(GpuPrecheckError):
-    """Normalized model name has no entry in GPU_VRAM_RANGES or KNOWN_UNRANGED."""
+    """Normalized model name has no entry in GPU_VRAM_SIZES_MB or KNOWN_UNRANGED."""
 
 
 class VramRangeMismatchError(GpuPrecheckError):
@@ -40,7 +40,7 @@ def precheck_gpu_spec(gpu_model: Optional[str], gpu_capacity_mb: int) -> None:
 
     Raises:
         MissingGpuFieldError: gpu_model empty or gpu_capacity_mb <= 0.
-        UnsupportedGpuModelError: normalized model not in GPU_VRAM_RANGES.
+        UnsupportedGpuModelError: normalized model not in GPU_VRAM_SIZES_MB.
         VramRangeMismatchError: VRAM falls in none of the expected windows.
 
     Returns:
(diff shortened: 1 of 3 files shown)


arhangel66 (approved):
looks good. checked with real executors in DB

PR #1049, opened 2026-05-28
DAH-2158: Exclude paused executors from incentives

pixel29913:
## Describe your changes

Accept `new_rentals_paused_executor_ids` from the backend rented-executors response as an incentive-exclusion signal, while continuing to validate all executors returned by miners.

[DAH-2158](https://app.notion.com/p/Disable-node-for-new-rentals-after-current-rental-ends-36cb8bfdbde981f597e7ddec058dd776)

## Design Notes

- The validator no longer filters paused executors before scheduling validation tasks.
- WebSocket and REST miner paths still call `create_task` for every executor returned by the miner.
- `JobResult` now carries `is_new_rentals_paused`.
- Paused and unrented executors are excluded from mining score and rental-share incentive pools.
- Paused and rented executors continue through rented scoring so active rentals are not interrupted.
- Paused and unrented executors are also excluded from GPU denominator totals, matching spot-tier reward math.
- `DefaultIncentive` has a fallback guard for direct legacy/default algorithm usage; the active `rental_price` algorithm handles this before calling into the default scoring path.

## Testing

- Focused validator tests passed.
- Covers backward-compatible parsing when backend responses do not include paused executor IDs.
- Covers parsing paused executor IDs when the backend sends them.
- Covers both WebSocket and REST miner paths scheduling paused executors for validation.
- Covers paused and unrented executors earning zero incentive, not entering rental-share eligibility, and not diluting denominators.

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/core/validator.py
+++ b/neurons/validators/src/core/validator.py
@@ -311,6 +311,9 @@ class Validator:
                                         and job_result.gpu_count
                                         and (job_result.score > 0 or job_result.job_score > 0)
                                         and not job_result.is_spot
+                                        and not (
+                                            job_result.is_new_rentals_paused and not job_result.is_rented
+                                        )
                                     ):
                                         total_gpu_model_count_map[job_result.gpu_model] = total_gpu_model_count_map.get(job_result.gpu_model, 0) + job_result.gpu_count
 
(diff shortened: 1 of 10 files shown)


arhangel66 (approved):
looks good

PR #1050, opened 2026-05-29
DAH-2164 - Run stale container cleanup before port checks

fortunelucky777:
## Describe your changes

Orphaned rental-container cleanup is moved out of `TenantEnforcementCheck` into a new, dedicated `StaleContainerCleanupCheck` that runs **before** the port checks in the verification pipeline.

**Problem:** When a pod container outlives its rental — most clearly when a pod is marked `BROKEN_BY_PROVIDER`, which the platform deliberately does not tear down — the orphan keeps binding the rental port range. The backend reports the executor as NOT rented, so:

- `PortConnectivityCheck` tries to bind those ports, Docker rejects with "port is already allocated" → 0 working ports, and
- `PortCountCheck` (`fatal=True`) fails on `not is_rented and port_count < MIN_PORT_COUNT` and halts the pipeline.

The only place that ran the cleanup was `TenantEnforcementCheck`, which sits *after* those port checks. So the cleanup was never reached, the orphan was never removed, and the executor was stuck at score 0 every cycle with no path to self-heal.

**Solution:** Run the cleanup ahead of the port checks so orphaned (non-rented) containers older than the grace window are force-removed, freeing the ports for the very next port probe in the same cycle. Containers present in `rented_data` (real tenants + the filler) are never touched. The check is non-fatal and best-effort — a failure here must not change the executor's verdict. It is skipped in the GPU-split pipeline because it removes containers on the executor.

### Changes

- Add `StaleContainerCleanupCheck` (`executor.cleanup.stale_containers`, non-fatal) that calls `container_cleanup.cleanup(...)` and reports removed count/names via a `CheckResult` event.
- Add `StaleContainerCleanupMessages.CLEANED` message template (`STALE_CLEANUP_DONE`).
- Insert the check after `CollateralCheck` and before `PortConnectivityCheck` in the standard pipeline; leave it commented-out in the GPU-split pipeline.
- Remove the inline cleanup from `TenantEnforcementCheck` (now redundant).
- Export the new check from `checks/__init__.py`.
- Add `test_stale_container_cleanup_check.py` (5 tests, all passing).

DAH-2164

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/services/task/checks/__init__.py
+++ b/neurons/validators/src/services/task/checks/__init__.py
@@ -12,12 +12,13 @@ from .machine_spec_scrape import MachineSpecScrapeCheck
 from .nvml_digest import NvmlDigestCheck
 from .port_connectivity import PortConnectivityCheck
 from .port_count import PortCountCheck
-from .rented_machine import TenantEnforcementCheck
 from .rental_verification import RentalVerificationCheck
-from .tdx_host import TdxHostCheck
+from .rented_machine import TenantEnforcementCheck
 from .score import ScoreCheck
-from .start_gpu_monitor import StartGPUMonitorCheck
 from .spec_change import SpecChangeCheck
+from .stale_container_cleanup import StaleContainerCleanupCheck
+from .start_gpu_monitor import StartGPUMonitorCheck
+from .tdx_host import TdxHostCheck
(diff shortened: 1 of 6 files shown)


arhangel66 (approved):
looks good

PR #1055, opened 2026-06-03
DAH-2122: add create_container SSH reset mitigation

pixel29913:
## Summary
- Add AsyncSSH keepalives to the rental `create_container` SSH session.
- Track the active create_container step and propagate `failure_step` in `FailedContainerRequest`.
- Add a fixed 3 hour timeout for the docker pull command.
- Add focused DockerService tests for keepalives, docker pull timeout, and `docker_pull` failure_step.

## Reasoning
We are investigating `[Errno 104] Connection reset by peer` failures during customer rental `create_container`. The strongest evidence so far points to resets happening while the validator is waiting on a long docker pull over SSH, but the old failure path only reported a generic create_container failure. That made it hard to separate docker pull resets from failures in volume setup, container startup, verification, or cleanup.

The new `failure_step` field keeps this simple: the validator records the current create_container step locally, and when the flow fails it sends that step with the failed-container event. Backend and stats can then classify failures without parsing fragile free-form error strings or assuming every SSH reset happened during docker pull.

The SSH keepalives are a immediate mitigation for long quiet pulls. Some image downloads can spend a long time without producing useful stdout, while the validator is still holding the SSH session open. Keepalives at 30 seconds with 4 missed replies let AsyncSSH detect a dead peer after about 2 minutes and keep intermediate network/NAT paths from treating the connection as idle. This does not retry the operation or change executor-side docker behavior; it only makes the existing SSH session healthier and easier to diagnose when it dies.

The 3 hour docker pull timeout is a hard upper bound for the pull step. It avoids leaving the validator stuck indefinitely on a stalled pull while staying high enough to avoid cutting off very large legitimate image downloads.

### DIFF
--- a/neurons/validators/src/payload_models/payloads.py
+++ b/neurons/validators/src/payload_models/payloads.py
@@ -432,6 +432,7 @@ class FailedContainerRequest(ContainerBaseResponse):
     error_type: FailedContainerErrorTypes = FailedContainerErrorTypes.ContainerCreationFailed
     msg: str
     error_code: FailedContainerErrorCodes | None = None
+    failure_step: str | None = None
 
 
 class DuplicateExecutorsResponse(BaseModel):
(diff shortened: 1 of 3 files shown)


fortunelucky777 (approved):

arhangel66 (approved):
looks good

PR #1056, opened 2026-06-04
DAH-2122 add docker pull stall observability

pixel29913:
## Summary

Follow-up to #1055. Adds Docker pull stall observability during container creation without changing command cancellation, retry, or failure behavior.

## What changed

- Drain remote command stdout and stderr concurrently in `DockerService._stream_process_output`.
- Add a Docker-pull-only quiet-output warning after 15 minutes with no stdout/stderr activity.
- Recompute warning timing from the latest output time, so progress output resets the quiet timer accurately.
- Keep the existing Docker pull hard timeout behavior unchanged.

## Why

The SSH reset investigation showed failures around long-running `docker pull` during container creation. This adds visibility into whether a pull has gone quiet before it eventually fails or times out, while keeping execution semantics unchanged.

The stdout/stderr concurrency change matters because Docker can write meaningful progress and errors to stderr. Reading stdout to EOF before stderr can delay stderr visibility and make quiet-output detection misleading.

## Testing

- `PYTHONPATH=datura:neurons/validators/src neurons/validators/.venv/bin/python -m pytest neurons/validators/tests/test_docker_service.py`
- `PYTHONPATH=datura:neurons/validators/src neurons/validators/.venv/bin/python -m py_compile neurons/validators/src/services/docker_service.py neurons/validators/tests/test_docker_service.py`
- `git diff --cached --check`
- `git diff --check`

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -101,7 +101,9 @@ _DOCKER_NO_SUCH_CONTAINER_PHRASE = "No such container"
 # are quiet. With 30s/4, AsyncSSH declares a dead peer after about 2 minutes.
 _CREATE_CONTAINER_SSH_KEEPALIVE_INTERVAL_SEC = 30
 _CREATE_CONTAINER_SSH_KEEPALIVE_COUNT_MAX = 4
-_DOCKER_PULL_TIMEOUT_SECONDS = 3 * 60 * 60 # 3 hours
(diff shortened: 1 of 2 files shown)


arhangel66 (approved):
Looks good, but it requires some careful deployment. Real testing after deployment and monitoring for a few days afterward will be necessary.

fortunelucky777:
Implemented [here](https://github.com/Datura-ai/lium-io/pull/1075)

PR #1058, opened 2026-06-05
DAH-2186, add executor auth timeout logs

arhangel66:
## Summary

Add a bounded timeout and minimal breadcrumbs around executor authenticated POST dispatch so `/upload_ssh_key` and `/remove_ssh_key` failures are visible instead of silently hanging after miner authentication.

### Task Link

[DAH-2186](https://app.notion.com/p/Add-executor-SSH-key-validation-breadcrumbs-376b8bfdbde9816c8204fb6d379a7cbc)

---

## Problem

A live rented executor remained healthy enough to answer `/ping`, but authenticated SSH-key validation requests logged `Auth successful` and then produced no route-level success, no `200 OK`, and no terminal error. This caused a false-positive `EXECUTOR_INACTIVE_MID_RENTAL` penalty that had to be reverted manually.

## Solution

Wrap the post-auth `call_next(request)` path in an 8 second timeout. If the authenticated downstream route does not return, executor now responds with structured `504` JSON before the central miner's 10 second HTTP timeout. Add route breadcrumbs for `/upload_ssh_key` and `/remove_ssh_key` so the next incident shows whether the route was entered and which route-level step completed last.

## Changes

- Add `EXECUTOR_AUTHENTICATED_REQUEST_TIMEOUT` structured `504` response from `MinerMiddleware` when downstream route dispatch hangs after miner auth.
- Log authenticated request completion with status code and duration.
- Add minimal route-stage logs for SSH-key consistency validation, validator signature validation, and miner service calls.
- Add regression coverage for authenticated downstream timeout returning structured `504`.

## Deployment Steps

Use GitHub Action.

## Review Request

- [x] Just code review

## Other PRs

None.

## Risks

- Authenticated executor POST handlers that legitimately take more than 8 seconds will now return `504` instead of waiting until the central miner's 10 second timeout.
- `/ping`, `/hardware_utilization`, and `/containers` remain excluded from `MinerMiddleware` and are not affected.

### Test Case 1

**Actions**

Run `pdm run pytest tests/test_upload_ssh_key_auth.py -q` in `neurons/executor`.

**Expected Output**

All SSH-key auth tests pass, including the new authenticated timeout regression.

### Test Case 2

**Actions**

Run `pdm run pytest -q` in `neurons/executor`.

**Expected Output**

Executor test suite passes.

### Test Case 3

**Actions**

Run `scripts/agent/agentctl verify lium-io --quick --quiet` from the workspace root.

**Expected Output**

Repo quick verification passes.

## Checklist

- [x] Risks are documented
- [x] Tests are described

### DIFF
--- a/neurons/executor/src/middlewares/miner.py
+++ b/neurons/executor/src/middlewares/miner.py
@@ -13,7 +13,7 @@ from core.logger import _m, get_logger
 
 logger = get_logger(__name__)
 
-AUTHENTICATED_REQUEST_TIMEOUT_SECONDS = 8
+AUTHENTICATED_REQUEST_TIMEOUT_SECONDS = 30
 
 
 class MinerMiddleware(BaseHTTPMiddleware):
--- a/neurons/executor/src/middlewares/miner.py
+++ b/neurons/executor/src/middlewares/miner.py
@@ -1,4 +1,7 @@
+import asyncio
 import re
+import time
+
 import bittensor
 from fastapi.responses import JSONResponse
 from payloads.miner import MinerAuthPayload
@@ -9,6 +12,8 @@ from core.config import settings
 from core.logger import _m, get_logger
 
 logger = get_logger(__name__)
+
+AUTHENTICATED_REQUEST_TIMEOUT_SECONDS = 8
 
 
 class MinerMiddleware(BaseHTTPMiddleware):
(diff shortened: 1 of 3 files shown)


pixel29913 (approved):

neurons/executor/src/middlewares/miner.py
```diff
@@ -10,6 +13,8 @@
 
 logger = get_logger(__name__)
 
+AUTHENTICATED_REQUEST_TIMEOUT_SECONDS = 8
```
pixel29913: What's the rationale behind setting timeout to 8 seconds?  isn't that bit less? 
should we may be increase it to be 30?

arhangel66: Assumed it should <1s... but it's ok to extend to 30. thank you.

pixel29913 (approved):

PR #1061, opened 2026-06-10
DAH-2171- Plumb TFLOPS metrics through validation pipeline and seal matrix verifier result

fortunelucky777:
## Describe your changes

Two related changes to the matrix GPU-verification path.

**1. TFLOPS metrics plumbing.** Thread the GPU metrics dict produced by the matrix-validation result through the task pipeline so it reaches the result handler / specs:
- `task/pipeline.py`: add `gpu_metrics: dict | None` to the pipeline state.
- `task/checks/capability.py`: store `result.metrics` into `state.gpu_metrics` and include it in the capability event.
- `task/result_handler.py`: surface `gpu_metrics` in the emitted specs.
- Tests: `test_capability_check.py`, `test_matrix_validation_service_precheck.py`, and new `test_result_handler_gpu_metrics.py`.

**2. Seal the verifier result against stdout forgery (anti-spoof).** The validator runs `decrypt_challenge.py` on the miner-controlled executor and trusts what it prints to stdout. The recovered UUID is already unforgeable (a secret nonce the miner can't print without genuinely decrypting on the real GPU), but the TFLOPS `metrics` were emitted as **plaintext** and re-serialized by the Python wrapper — so a miner who edits only that (untrusted) wrapper could keep the real UUID while **injecting inflated metrics**.

Fix: the executor's `.so` now emits a single **AES-256-GCM authenticated blob** of `{uuid, metrics}`, keyed by the per-challenge matmul-cell key that the genuine `.so` and the validator already share and that is never exported. The validator decrypts + authenticates it with its generate-side object and reads the metrics from *inside* the blob, **rejecting** (fail-closed) any blob that fails authentication. The Python wrapper becomes a dumb pipe: it can neither fabricate a valid blob (no key) nor alter the metrics inside it (GCM tag), and it can't get the key without doing the genuine GPU work. Mirrors the `verifyx` flow; no transport/architecture change.
- `executor/src/decrypt_challenge.py`: emit the sealed blob (legacy `uuid`/`metrics` kept so pre-sealing validators keep working during independent deploys).
- `validators/.../matrix_validation_service.py`: `unsealResult` wrapper; a **per-call** verifier object that retains the seal key across the SSH round-trip (a shared object would race across concurrent validations at the `await`); and the anti-spoof gate that treats the sealed result as authoritative.

**Threat model:** defeats the *lazy script-swapper* (edits the Python wrapper but not the compiled `.so`). It does **not** defend against a miner who reverse-engineers/patches the `.so` to extract the key — that needs TEE/attestation (out of scope) and no pure-software check inside the miner's container can stop it.

**Dependency / deploy note:** the sealing path activates only once the rebuilt `libdmcompverify.so` (exporting `getSealedResult` / `unsealResult`) is deployed to `neurons/{executor,validators}/`. That `.so` is built from the companion celium-gpu-verifier PR (https://github.com/Datura-ai/celium-gpu-verifier/pull/11). Until both sides run the sealing-aware `.so` + scripts, the legacy plaintext path stays active (backward compatible). The validator-side crypto was verified in a standalone harness; full on-GPU round-trip is pending the rebuilt `.so`.

DAH-2201

- [ ] I wrote tests.


arhangel66 (approved):
overal it's good, but please keep in mind the defaultJob.
It seems today you check tflops in any way - defaultjob is active or not...
But if defaultjob is active - then tflops can be less, right?

PR #1063, opened 2026-06-11
DAH-2203, quote startup_commands as container argv in docker_service

arhangel66:
### Task Link

[DAH-2203](https://app.notion.com/p/Fix-startup_commands-host-breakout-RCE-strip-bypass-37cb8bfdbde981d3887ede8c64b36726)

---

## Problem

The validator's `docker_service.create_container` interpolates user-supplied `startup_commands` raw into the host-side `docker run ...` f-string, which runs via `/bin/sh -c` over SSH as **root on the executor host**. A newline-prefixed payload (`"\n\nsh /tmp/Jtd7.sh"`, observed in a real production attack) terminates the `docker run` line and executes the rest directly on the host.

The backend sanitizer is the first line of defense (Datura-ai/lium-io-backend#682), but the validator must not trust upstream input: this PR adds the second, defense-in-depth layer on the validator side.

## Solution

New `build_startup_command_args()` quotes `startup_commands` into a safe argv fragment before it reaches the host shell:

- The value is tokenized with `shlex.split` (honouring its own quoting), and each token is `shlex.quote`-d. The host shell can no longer interpret any metacharacter — the tokens become the **container's** command + args, never a host command.
- Legitimate quoted commands are preserved: `/bin/bash -c "service ssh start && tail -f /dev/null"` (the real top template) round-trips exactly; the `&&` runs inside the container.
- Break-out attempts collapse to harmless container arguments: a leading newline is consumed by tokenization, `;`-chaining becomes a literal argument.
- Empty values or unbalanced quotes (`shlex.split` raises `ValueError`) fall back to the image default command instead of risking a malformed host command.

## Changes

- `neurons/validators/src/services/docker_service.py`: added `build_startup_command_args()`; `create_container` uses it instead of raw f-string interpolation.
- `neurons/validators/tests/test_startup_command_quoting.py`: new test module — prod attack payload, `;`-chaining, legit quoted metacharacters, env expansion inside container shell, empty/unbalanced fallback, and a round-trip invariant over representative real templates.

## Deployment Steps

Use GitHub Action.

## Review Request

- [x] Just code review

## Other PRs

- Datura-ai/lium-io-backend#682 — first layer: backend `TemplateSanitizer` now checks metacharacters on the raw (non-stripped) value.

## Risks

- Behavior change for templates with **unbalanced quotes** in `startup_commands`: previously interpolated raw, now fall back to the image default command. Such values were already broken or dangerous.
- Quoting round-trip is covered by an invariant test over representative real templates (`bash -c "a && b | c"`, vllm/jupyter launch lines), so legitimate workloads keep their exact argv.
- The change is limited to `create_container`; other docker_service code paths are untouched.

### Test Case 1 — prod attack payload neutralized

**Actions**

`build_startup_command_args("\n\nsh /tmp/evil.sh")`

**Expected Output**

No raw `\n` in the result; the host `docker run` line cannot be terminated. Remainder reduces to harmless container arguments `["sh", "/tmp/evil.sh"]`.

### Test Case 2 — real top template preserved

**Actions**

`build_startup_command_args('/bin/bash -c "service ssh start && tail -f /dev/null"')`

**Expected Output**

Host shell re-tokenizes the fragment back into the original argv; `&&` runs inside the container.

### Test Case 3 — unbalanced quotes fall back

**Actions**

`build_startup_command_args('bash -c "unterminated')`

**Expected Output**

Empty string → container starts with the image default command.

### Test run

`pdm run pytest` in `neurons/validators` — 633 passed, 4 skipped locally (includes the new `test_startup_command_quoting.py`, 8 tests).

## Checklist

- [x] Risks are documented
- [x] Tests are described

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -141,6 +141,29 @@ def _should_repair_stale_mountpoint(
     )
 
 
+def build_startup_command_args(startup_commands: str | None) -> str:
+    """Quote user-supplied startup_commands into a safe argv fragment.
+
+    The fragment is appended to the host-side ``docker run ... <image>`` command
+    that runs via ``/bin/sh -c`` over SSH as root on the executor. The user value
+    is split into tokens (honouring its own quoting) and each token is
+    ``shlex.quote``-d, so the host shell cannot interpret any metacharacter
+    inside it: the tokens become the container's command + args, never a host
+    command. Legitimate quoted commands such as ``bash -c "a && b"`` are
+    preserved (the ``&&`` runs inside the container); break-out attempts such as
+    a leading newline collapse to harmless container arguments. Unbalanced
+    quotes or an empty value fall back to the image default command.
+    """
+    if not startup_commands or not startup_commands.strip():
+        return ""
+    try:
+        tokens = shlex.split(startup_commands)
+    except ValueError:
+        # Unbalanced quotes etc. — don't risk a malformed/unsafe host command.
+        return ""
+    return " ".join(shlex.quote(token) for token in tokens)
+
+
 class DockerService:
     def __init__(
         self,
@@ -1583,12 +1606,8 @@ class DockerService:
                     if custom_options and custom_options.environment
                     else "-e NVIDIA_DRIVER_CAPABILITIES=all"
                 )
-                startup_commands = (
-                    f"{custom_options.startup_commands}"
-                    if custom_options
-                    and custom_options.startup_commands
-                    and custom_options.startup_commands.strip()
-                    else ""
+                startup_commands = build_startup_command_args(
+                    custom_options.startup_commands if custom_options else None
                 )
 
                 container_name = self.get_container_name(payload)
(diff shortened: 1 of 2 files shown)


arhangel66:
## DAH-2203 — verification summary

### Approach
`startup_commands` is now passed through `build_startup_command_args` (`shlex.split` then `shlex.quote` per token) before interpolation into the `docker run` command — consistent with how container name, volumes, registry password and ssh pubkey are already quoted in this file.

An earlier metacharacter-denylist approach was dropped after a data check: a denylist would have rejected ~53% of existing templates (e.g. `bash -c "service ssh start && tail -f /dev/null"`, 214 downloads) because a regex cannot distinguish a metacharacter inside a quoted `bash -c "..."` (runs in the container) from one outside it. `shlex.quote` preserves all of them while ensuring the host shell treats the value strictly as the container's command + args.

### Tests
- `pdm run pytest neurons/validators/tests/test_startup_command_quoting.py` — 8 tests, all green: round-trip preservation of legitimate quoted commands + neutralization of break-out payloads.
- Independent review: clean, no blockers.

### Deploy status
- Image built+pushed (workflow green).
- **Not yet rolled to the staging-subnet `validator` deployment** — `validator_cd_staging.yml` builds+pushes only (no rollout step); the deployment is still pinned to the prior commit. A deployment image update is needed to verify end-to-end on staging.

### Verdict
Logic correct and unit-tested; pending staging rollout for end-to-end verification.

fortunelucky777 (approved):

PR #1065, opened 2026-06-12
DAH-2200 - Add phased min NVIDIA driver requirement (rented executors exempt)

fortunelucky777:
## Describe your changes

Adds a phased gate that penalises executors below `MIN_NVIDIA_DRIVER_VERSION` (default `580.65.06`, the r580 floor that ships CUDA 13.0). Mirrors the sysbox rollout: soft penalty (`PORTION_FOR_MIN_DRIVER_BEFORE = 0.2`) before `MIN_DRIVER_CUTOFF` (2026-07-15), full gate after.

Currently-rented executors are exempt from the gate — an active customer must not lose score because the miner has not yet upgraded the host driver. The exemption lives inside `get_min_driver_multiplier(..., is_rented=...)` so it's the single source of truth.

Driver version is read from `specs["gpu"]["driver"]` in the result handler and threaded through `JobResult.nvidia_driver_version`. The multiplier is logged and surfaced in `JobResult.driver_multiplier` for both the default and rental-price incentive paths.

Tests in `neurons/validators/tests/test_min_driver_gate.py` cover: compliant driver, unknown/missing driver fails open, soft-penalty window, full gate after cutoff, numeric (not lexical) version compare, tz-aware reference time normalisation, and the rented-exemption skip in both windows.

[DAH-2200](https://www.notion.so/Compute-SN-c27d35dd084e4c4d92374f55cdd293f2?p=f9b26856f1a6406892b5db46446260da&pm=s)

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/core/config.py
+++ b/neurons/validators/src/core/config.py
@@ -132,9 +132,11 @@ class Settings(BaseSettings):
 
     # Minimum NVIDIA driver requirement. Compared as a dotted version tuple against the
     # executor's reported gpu.driver (e.g. "580.95.05"). 580.65.06 is the r580 floor that
-    # ships CUDA 13.0. Non-compliant unrented executors are fully gated (multiplier 0);
-    # currently-rented executors are exempt.
+    # ships CUDA 13.0. Before MIN_DRIVER_CUTOFF providers get a grace period (no penalty);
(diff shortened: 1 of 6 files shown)


arhangel66 (approved):
The PR describes a 0.2 soft penalty before the cutoff, but this branch returns 1.0 and no corresponding setting exists. Update the PR description

PR #1072, opened 2026-06-16
DAH-2167: inspector integration

jam6099:
## Summary

Adds runtime inspection for **rented executors**. When a customer pod is on an executor, we start a **collector** that watches for suspicious activity that interferes with rental workloads. During validation, the validator SSHes in, runs **inspector**, and logs **malicious findings** or a clean result. Scoring is unchanged for now.

### Executor rental lifecycle

- Customer pod created → start collector
- Last customer pod removed → stop collector
- Filler pods skip collector start/stop

### Validation job

- New check: `InspectorRentedCheck`
- Only runs when the executor has **customer rented pods**
- SSH in, run inspector, get a report
- Before inspect, optionally ensures collector is running (idempotent; covers existing rentals after deploy without a fresh rent)
- Logged outcomes:
  - **Clean**: no malicious findings
  - **Malicious findings**:flagged events in the report (no score impact yet)
  - **Validation error**: inspect failed; pipeline continues
- **Warning** if collector started recently (within one validation job interval): observation window may be incomplete

## Review / Test plan

- Rent customer pod → collector starts on executor
- Delete last customer pod → collector stops
- Validation on rented executor → clean or malicious findings in inspector check logs
- `ENABLE_INSPECTOR=false` → no inspector check, no collector start/stop on lifecycle
- Filler-only / unrented executors skip inspector entirely


arhangel66 (changes requested):
InspectorValidationService shares one mutable session across concurrent executor checks — neurons/validators/src/services/inspector_validation_service.py

PipelineFactory injects a single InspectorValidationService (one InspectorValidator with one mutable session) into every context. MinerService validates a miner's executors concurrently (asyncio.gather in miner_service.py:276/:1499), and validate_rented_executor has multiple awaits between start_session() and close_session(). So with ≥2 rented executors in a batch, a second start_session() clobbers the first's session → wrong/closed session for handshake/generate/verify → spurious failures or misattributed reports. Enabled by default (ENABLE_INSPECTOR=True); test_inspector_validator_reused_across_validations only covers sequential reuse.

Fix: build the InspectorValidator/session locally inside validate_rented_executor, and add a concurrent regression test (asyncio.gather of two calls).

arhangel66 (approved):
The stop decision is made after `remove_rented_machine(...)` and relies on a fresh Redis read. If that helper removes the executor rental record for the deleted container, the collector can stop while other customer pods are still running. Please base the stop condition on the remaining pods after removing only this pod, and add a test with two customer pods on one executor where deleting one does not stop the collector.

----
Make sure it will keep working in case of gpu split case when several rented pods is active and one of it closing

jam6099:
> The stop decision is made after `remove_rented_machine(...)` and relies on a fresh Redis read. If that helper removes the executor rental record for the deleted container, the collector can stop while other customer pods are still running. Please base the stop condition on the remaining pods after removing only this pod, and add a test with two customer pods on one executor where deleting one does not stop the collector.
> 
> Make sure it will keep working in case of gpu split case when several rented pods is active and one of it closing

`remove_rented_machine` only drops the pod that was deleted, not the whole executor record. If another customer pod is still running on that machine, Redis still shows it as rented (via `_has_rented_customer_containers` and we leave the collector alone. We only stop the collector when that was the last customer pod on the executor.

PR #1073, opened 2026-06-17
DAH-2213, revive duplicate-executor detection: fix redis key, observe mode

arhangel66:
### Task Link

[DAH-2213 — Revive duplicate-executor detection in the validator](https://app.notion.com/p/Revive-duplicate-executor-detection-in-the-validator-37db8bfdbde98192a821fee08af0be16)

---

## Problem

`DuplicateExecutorCheck` has been a no-op since 2025-10-29 (commit `24bb8652`). The check reads the Redis set `duplicated_machine:set` — a constant re-declared locally inside the check — while the writer (`compute_client`) populates `duplicated_machines` (the shared constant in `redis_service`). The reader's key is never written, so `is_elem_exists_in_set` always returns `False` and the check always passes.

Result: the same executor can be announced by two miner hotkeys and both get scored. Confirmed in prod on 2026-06-17 (executor `bf6d4d9a-a7d7-4e6b-8fe3-ae741b185167` passed as "unique" under `5GivXCUF5pPSHZnRHupmgLTvRT6Q24iNZyxMvzFGwgAuz1HH` and `5DcgzqMmpRTdm6ajx4NbutvRGe257PRFgSavP8np6jqAEpDr` in the same batch). Earlier Jun 9-12 incident caused double emission and miner payouts landing on the wrong wallet.

## Solution

Import the shared `DUPLICATED_MACHINE_SET` from `redis_service` and remove the divergent local constant, so the reader and writer use the same key again — restoring the pre-`24bb8652` behavior (all copies of a duplicated machine fail the check).

To de-risk re-enabling a long-dead check, it ships behind an observe-mode flag `DUPLICATE_EXECUTOR_DRY_RUN` (default `true`). In observe mode the check logs a distinct event `EXECUTOR_DUPLICATE_OBSERVED` and does **not** change score. Setting the flag to `false` enforces (duplicate fails with score 0 and clears verified job info).

## Changes

- `core/config.py`: add `DUPLICATE_EXECUTOR_DRY_RUN` env flag (default `true` = observe).
- `checks/duplicate_executor.py`: read the shared `DUPLICATED_MACHINE_SET`; drop the local `duplicated_machine:set`; add observe branch (log + pass) vs enforce branch (fail + clear).
- `task/messages.py`: add `DUPLICATE_OBSERVED` template (`EXECUTOR_DUPLICATE_OBSERVED`, severity `warning`).
- `tests/test_duplicate_executor_check.py`: fix the assertion that hardcoded the wrong key; add observe-mode, enforce-mode, and a writer→reader dual-registration regression test.

## Deployment Steps

Use GitHub Action. Ships in observe mode (`DUPLICATE_EXECUTOR_DRY_RUN=true`). After a clean observe window (no false positives), flip `DUPLICATE_EXECUTOR_DRY_RUN=false` in prod env to enforce.

## Review Request

- [x] Just code review

## Other PRs

None. (Backend `get_executors_with_duplicate_gpu_uuids` already returns all copies — no compute-app change needed.)

## Risks

- When flipped to enforce, **all** copies of a duplicated machine fail (no canonical-keep) — a legit provider migration that briefly announces one machine under two hotkeys gets score 0 on both until the conflict clears. Mitigated by the observe-mode default and a shadow window before enforcing.
- If the observe flag is left on indefinitely, dedup stays disabled (same as the original bug). Tracked in the task's post-deploy checklist.
- Scope limited to the validator's duplicate check; no change to how the duplicate set is built.

### Test Case 1 — Observe mode (default)
**Actions**: duplicate pair present in `duplicated_machines`, `DUPLICATE_EXECUTOR_DRY_RUN=true`.
**Expected Output**: `passed=True`, event `EXECUTOR_DUPLICATE_OBSERVED`, no `clear_verified_job_info`.

### Test Case 2 — Enforce mode
**Actions**: same setup, `DUPLICATE_EXECUTOR_DRY_RUN=false`.
**Expected Output**: `passed=False`, event `EXECUTOR_DUPLICATE`, `clear_verified_job_info=True`; the check reads key `duplicated_machines`.

### Test Case 3 — Dual registration (regression guard)
**Actions**: one executor announced by two hotkeys → both pairs written under `duplicated_machines`; enforce mode.
**Expected Output**: both registrations fail. Guards against the reader/writer key diverging again.

Full validator suite: **687 passed, 4 skipped**.

## Checklist

- [x] Risks are documented
- [x] Tests are described

### DIFF
--- a/neurons/validators/src/services/task/checks/duplicate_executor.py
+++ b/neurons/validators/src/services/task/checks/duplicate_executor.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
+from core.config import settings
+from services.redis_service import DUPLICATED_MACHINE_SET
+
 from ..messages import DuplicateExecutorMessages as Msg, render_message
 from ..pipeline import CheckResult, Context
 
-DUPLICATED_MACHINE_SET = "duplicated_machine:set"
 
 class DuplicateExecutorCheck:
     """Ensure a miner is not registering the same executor UUID multiple times.
@@ -23,14 +25,24 @@ class DuplicateExecutorCheck:
         )
 
         if is_duplicate:
+            what = {
+                "executor_uuid": ctx.executor.uuid,
+                "miner_hotkey": ctx.miner_hotkey,
+            }
+            if settings.DUPLICATE_EXECUTOR_DRY_RUN:
+                event = render_message(
+                    Msg.DUPLICATE_OBSERVED,
+                    ctx=ctx,
+                    check_id=self.check_id,
+                    what=what,
+                )
+                return CheckResult(passed=True, event=event)
+
             event = render_message(
                 Msg.DUPLICATE,
                 ctx=ctx,
                 check_id=self.check_id,
-                what={
-                    "executor_uuid": ctx.executor.uuid,
-                    "miner_hotkey": ctx.miner_hotkey,
-                },
+                what=what,
             )
             return CheckResult(
                 passed=False,
--- a/neurons/validators/src/services/task/checks/duplicate_executor.py
+++ b/neurons/validators/src/services/task/checks/duplicate_executor.py
@@ -25,7 +25,7 @@ class DuplicateExecutorCheck:
         )
 
         if is_duplicate:
-            what = {
+            what: dict[str, str] = {
                 "executor_uuid": ctx.executor.uuid,
                 "miner_hotkey": ctx.miner_hotkey,
             }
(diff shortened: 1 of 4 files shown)


taiberium (approved):

neurons/validators/src/services/task/checks/duplicate_executor.py
```diff
@@ -23,14 +25,24 @@ async def run(self, ctx: Context) -> CheckResult:
         )
 
         if is_duplicate:
+            what = {
```
taiberium: Maybe we need a type declaration here?

fortunelucky777 (approved):
Approving. The fix is the right shape and minimal — pulling the key from the shared constant means reader and writer can no longer drift apart the way they did since 2025-10-29, and shipping behind an observe-mode default keeps the merge itself free of production scoring changes.

PR #1074, opened 2026-06-18
DAH-2233 - Sandbox custom-Dockerfile pod builds in isolated sysbox DinD with egress firewall

fortunelucky777:
## Describe your changes

Custom-dockerfile pods previously built with `docker build --network=none` directly on the executor **host daemon**, so `RUN` steps could not fetch dependencies. This moves the build into a **throwaway sysbox Docker-in-Docker container with network enabled**, then exports the image and loads it onto the host for the rental `docker run`.

**Isolation / security (both threats from the review):**
- **Host-escape blast radius** — the build runs under `--runtime=sysbox-runc` (user-namespaced: root inside maps to an unprivileged host UID). Preflight hard-fails with `build_sysbox_unavailable` if `sysbox-runc` is absent — it never silently falls back to `runc`.
- **Network egress abuse** — a host-side egress firewall (`DOCKER-USER`, backend-aware nft/legacy helper, scoped to the DinD container IP) blocks cloud metadata (`169.254.0.0/16`) and RFC1918 while leaving public internet open. If the firewall cannot be applied the build is aborted (`build_egress_setup`) — it never runs open.
- The DinD container and its egress rules are always torn down (`finally`); the orphan sweep now also reaps leftover `lium-dind-build-*` containers.

**Build success/failure detection:** BuildKit streams progress to stderr, which `execute_and_stream_logs` reads as failure, so the inner build redirects output to stdout and emits a stderr marker only on non-zero exit.

**Changes**
- `docker_service.py` — rewrote `_custom_build_image` for the DinD flow; added `_dind_container_name`, `_write_dockerfile_into_dind`, `_parse_egress_block_cidrs`, `_egress_filter_script`, `_teardown_dind_build`.
- `config.py` — new tunables: DinD image / cpus / memory, ready timeout, egress block CIDRs.
- `custom_build_orphan_sweep.py` — sweep leftover `lium-dind-build-*` containers.
- Tests — updated A.5/A.6/A.9 for the new flow; added A.12–A.15 (sysbox-unavailable, egress failure, always-teardown, export failure) and a DinD orphan-sweep case.

**Verification**
- Full validator suite: **690 passed, 4 skipped**.
- The exact generated command sequence was run end-to-end on a live sysbox executor: internet works during build (`pypi=200`), metadata blocked (`169.254.169.254` → blocked), image loaded and ran on host, teardown left zero rules.

**Rollout note:** every executor host must have `sysbox-runc` installed and the `daturaai/dind` image present. The preflight fails safe (custom builds error cleanly) on hosts that lack either, until they are provisioned.

[DAH-2233 - Sandbox custom-Dockerfile pod builds (sysbox/rootless-BuildKit) instead of host docker build --network=none](https://app.notion.com/p/Sandbox-custom-Dockerfile-pod-builds-sysbox-rootless-BuildKit-instead-of-host-docker-build-netwo-382b8bfdbde981b89210f5acde98015e)

- [x] I wrote tests.
- [ ] Need to take care of performance? Build now adds DinD startup + image save/load (seconds–minutes + transient 2x image disk for large CUDA images); bounded by existing build timeout and new `--cpus`/`--memory` limits.

### DIFF
--- a/neurons/validators/src/core/config.py
+++ b/neurons/validators/src/core/config.py
@@ -188,6 +188,36 @@ class Settings(BaseSettings):
         env="CUSTOM_DOCKERFILE_MAX_BYTES", default=65_536,
         description="Defense-in-depth cap on dockerfile_content size; route is authoritative.",
     )
+    # DAH-2211 — internet-enabled custom builds run inside a throwaway sysbox
+    # Docker-in-Docker container (NOT on the host daemon) so the build can
+    # `--network` out for deps while a build-time escape stays confined to an
+    # unprivileged-on-host (user-namespaced) container. Egress is firewalled
+    # host-side to block cloud metadata + RFC1918. See the DAH-2211 build flow.
+    CUSTOM_DOCKERFILE_DIND_IMAGE: str = Field(
+        env="CUSTOM_DOCKERFILE_DIND_IMAGE", default="daturaai/dind:0.0.1",
+        description="Sysbox DinD image used to build custom-dockerfile pods in isolation.",
+    )
+    CUSTOM_DOCKERFILE_DIND_CPUS: str = Field(
+        env="CUSTOM_DOCKERFILE_DIND_CPUS", default="4",
+        description="--cpus limit for the throwaway DinD build container.",
+    )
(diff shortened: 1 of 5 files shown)


arhangel66 (approved):
looks good

PR #1075, opened 2026-06-18
DAH-1524 - Reduce cached-template pod deploy time and add deploy profiling

fortunelucky777:
## Describe your changes

Reduce deployment time for pods that use **cached templates**, and add deployment profiling so the impact is measurable. Validator/connector-only change (`neurons/validators`), backward compatible.

- **Skip the redundant `docker pull`.** In `DockerService.create_container`, probe `docker image inspect` (fail-open, `check=False`) before pulling. If the image is already present locally — the cached-template case — skip the pull; otherwise pull exactly as before. The custom-Dockerfile build path is untouched. Mutable tags rely on the executor's existing background cache refresh for freshness (accepted tradeoff).
- **Gate the post-creation SSH-server install** behind a new optional `ships_sshd` flag on `ContainerCreateRequest`. When truthy, the validator skips the sshd bootstrap and re-ensures `~/.ssh` so key injection still works. **Ships inert:** defaults to `None`, nothing sets it yet, so behavior is unchanged until `lium-io-backend` populates it for self-hardening templates (follow-up). The flag docstring documents the four guarantees the skip surrenders (daemon start, host keys, `PasswordAuthentication no` hardening, 30s restart watchdog).
- **Remove the vestigial 10s cleanup sleep from the critical path.** The container-cleanup step was dominated by a hardcoded `asyncio.sleep(10)` in `clean_existing_containers`, added in early 2025 as a coarse hedge against a port-allocation race. That race is now covered by explicit, bounded mechanisms (`force_remove_health_checks`, `wait_for_port_check_containers` before `docker run`, and the 90s `_run_docker_create_with_port_retry` budget), and stale-container GC also runs every verification cycle via `StaleContainerCleanupCheck`. The **GC itself stays inline** (it frees ports/volumes the new pod needs); only the blind sleep is dropped (`sleep` defaults to 0). Staging profiling showed this step at ~10.2s — virtually all of it the sleep.
- **Deployment profile summary log + per-phase container-creation breakdown.** Emit one structured, Loki-queryable `logger.info` at the end of a successful deploy with each step's duration (ms) and the total, reusing the existing `profilers` list. Skipped steps appear at ~0ms so before/after is visible. The previously single *"Docker container creation"* bucket — which blended local volume creation, GPU device probing (`build_gpu_flags`' 2-3 serial SSH probes), the port-check wait, the bare `docker run`, and the post-run running-state poll into one number, so a slow GPU probe or port-check wait read as a slow `docker run` — is now split into five distinct steps so the real bottleneck is attributable:

  | Step (in deploy order) | What it measures |
  |---|---|
  | `Docker volume creation step finished` | local volume sizing + creation (SSH-bound) |
  | `GPU device probe step finished` | `build_gpu_flags` device-minor / shared-node / `nvidia-smi -q -x` probes |
  | `Port-check wait step finished` | pre-run wait on live backend `health_check_*` probes |
  | `Docker run step finished` | the bare `docker run` (NVIDIA `--gpus` prestart hook, +sysbox/storage-opt) |
  | `Container running check step finished` | post-run running-state poll |

  Resets stay contiguous so the end-to-end total is unchanged. Together with the existing buckets, a deploy now profiles the full path: *Requested from backend → Started in subnet → Port mappings generated → SSH connection → Docker login → (Custom build **or** Docker pull) → Container cleaning → Docker volume creation → GPU device probe → Port-check wait → Docker run → Container running check → SSH service install → Adding public keys → Finished in subnet.*

- **Typed `ProfilerStep` schema for each step.** The profiler is no longer a free-form `list[dict]`: each step is a `ProfilerStep` model whose `name` is a `ProfilerStepName` **string enum**, and `ContainerCreated.profilers` is now `list[ProfilerStep]` — so the step shape is part of the response contract and an unknown/typo'd step name fails validation instead of slipping through. **Wire-compatible:** a `@model_serializer` reproduces the exact pre-existing JSON (`name`+`duration`; `name`+`timestamp` for the anchor; `+skipped` only when `True`), so `lium-io-backend` receives byte-identical payloads and needs no change. The enum values are the unchanged step strings (they double as Loki query keys). To later emit the fuller self-describing shape on the wire, drop the serializer — a one-line change to coordinate with the backend.

**Files:** `payload_models/payloads.py` (`ProfilerStep`/`ProfilerStepName` schema + typed `profilers` field), `services/docker_service.py` (pull-skip, sshd-skip gate, cleanup-sleep removal, profile summary + per-phase container-creation split, typed step construction), plus tests.

## Measured impact (staging)

Before/after deploy profiles from the "Deployment profile summary" log — cached `daturaai/pytorch` template, RTX A6000. (Both captures predate the per-phase split `d13e8656`, so `docker run` still shows as one line.)

| Step | Before | After | |
|---|---:|---:|---|
| Started in subnet | 508 ms | 368 ms | |
| Port mappings generated | 40 ms | 12 ms | |
| SSH connection established | 169 ms | 159 ms | |
| Docker login | 0 ms | 0 ms | |
| Docker pull | 127 ms · `skipped` | 115 ms · `skipped` | pull-skip — `inspect` probe only, no pull (active in both captures) |
| **Container cleaning** | **10,196 ms** | **186 ms** | ✅ 10s `asyncio.sleep` removed — the headline win |
| Docker container creation (`docker run`) | 3,223 ms | 3,896 ms | now the dominant remaining cost (~74%) |
| SSH service installation | 1,669 ms | 124 ms | sshd-skip path exercised (`ships_sshd` set for this capture; ships inert by default) |
| Adding public keys | 121 ms | 112 ms | |
| Finished in subnet | 288 ms | 273 ms | |
| **Total** | **~16,341 ms (~16.3 s)** | **~5,245 ms (~5.2 s)** | **~11.1 s faster (~68%)** |

- **Container cleaning: 10,196 → 186 ms (~10.0 s saved)** — the dominant win, from dropping the blind 10s sleep. It was ~62% of the old deploy; gone.
- **Docker pull** stayed skipped in both captures (probe-only, ~120 ms; no full pull) — confirms the pull-skip fires for cached templates.
- **`docker run` is now the dominant remaining cost (~74%, ~3.9 s).** The per-phase split (`d13e8656`) will break that line into volume creation / GPU device probe / port-check wait / bare run / running-state poll in the next profile to find the next bottleneck.
- **SSH service installation: 1,669 → 124 ms** in the "after" capture, which exercised the `ships_sshd` skip path (124 ms ≈ re-ensuring `~/.ssh` only, not an apt install). On the default path (`ships_sshd` unset) this step still runs the full bootstrap.

**Testing:** full validator suite passes — `710 passed, 4 skipped`. New `tests/test_deploy_optimizations.py` covers pull skip/run/fail-open + `check=False`, sshd skip/install (incl. a default-path regression guard), the summary log (emitted/skipped-marking/failure-path/mixed-shape-safe), payload backward-compat, a guard that the deploy path no longer requests a blocking cleanup sleep, and the `ProfilerStep` schema (enum rejects unknown names, serializer preserves the historical wire shape, `ContainerCreated` JSON round-trip). Updated `test_custom_dockerfile_build.py` + `test_docker_service.py` for the new probe; regenerated the golden snapshot.

**Risks:** Low. Pull-skip is fail-safe (probe errors fall through to a pull; a bad cached image still routes to `FailedContainerRequest`). sshd-skip is inert until the backend opts in. The summary log is additive and reads typed `ProfilerStep` attributes (cannot raise). The per-phase profiling split is measurement-only (timestamps + resets stay contiguous, end-to-end total unchanged). The `ProfilerStep` typing is wire-preserving — the `@model_serializer` keeps the JSON sent to `lium-io-backend` byte-identical, so it carries no backend-facing risk. The cleanup-sleep removal keeps the GC inline and all DAH-1991/2018/2133 race mitigations intact; recommended staging check is an orphan-port stress test to confirm the GC (not the sleep) is the protection.

**Follow-up:** `lium-io-backend` must populate `ships_sshd` (gated to a validated allowlist of self-hardening templates) for the sshd-skip to take effect on the default path.

DAH-1524

- [x] I wrote tests.
- [x] Need to take care of performance? (this PR reduces cached-template deploy time ~16.3 s → ~5.2 s on staging, ~68% faster — see "Measured impact"; mainly the ~10s cleanup-sleep removal + pull-skip)


arhangel66 (approved):
looks good

PR #1079, opened 2026-06-23
DAH-2205: Move rental Docker hot path to SDK over SSH

pixel29913:
## Summary

Move the validator's normal rental Docker hot path from host-side Docker CLI shell strings to Docker SDK calls over `ssh://`.

This PR keeps the implementation intentionally narrow: one rental SDK helper module, structured request specs for the Docker operations we use, and no runtime feature flag or dual CLI/SDK path for migrated operations.

## What changed

- Added `docker[ssh]>=7.1.0` to validator dependencies.
- Added `services/rental_docker_sdk.py` for Docker SDK over SSH connection setup and rental Docker operations.
- Added structured SDK operation logging in `services/rental_docker_observability.py`.
- Migrated normal rental create operations to SDK data/API calls:
  - registry login;
  - image pull;
  - container create/start;
  - startup command argv mapping;
  - ports, env, volumes, restart policy, runtime, GPU/device settings, CPU/memory, storage opts, shm size, and entrypoint.
- Migrated create-time local Docker volume creation to SDK `create_volume(...)`; loopback plugin installation remains CLI/SSH setup for now.
- Migrated create/start SSH bootstrap exec calls through SDK exec.
- Migrated public-key add/remove and `/etc/environment` persistence through SDK exec with stdin data.
- Migrated start, stop, rental container remove, delete-time image prune, and volume removal through SDK calls.
- Hid Docker registry passwords from payload `repr(...)` output.
- Removed dead shell helpers no longer used after migration: the legacy Docker login shell builder and unused `setup_ssh_access`.
- Added unit/security guardrails and opt-in local Docker integration coverage, including local SDK create/remove volume coverage.

## Security notes

The migrated rental hot path passes user-controlled image, credential, env, startup, public-key, volume-name, and container-name values as SDK data or stdin instead of interpolating them into host-side Docker shell command text.

Credential regression coverage includes:

- the exact old malicious Docker password payload;
- the exact old malicious Docker username payload;
- production-like legitimate passwords containing shell metacharacters.

Public-key setup/removal coverage verifies key contents are passed via stdin and the local Docker smoke test verifies add/remove actually changes `/root/.ssh/authorized_keys`.

## SSH host-key requirement

`ExecutorSSHInfo.ssh_host_key` remains optional at the shared protocol/model level because that model is also used by placeholder/error/debug records. The migrated rental Docker SDK path treats it as mandatory because Docker SDK over SSH needs operation-scoped `known_hosts` material to verify the executor SSH server identity.

If a real rental operation reaches the SDK path without `ssh_host_key`, it fails closed before Docker SDK connection or Docker operations. Create returns the existing failed-container response shape with `failure_step="docker_sdk_ssh_host_key"`; lifecycle/key operations return their existing operation-specific failed response shape. There is no Docker CLI fallback.

We checked production validator logs before keeping this strict requirement: completed validation rows showed real executors providing `ssh_host_key`, and no real completed executor rows missing it were found in the sampled 24h/7d/30d windows. The only broad `ssh_host_key: null` matches were placeholder/error monitoring records, not connectable executor responses.

If compatibility is ever needed for a real executor missing this field, the fallback should stay inside the SDK factory: fetch the remote SSH server key, log `host_key_source=compatibility_remote_fetch`, write that key into the operation-scoped temp `known_hosts`, and continue through Docker SDK. It should not fall back to host-side Docker CLI shell commands.

## Not migrated in this PR

These paths still contain Docker CLI usage and are intentionally left for follow-up because they are outside the original normal rental hot path:

- custom Dockerfile build internals, including DinD start/remove, DinD exec, inner build, image export/load, and egress helper containers;
- Jupyter setup;
- S3FS/external-volume plugin management and volume setup;
- loopback Docker volume plugin installation for local volumes; create-time local volume creation itself is migrated to SDK;
- broad cleanup helpers;
- backup/restore helpers;
- diagnostics such as container inspect/log capture on failure.

Custom Dockerfile build is rental-create-triggerable and should remain visible follow-up inventory, but it is not included in this scoped migration.

## Validation

From `neurons/validators`:

```bash
pdm run python -W ignore -m pytest tests -q --tb=short
```

Result: `759 passed, 8 skipped`.

```bash
RUN_RENTAL_DOCKER_SDK_INTEGRATION=1 RUN_RENTAL_DOCKER_SDK_SSH_INTEGRATION=1 pdm run python -W ignore -m pytest tests/integration/test_rental_docker_sdk_local.py -q --tb=short
```

Result: `4 passed`.

Also passed:

```bash
python -m py_compile neurons/validators/src/services/rental_docker_sdk.py neurons/validators/src/services/docker_service.py
git diff --check
```

Targeted grep found no references to the removed login shell helper, removed `setup_ssh_access`, or old SDK feature flag.

## Staging validation

Staging validation completed for the migrated rental hot path. The run covered successful rental creation, SDK-backed Docker operations, SSH/bootstrap behavior, env/startup handling, reboot/delete cleanup, local volume creation/removal, and negative security replay cases that are reachable through the public staging API. No PR-scoped runtime blocker was found.

### DIFF
--- a/neurons/validators/pdm.lock
+++ b/neurons/validators/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "dev"]
 strategy = ["inherit_metadata"]
 lock_version = "4.5.0"
-content_hash = "sha256:887342ee173e6fba344d288cc9dfd7ae8b01154e9d1f172db77b554df636cded"
+content_hash = "sha256:d24a3d057bc9c825d6ab5680ea10aa99d0c5f1c996ac2154060c0ad1bbf4f506"
 
 [[metadata.targets]]
 requires_python = ">=3.11,<3.13"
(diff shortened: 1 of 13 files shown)


fortunelucky777 (dismissed):

fortunelucky777 (approved):
Approving — Docker SDK rental hot-path migration looks good.

jam6099 (approved):

PR #1084, opened 2026-06-24
feat: DAH-2265 — verify cached image is digest-current (+ plan 2/3)

fortunelucky777:
## Describe your changes

Validator-side DAH-2265 work. The headline addition is **image-digest verification**: the validator now flags provider nodes that serve **stale content under an unchanged tag**. It also carries two earlier DAH-2265 plans that were sitting in the working tree.

**Image digest verification (new)**
- `CachedTemplateVerificationCheck` probe now reads the executor's `RepoDigests` (instead of exit-status only) and compares the local digest against the backend's published manifest digest (served by the companion backend PR).
- New advisory signal `recommended_image_digest_match` (`bool | None`) published to `executor.specs`:
  - `True` — cached image's RepoDigest matches the published manifest digest.
  - `False` — cached but **differs** → node holds stale content under the tag (the case we want to catch).
  - `None` — not compared (not cached / no backend digest / unreadable RepoDigest). **Strict fail-open.**
- Backend↔executor comparison is apples-to-apples: DockerHub **manifest-list** digest ↔ executor `RepoDigest` (bare `sha256:…`, normalized to tolerate `repo@sha`). Matching is strictly per-repo (guards the same-sha/different-repo false-match).
- `fatal=False`, always `passed=True` — **no scoring or `active` impact** (advisory only).
- `compute_requests.py`: `DefaultDockerImage` gains `docker_image_digest`. `messages.py`: `DIGEST_MATCH/MISMATCH/SKIPPED`. `result_handler.py`/`pipeline.py`: state field + publish-when-not-None.

**Plan 2 — cached-template verification** (base check this digest work extends): registers `CachedTemplateVerificationCheck` and publishes `recommended_image_cached`.

**Plan 3 — sparse loopback volume**: full-node rentals (`disk_share >= 1.0`) get `sparse=true` for flat-time volume creation; partial/legacy rentals stay preallocated to keep the DAH-2183 fresh-sizing math balanced.

[DAH-2265](https://www.notion.so/DAH-2265)

Companion PR (backend, serves the digest): https://github.com/Datura-ai/lium-io-backend/pull/712

- [x] I wrote tests.

  Validator suite **750 passed / 4 skipped**; digest check file covers all five truth-table branches incl. the same-sha/different-repo guard, the `repo@sha` normalization, and fail-open on unreadable RepoDigests. New files `ruff@0.5.1`-clean; zero net-new lint on edited files.

  No new hot-path cost: the digest read reuses the existing single `docker image inspect` probe (now `--format "{{json .RepoDigests}}"`); no extra SSH round-trips.

## Notes for reviewers
- **Advisory boundary:** the digest signal cannot affect score or node liveness; a wrong/missing digest is visibility noise only.
- **Backend digest is populated at push time** (`docker buildx imagetools inspect`). Until then it is `None` and the validator correctly reports `None` (skip) — no false flags.
- Out of scope (separate follow-ups): the remediation/"force re-pull" action, and surfacing `recommended_image_digest_match=False` in the provider portal.

### DIFF
--- a/neurons/validators/src/clients/backend_client.py
+++ b/neurons/validators/src/clients/backend_client.py
@@ -5,6 +5,7 @@ import json
 import logging
 import time
 from typing import Any, ClassVar, TypeVar
+from urllib.parse import urlencode
 
 import aiohttp
 import bittensor
@@ -13,6 +14,8 @@ from tenacity import retry, retry_if_result, stop_after_attempt, wait_fixed
 
 from core.utils import _m, get_extra_info
 from protocol.vc_protocol.compute_requests import (
+    DefaultDockerImage,
+    DefaultDockerImagesResponse,
     ExecutorHealthCheckResponse,
     PodRentalActiveResponse,
     RentedExecutorsResponse,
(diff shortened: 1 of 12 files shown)


arhangel66 (approved):
Approved, but please keep in mind - you do the request to get default image for every node.
so, it will be 300+ requests. Every request in backend use the redis and some logic inside.
Theoretically it's no problem, but in case of 3000 nodes it could be a problem.

maybe will need to think about named cach with TTL or so.

PR #1089, opened 2026-06-26
DAH-2274 - Read burn emission from shared config

fortunelucky777:
## Describe your changes

Wire the validator to read the burning-emission value from shared config instead of a hardcoded constant (DAH-2274), giving the value a single source of truth — the backend `/v1/shared-config` endpoint.

- New `core.config.get_total_burn_emission()`: reads `shared_client.config.total_burn_emission`, **clamps to `[0, 1]`**, and falls back to the packaged lium-core default when missing/invalid — so a misconfigured or stale endpoint cannot push a bad burn share into on-chain weights.
- `burn_service.py`, `rental_price.py`, `default.py`, `utils.py` now use the accessor instead of the local const.
- `services/const.py TOTAL_BURN_EMISSION` is no longer an independent literal — it derives from the lium-core default (offline fallback, retained for tests).
- Removed a now-unused import in `base.py`.
- Added `tests/test_total_burn_emission.py` (clamp + fallback).

Net on-chain burn is unchanged (the backend serves 0.87). Requires the backend PR deployed first.

Tests: burn/incentive suites + new accessor test — **113 passed, 4 skipped**. Ruff 0.5.1 net-clean (no new violations vs baseline).

Part of DAH-2274 — related PRs (same branch in each repo):
1. lium-io-backend#719 — https://github.com/Datura-ai/lium-io-backend/pull/719 _(authority: serve 0.87)_
2. lium-io#1089 — https://github.com/Datura-ai/lium-io/pull/1089 _(validator reads shared config)_
3. lium-miner-portal-frontend#142 — https://github.com/Datura-ai/lium-miner-portal-frontend/pull/142 _(portal label)_

Deploy in the order above.

[DAH-2274](https://www.notion.so/DAH-2274)

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/services/const.py
+++ b/neurons/validators/src/services/const.py
@@ -1,5 +1,3 @@
-from lium_core.shared_config.defaults import DEFAULT_SHARED_CONFIG
-
 MIN_JOB_TAKEN_TIME = 20
 
 GPU_MODEL_RATES = {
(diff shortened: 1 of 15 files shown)


neurons/validators/src/services/const.py
```diff
@@ -110,9 +110,11 @@
...
 # Emission split between the rented "mining" pool and the unrented + burn pool.
-# TOTAL_BURN_EMISSION caps the unrented-rental + burn pool; the remainder
-# (1 - TOTAL_BURN_EMISSION) is the rented mining pool. Raising the rented pool
-# from 0.09 to 0.13 (DAH-2273) lowers this cap from 0.91 to 0.87.
+# The LIVE value is sourced from shared config at runtime via
+# core.config.get_total_burn_emission() (DAH-2274) — the backend is the source of
+# truth. This constant is the validator's expected production value: it is what the
+# tests assert against and what the test harness serves (see tests/conftest.py).
+# (1 - TOTAL_BURN_EMISSION) is the rented mining pool.
 TOTAL_BURN_EMISSION = 0.87
```
pixel29913: if it's for testing, wouldn't it be better to mock the api call for backend and use this value from there?

pixel29913: not a blocker, but this could be confusing in future if such a value shows up on searching burn_emission etc.

pixel29913 (approved):

PR #1090, opened 2026-06-26
DAH-2291 - Fix alpha price query after finney Swap pallet migration

fortunelucky777:
## Describe your changes

**Production incident:** the validator's alpha-token-price query was failing with
`Storage function "Swap.AlphaSqrtPrice" not found`.

**Root cause:** this is a chain-runtime schema change, not a code or network bug. `bittensor==9.10.1`'s `subtensor.get_subnet_price(netuid)` reads the `Swap.AlphaSqrtPrice` storage item directly. The finney runtime upgraded and migrated its `Swap` pallet to a new AMM, removing `AlphaSqrtPrice` (the pallet now exposes `FeeRate`, `HasMigrationRun`, `PalSwapInitialized`, `ScrapReservoirAlpha`, `SwapBalancer`). The storage lookup therefore fails. Newer SDKs (9.12.x) confirm the diagnosis — they rewrote `get_subnet_price` to use the `SwapRuntimeApi.current_alpha_price` runtime API.

**Fix:** route `SubtensorClient.get_alpha_rate()` through `subtensor.subnet(netuid)` instead of `get_subnet_price()`. In 9.10.1, `subnet()` attempts `get_subnet_price()` and, when that storage query raises, **falls back internally to the `tao_in / alpha_in` reserve ratio** — returning the same alpha price in TAO. This restores the query on the current runtime **without a bittensor SDK upgrade** and is more resilient (built-in fallback rather than one hardcoded storage path).

- `get_alpha_rate()` returns the identical value the old path produced, so the downstream `epoch_subnet_emission = TEMPO * tao_price * alpha_rate` math is unchanged.
- Verified live against finney with the production-pinned stack (`bittensor==9.10.1`, `async-substrate-interface==1.6.3`, `scalecodec==1.2.11`):
  - Old path: `get_subnet_price(51).tao` → `SubstrateRequestException: Storage function "Swap.AlphaSqrtPrice" not found`
  - New path: `subnet(51).price.tao` → `0.0543…` TAO ✓ (agrees with `SwapRuntimeApi.current_alpha_price` and the `SubnetTAO/SubnetAlphaIn` ratio)
- Added `tests/test_subtensor_client_alpha_rate.py`. Test suite: 3 new + 47 price-provider + 59 incentive-flow tests pass.

[DAH-2291](https://www.notion.so/DAH-2291)

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/clients/subtensor_client.py
+++ b/neurons/validators/src/clients/subtensor_client.py
@@ -661,5 +661,13 @@ class SubtensorClient:
         return cls._subtensor
 
     def get_alpha_rate(self) -> float:
-        price = self.subtensor.get_subnet_price(netuid=self.netuid)
-        return price.tao
+        # `subtensor.get_subnet_price()` reads the `Swap.AlphaSqrtPrice` storage item,
+        # which the finney runtime removed when the `Swap` pallet was migrated to a new
+        # AMM — querying it now raises `Storage function "Swap.AlphaSqrtPrice" not found`.
+        # `subtensor.subnet()` is resilient: it attempts `get_subnet_price()` and, on that
+        # failure, falls back to the `tao_in / alpha_in` reserve ratio, yielding the same
+        # alpha price in TAO. This keeps the query working without a bittensor SDK upgrade.
+        subnet = self.subtensor.subnet(netuid=self.netuid)
+        if subnet is None or subnet.price is None:
+            raise RuntimeError(f"No subnet price available for netuid {self.netuid}")
+        return subnet.price.tao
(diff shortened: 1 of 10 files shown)


fortunelucky777:
Superseded by #1091. This PR was branched off a base that carried unrelated DAH-2274 commits (`212a2c44`, `a748003b` — burn-emission shared-config work), so its diff against `main` pulled in `config.py`, `burn_service.py`, `rental_price.py`, `const.py`, etc. #1091 contains only the DAH-2291 fix (subtensor_client.py + its regression test), cherry-picked onto a clean `main` base. Closing this one.

PR #1092, opened 2026-06-29
DAH-2167: memory leak and enable inspector feature

jam6099:
fix memory leak by loading the library once only


pixel29913 (approved):

arhangel66 (approved):
LGTM — memory-leak fix looks correct (shared CDLL + per-validator sessions, nice interleaved-session test).

One non-blocking concern before/after merge: the eager library load in InspectorValidationService.__init__ (sha256_from_path + InspectorValidator.load_library) now runs unconditionally, because the service is constructed in PipelineFactory.__init__ → TaskService.__init__ regardless of ENABLE_INSPECTOR.

PR #1095, opened 2026-06-29
DAH-2265 - Fail verification for uncached recommended template after cutoff

fortunelucky777:
## Describe your changes

Makes `CachedTemplateVerificationCheck` **conditionally critical** so that, after a cutoff, an executor that hasn't pre-pulled the recommended default image fails verification early — and the provider sees the reason.

Before `CACHED_TEMPLATE_CUTOFF` (default `2026-07-15 12:00 UTC`) the check stays **advisory**: it publishes `recommended_image_cached` / `recommended_image_digest_match` to `executor.specs` and never changes score (unchanged behavior). On/after the cutoff, the two bad signals — image not pre-pulled (`NOT_CACHED`) or stale content under an unchanged tag (`DIGEST_MISMATCH`) — return `passed=False`. Since the check is `fatal`, the pipeline halts early: the executor scores 0, and the failure event becomes `log_text`, which is published to the provider via `full_log_text`. It still **fails open** on every uncertainty (`None` — unknown GPU/driver, backend unreachable, unreadable digest), regardless of the cutoff.

**Why a fatal check instead of an incentive-layer gate:** an earlier revision gated this in the incentive layer (mirroring the Discord cutoff). The fatal-check approach is simpler and, crucially, **surfaces the reason to the provider** — the incentive gate only logged it validator-side. It's also naturally safe for active rentals: the check runs *after* `TenantEnforcementCheck`, which halts the pipeline for rented executors, so only **unrented** executors are ever gated. A customer's active rental can't be failed by this.

**Changes**
- `core/config.py` — new `CACHED_TEMPLATE_CUTOFF` constant.
- `services/task/checks/cached_template_verification.py` — `fatal=True`; `run()` returns `passed=False` only when after the cutoff AND the signal is `NOT_CACHED`/`DIGEST_MISMATCH`, rendering the event with a `severity="error"` override. Advisory specs are still published either way.
- `services/task/messages.py` — updated `CachedTemplateMessages` docstring.

**Tests** (`tests/test_cached_template_verification.py`)
- Autouse fixture pins the cutoff to advisory mode by default, so the existing 16 advisory tests are wall-clock-independent.
- New tests: fatal `NOT_CACHED` / `DIGEST_MISMATCH` after cutoff (incl. `severity == "error"` and that specs are still published), `CACHED` / `DIGEST_MATCH` pass after cutoff, fail-open uncertainty after cutoff, `DIGEST_SKIPPED` pass after cutoff, and advisory-before-cutoff.

Verification suite: 23 passed. Full validators suite (excluding the pre-existing-flaky `prod_snapshot`): 809 passed, 4 skipped.

DAH-2265

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/core/config.py
+++ b/neurons/validators/src/core/config.py
@@ -137,7 +137,7 @@ class Settings(BaseSettings):
     # unrented executor that has not pre-pulled the recommended default image, or holds stale
     # content under the same tag, fails verification early (score 0, reason surfaced to the
     # provider). Fails open: an unmeasured signal (None) is never penalised.
-    CACHED_TEMPLATE_CUTOFF: datetime = datetime(2026, 7, 8, 12, 0, 0)
+    CACHED_TEMPLATE_CUTOFF: datetime = datetime(2026, 7, 9, 12, 0, 0)
 
     # Minimum NVIDIA driver requirement. Compared as a dotted version tuple against the
     # executor's reported gpu.driver (e.g. "580.95.05"). 580.65.06 is the r580 floor that
--- a/neurons/validators/src/core/config.py
+++ b/neurons/validators/src/core/config.py
@@ -131,6 +131,14 @@ class Settings(BaseSettings):
     SYSBOX_RENTED_CUTOFF: datetime = datetime(2026, 4, 3, 12, 0, 0)
     DISCORD_INCENTIVE_CUTOFF: datetime = datetime(2026, 6, 15, 12, 0, 0)
 
+    # DAH-2265: cached-template requirement. Before the cutoff the CachedTemplateVerificationCheck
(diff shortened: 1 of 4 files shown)


arhangel66 (changes requested):
How many nodes will be affected to this action? can we an exception for rented nodes? or it's 0 affected nodes expected - then I will approve

fortunelucky777:
> How many nodes will be affected to this action? can we an exception for rented nodes? or it's 0 affected nodes expected - then I will approve

No nodes will be affect right now because I set cutoff date to 7/8. If we assume that cutoff date is passed, only 1 node. 


image

arhangel66 (approved):
looks good

PR #1098, opened 2026-07-01
DAH-2243 encrypted local volumes via gocryptfs

jam6099:
Implement encrypted volumes for rental containers


arhangel66 (changes requested):
Requesting changes — the security guarantee is not fail-safe yet.

1) VOLUME_MASTER_SECRET defaults to the public string "volume-secret", and
   volume_encryption_enabled() treats any non-empty value as "on". So without
   explicit prod config we create "encrypted" volumes whose passphrase =
   HKDF(secret, pod_id) is reproducible from source + a non-secret pod_id.
   Please drop the default and fail startup when encryption is on but no
   strong secret is provisioned. Note it's a single validator-wide secret
   guarding every volume — worth calling out in the threat model.

2) _image_has_encrypted_volume_label() fails open: a missing label OR a failed
   `docker image inspect` silently falls back to a plaintext volume while
   creation still succeeds. Please fail closed when encryption is required, or
   gate plaintext behind an explicit rollout policy and surface the mode to the
   caller (don't conflate "image unsupported" with "couldn't check").

Heads-up: the derived passphrase is delivered into the container on the miner
host, so this protects data-at-rest against disk theft / co-tenants, not
against a root-level host operator — fine if that's the intended model, just
let's make it explicit.

arhangel66 (changes requested):
We settled this in Discord after your fio numbers: encryption should be a rent-time toggle in the UI/SDK and API, on by default. ~38% read / 37.5% write overhead is enough that an IO-bound customer needs a way out. Chain is `ui/sdk → /executors/{uuid}/rent → CreateContainerRequest → docker_service`.

Right now there's no per-rent switch. `_should_encrypt_local_volume()` decides everything on the validator side, and the only levers are `ENABLE_VOLUME_ENCRYPTION` (all-or-nothing per validator) and the image label. Neither is the renter's choice.

To approve I need the flag on `ContainerCreateRequest`, defaulting to on. Backend and UI can follow later.

arhangel66 (changes requested):
Request changes — three blocking items, plus two things worth confirming before rollout.

**1. Encrypted-volume setup sits outside the cleanup boundary (docker_service.py:4198)**

`setup_encrypted_local_volume` runs after the successful `docker run` but before the post-run
`try` whose `except` (4326) calls `cleanup_failed_container_creation`. Any init/mount/verify
failure therefore falls through to the outer handler at 4415, which only logs and returns
`FailedContainerRequest` — the running container and the created volume stay on the executor,
holding GPUs and ports until the next create on that machine sweeps them via
`clean_existing_containers`. Please move the step inside the cleanup boundary and add a
regression test asserting container + volume removal when `setup_encrypted_local_volume` raises.

**2. `ENABLE_VOLUME_ENCRYPTION` defaults to True and makes a new external secret mandatory (config.py:334)**

`settings = Settings()` is evaluated at import, and the model validator rejects a missing or
short `VOLUME_MASTER_SECRET`. Nothing in this repo provisions it: no compose file, no env
template, no docs mention the variable — the only place it appears is `conftest.py`, which had
to be patched so the test suite could import at all. A validator upgrade without prior
provisioning will fail to start even though the backend never asks for encryption. Please
default the flag to false for a staged rollout, or land the provisioning first and confirm it
across all validator environments.

**3. Plaintext mount path is hardcoded to /root on every path except create**

`create_container` derives `local_volume_path` from `custom_options.volumes` (3771), but
`start_container` remounts at a hardcoded `/root` (4666) and `teardown_encrypted_local_volume`
defaults to `/root` (1894, used by both delete and the failed-create cleanup). A template with a
non-`/root` volume gets its data at `/root` after a stop/start, while the workload sees an empty
original path; unmount on teardown silently no-ops (only the `pkill` saves it). Please persist
the destination or derive it from durable container metadata, and cover it with a non-`/root`
restart test — the current test asserts `plaintext_path="/root"`, which locks the bug in.

**4. Backup/restore fails open on ciphertext (workspace_mount.py: `detect_volume_access`)**

Encryption is detected only by finding a *running* container with a `/lium-cipher` mount. If the
container is not running at that moment (restart window, while `Pod.status` is still RUNNING),
the helper silently falls back to mounting the raw volume at `/workspace` and tars the
ciphertext. When `backup_path` is the volume root that path *succeeds* and stores a garbage
backup; a later restore would extract ciphertext over the plaintext mount. This should fail
closed — e.g. if `gocryptfs.conf` exists at the volume root and no running container is found,
abort instead of falling back.

**5. Two things to confirm rather than change**

- Threat model: the passphrase is delivered to the miner's host (`docker exec -i` over SSH) and
  gocryptfs runs inside a container on that host, so a root miner can capture the key and read
  the plaintext mount via the container's mount namespace. The property delivered is
  at-rest protection against a passive miner, not confidentiality against the host. Worth
  stating explicitly in the ticket so the feature isn't read as more than it is.
- Key rotation: no key version is stored anywhere. Rotating `VOLUME_MASTER_SECRET` leaves every
  existing volume with a `gocryptfs.conf` that init skips and mount then rejects, so every
  encrypted pod becomes unrestartable. If rotation is ever expected, it needs a plan now.

Also worth double-checking against the image the label gates on: the setup script only inits and
mounts (despite the "Installing..." log text), so gocryptfs must already be in the image, and
`-nonempty` behaviour differs across gocryptfs versions — an unknown flag means the mount fails.

jam6099:
@arhangel66 recent commits should address points 1, 3, 4. Regarding #2, the secret is already set for deployment so it's not a concern.

Additionally, I improved the security model by not requiring the passphrase via stdin. Key rotation is out of scope for now.

arhangel66 (approved):
Two more fail-open gaps in the same vein as round 3's point 4 — same "diagnostic failure reads as no encryption" pattern.

1) _encrypted_local_volume_name still swallows inspect failures (docker_service.py:1942)

It catches every exception from mount_source_for_destination and returns None. In start_container, if encrypted_volume_name: treats that identically to "not encrypted" — a transient Docker inspect failure silently skips the gocryptfs remount, and the rental continues on the container's ephemeral layer instead of the persistent volume. Please propagate inspect errors and reserve None for a confirmed absence of the /lium-cipher mount.

2) _volume_has_gocryptfs_conf still fails open on probe errors (workspace_mount.py)

Any non-zero docker run exit is read as "no gocryptfs.conf" — that covers a missing daturaai/aws-cli image or a Docker daemon error, not just a genuine test -f miss. detect_volume_access runs before pull_aws_cli, so a fresh executor can hit this on first use. Please distinguish probe-execution failure from a real test -f absence, and fail closed when it can't be determined.

Approving — narrower blast radius than the earlier rounds and the new check already covers the common "stopped container" case, but these should get closed before this rolls out broadly given what's at stake (silent data loss / garbage backups).

pixel29913 (approved):

fortunelucky777 (approved):

PR #1101, opened 2026-07-02
DAH-2272 - Replace flat port-check sleep with bounded time-budget poll

fortunelucky777:
## Describe your changes

Fixes the 60-120s "Started in subnet" stall (DAH-2272) caused by `wait_for_port_check_containers`'s flat `asyncio.sleep(retry_delay)` when a leftover validator DinD probe or backend `health_check_*` container lingers on the executor.

- Bounds the DinD probe's SSH connect/login (12s) and inner `docker run` (30s) so it can't hang (`dind_probe.py`).
- Replaces the flat-sleep retry loop with a `time.monotonic()`-deadline poll loop (~18s early-caller / ~20s late-caller budgets), converting `wait_for_port_check_containers` to a keyword-only `budget_sec`/`poll_interval_sec` signature (no back-compat shim — only 2 internal callers, forces every call site to be migrated rather than silently drifting).
- Adds structured-log telemetry (`container_age_sec`, fetched concurrently and timeout-bounded so a slow inspect can't stall the hot path, clamped against clock skew) so a future PR can data-gate a filter-narrowing decision.
- Gives the early caller its own `wait_ms` timing log, additive to its existing logs.
- **Preserved as-is:** the force-remove backstop at budget expiry still targets both `container_{hotkey}_*` and `health_check_*` — narrowing this was the original design's proposal but was overturned during review as unsafe. The validator-side cleanup for `health_check_*` runs in a different, unsynchronized pipeline, and the 90s port-retry backstop can only wait out a foreign `health_check_*`, never clear it.
- **Explicitly deferred, not part of this PR:** age-scoped early force-remove (gated on this PR's telemetry), an eager-force-remove+suppress-signal alternative (reviewed and rejected), and a follow-up spike on `PortConnectivityCheck` consulting Redis `renting_in_progress` state.

Process: deep-interview → ralplan consensus (Planner/Architect/Critic, 2 iterations) → autopilot implementation, followed by 3 parallel post-implementation reviews (functional/security/code-quality); findings from those reviews are fixed in this PR. Full design writeup + ADR: `.omc/plans/dah-2272-pr1-port-check-wait.md`.

**Testing:** full `neurons/validators` suite green (871 passed, 8 skipped; 6 pre-existing failures confirmed unrelated via stash-and-compare against `main` — missing `docker`/`paramiko` modules in the test venv and one stale golden snapshot). No staging deploy or Notion status update as part of this PR (scoped to code + tests + PR).

[DAH-2272 — Investigate Cache Templates Deployment Latency](https://app.notion.com/p/Investigate-Cache-Templates-Deployment-Latency-38ab8bfdbde98184b747e537fd323494) (see "Design plan — eliminating the 60s port-check wait" section)

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/clients/backend_client.py
+++ b/neurons/validators/src/clients/backend_client.py
@@ -244,6 +244,7 @@ class BackendClient:
         miner_hotkey: str,
         container_port: int,
         executor_id: str | None = None,
+        rental_in_progress: bool = False,
     ) -> ExecutorHealthCheckResponse | None:
         """Check executor health via backend API.
 
@@ -253,6 +254,9 @@ class BackendClient:
             miner_hotkey: Miner hotkey (SS58 address)
             container_port: Container port to check
             executor_id: Executor ID (optional)
+            rental_in_progress: True if this validator already sees an active customer rental for the
+                executor. Lets the backend skip the container-creating check immediately; the backend
+                still re-checks the DB itself when this is False.
 
         Returns:
             ExecutorHealthCheckResponse if successful, None otherwise
@@ -265,6 +269,7 @@ class BackendClient:
             "miner_port": miner_port,
             "miner_hotkey": miner_hotkey,
             "container_port": container_port,
+            "rental_in_progress": rental_in_progress,
         }
 
         if executor_id:
--- a/neurons/validators/src/core/docker_utils.py
+++ b/neurons/validators/src/core/docker_utils.py
@@ -1,3 +1,5 @@
+import shlex
+
 from services.const import DOCKER_DIND_IMAGE
 
 
@@ -75,7 +77,7 @@ class DockerCommand:
     def inspect_created_timestamp(container_id: str) -> str:
         """Build docker inspect command to get creation timestamp in seconds."""
         return (
-            f"/usr/bin/docker inspect {container_id} "
+            f"/usr/bin/docker inspect {shlex.quote(container_id)} "
             "--format '{{json .Created}}' | "
             "xargs -I {} date -d {} +%s"
         )
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -1139,25 +1139,47 @@ class DockerService:
                     # as before; do NOT narrow this to hotkey-only (see DAH-2272 ADR).
                     names = [n for n in container_names.strip().split("\n") if n]
 
-                    container_ages_sec: dict[str, int | None] = {}
-                    for name in names:
-                        try:
-                            created = await client.run(DockerCommand.inspect_created_timestamp(name))
+                    # Advisory telemetry only (feeds a future, human-reviewed
+                    # PR-2 filter-narrowing decision) — deliberately use the
+                    # validator's LOCAL clock instead of a second SSH
+                    # round-trip to the remote host's clock, to avoid doubling
+                    # the SSH cost per lingering container. Clock skew is an
+                    # accepted tradeoff for this field; the result is clamped
+                    # to 0 as a floor so skew can't produce a confusing
+                    # negative value.
+                    #
+                    # Fan the per-container `docker inspect` calls out
+                    # CONCURRENTLY (rather than serially) so this stays
+                    # bounded on the rental-creation hot path even when
+                    # several containers are lingering, and cap the whole
+                    # fan-out with an outer timeout so a hung remote inspect
+                    # can't stall the force-remove path indefinitely.
+                    container_ages_sec: dict[str, int | None] = dict.fromkeys(names)
+                    try:
+                        inspect_results = await asyncio.wait_for(
+                            asyncio.gather(
+                                *[
+                                    client.run(DockerCommand.inspect_created_timestamp(name))
+                                    for name in names
+                                ],
+                                return_exceptions=True,
+                            ),
+                            timeout=5,
+                        )
+                        for name, created in zip(names, inspect_results, strict=True):
+                            if isinstance(created, BaseException):
+                                container_ages_sec[name] = None
+                                continue
                             if created.exit_status == 0 and created.stdout.strip():
-                                # Advisory telemetry only (feeds a future,
-                                # human-reviewed PR-2 filter-narrowing decision) —
-                                # deliberately use the validator's LOCAL clock
-                                # instead of a second SSH round-trip to the
-                                # remote host's clock, to avoid doubling the
-                                # SSH cost per lingering container. Clock skew
-                                # is an accepted tradeoff for this field.
-                                container_ages_sec[name] = int(
-                                    time.time() - int(created.stdout.strip())
+                                container_ages_sec[name] = max(
+                                    0, int(time.time() - int(created.stdout.strip()))
                                 )
                             else:
                                 container_ages_sec[name] = None
-                        except Exception:
-                            container_ages_sec[name] = None
+                    except TimeoutError:
+                        # Outer cap fired — whichever containers didn't get a
+                        # response in time stay None rather than blocking.
+                        pass
 
                     logger.warning(
                         _m(
@@ -3261,7 +3283,7 @@ class DockerService:
                 )
 
                 # DAH-1524: the pre-run wait can block on live backend health_check_*
-                # probes (up to retry_delay); keep it out of the docker-run measurement.
+                # probes (up to budget_sec); keep it out of the docker-run measurement.
                 profilers.append(ProfilerStep.since(ProfilerStepName.PORT_CHECK_WAIT, prev_timestamp))
                 prev_timestamp = now_ms()
 
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -50,6 +50,7 @@ from payload_models.payloads import (
 from protocol.vc_protocol.compute_requests import RentedMachine
 
 from core.config import settings
+from core.docker_utils import DockerCommand
 from core.utils import _m, get_extra_info, retry_ssh_command
 from services.ssh_connect_timing import connect_with_phase_timing
 from services.const import (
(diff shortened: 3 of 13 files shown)


fortunelucky777:
## Race-condition validation for the probe/rental fix — methodology, parsing, and results

Beyond the unit tests in this PR, I validated the DAH-2272 fix under **controlled concurrent interleavings** between the two actors that actually contend here:

- **Validator verification** — `PortConnectivityCheck → ExecutorConnectivityService → ConnectivityOrchestrator → DindProbe → DindVerifier`, which creates the DinD/port-check probe `container_{hotkey}_{port}` on the executor.
- **Connector rental** — `create_container → wait_for_port_check_containers` (the force-remove this PR changes), plus `add_pending_pod` / `remove_pending_pod`.

The unit tests mock the connectivity boundary and set `renting_in_progress` statically, so they can't exercise the *cross-process ordering* that the tolerate branch depends on. The harness below does, deterministically. **The harness scripts are intentionally not committed** (throwaway analysis tooling); this comment is the record of the approach and results.

---

### Testing approach — cross-process `debugpy` orchestration

Two **real OS processes** (validator + connector) run under `debugpy`, each paused at a controlled breakpoint by a small orchestrator that speaks the Debug Adapter Protocol. The orchestrator confirms both are paused, then resumes them in a chosen order with a controlled delay — reproducing a specific interleaving on demand.

**Why no real backend / miner / executor / validator-neuron is needed.** This race is not a four-party interaction; it's a two-party contention over shared state, triggered by a request. Each real component maps cleanly:

| Component | Role in this race | In the harness |
|---|---|---|
| Backend | sends the create request; `get_all_rented_executors()` | synthetic payload + stub (the callback is only hit on the failure path, which the happy path never enters) |
| Miner neuron | pass-through relay | not a contention participant → omitted |
| Validator neuron | **runs the verification code under test** | the real chain runs in a focused process (no chain-sync / weights / websocket scaffolding) |
| Executor (Docker daemon) | **the shared resource** — container names + host ports | a fake daemon: a **Redis-backed container/port table** behind a fake `ssh_client` |
| Redis (`renting_in_progress`) | the flag the tolerate depends on | **real Redis** — the flag's timing/atomicity are themselves under test |

**Seam thesis:** production code reaches the outside world only through four seams — `ssh_client.run()`, module-level `asyncssh.connect()`, `redis_service`, and `backend_client`. We substitute *only* at those seams; everything between them runs unmodified.

**What runs as real code (not mocked):** `PortConnectivityCheck.run` (incl. the tolerate branch), `ExecutorConnectivityService.verify_ports`, `ConnectivityOrchestrator.verify` (the `sysbox_runtime=False`-on-probe-fail logic), `DindProbe`/`DindVerifier.verify`, `DockerService.wait_for_port_check_containers`, `RedisService.add_pending_pod / remove_pending_pod / renting_in_progress`, and the port-allocated retry policy (constants + phrase list imported from `docker_service`). Substituted: the `ssh_client`, the inner `asyncssh.connect`, the network-only `PortProbe` leaf, and `asyncio.sleep` (sped up).

**Why fake the executor rather than use a real one:** a real box can't be made to collide a real rental with a real probe's inner-SSH window on demand — determinism requires controllable shared state. The real executor's only contribution (container/port lifecycle over SSH) is exactly what the fake models.

**Credibility gate (run before any scenario):** the *real* `wait_for_port_check_containers` is pointed at the fake — it finds and force-removes seeded probes and returns the exact production messages (`"Port check containers forcefully removed"` / `"No port check containers found"`). This proves the fake↔real contract before drawing any conclusions.

**The breakpoints (1:1 with the interleaving):**
- validator pauses **right after** the DinD probe container is created (before the inner sysbox SSH);
- connector pauses **right before** the force-remove fires;
- orchestrator verifies both paused, resumes **connector first**, lets it finish its critical section, then resumes the validator.

---

### How results are parsed (verification levels → PASS/FAIL)

Each scenario is judged against **4 observability levels**, and the runner compares observed vs. an expected table to emit PASS/FAIL:

| Level | Source | What it establishes |
|---|---|---|
| **L1** debugpy `stopped` events | orchestrator/DAP | both processes truly paused at the intended seams |
| **L2** `renting_in_progress` flag | `pending_pods_prefix` hash on real Redis | whether the tolerate precondition holds at the moment of truth |
| **L3** executor container/port table | fake daemon | probe created → force-removed → rental created |
| **L4** real production output | `dind_probe.py` / `docker_service.py` logs + the returned `CheckResult` | the real code took the branch we claim |

L2 + L3 are snapshotted at two moments — **both-paused** and **pre-validator-resume** (the crux). The pass predicate then reads back the `CheckResult` fields the validator emits (`passed`, `probe_sysbox_runtime`, `final_state_sysbox_runtime`, `downgrade_tolerated`) plus the connector's flag/retry state. Because the interleaving is *forced* by the orchestrator (resume order + waiting on the connector's critical section) rather than left to timing, verdicts are deterministic — the full suite was run 3× with identical results.

---

### Scenarios & results

| # | Scenario | Verdict | What it establishes |
|---|---|---|---|
| **S1** | Baseline tolerate | ✅ PASS | probe force-removed mid-flight while flag set → downgrade **tolerated** → miner keeps `sysbox_runtime=True`. **The fix works in the common case.** |
| **S2** | Flag cleared early (G1) | ✅ PASS | connector clears the flag before the validator reads it → downgrade **recorded** → one-cycle sysbox mis-score. **Residual window.** |
| **S3** | Redis lost update (G4) | ✅ PASS | concurrent `add_pending_pod` (non-atomic RMW) → a pending-pod entry is **clobbered** → `renting_in_progress` wrongly false. Sequential control keeps both. **Residual window.** |
| **S4** | Port-alloc race (R4) | ✅ PASS | rental collides on the just-freed probe port → the real 90s same-command retry **recovers**. **Backstop works.** |

<details>
<summary>Captured evidence — S1 (baseline tolerate)</summary>

```
[orch] validator PAUSED at probe_created   /   connector PAUSED at about_to_cleanup   /   ✔ BOTH paused
[monitor @ both-paused]          renting_in_progress=True  | executor_containers={'container_..._9000': 9000}   ← probe up
[monitor @ pre-validator-resume] renting_in_progress=True  | executor_containers={'pod_pod-race-1': 9001}        ← probe GONE, flag STILL set

# real dind_probe.py: "DinD start" → "DinD container created" → "DinD check failed"
#   error = "no live DinD container on port 9000 (probe was removed)"   ← the downgrade is genuine
# real docker_service.py: port_check_force_removed ; wait_for_port_check_containers -> (True, 'Port check containers forcefully removed')

[validator] RESULT: {passed:true, probe_sysbox_runtime:false, final_state_sysbox_runtime:true, downgrade_tolerated:true}
```
</details>

<details>
<summary>Captured evidence — S2 (G1: flag cleared early → downgrade recorded)</summary>

```
[monitor @ pre-validator-resume] renting_in_progress=False | executor_containers={'pod_pod-race-1': 9001}   ← flag CLEARED
[connector] remove_pending_pod (rental completed) — flag cleared
[validator] RESULT: {passed:true, probe_sysbox_runtime:false, final_state_sysbox_runtime:false, downgrade_tolerated:null}
```
Same code as S1; the *only* variable changed is the flag state at the crux (True→False), which flips the outcome — confirming it's the real tolerate branch, not an artifact.
</details>

<details>
<summary>Captured evidence — S3 (G4: Redis lost update) + control</summary>

```
[orch] A paused after read, before write ; B paused after read, before write — both hold stale reads
[monitor @ both-paused] pending_pods=[]          ← both read empty
[monitor @ after A's write] pending_pods=['pod-A']
[orch] surviving pending pods = ['pod-B']         ← pod-A CLOBBERED by B's stale write
# sequential control: surviving = ['pod-A','pod-B']  ← both survive
```
</details>

<details>
<summary>Captured evidence — S4 (R4: port-alloc retry)</summary>

```
[monitor @ pre-validator-resume] executor_containers={'pod_pod-race-1': 9000}   ← rental lands on the reused probe port
[connector] rental port-allocated (attempt 1); sleeping 5s then retrying         ← real _PORT_ALLOCATED_* constants
[connector] rental container created after 1 retry(ies)
```
</details>

---

### Findings (calibrated — both transient/self-healing, not blockers)

**This PR is sound: S1 shows the tolerate works in the common case, and S4 shows the retry backstop holds.** Two residual windows are worth noting as potential follow-ups (neither is exercised by existing tests):

1. **G1 — the tolerate hinges on a cross-process flag-read winning a race.** The tolerate only fires while `renting_in_progress` is true, but the connector clears that flag at the end of `create_container` (`remove_pending_pod`, `docker_service.py:2988` / `compute_client.py:871`). If the rental completes before the validator's killed-probe result reaches the flag-read, the sysbox downgrade is recorded for that cycle. Narrow in practice (the validator's flag-read is sub-second after the probe dies, while the connector's flag-clear spans the whole `docker run` + bootstrap + finalize), and it self-heals on the next verification cycle — S2 forces the worst-case ordering only to prove the failure mode exists. *Possible hardening:* hold the flag until post-rental verification settles, or give the tolerate a small grace window for a just-completed rental.

2. **G4 — `add_pending_pod` / `remove_pending_pod` are a non-atomic read-modify-write** on a single hash field with no lock (`redis_service.py:228`). Concurrent pod creations on the same `(hotkey, executor)` can lose an entry, which feeds directly into G1 for that rental. `RedisService.acquire_executor_lock` already exists but isn't used here — a natural fix (or use atomic Redis ops).

---

### Scope / limitations

This proves the **logical/state** race with the real code paths. It does **not** exercise wire-level `docker rm -f` host-port release timing on a real dockerd — that would be a staging follow-up. The fake models only the docker behaviors the race touches (e.g. the DAH-2018 name-reservation-on-port-bind-fail nuance is not modeled, as it's irrelevant to the R4 retry-recovery assertion).

arhangel66 (approved):
Approving — the code is coherent and well-justified. One non-blocking fixup: the PR description doesn't match the code.

The body describes a deadline poll loop (budgets, `container_age_sec` telemetry, force-remove only as a backstop) and lists eager force-remove as an alternative that was "reviewed and rejected". The diff actually does immediate eager force-remove on first match — no poll, no budget — which is that "rejected" alternative. Docstrings, the `PortConnectivityCheck` tolerate, and tests are all built around eager-remove, so it reads as a stale description. Please update the PR body (and the DAH-2272 ADR).

Two small follow-ups (non-blocking):
- `wait_for_port_check_containers` ignores the `docker rm -f` result and always returns success — worth checking exit status/stderr (mitigated by the 90s retry).
- `health_check_*` is matched hotkey-agnostically and killed immediately — a test for a concurrent foreign `health_check_*` would pin down behavior on a shared host.

PR #1105, opened 2026-07-02
DAH-2265 - Default-image deploy: restore sshd skip, hand Jupyter to the image (JUPYTER_PASSWORD), skip docker login

fortunelucky777:
## Describe your changes

Two deploy-path optimizations for the **default docker image** rental: the validator should not redo work the default image (or the executor's cache) has already done. Both are inert for every other image.

### 1. Restore the default-image sshd skip, hand Jupyter to the image

The default Docker image / default cache template ships its own `start.sh`, which starts `sshd` unconditionally and can launch Jupyter itself. The `ships_sshd` flag on `ContainerCreateRequest` exists so the validator can skip the work the image already does — but commit `2345af85` removed the skip, so the validator's sshd bootstrap ran on **every** rental regardless of the flag (the flag became a no-op log line).

This restores the skip and extends it to Jupyter:

- **SSH** — when `ships_sshd` is truthy, skip `install_open_ssh_server_and_start_ssh_service` (and its `sshd_bootstrap.sh`); only `mkdir -p ~/.ssh` so public-key injection still works. The `else` branch keeps the full bootstrap for the default/`None`/`False` path.
- **Jupyter** — when `ships_sshd` **and** `enable_jupyter` are set, skip `run_jupyter` and instead forward a validator-chosen **`JUPYTER_PASSWORD`** as a `docker run -e` env var so the image's `start.sh` launches Jupyter itself. It must be set at `docker run` time — `start.sh` (the image `CMD`) reads the environment once at startup, so a post-create write to `/etc/environment` would be too late. The same token is reused to build the returned `jupyter_url`.

**Why `JUPYTER_PASSWORD` and not `ENABLE_JUPYTER`** — an earlier revision of this PR forwarded `ENABLE_JUPYTER=true` + `JUPYTER_TOKEN` + `JUPYTER_PORT`. **The images read none of those.** Every image in `computenet-docker-images` gates on exactly one variable:

```bash
start_jupyter() {
    if [[ $JUPYTER_PASSWORD ]]; then
        nohup jupyter lab ... --port=8888 --ServerApp.token=$JUPYTER_PASSWORD ... &> /jupyter.log &
```

`ENABLE_JUPYTER` appears nowhere in that repo, and the in-container port is hardcoded to 8888, so `JUPYTER_PORT` is ignored too. The gate therefore never fired: **Jupyter never started, while the port was still mapped and a Jupyter URL was still returned** — the pod came up `Running` and the URL reset the connection. Nothing on the deploy path checks that Jupyter actually came up, so the failure was invisible until a user clicked the link.

Caught by an end-to-end staging test across all three available nodes (H100 PCIe, RTX A4000, RTX A6000). On every one: `ENABLE_JUPYTER=true` present in PID 1's environment, `jupyter` installed and working (4.6.1), **no jupyter process, nothing listening on 8888** (only `sshd` on 22), and no `/jupyter.log` — i.e. `start_jupyter()` was a silent no-op. SSH succeeded on all three.

`JUPYTER_PASSWORD` is also the name the validator's own `run_jupyter.sh` takes (`--password=`), so both paths now agree on one contract. The image-managed path is additionally guarded on the mapped docker port being 8888 — `start.sh` always binds 8888, so if the mapping ever moved off it we fall back to `run_jupyter` rather than silently serving nothing.

The default path (`ships_sshd` None/False) is byte-identical to before: full bootstrap + `run_jupyter`, no env forwarding. The flag is still inert in production until `lium-io-backend` starts sending `ships_sshd=true`.

Prod impact once it does: `SSH service installation` is the single biggest step on default-image deploys — **11.8s median, 36.8s p90** (14 days of `rental_history.profilers`).

### 2. Skip the docker login for default images, even when credentials are passed

The default cache-template images are **public** Docker Hub refs, and the pull itself is skipped 99.1% of the time because the image is already cached on the executor — so a registry login gates nothing on that path. But the backend attaches credentials whenever the renter has *any* saved: the default templates carry no `docker_credential_id`, and `get_docker_credentials(user_id, None)` drops the id filter and returns the user's first credential row. Over the last 14 days:

| default-image deploys | did a real login | median | p90 | max |
|---|---|---|---|---|
| 1627 | 357 (22%) | 1195 ms | 1667 ms | 8230 ms |

Those 357 correlate 1:1 with "the renter has a saved docker credential" — 357/357. That is ~382s of pure wait per fortnight, spent authenticating in order to pull an image that needs no auth and usually is not pulled at all.

The validator now decides this itself — no backend change, no protocol change:

- **Skip the login** when the rental is for a default image, even when `docker_username`/`docker_password` are present.
- **Mark the profiler step `skipped`** when no login ran. Previously the step was appended unconditionally, so a no-credential deploy emitted a 0 ms `Docker login step finished` that reads as real work. Same convention `DOCKER_PULL` already uses; the step name is unchanged (it is a Loki/backend query key).
- **`ships_sshd` *is* the "renter selected a default image" signal** — the validator does not re-derive it. The backend already made this decision: `ships_sshd=is_cached`, where `is_cached = is_cached_template_image(...)` resolves the recommended image for *that executor's* GPU + driver, and is forced to `False` for custom builds (whose `FROM` may pull a private base image, so they keep the login). An earlier revision of this PR added a validator-side `is_default_docker_image()` predicate over the flat shared-config list; that duplicated the backend's decision from a *different source* and could disagree with it — skipping a login that was actually needed. It's gone.

Legacy paths are unaffected: retry/filler rentals omit `ships_sshd` (`None`), which is not a default-image signal, so they still log in exactly as today.

**Files**
- `neurons/validators/src/services/docker_service.py` — the sshd/Jupyter skip + `JUPYTER_PASSWORD` forwarding, and the docker-login skip, in `create_container`.
- `neurons/validators/src/payload_models/payloads.py` — updated `ships_sshd` field docs.
- `neurons/validators/tests/test_deploy_optimizations.py` — tests for both skips plus default/legacy-path regression guards.

**Verified on staging** (all three available nodes — H100 PCIe, RTX A4000, RTX A6000): rented with the default template + Jupyter + an SSH key, all reached `Running`, SSH succeeded on each, and Jupyter now returns **HTTP 200** on `/lab` and `/api/status` where it previously reset the connection. Inside each container: `JUPYTER_PASSWORD` is the only Jupyter variable in PID 1's env, `jupyter-lab` is running, 8888 is listening, and `/jupyter.log` exists — i.e. the image's `start.sh` gate actually fired.

**Tests**: full validator suite green — **1072 passed, 8 skipped**. New coverage: `JUPYTER_PASSWORD` forwarded (and the returned URL carrying the same token) on the `ships_sshd` + `enable_jupyter` path; the default path still runs `run_jupyter` and forwards no `JUPYTER_PASSWORD`; login skipped when `ships_sshd` with credentials present; a non-default image with credentials still logs in; `ships_sshd=None` (retry/filler) still logs in; profiler step flagged `skipped` when no credentials are sent; a custom build still logs in.

DAH-2265

- [x] I wrote tests.
- [ ] Need to take care of performance? Net negative cost. On the `ships_sshd` path it removes a redundant in-container sshd install/bootstrap and the `run_jupyter.sh` copy+exec; on the default-image path it removes a Docker Hub auth round-trip (~1.1s median) from 22% of deploys.

### DIFF
--- a/neurons/validators/src/payload_models/payloads.py
+++ b/neurons/validators/src/payload_models/payloads.py
@@ -287,9 +287,10 @@ class ContainerCreateRequest(ContainerBaseRequest):
     #       (d) installs a 30s self-heal restart watchdog (sshd_bootstrap.sh:119-156)
     #     The validator then only re-ensures ~/.ssh for key injection.
     #   * Jupyter: skip the validator's run_jupyter path. Instead, when `enable_jupyter`
-    #     is also set, the validator forwards ENABLE_JUPYTER (+ a validator-chosen
-    #     JUPYTER_TOKEN and JUPYTER_PORT) as `docker run -e` env vars so the image's
-    #     start.sh launches Jupyter itself.
+    #     is also set, the validator forwards a validator-chosen JUPYTER_PASSWORD as a
+    #     `docker run -e` env var so the image's start.sh launches Jupyter itself and
+    #     serves it as that token. JUPYTER_PASSWORD is the only Jupyter variable these
+    #     images read (start.sh gates on it and hardcodes the in-container port 8888).
     # None/False (default) preserves the always-bootstrap + run_jupyter behavior.
     # Populated by lium-io-backend in a follow-up; ships inert here (no producer sets it yet).
     ships_sshd: bool | None = None
--- a/neurons/validators/src/payload_models/payloads.py
+++ b/neurons/validators/src/payload_models/payloads.py
@@ -272,18 +272,26 @@ class ContainerCreateRequest(ContainerBaseRequest):
     # builds the image from this Dockerfile on the executor host instead of pulling
     # `docker_image`. None or "" keeps the normal image-pull path.
     dockerfile_content: str | None = None
-    # DAH-1524 introduced this flag ("skip the bootstrap for cached templates");
-    # its semantics have since changed TWICE — do not trust older comments:
-    #   - 2345af85 reverted the skip: the sshd bootstrap ALWAYS runs, because a
-    #     renter's custom startup_commands replaces the image CMD, so the image
-    #     alone can never be trusted to bring sshd up.
-    #   - DAH-2341 made the bootstrap converge with a self-starting image
-    #     instead of racing it (grace-wait, adopt a running sshd, no keygen
-    #     when adopted; see sshd_bootstrap.sh header).
-    # Today ships_sshd=True (set by lium-io-backend for cached recommended
-    # images) means exactly one thing: a bootstrap failure is FATAL to the
-    # deploy (RuntimeError -> FailedContainerRequest). None/False keeps
-    # bootstrap failures lenient (logged, deploy continues).
+    # DAH-1524 / DAH-2265 (default-image / cached-template deploy): set truthy by the
+    # backend when the user selects the default Docker image or default cache template.
+    # Such images run their own start.sh, so the validator skips work the image already
+    # does:
+    #   * SSH: skip the ENTIRE post-creation sshd bootstrap
+    #     (install_open_ssh_server_and_start_ssh_service -> sshd_bootstrap.sh), which
+    #     otherwise provides FOUR guarantees, ALL surrendered when this flag is set —
+    #     the image ASSERTS it provides them itself:
+    #       (a) starts the sshd daemon                 (sshd_bootstrap.sh:110-117, main 166)
+    #       (b) generates host keys via `ssh-keygen -A` (sshd_bootstrap.sh:106)
+    #       (c) hardens config: PasswordAuthentication/KbdInteractive/ChallengeResponse no
+    #                                                  (sshd_bootstrap.sh:89-99)
+    #       (d) installs a 30s self-heal restart watchdog (sshd_bootstrap.sh:119-156)
+    #     The validator then only re-ensures ~/.ssh for key injection.
+    #   * Jupyter: skip the validator's run_jupyter path. Instead, when `enable_jupyter`
+    #     is also set, the validator forwards ENABLE_JUPYTER (+ a validator-chosen
+    #     JUPYTER_TOKEN and JUPYTER_PORT) as `docker run -e` env vars so the image's
+    #     start.sh launches Jupyter itself.
+    # None/False (default) preserves the always-bootstrap + run_jupyter behavior.
+    # Populated by lium-io-backend in a follow-up; ships inert here (no producer sets it yet).
     ships_sshd: bool | None = None
     # DAH-2356: per-GPU power caps for the Lium PEARL default-job (FILLER) container. Applied
     # host-side via `nvidia-smi -pl` (clamped to the GPU's live min/max) before the filler starts;
--- a/neurons/validators/src/payload_models/payloads.py
+++ b/neurons/validators/src/payload_models/payloads.py
@@ -291,8 +291,16 @@ class ContainerCreateRequest(ContainerBaseRequest):
     #     `docker run -e` env var so the image's start.sh launches Jupyter itself and
     #     serves it as that token. JUPYTER_PASSWORD is the only Jupyter variable these
     #     images read (start.sh gates on it and hardcodes the in-container port 8888).
+    # CAVEAT — this only holds when the image's own start.sh actually runs. start.sh is
+    # the image CMD, exec'd by the ENTRYPOINT (/pytorch-entrypoint.sh). A renter-supplied
+    # `startup_commands` is passed as the container command and REPLACES that CMD, and a
+    # renter `entrypoint` replaces the ENTRYPOINT — either way start.sh never runs and the
+    # image starts no sshd/Jupyter. The validator therefore only takes the skip path when
+    # BOTH are empty (see docker_service.create_container `image_manages_services`); with
+    # an override present it falls back to the full bootstrap + run_jupyter.
     # None/False (default) preserves the always-bootstrap + run_jupyter behavior.
-    # Populated by lium-io-backend in a follow-up; ships inert here (no producer sets it yet).
+    # Set by lium-io-backend as `ships_sshd=is_cached` (DAH-1524) for default/cached-template
+    # images, independent of startup_commands — so the skip path is reachable in production.
     ships_sshd: bool | None = None
     # DAH-2356: per-GPU power caps for the Lium PEARL default-job (FILLER) container. Applied
     # host-side via `nvidia-smi -pl` (clamped to the GPU's live min/max) before the filler starts;
--- a/neurons/validators/src/services/default_docker_image_digest_service.py
+++ b/neurons/validators/src/services/default_docker_image_digest_service.py
@@ -51,32 +51,6 @@ def _shared_config_image_refs() -> tuple[str, ...]:
     return tuple(refs)
 
 
-def is_default_docker_image(image_ref: str | None) -> bool:
-    """True when ``image_ref`` is one of the default cache-template images.
-
-    These images are public on Docker Hub and pre-cached on the executors, so a
-    deploy that uses one needs no registry credentials — see the login skip in
-    ``DockerService.create_container``. Matching is exact against the shared-config
-    ``repo:tag`` refs, which is what the backend sends for a default-image rental.
-
-    Fail-closed: this runs on the rental deploy path, so any failure to read the
-    shared config returns False — the deploy then logs in exactly as it does today,
-    rather than a rental failing on an unreadable optimization hint.
-    """
-    if not image_ref:
-        return False
-    try:
-        return image_ref in _shared_config_image_refs()
-    except Exception:
-        logger.warning(
-            "Could not read default docker images from shared config; "
-            "treating %s as a non-default image",
-            image_ref,
-            exc_info=True,
-        )
-        return False
-
-
 async def fetch_registry_digest(session: aiohttp.ClientSession, image_ref: str) -> str | None:
     """Return the registry manifest digest for ``repository:tag``, or None on any error."""
     repository, _, tag = image_ref.partition(":")
--- a/neurons/validators/src/services/default_docker_image_digest_service.py
+++ b/neurons/validators/src/services/default_docker_image_digest_service.py
@@ -51,6 +51,32 @@ def _shared_config_image_refs() -> tuple[str, ...]:
     return tuple(refs)
 
 
+def is_default_docker_image(image_ref: str | None) -> bool:
+    """True when ``image_ref`` is one of the default cache-template images.
+
+    These images are public on Docker Hub and pre-cached on the executors, so a
+    deploy that uses one needs no registry credentials — see the login skip in
+    ``DockerService.create_container``. Matching is exact against the shared-config
+    ``repo:tag`` refs, which is what the backend sends for a default-image rental.
+
+    Fail-closed: this runs on the rental deploy path, so any failure to read the
+    shared config returns False — the deploy then logs in exactly as it does today,
+    rather than a rental failing on an unreadable optimization hint.
+    """
+    if not image_ref:
+        return False
+    try:
+        return image_ref in _shared_config_image_refs()
+    except Exception:
+        logger.warning(
+            "Could not read default docker images from shared config; "
+            "treating %s as a non-default image",
+            image_ref,
+            exc_info=True,
+        )
+        return False
+
+
 async def fetch_registry_digest(session: aiohttp.ClientSession, image_ref: str) -> str | None:
     """Return the registry manifest digest for ``repository:tag``, or None on any error."""
     repository, _, tag = image_ref.partition(":")
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -58,6 +58,7 @@ from services.const import (
     POD_CONTAINER_PREFIX,
     PREFERRED_POD_PORTS,
 )
+from services.default_docker_image_digest_service import is_default_docker_image
 from services.gpu_power_limit import (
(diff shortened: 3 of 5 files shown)


fortunelucky777:
## 🚫 DO NOT DEPLOY — blocked

This change makes the validator **stop bootstrapping sshd (and stop running `run_jupyter`) for `ships_sshd` rentals** and instead rely on the image's own `start.sh` to start sshd unconditionally and to launch Jupyter from `ENABLE_JUPYTER`. That is only safe once the image actually guarantees those behaviors on every node.

**Blocked until both are true:**

- [ ] The new pytorch Docker image is verified safe — `start.sh` starts sshd unconditionally and gates Jupyter on `ENABLE_JUPYTER` (confirmed in `computenet-docker-images`).
- [ ] All nodes are running **our** pytorch image, pinned to our published digest — i.e. every executor pulls/uses the verified digest, not a stale or arbitrary image.

**Why:** if this ships before the fleet is on the verified digest, any node whose default image does *not* start sshd itself would lose SSH access, because the validator no longer installs/starts sshd on the `ships_sshd` path. The flag is also still inert until `lium-io-backend` starts sending `ships_sshd=true`, so there is no rush to merge/deploy ahead of the image rollout.

Keeping as **draft** until the two boxes above are checked.

fortunelucky777:
## Production evidence — this is exactly the cost this PR removes (and it's already active in prod)

Investigated two slow customer rentals on **2026-07-09** (provider `5Eh4hsGH…`, image `daturaai/pytorch:2.12.0-py3.12-cuda13.0.2-devel-ubuntu24.04-dind`). Both were dominated by the `SSH service installation` profiler step:

| Rental (`rental_history.id`) | pod_id | node | everything before SSH-install | **SSH service installation step** | total deploy |
|---|---|---|---|---|---|
| `4af70cfa` | `d719d093` | `8cb9ffb1` (91.224.44.222:8015) | ~14.0s | **82.836s** | 96.5s |
| `e2c917b0` | `89a02692` | `d248f914` (91.224.44.85:8006) | ~10.8s | **65.728s** | 77.0s |

That single step is >85% of each deploy. Numbers agree exactly between `rental_history.profilers` and the connector `Deployment profile summary` log.

### The step is misnamed — most of it is Jupyter, not sshd

The `SSH service installation` profiler ([`docker_service.py` ~L3377](../blob/main/neurons/validators/src/services/docker_service.py)) bundles *add-keys → sshd bootstrap → run_jupyter*. Isolating the connector `miner_driver` timeline for pod `d719d093`:

| Sub-op | Window (UTC) | Duration | What it is |
|---|---|---|---|
| Key injection | 08:38:11 → 08:38:14 | ~3s | cheap |
| **sshd bootstrap** | 08:38:14 → 08:38:29 | **~15s** | `sshd_bootstrap.sh` — mostly the fixed ~10s grace-wait for the image's own sshd + `ssh-keygen -A` + verify budget |
| **`run_jupyter.sh`** | 08:38:32 → 08:39:34 | **~62s** | rebuilds the whole Jupyter stack from scratch |

Rental 2 is the same shape: bootstrap 08:38:01→08:38:15 (~14s), then `Running jupyter from volume` 08:38:17 → summary 08:39:04 = **~47s** on Jupyter.

### What `run_jupyter.sh` does on every deploy (the ~62s / ~47s)

`neurons/executor/run_jupyter.sh` unconditionally, per rental, with nothing cached:
1. `python3 -m venv /opt/jupyter-env` (fresh venv)
2. `pip install --upgrade pip` (uninstalls 24.0, installs 26.1.2)
3. `pip install jupyter jupyterlab==4.5.7 jupyter-server==2.18.0 notebook==7.5.6 ipykernel==7.2.0 matplotlib numpy pandas requests`

Literal proof from `pod_status.logs` (326 install-output lines for rental 1, 163 for rental 2):
```
Creating virtual environment at /opt/jupyter-env...
Upgrading pip... → Successfully installed pip-26.1.2
Installing Jupyter and Python packages...
Collecting jupyter / jupyterlab==4.5.7 / jupyter-server==2.18.0 / notebook==7.5.6 / ipykernel==7.2.0 …
```

### Why this PR is the right fix — and is recovering time *today*, not latently

For both deploys the connector logged **`Running SSH bootstrap for template that ships sshd`**, which is emitted only inside `if payload.ships_sshd:`. So **`ships_sshd` was already truthy in production on 2026-07-09**, yet the current (pre-PR) code still ran the full sshd bootstrap *and* `run_jupyter` — the exact no-op-flag regression from `2345af85` this PR reverses.

Net effect of this PR on these rentals:
- **SSH** skip removes the ~15s / ~14s redundant bootstrap (the image's `start.sh` already starts sshd).
- **Jupyter** skip + `ENABLE_JUPYTER=true` env-forwarding removes the ~62s / ~47s `run_jupyter.sh` pip build (the image's `start.sh` launches Jupyter itself).

Expected outcome: these deploys drop from **~1.6 min / ~1.3 min to ~15–25s**.

One note for the PR description: the "flag is still inert in production until lium-io-backend starts sending `ships_sshd=true`" line looks out of date — the backend was already sending `ships_sshd=true` here (the connector took the ships-sshd branch), so this regression is actively costing ~60–80s per rental in prod right now. Worth landing sooner rather than later.

fortunelucky777:
## E2E re-test on staging — single node, post-deploy ✅

Re-ran the end-to-end rental on **staging** against the currently deployed validator build (the earlier 3 pods were deployed by the pre-fix build, so they were torn down first and this pod was rented fresh). One node, default template, Jupyter enabled, `waris` SSH key.

**Result: all checks pass.** Jupyter now actually starts — the failure this PR fixes is gone.

### Node under test

| | |
|---|---|
| Pod | `Kind Ocean Thunder` — `aaee27f8-4614-4aca-95d4-d7702d9545da` |
| Executor | `0d2f294b-3a7d-474d-b97d-d4228e7c26a8` (1× **NVIDIA H100 PCIe**, Montreal) |
| Template | `daturaai/pytorch:2.12.0-py3.12-cuda13.0.2-devel-ubuntu24.04-dind` (default cache template, `Fast deploy` + `DinD`) |
| Endpoints | SSH `185.216.21.144:40009` · Jupyter `185.216.21.144:40008` |

### Results

| Check | Result |
|---|---|
| Rental created | ✅ |
| Reached `RUNNING` | ✅ (~1 min) |
| SSH with `waris` key | ✅ `SSH_OK host=deb23cd8e507`, `nvidia-smi` → `NVIDIA H100 PCIe` |
| Jupyter reachable | ✅ `GET /lab?token=<valid>` → **HTTP 200**, JupyterLab Launcher renders in-browser |
| Jupyter auth enforced | ✅ no token → **302** to login; wrong token → **302**; `/api/status` unauthenticated → **403**, with token → **200** |
| Errors | none |

### The three code paths this PR touches, verified in the live container

**1. Jupyter is image-managed via `JUPYTER_PASSWORD`** — PID 1's environment contains **only** `JUPYTER_PASSWORD`:

```
JUPYTER_PASSWORD: PRESENT
ENABLE_JUPYTER:   absent
JUPYTER_TOKEN:    absent
JUPYTER_PORT:     absent
```

The image's `start.sh` gate fired: `jupyter-lab` is running (pid 612), bound on `0.0.0.0:8888`, and `/jupyter.log` exists. The token the validator generated is the one Jupyter is actually serving — the running process carries `--ServerApp.token=b45ef52c…`, byte-identical to the `token=` in the portal's Jupyter URL. That end-to-end token match is what was broken before: the URL was returned but nothing was listening.

```
root 612 /usr/local/bin/jupyter-lab --allow-root --no-browser --port=8888 --ip=* \
  --ServerApp.token=b45ef52c… --ServerApp.preferred_dir=/workspace
LISTEN 0 128 0.0.0.0:8888  users:(("jupyter-lab",pid=612,fd=6))
-rw-r--r-- 1 root root 3472 Jul 14 18:44 /jupyter.log
```

**2. sshd bootstrap skipped, image's sshd used** — the validator's `sshd_bootstrap.sh` is **absent** from the container, and `sshd` (pid 586) has **PPID 1 = `/bin/bash /start.sh`**, i.e. it was started by the image, not by the validator. Public-key injection still works: `authorized_keys` has the one `waris` key, and SSH auth succeeded with it.

**3. Docker login** — no `Docker login` step appears in the validator logs for this deploy window, consistent with the `ships_sshd` skip. (Caveat: this is log-absence, not a positive assertion — the authoritative check is the `rental_history.profilers` step, which I couldn't reach from here as the staging DB tunnel wasn't up. Worth a spot-check on the profiler before merge.)

### Notes

- Teardown → rentable is **not** instant: an executor only re-lists after the validator's next block-cadence job re-verifies it (~15 min here, `remaining_blocks` counted 71 → 0). Not a regression, just worth knowing when re-testing.
- Private key contents were never displayed or copied; only the key path was used.
- The test pod is **still running and billing ($1.20/hr)** — say the word and I'll delete it.

<sub>Verified against the deployed staging validator on 2026-07-14 ~18:44 UTC.</sub>

arhangel66 (changes requested):
Thanks for the solid work on the Docker login / SSH / Jupyter optimizations — the reasoning is clear and the test coverage is nice. I have one blocking concern before we merge.

**Non-empty `startup_commands` break the `ships_sshd` fast path** — the container comes up with no working SSH and a dead Jupyter URL.

Here's the breakdown:
- A non-empty `startup_commands` is passed as `ContainerRunSpec.command`, which replaces the image's `CMD`.
- In our default cache-template images, `start.sh` (the script that runs `service ssh start` and launches Jupyter) lives in `CMD`, not `ENTRYPOINT`. The `pytorch-entrypoint.sh` in `ENTRYPOINT` only starts dockerd and then runs `exec "$@"`. The image Dockerfiles actually document this on purpose — the entrypoint was split out "because lium's validator appends the renter's startup command, which replaces CMD."
- This PR now unconditionally skips both the validator's `sshd` bootstrap and `run_jupyter` when `ships_sshd` is true.
- **Result:** when `ships_sshd=True` and `startup_commands` is set, `start.sh` never runs. The container reaches Running with ports mapped and a `jupyter_url` returned, but `sshd` is down and the Jupyter URL resets the connection.

Two notes on impact:
- This is reachable in production today, not just in a follow-up: the backend already sets `ships_sshd=is_cached` in `executor.py` (`DAH-1524`, on master), independent of `startup_commands`. The new comment's "no producer sets it yet" looks out of date.
- We've been here before — `commit 2345af85` reverted this exact skip for the same reason ("a renter's custom startup_commands replaces the image CMD").

**Suggested direction:** only take the new image-managed path when `startup_commands` is empty; when it's present, fall back to the existing validator-managed path (bootstrap + `run_jupyter`). A regression test for `ships_sshd=True` + non-empty `startup_commands` covering both SSH and Jupyter would round it out.

Happy to pair on this if it helps. Thanks again for the solid work!

fortunelucky777:
@arhangel66 you're right, and thank you — this was a real, prod-reachable break. Fixed in `3120ee43`.

I verified every link in your chain at source before touching anything:

- **`command` replaces `CMD`** — `rental_docker_sdk._run_container_sync` calls `create_container(command=list(spec.command) or None, ...)`, and `spec.command = build_container_command_argv(custom_options.startup_commands)`. Non-empty `startup_commands` ⇒ the container command is overridden.
- **`start.sh` is `CMD`, not `ENTRYPOINT`** — `templates/pytorch/Dockerfile`:
  ```dockerfile
  # Start dockerd from ENTRYPOINT so the nested daemon survives a CMD override
  # (lium's validator appends the renter's startup command, which replaces CMD).
  ENTRYPOINT [ "/pytorch-entrypoint.sh" ]
  CMD [ "/start.sh" ]
  ```
  `pytorch-entrypoint.sh` starts dockerd then `exec "$@"` — so with a `startup_commands` override it execs *that*, never `/start.sh`. And `/start.sh` is the only thing that runs `service ssh start` (line 48) and gates Jupyter on `JUPYTER_PASSWORD`.

So `ships_sshd=True` + non-empty `startup_commands` ⇒ no sshd, dead Jupyter, exactly as you described — and `2345af85` reverted this same skip for the same reason.

**One thing I extended beyond your minimum:** a renter **`entrypoint`** override is the same failure mode — it replaces `/pytorch-entrypoint.sh`, which is what execs the CMD, so `start.sh` won't run even with the default CMD. So rather than gate only on empty `startup_commands`, the skip now requires **both** to be absent:

```python
image_manages_services = bool(
    payload.ships_sshd
    and not (custom_options.startup_commands and custom_options.startup_commands.strip())
    and not (custom_options.entrypoint and custom_options.entrypoint.strip())
)
```

Both the sshd-bootstrap skip and the image-managed-Jupyter path key off `image_manages_services`; when either override is present, we fall back to the full validator-managed `bootstrap + run_jupyter`. The docker-login skip is untouched on purpose — it depends only on the image being a public default, not on which CMD ends up running.

**Tests** (as you suggested — SSH *and* Jupyter):
- `test_ships_sshd_with_startup_commands_runs_bootstrap_and_run_jupyter` — `ships_sshd=True` + `enable_jupyter=True` + `startup_commands`, port mapped to **8888** so the *only* thing forcing the fallback is the override (not the port guard): asserts both `install_open_ssh…` and `run_jupyter` run and no `JUPYTER_PASSWORD` is forwarded.
- `test_ships_sshd_with_entrypoint_override_runs_bootstrap` — same via the entrypoint override.

Full validator suite: **1074 passed, 8 skipped** (was 1072; +2). Also corrected the stale "no producer sets `ships_sshd` yet" line in `payloads.py` you flagged — it's set as `ships_sshd=is_cached` (DAH-1524) today, independent of `startup_commands`.

fortunelucky777:
@arhangel66 , Ty for your feedback and I covered it with latest commits.

arhangel66 (approved):
approved

PR #1107, opened 2026-07-03
DAH-2341 - Make rental SSH bootstrap race-safe against self-starting images

fortunelucky777:
## Describe your changes

Fixes the race between a rental image's own SSH setup (entrypoint `/start.sh`: host-key generation + `service ssh start`) and the validator's post-create `sshd_bootstrap.sh`. In the staging repro (pod `93b13164-863a-4f2d-8e27-c0f3b32c98a2`, custom `fortunelucky777/pytorch:...-dind` on the L40), the validator's `ssh-keygen -A` created a host key between the image's `[ ! -f ]` check and its per-type `ssh-keygen -t rsa -f`, which then blocked PID 1 on `Overwrite (y/n)?` — pod RUNNING, ports published, SSH connection refused.

The bootstrap now **converges on an invariant** ("host keys exist and sshd is listening on :22") instead of racing to create it:

- **Adopt, don't fight**: if sshd is already running, adopt it — no keygen at all. If the image ships an sshd binary, wait a grace period (default 10s) for the image to start it before the fallback takes over. Images without an sshd binary skip the wait entirely (no latency change for them).
- **Fallback under a lock**: own bring-up (keygen -A, hardening, start) runs under `mkdir /run/lium-ssh-setup.lock`, shared with the patched lium image `start.sh` (companion PR). A failed sshd start is re-checked — losing a benign bind race is success, not failure.
- **Exit code = invariant**: the script verifies sshd is running AND a `:22` listener exists (`/proc/net/tcp{,6}`) before exiting 0; command exit codes no longer produce false-fatal `RuntimeError("SSH bootstrap failed")` for `ships_sshd` templates.
- **Hardening determinism**: when config is hardened after an adopted sshd was already up, the master gets SIGHUP via its pidfile so `PasswordAuthentication no` actually loads. Read-only `/etc/ssh` no longer aborts the adopt path.
- **Keys before bootstrap** (`docker_service.py`): the customer's public keys are plain data (mkdir + append, now with `chmod 700`) and are injected *before* the bootstrap, so whichever sshd comes up first already finds them.
- Hostile-image hygiene: `docker exec` inherits image ENV, so the `LIUM_*` timing knobs are integer-validated and capped (grace ≤300s, verify ≤120s, lock ≤300s).
- Updated the stale `ships_sshd` docstring in `payloads.py` (it still described the reverted DAH-1524 "skip" semantics); today the flag only escalates bootstrap failure to fatal.

**Latency note**: images that ship an sshd binary but never self-start it (including today's default `daturaai/pytorch` tags, whose `start.sh` gates SSH on `PUBLIC_KEY`) pay up to +10s in `SSH_SERVICE_INSTALLATION` (grace window). This self-resolves for the rebuilt self-starting images (companion PR): the wait exits the moment sshd appears. Follow-ups deliberately out of scope: backend declaring per-tag sshd behavior, and a host-side reachability probe of the mapped port before `ContainerCreated` (the in-container `:22` listener check covers this task's failure class; port publication was already correct in the repro).

Companion PR: Datura-ai/computenet-docker-images — race-safe `setup_ssh` (ssh-keygen -A, shared lock, never kill PID 1) + stale nested-dockerd pidfile cleanup that caused the post-crash restart loop.

**Test plan** (from the task spec):
- [x] Unit tests for the new decision flow — `tests/test_sshd_bootstrap_script.py` runs the real script through shims: adopt-running (no keygen), adopt-during-grace, fallback bring-up, no-wait-without-binary, grace-0, bind-race tolerance, exit-1 only when sshd never serves, listener-required, lock-timeout. 10/10 passed.
- [x] Full validator suite: 892 passed, 8 skipped (includes updated key-before-bootstrap ordering test).
- [x] Forced-concurrency smoke (image `setup_ssh` vs bootstrap, 30 rounds, `--grace 0` worst case): 0 keygen collisions, both sides exit 0, single sshd, 3 host keys — every round.
- [ ] Post-deploy (staging): re-run the failed custom-image case on L40 (`fortunelucky777/pytorch:2.12.0-py3.12-cuda12.8-devel-ubuntu24.04-dind`) → RUNNING + direct SSH OK.
- [ ] Post-deploy (staging): default-template smoke across A6000 38.80.122.73 / L40 38.80.122.143 / A4000 149.36.1.151.

[DAH-2341 — Fix rental SSH bootstrap race with images that start sshd](https://app.notion.com/p/Fix-rental-SSH-bootstrap-race-with-images-that-start-sshd-392b8bfdbde98176b12ae7833ddf2d70)

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/payload_models/payloads.py
+++ b/neurons/validators/src/payload_models/payloads.py
@@ -266,19 +266,18 @@ class ContainerCreateRequest(ContainerBaseRequest):
     # builds the image from this Dockerfile on the executor host instead of pulling
     # `docker_image`. None or "" keeps the normal image-pull path.
     dockerfile_content: str | None = None
-    # DAH-1524 (cached-template deploy): when truthy, the validator skips the
-    # ENTIRE post-creation sshd bootstrap (install_open_ssh_server_and_start_ssh_service
-    # -> sshd_bootstrap.sh). That bootstrap normally provides FOUR guarantees, ALL of
-    # which are surrendered when this flag is set:
-    #   (a) starts the sshd daemon                 (sshd_bootstrap.sh:110-117, main 166)
-    #   (b) generates host keys via `ssh-keygen -A` (sshd_bootstrap.sh:106)
-    #   (c) hardens config: PasswordAuthentication/KbdInteractive/ChallengeResponse no
-    #                                              (sshd_bootstrap.sh:89-99)
-    #   (d) installs a 30s self-heal restart watchdog (sshd_bootstrap.sh:119-156)
-    # Setting ships_sshd=True ASSERTS the image itself provides all four; the validator
-    # then only re-ensures ~/.ssh for key injection. None/False (default) preserves
-    # today's always-bootstrap behavior. Populated by lium-io-backend in a follow-up;
-    # ships inert here (no producer sets it yet).
+    # DAH-1524 introduced this flag ("skip the bootstrap for cached templates");
+    # its semantics have since changed TWICE — do not trust older comments:
+    #   - 2345af85 reverted the skip: the sshd bootstrap ALWAYS runs, because a
+    #     renter's custom startup_commands replaces the image CMD, so the image
+    #     alone can never be trusted to bring sshd up.
+    #   - DAH-2341 made the bootstrap converge with a self-starting image
+    #     instead of racing it (grace-wait, adopt a running sshd, no keygen
+    #     when adopted; see sshd_bootstrap.sh header).
+    # Today ships_sshd=True (set by lium-io-backend for cached recommended
+    # images) means exactly one thing: a bootstrap failure is FATAL to the
+    # deploy (RuntimeError -> FailedContainerRequest). None/False keeps
+    # bootstrap failures lenient (logged, deploy continues).
     ships_sshd: bool | None = None
 
 
--- a/neurons/validators/src/services/assets/sshd_bootstrap.sh
+++ b/neurons/validators/src/services/assets/sshd_bootstrap.sh
@@ -424,6 +424,12 @@ prepare_sshd_runtime
 if ! start_sshd_if_needed; then
     log "Fallback sshd start failed"
 fi
+# start_sshd_if_needed may have *adopted* an sshd that appeared concurrently
+# (either before its running-check or by losing the bind race) — that daemon
+# loaded the pre-hardening config, so nudge it like the converge path does.
+# No-op when the config did not change; redundant-but-harmless SIGHUP when
+# the daemon we just started already read the hardened config.
+reload_sshd_if_config_changed
 release_lock
 spawn_watchdog
 
--- a/neurons/validators/src/services/assets/sshd_bootstrap.sh
+++ b/neurons/validators/src/services/assets/sshd_bootstrap.sh
@@ -1,10 +1,70 @@
 #!/bin/sh
+# Rental-container SSH bootstrap (DAH-2341: race-safe against images that
+# start sshd themselves).
+#
+# The container image's own entrypoint may generate host keys and start sshd
+# concurrently with this script (e.g. runpod-style /start.sh). This script
+# therefore never assumes it owns sshd bring-up; it converges on an invariant
+# instead: "host keys exist and sshd is listening on :22". Who got there
+# first does not matter.
+#
+# Decision flow:
+#   1. sshd already running          -> converge (no keygen), verify, exit.
+#   2. sshd binary present           -> wait up to a grace period for the
+#      image to start sshd itself; if it appears -> converge, verify, exit.
+#   3. fallback: install sshd if missing, then (under a lock shared with
+#      lium images' /start.sh) generate missing host keys, harden config,
+#      start sshd. A failed start is re-checked: if sshd is running anyway
+#      (lost a benign start race), that is success.
+#   4. verify: sshd running AND listening on port 22 — the exit code of this
+#      script reflects the invariant, not individual command exit codes.
+#
+# Callers treat a non-zero exit as bootstrap failure (fatal for templates
+# with ships_sshd=True), so exit 1 only when sshd is genuinely not serving.
 set -eu
 
-WATCHDOG_PIDFILE="/run/sshd-watchdog.pid"
+# LIUM_* environment overrides exist for tests (which cannot write /run or
+# /root) and for operational tuning; production execs never set them.
+RUN_DIR="${LIUM_RUN_DIR:-/run}"
+SSH_DIR="${LIUM_SSH_DIR:-/root/.ssh}"
+SSHD_CONFIG="${LIUM_SSHD_CONFIG:-/etc/ssh/sshd_config}"
+
+WATCHDOG_PIDFILE="$RUN_DIR/sshd-watchdog.pid"
 WATCHDOG_LOG="/tmp/sshd-watchdog.log"
 SLEEP_SECONDS=30
 SCRIPT_PATH="$0"
+
+# Non-negative integer (capped) or the fallback — the timing knobs feed
+# `[ -lt/-ge ]` tests that would abort the script under `set -eu` on malformed
+# input, and `docker exec` inherits the image's ENV, so a hostile image must
+# not be able to stretch the wait loops via oversized LIUM_* values.
+int_or() {
+    case "${1:-}" in
+        *[!0-9]* | "")
+            printf '%s' "$2"
+            return 0
+            ;;
+    esac
+    if [ "$1" -gt "$3" ]; then
+        printf '%s' "$3"
+    else
+        printf '%s' "$1"
+    fi
+}
+
+# Shared with lium images' /start.sh setup_ssh — keep the path in sync.
+LOCK_DIR="$RUN_DIR/lium-ssh-setup.lock"
+LOCK_TIMEOUT_SECS="$(int_or "${LIUM_SSH_LOCK_TIMEOUT_SECS:-}" 60 300)"
+LOCK_HELD=0
+
+# Grace period to let a self-starting image bring sshd up before the fallback
+# takes over. Resolved in main: --grace flag > LIUM_SSHD_GRACE_SECS > default
+# (only applies when an sshd binary ships in the image — an image without
+# sshd cannot self-start it, so waiting would be pure latency).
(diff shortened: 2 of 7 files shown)


fortunelucky777:
# DAH-2341 — Verification evidence

**PRs**: validator side [Datura-ai/lium-io#1107](https://github.com/Datura-ai/lium-io/pull/1107) (commit `e09e60fc`, base `main`) · image side [Datura-ai/computenet-docker-images#14](https://github.com/Datura-ai/computenet-docker-images/pull/14) (commit `3df0bc8`, base `feat/pytorch-dind-image`). Evidence recorded 2026-07-03 UTC.

## Failing baseline (pre-fix, from the task's staging reproduction)

Recorded in the Notion spec (staging, Docker SDK rental path):

- Image: `fortunelucky777/pytorch:2.12.0-py3.12-cuda12.8-devel-ubuntu24.04-dind` (custom clone, self-starting sshd)
- Executor: L40 / 38.80.122.143, Pod `93b13164-863a-4f2d-8e27-c0f3b32c98a2`
- Pod status RUNNING, ports published correctly, but `ssh root@38.80.122.143 -p 20299` → **connection refused after 26 attempts**
- Container log tail: `Setting up SSH...` → `/etc/ssh/ssh_host_rsa_key already exists.` → `Overwrite (y/n)?` (PID 1 killed by the keygen TOCTOU against the validator bootstrap), then a restart loop on the stale nested-dockerd pidfile (`process with PID 108 is still running`).
- Control: default template (old, non-self-starting image) passed 15/15 — the bootstrap won the race uncontested there.

## Post-fix evidence (local, pre-deploy)

1. **Unit/behavioral tests — new decision flow** (`neurons/validators/tests/test_sshd_bootstrap_script.py`, shim-based, runs the real script):
   10/10 passed — adopt-running (no keygen), adopt-during-grace, fallback bring-up, no-grace-when-no-sshd-binary, grace-0 override, bind-race tolerance, exit-1 only when sshd never serves, listener-on-:22 required, lock-timeout proceeds.
2. **Full validator suite**: 892 passed, 8 skipped (includes updated `test_deploy_optimizations.py` key-before-bootstrap ordering test and `test_docker_service.py` script-content tests).
3. **Forced-concurrency harness** (30 rounds, patched image `setup_ssh` vs patched validator bootstrap running simultaneously with `--grace 0` — the worst case): `ALL 30 ROUNDS PASSED: no concurrent keygen, both sides exit 0, single sshd, 3 host keys`. A shim ssh-keygen flags any overlapping keygen; a shim sshd/service exits 1 on bind conflict like the real ones.
4. **Lint**: new files ruff-0.5.1-clean; edited files have identical violation counts to the origin/main baseline (zero net-new).

## Post-deploy verification plan (from the Notion spec)

- Re-run the staging custom-image case that failed on L40 (`fortunelucky777/pytorch:2.12.0-...-dind`) → expect RUNNING + direct SSH OK on first probe.
- Re-run the default-template smoke across available staging executors (A6000 38.80.122.73, L40 38.80.122.143, A4000 149.36.1.151) → expect no regression; note deploy profiler `SSH_SERVICE_INSTALLATION` gains up to ~10s on images that ship an sshd binary but never self-start it (grace window; self-resolves once the rebuilt self-starting default images roll out).
- Grafana: `{namespace="subnet", container="connector"} |= "SSH bootstrap"` — expect "adopting existing daemon" / "Image-provided sshd came up" lines and no `SSH bootstrap script finished with errors` on daturaai images.

fortunelucky777:
## Staging verification — Layer 1 alone fixes the SSH bootstrap race

Deployed this branch to staging (connector rolled over `18:36 UTC`) and re-ran the failing scenario. Throughout, the image is the **unpatched racy `fortunelucky777/pytorch:2.12.0-...-dind`** (`/start.sh` still has the 3× prompt-capable `ssh-keygen -t`, no `-A`), so everything below is attributable purely to the validator change.

### Method

Rented the racy clone image on the L40 `38.80.122.143` — the exact node where the original hard failure (`Overwrite (y/n)?` → PID-1 death → nested-dockerd restart loop) occurred. Then rebooted it 5 more times. A staging reboot **recreates the container** (new container id → `/etc/ssh` reset → the image's keygen runs fresh), so each reboot is a genuine window-1 race, not a weaker restart.

### Black-box — 6 distinct fresh containers, SSH healthy on every one

| wave | container | SSH | host keys |
|---|---|---|---|
| fresh deploy | `8a4747d2193c` | ✅ | 3 |
| reboot 1 | `37b2914667ca` | ✅ | 3 |
| reboot 2 | `4c8cabc4ad34` | ✅ | 3 |
| reboot 3 | `3396c62f2d2d` | ✅ | 3 |
| reboot 4 | `4ec5b2425451` | ✅ | 3 |
| reboot 5 | `d48469162ce6` | ✅ | 3 |

### White-box — connector logs, every wave adopted

`services.rental_docker_observability` / `exec_run_ssh_bootstrap_script`:

```
18:42:51 exit=0  Waiting up to 10s for image-provided sshd | Image-provided sshd came up; adopting it | Sent SIGHUP to sshd (pid 653) to load hardened config | Started sshd watchdog with pid 686
19:05:38 exit=0  Image-provided sshd came up; adopting it → SIGHUP → watchdog
19:08:09 exit=0  Image-provided sshd came up; adopting it → SIGHUP → watchdog
19:09:29 exit=0  Image-provided sshd came up; adopting it → SIGHUP → watchdog
19:10:50 exit=0  Image-provided sshd came up; adopting it → SIGHUP → watchdog
19:11:41 exit=0  Image-provided sshd came up; adopting it → SIGHUP → watchdog
```

**`adopting it`: 6/6 · `exit_status=0`: 6/6 · `Overwrite`/bootstrap errors: 0** across the whole window.

### Why it works (and why it's not luck)

The old validator's immediate `ssh-keygen -A` is what filled the image's `[ ! -f key ]` → keygen check-to-act gap, so the image's keygen sometimes hit an existing file → interactive `Overwrite (y/n)?` → EOF → `set -e` kills PID 1. The new bootstrap **grace-waits and adopts the image's own sshd** instead of racing it, so it never creates keys in that gap — there is no longer a coin to flip. The `SIGHUP` line also confirms the hardened config is now actually reloaded (closes the silent hardening-drift face).

### Additional note

Also captured a **deterministic** proof of the mechanism inside a live pod (independent of timing): running the image's exact `ssh-keygen -t rsa -f KEY -q -N '' </dev/null` against a key that a prior `ssh-keygen -A` created reproduces `Overwrite (y/n)?` → exit 1 every time. That exercises the **image** code, which this PR doesn't touch — which is exactly why the image-side hardening (computenet-docker-images #14) is still worth landing as defense-in-depth (residual >10s-grace fallback window, stale `/var/run/docker.pid` restart-loop stickiness, and third-party images we don't rebuild).

### Conclusion

**Layer 1 (this PR) alone resolves the reported issue** — verified end-to-end on the original-failure node, n=6 waves, 100% healthy, 0 races. The image fix is complementary hardening, not a prerequisite.

pixel29913 (changes requested):
Nice work Adopt-before-keygen, key injection before bootstrap, and final sshd verification all move this in the right direction. Two things I think are worth tightening before merge.

1. Create-path gate vs. the new bootstrap contract  https://github.com/Datura-ai/lium-io/blob/e09e60fc7c4f5f37372c0dab1c3efb35a26ab173/neurons/validators/src/services/docker_service.py#L3333-L3334

The bootstrap now verifies “sshd running AND listening” and returns that faithfully, but the create path only escalates to a failed create when `payload.ships_sshd and not ssh_bootstrap_ok`.

For custom / Dockerfile-built rentals (`ships_sshd` usually `None`/`False`), the bootstrap can correctly return `False` after grace + fallback, yet `create_container` can still return `ContainerCreated`. That preserves the user-visible failure mode DAH-2341 targets: _pod marked RUNNING while customer SSH is refused._

Could we fail the create when ssh_bootstrap_ok is false for customer rentals? Since current_step = "ssh_bootstrap" is already set, this should produce the right failure step.

2. Fallback adopt path may not reload hardened config `sshd_bootstrap.sh:424` (fallback `start_sshd_if_needed`), vs. `converge_on_running_sshd at :283`

In the fallback path, `prepare_sshd_runtime` can harden `sshd_config`, then `start_sshd_if_needed` can observe/adopt an sshd that appeared concurrently. That path returns success but does not call `reload_sshd_if_config_changed`; only `converge_on_running_sshd` does. 

If sshd started before/while hardening was written, `PasswordAuthentication no` may be on disk but not loaded, while the script still exits 0.

May be reload on the fallback adopt/success path when `CONFIG_CHANGED=1`.

Mainly hoping to align the outer create invariant with the new bootstrap invariant: customer rental create succeeds only once SSH is actually serving.

fortunelucky777:
@pixel29913 Thanks for the careful review — re point 1 (create-path gate vs. the new bootstrap invariant):

Your reading is right: the gate only escalates for `ships_sshd=True`, so for custom / Dockerfile rentals a bootstrap that correctly returns false still yields `ContainerCreated`. I'd like to keep that lenient path as-is in this PR, though, for two reasons:

1. **It would turn today's succeeding deploys into hard failures.** The fallback installs openssh via `apt-get`/`apk`/`dnf`/`yum`; a custom image with no supported package manager fails the bootstrap on `main` today too, and its deploy currently succeeds. Some of those rentals are legitimately SSH-less — customers running a service via `startup_commands` + exposed ports, and Jupyter doesn't depend on the container sshd either (it's set up via the host SSH client). Escalating unconditionally for non-`ships_sshd` rentals makes "SSH must serve" a hard product requirement for every customer rental, which I don't think we've decided — I'd want product/team sign-off before making that class of image un-rentable.

2. **The DAH-2341 failure mode is the race, and that's fixed at the root.** The repro was a *working* image whose sshd got wedged by our concurrent keygen — that can no longer happen regardless of the gate. What the lenient path still allows through is images where sshd genuinely cannot serve, which is a different (pre-existing) failure class.

One more wrinkle: `WorkloadKind.FILLER` also flows through `create_container` and has no customer SSH user, so any escalation would need a filler exclusion on top of the `ships_sshd` check.

My suggestion: handle this via the follow-up already noted in the description — backend declaring per-tag/per-rental sshd expectations — so the escalation is driven by an explicit "this rental needs SSH" signal rather than inferred. If the consensus is that SSH is a hard requirement for every customer rental, I'm happy to add the escalation here with a `workload_kind != FILLER` guard instead; just say the word.

Point 2 is a fair catch — pushing a fix that calls `reload_sshd_if_config_changed` on the fallback adopt/success path.

pixel29913 (approved):

PR #1110, opened 2026-07-06
DAH-2346: forward per-executor incentive + default_job_owner to backend

taiberium:
## What
Forwards the per-executor `incentive` and `default_job_owner` (already on `JobResult`) to the backend over the existing validator→backend path. They were computed each cycle but dropped when building the `MACHINE_SPEC_CHANNEL` payload, so the backend never saw them.

## Changes
- `services/miner_service.py` — add `incentive` + `default_job_owner` to the `MACHINE_SPEC_CHANNEL` publish dict.
- `clients/compute_client.py` — pass both into `ExecutorSpecRequest(...)` (tolerant `.get()`).
- `protocol/vc_protocol/validator_requests.py` — add the two nullable fields to `ExecutorSpecRequest`.
- Test: `tests/test_publish_incentive.py` asserts the published payload includes both fields.

## Why
Backend needs to persist unrented incentive to TimescaleDB for the Default-Job revenue/incentive dashboard (DAH-2344). `default_job_owner` lets the dashboard isolate Lium's idle nodes.

## Deploy safety
Fields are optional with defaults and the backend model ignores extras → deploy order between this repo and lium-io-backend does not break the integration. **Paired with lium-io-backend PR (DAH-2346)** which reads and stores them.

### DIFF
--- a/neurons/validators/src/clients/compute_client.py
+++ b/neurons/validators/src/clients/compute_client.py
@@ -319,6 +319,8 @@ class ComputeClient:
                             price_per_gpu=data["price_per_gpu"],
                             collateral_deposited=data["collateral_deposited"],
                             ssh_pub_keys=data["ssh_pub_keys"],
+                            incentive=data.get("incentive"),
+                            default_job_owner=data.get("default_job_owner"),
                         )
 
                         async with self.lock:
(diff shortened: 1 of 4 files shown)


taiberium:
Abandoning this effort — incentive work will be handled separately by someone else. Closing and deleting the branch.

PR #1111, opened 2026-07-06
DAH-2329, read soft_limit_price_rate from shared config

arhangel66:
## What

The validator's unrented soft price limit multiplier was a local constant
`SOFT_LIMIT_PRICE_RATE = 1.1` in `incentive/rental_price.py`, duplicating the
same value published by compute-app in `core/constants.py`. This makes the
validator consume it from shared config instead, so there is a single source of
truth (compute-app publishes it via `/v1/shared-config`).

## Changes

- `incentive/rental_price.py`: drop the local `SOFT_LIMIT_PRICE_RATE` constant;
  read `shared_client.config.soft_limit_price_rate` at each use site (mirrors how
  `machine_prices_p90` is already consumed).
- `core/config.py`: update the comment referencing the old constant.
- `pyproject.toml` + `pdm.lock`: bump `lium-core` to `>=0.1.5` (the release that
  adds the `soft_limit_price_rate` field, backward-compatible default 1.1).

No behavior change: the published value is 1.1, identical to the dropped constant.

## Testing

`pdm run pytest` — 850 passed, 8 skipped locally (incl. rental price incentive
flow + prod snapshot; the value 1.1 is unchanged so snapshots are identical).

## Related

- lium-core 0.1.5 (adds `soft_limit_price_rate`) — released to PyPI.
- compute-app DAH-2329 (#729) publishes the field and reads it via a single
  in-process SharedConfig surface.

### DIFF
--- a/neurons/validators/pdm.lock
+++ b/neurons/validators/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "dev"]
 strategy = ["inherit_metadata"]
 lock_version = "4.5.0"
-content_hash = "sha256:d24a3d057bc9c825d6ab5680ea10aa99d0c5f1c996ac2154060c0ad1bbf4f506"
+content_hash = "sha256:ff6cbbf6ebdde7685ac77880b448c3dea112370c47d69cf26597dad37c257760"
 
 [[metadata.targets]]
 requires_python = ">=3.11,<3.13"
@@ -1245,7 +1245,7 @@ files = [
 
 [[package]]
 name = "lium-core"
-version = "0.1.4"
+version = "0.1.5"
 requires_python = ">=3.11"
 summary = "Shared core library for Lium platform"
 groups = ["default"]
@@ -1254,8 +1254,8 @@ dependencies = [
     "requests>=2.31.0",
 ]
 files = [
-    {file = "lium_core-0.1.4-py3-none-any.whl", hash = "sha256:9a982cbf8f6bb71fd1528e97b97e52083ce37e78b870c530d9d74fe5823c35e0"},
-    {file = "lium_core-0.1.4.tar.gz", hash = "sha256:d96692168011cb278478a5d044e4d62ac8994309551da9097d1d095e64b9fb7e"},
+    {file = "lium_core-0.1.5-py3-none-any.whl", hash = "sha256:62b718d9dfb029c897330765d769a0af1a69fc54cc43fe2cd11df96a55fa5b87"},
+    {file = "lium_core-0.1.5.tar.gz", hash = "sha256:0fda76ca77cee5a5376efef3bb0fda5efc63d54a57725086bf635210a61e8b7e"},
 ]
 
 [[package]]
--- a/neurons/validators/pyproject.toml
+++ b/neurons/validators/pyproject.toml
@@ -15,7 +15,7 @@ dependencies = [
     "celium-collateral==1.1.6",
     "databases>=0.9.0",
     "datura @ file:///${PROJECT_ROOT}/../../datura",
-    "lium-core>=0.1.4",
+    "lium-core>=0.1.5",
     "fastapi>=0.110.3",
     "pydantic>=2.11.7",
     "pydantic-settings>=2.10.1",
--- a/neurons/validators/src/core/config.py
+++ b/neurons/validators/src/core/config.py
@@ -176,7 +176,7 @@ class Settings(BaseSettings):
     SSH_DEBUG_LOGGING: bool = Field(env="SSH_DEBUG_LOGGING", default=True, description="Enable verbose asyncssh SSH handshake debug logging")
 
     # DAH-2250 — unrented incentive soft price limit. When True, an unrented executor
-    # whose price_per_gpu exceeds market p90 * SOFT_LIMIT_PRICE_RATE loses the unrented
+    # whose price_per_gpu exceeds market p90 * soft_limit_price_rate loses the unrented
     # rental incentive while staying active. When False, the breach is only logged
     # (shadow mode) so prod impact can be observed before enforcing.
     ENABLE_UNRENTED_SOFT_PRICE_LIMIT: bool = Field(env="ENABLE_UNRENTED_SOFT_PRICE_LIMIT", default=False)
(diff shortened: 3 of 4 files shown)


surcyf123:
Triage: @arhangel66 — draft, 64 days idle, CI failing, conflicting (DAH-2329 soft_limit_price_rate from shared config). Still wanted?

surcyf123:
Value audit (SO §69, Rustam asked 15 Sep 11:30Z to re-check every open PR): Mikhail: no behavior change (1.1 stays 1.1) and it needs a lium-core bump; draft 10 weeks; fold the shared-config read into the next rental_price change instead; propose closing

Numbers: size +11/-15 files=4; value 0 behavior change; risk money.

surcyf123:
**Fresh review of `39b7e40d`: findings, 2 low. No medium or high, so nothing blocks.**

The head didn't move: I reviewed `39b7e40d`, as expected.

### Findings

**1. low: the PR as a whole. lium-io#1445 duplicates this change.**
#1445, opened 11:14Z, says it "takes over #1111 (DAH-2329)". Its `src/` diff against this head is a single type annotation (`rate: float =` at `rental_price.py:262`). lium-platform#648's body at `4cf9a2e` now names #1445 as the validator half. Two open PRs edit the same four lines, so whichever merges second will conflict, and the flip gate in #648 points at only one of them.
*Fix:* decide which one ships. Close #1111 in favour of #1445, or close #1445 and point #648 back here. There's no code change to make either way.

**2. low: `neurons/validators/tests/test_unrented_soft_price_limit.py:100` (and lines 102, 113 and 122), plus the body's "Pairs with" section. They describe 1.5, but #648 now serves 1.25.**
The comment says "compute-app serves 1.5 with PRICE_LIMITS_RAISED_ENABLED on (lium-platform#648)", but #648's raised soft rate is 1.25 since `4cf9a2e`. The tests are still correct, because they pin whatever rate is served. Only the comment and the body are stale.
*Fix:* use 1.25 in the comment and test values (threshold 2.0 × 1.25 = 2.50, over at 2.51, log "x 1.25)"), and update the body. Optionally, add a test that a served config without `soft_limit_price_rate` gates at 1.1.

### Checks (all at `39b7e40d`)
- **Rebase resolution:** against the merge base `11ed407d`, `pyproject.toml` and `pdm.lock` are unchanged; main's side was taken, and lium-core resolves to 0.1.8 from PyPI. In `rental_price.py`, only the `SOFT_LIMIT_PRICE_RATE` block is removed, and main's DAH-2520/2546/2715 gate constants and models are kept whole.
- **The hidden conflict:** main's enforcement path at `rental_price.py:940` now passes `shared_client.config.soft_limit_price_rate` to `no_payout_because_price_above_market_soft_limit`. The gate (`:256`) and the structured log (`:262`) read it too. Nothing in the repo still references `SOFT_LIMIT_PRICE_RATE`.
- **Main has moved since the rebase:** 5 commits, including #1414, which adds 76 lines to `rental_price.py`. Merging `origin/main` at `5754d83a` is clean. The merged tree has no leftover constant; the enforcement call is at `:1015`.
- **Fallback when the field is missing:** in PyPI lium-core 0.1.8, `SharedConfig.soft_limit_price_rate` defaults to 1.1, and `DEFAULT_SHARED_CONFIG.soft_limit_price_rate` is 1.1. Checked by hand: a served config without the field validates to 1.1, and a served 1.25 stays 1.25. If the first fetch fails, the client uses `DEFAULT_SHARED_CONFIG`; if a refresh fails, it keeps the last good config. So with #648's flags off, or against an older backend, behaviour is unchanged.
- **Compatibility with #648's flags:** only `soft_limit_price_rate` (`PRICE_LIMITS_RAISED_ENABLED`) is read here. The hard flag (`machine_max_price_rate`) is already read live at `score_calculator.py:56`. In a scratch test at #648's current 1.25, the threshold is p90 × 1.25 and the log quotes "x 1.25)"; that plus the existing file passed 24 of 24.
- **Validator suite with the CI env** (`.github/workflows/test.yml`: the wallet names, sqlite URIs, `ENABLE_TDX_ATTESTATION=True`, `TDX_VERIFIER_URL`), run on Linux with Python 3.11 and `pdm install`: the head gives 2813 passed and 15 skipped. #1111 merged into main `5754d83a` gives 2883 passed and 15 skipped.
- **CI (Actions runs at `39b7e40d`):** Tests succeeded (Validators Tests, both ruff jobs, `tests-ok`), and CodeQL succeeded. The PR is a draft.

surcyf123:
@arhangel66 quick question on this one: Rustam opened [#1445](https://github.com/Datura-ai/lium-io/pull/1445) (DAH-2329, "takes over #1111, rebuilt on current main"). It makes the same change: the validator reads `soft_limit_price_rate` from shared config at every use and drops the local 1.1 constant.

Does #1445 replace this PR? If yes, you can close #1111 (the loop won't close your PR). If you'd rather keep this one, say so and the loop will note it on #1445.

Context: the loop pushed only a rebase here today (39b7e40d4); your commit b625230d is unchanged. The soft rate itself comes from lium-platform#648 and stays at p90 × 1.5, as Fish approved on 23 Sep.

neurons/validators/tests/test_unrented_soft_price_limit.py
```diff
@@ -91,6 +91,37 @@ def test_is_over_soft_price_limit_at_threshold_is_not_over(monkeypatch):
     assert over is False
 
 
+def _set_soft_rate(monkeypatch, rate: float) -> None:
+    new_cfg = shared_client.config.model_copy(update={"soft_limit_price_rate": rate})
+    monkeypatch.setattr(shared_client, "_config", new_cfg)
+
+
+def test_the_threshold_is_the_served_soft_limit_price_rate(monkeypatch):
+    # compute-app serves 1.5 with PRICE_LIMITS_RAISED_ENABLED on (lium-platform#648): threshold 2.0 * 1.5 = 3.0
```
taiberium: **NIT:** This public repo comment names a private PR, an internal flag and the old name compute-app. Write "the backend serves 1.5 when the raised soft limit is on".

taiberium (approved):

PR #1112, opened 2026-07-06
DAH-2340, publish structured zero-incentive reasons on MACHINE_SPEC_CHANNEL

taiberium:
### Task Link

[DAH-2340](https://app.notion.com/p/Incentive-data-contract-pass-structured-zero-incentive-reasons-from-validator-to-backend-392b8bfdbde98116b42fe7a7f5e8288f)

---

## Problem

The validator already explains zero incentive in provider-facing `log_text`, but downstream accounting and Portal UI must not parse prose to recover those reasons. Each catalogued provider eligibility or exclusion reason needs to travel as structured cycle data alongside the incentive calculation.

## Solution

The same `MinerLogLine` object now drives both outputs:

- `JobResult.record_incentive_log(line)` preserves the existing human-readable log.
- Lines carrying a `ZeroIncentiveReason` also produce a typed `IncentiveReason` with a stable code, provider-facing message, and explicit context.
- `publish_machine_specs` sends JSON-safe `incentive_reasons` on `MACHINE_SPEC_CHANNEL`.
- The Redis-to-WebSocket bridge forwards the field through `ExecutorSpecRequest`.

Payload semantics:

- Missing key / `null`: the publisher did not report structured reasons.
- `[]`: the cycle reported no catalogued zero-incentive reason.
- Non-empty list: structured zero-incentive reasons recorded for that cycle.

The active scoring flow short-circuits when a causal exclusion is sufficient, so it currently records at most one reason per executor and cycle. The list-shaped contract remains additive for future scoring behavior. `log_text` remains unchanged for backward compatibility.

## Changes

- Adds the typed `IncentiveReason` wire model to the existing incentive reason catalog.
- Records structured reasons on `JobResult` without including them in generic model dumps.
- Routes all ten catalogued provider eligibility and exclusion reasons in the active `rental_price` flow through `record_incentive_log`.
- Publishes and forwards `incentive_reasons` without exposing internal observability fields.
- Serializes reason context in Pydantic JSON mode so future datetime, UUID, or enum values cannot break Redis publishing.
- Preserves the newer attestation fields added to `ExecutorSpecRequest` since this PR was opened.
- Covers the newer disk-to-VRAM exclusion, which otherwise would have remained text-only.

## Current Consumer

The backend side is already deployed:

- `lium-io-backend#778` accepts the additive field and stores it in `incentive_per_validator_cycle.zero_incentive_reasons`.
- `lium-miner-portal#156` exposes cycle reasons through the Provider Earnings API.
- `lium-miner-portal-frontend#151` renders them in node timelines and expanded epoch details.

## Deployment

Deploy through the normal validator GitHub Action. The deployed backend is already backward-compatible with validators that omit the field and ready to consume this payload.

## Risks

- Low: scoring and incentive calculations are unchanged.
- The payload change is additive.
- Existing provider-facing log text is preserved.
- Reason codes remain an append-only contract; messages and context may evolve.

## Verification

- `pdm run pytest tests/test_incentive_reasons_payload.py tests/test_miner_incentive_log.py tests/test_unrented_insufficient_disk.py -q`
  - `62 passed`
  - Exercises the actual publisher payload through the Redis-to-WebSocket bridge.
  - Verifies non-primitive reason context is converted to JSON-safe values.
- `pdm run pytest tests/ -q --tb=short`
  - `1282 passed, 8 skipped`
  - Three unrelated `test_sshd_bootstrap_script.py` harness tests fail in the local macOS environment; this PR does not touch that script or its tests. Linux CI remains the authoritative full-suite result.
- `git diff --check`

## Review Request

- [x] Code review

## Checklist

- [x] PR description is current
- [x] Risks are documented
- [x] Focused tests cover the real publisher and bridge paths
- [x] Current reason catalog is covered

### DIFF
--- a/neurons/validators/src/protocol/vc_protocol/validator_requests.py
+++ b/neurons/validators/src/protocol/vc_protocol/validator_requests.py
@@ -8,6 +8,8 @@ import bittensor
 import pydantic
 from datura.requests.base import BaseRequest
 
+from incentive.miner_incentive_log import IncentiveReason
+
 
 class RequestType(enum.Enum):
     AuthenticateRequest = "AuthenticateRequest"
@@ -53,20 +55,6 @@ class AuthenticateRequest(BaseValidatorRequest):
             timestamp=int(time.time()),
         )
         return cls(payload=payload, signature=f"0x{keypair.sign(payload.blob_for_signing()).hex()}")
-
-
-class IncentiveReason(pydantic.BaseModel):
-    """One structured reason an executor earns 0 subnet incentive (DAH-2340).
-
-    `reason` is a stable, append-only machine-readable code the backend keys off;
-    `message_for_miner` is free text. Extra miner_log_fields ride along via
-    extra='allow', keeping the contract additive without a schema change.
-    """
-
-    model_config = pydantic.ConfigDict(extra="allow")
-
-    reason: str
-    message_for_miner: str
 
 
 class ExecutorSpecRequest(BaseValidatorRequest):
--- a/neurons/validators/src/protocol/vc_protocol/validator_requests.py
+++ b/neurons/validators/src/protocol/vc_protocol/validator_requests.py
@@ -101,7 +101,8 @@ class ExecutorSpecRequest(BaseValidatorRequest):
     attestation_digest: str | None = None
     tdx_attestation_passed: bool | None = None
     gpu_attestation_passed: bool | None = None
-    incentive_reasons: list[IncentiveReason] = []
+    # None = validator predates DAH-2340 (key absent); [] = evaluated, earns normally
+    incentive_reasons: list[IncentiveReason] | None = None
 
 
 class RentedMachineRequest(BaseValidatorRequest):
--- a/neurons/validators/src/protocol/vc_protocol/validator_requests.py
+++ b/neurons/validators/src/protocol/vc_protocol/validator_requests.py
@@ -55,6 +55,20 @@ class AuthenticateRequest(BaseValidatorRequest):
         return cls(payload=payload, signature=f"0x{keypair.sign(payload.blob_for_signing()).hex()}")
 
 
+class IncentiveReason(pydantic.BaseModel):
+    """One structured reason an executor earns 0 subnet incentive (DAH-2340).
+
+    `reason` is a stable, append-only machine-readable code the backend keys off;
+    `message_for_miner` is free text. Extra miner_log_fields ride along via
+    extra='allow', keeping the contract additive without a schema change.
+    """
+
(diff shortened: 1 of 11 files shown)


arhangel66 (approved):
looks good

pixel29913:
## Update for re-review

This PR was previously reviewed and approved, but the branch has since been rebased onto current `main` and received two follow-up commits. The current head is `6a9213af`.

### Changes since the earlier review

- Rebased the original work onto current `main`, preserving the newer `ExecutorSpecRequest` attestation fields alongside the optional `incentive_reasons` contract.
- Audited the active scoring flow against all 10 catalogued zero-incentive codes.
- Fixed the newer insufficient-disk-for-VRAM exclusion so it also records its structured reason through the centralized `record_incentive_log` path.
- Kept the wire semantics explicit: `None` means the publisher did not report the field, `[]` means it reported no catalogued reason, and a populated list carries the structured reason(s).
- Changed reason serialization to `model_dump(mode="json")` so future JSON-compatible context values such as enums, UUIDs, or datetimes cannot break Redis publication.
- Added regression coverage that exercises the real Redis bridge into `ExecutorSpecRequest`; removing the forwarding field now fails the test instead of silently degrading every cycle to "not reported."

The deployed backend contract already accepts, stores, and exposes these reasons, so the older backend PR was closed as superseded.

### Staging verification

The current head was deployed to staging and verified through the complete path:

`validator -> Redis -> connector -> backend -> TimescaleDB -> Provider Earnings API/UI`

A real cycle stored `nvidia_driver_below_minimum`. We then forced a second cycle for the same node with `provider_discord_not_connected`; both consecutive structured reasons were stored independently and surfaced correctly. The temporary staging provider state was restored afterward.

### Verification

- Focused validator reason/publisher/bridge suite: 62 passed
- Validator Tests: passed
- Executor Tests: passed
- CodeQL: passed
- PR is currently clean and mergeable

Please review the current head rather than relying only on the earlier approval, particularly commits `bb316cf1` and `6a9213af`.

neurons/validators/src/protocol/vc_protocol/validator_requests.py
```diff
@@ -8,6 +8,8 @@
 import pydantic
 from datura.requests.base import BaseRequest
 
+from incentive.miner_incentive_log import IncentiveReason
```
taiberium, line 11: **NIT:** this makes `protocol` depend on the `incentive` package. Define `IncentiveReason` here and have `miner_incentive_log` import it, like the backend does.

taiberium (commented):
Comments.

arhangel66 (approved):
looks good

PR #1115, opened 2026-07-07
DAH-2356: Lium PEARL GPU power cap (70%, min-clamped) — apply, validator exemption, pre-cap restore (executor)

taiberium:
## DAH-2356 — Executor half: apply the 80% GPU power cap, keep the node validatable & rentable, restore pre-cap on stop

**Draft.** Pairs with backend Datura-ai/lium-io-backend#742 (computes per-GPU watts, sends `gpu_power_limits`). Mechanism proven live on staging by PoC #1120 and independently re-verified on the real A4000 host (`149.36.1.151`): set 140→127→140 W from the privileged executor container; a non-privileged renter got `Insufficient Permissions`.

### Three parts

**1. Apply the cap (host-side).** `services/gpu_power_limit.py`: on FILLER create, query each GPU's live `[current, min, max, default]`, clamp the backend target into `[min, max]`, and `nvidia-smi -pl` it. Best-effort (logs, never raises; 30s timeout per call so a hung driver can't stall deploy).

**2. Validator keeps the node valid & rentable (the key change).** At 80% the cap trips `GpuPowerLimitCheck` (zeroes score below 0.90 while unrented). The check now **skips the penalty when the node runs a Lium-owned default job** — `rented_data.get_default_job_owner(executor) == "lium"` (new `DEFAULT_JOB_OWNER_LIUM` const, new `GPU_POWER_LIMIT_SKIPPED_LIUM_FILLER` reason). Same active-filler-skip pattern as `capability`/`verifyx`. Scoped to `owner == "lium"` **only** — a miner's own default job gets no pass, so providers can't underpower behind a miner job. The node keeps score → stays rentable; when a renter arrives the filler yields, power is restored, and the check applies normally again.

**3. Restore the pre-cap limit on stop (not stock default).** On apply we record each GPU's current (pre-cap) watts in validator Redis keyed by `pod_id`; on delete we set exactly those back and clear the key. **No-op when nothing was stored**, so fillers we never capped (miner default jobs) are left untouched — also fixes the earlier reset-overreach.

### Tests
```
pytest tests/test_gpu_power_limit.py tests/test_gpu_power_limit_check.py   # 21 new/updated
pytest   # full validator suite: 945 passed, 8 skipped
```
Golden wire-snapshot regenerated for the new `gpu_power_limits` field.

### Known edge (follow-up)
If the validator restarts between a PEARL filler's create and delete, the Redis restore key is lost → that filler's GPUs stay at 80% until a reconciliation sweep ("unrented + no Lium filler ⇒ default"). Not included here.


taiberium:
Mechanism proven live on staging by @colleague's PoC #1120 (2026-07-07): the validator's asyncssh session into the **privileged** executor container (`pid: host`, `NVIDIA_DRIVER_CAPABILITIES=all`) has full NVML admin, `nvidia-smi -pl` applies driver-wide, and renter pods (not privileged) can't undo it. This PR is the productionized form the PoC's 'real implementation' notes recommend: backend sends per-GPU-UUID watts in ContainerCreateRequest, validator applies + clamps, stateless restore reads `power.default_limit` on delete. Adopted the PoC's `timeout=30` on every nvidia-smi call.

arhangel66 (changes requested):
Using `Redis` to save the miner-side power limit is an interesting and working solution. However, I do not see any protection against failure cases. What happens if `delete_container` fails, the SSH connection breaks, or another exception occurs?

It seems we could end up with a situation where a reduced power limit gets **stuck indefinitely**, and we would then exclude the executor because of the low `PL`.

To make this robust, I propose we establish a few invariants and add two safety nets.

### Proposed invariants for the `Redis` state

* When changing the power limit, save the original `PL` to `Redis` using `gpu_uuid` as the key, not `pod_id`.
* The saved value must be frozen: never updated, only deleted — and only after the original limit has been successfully restored.

### Proposed safety nets

1. **In the validator**, when validating the `PL`: if the `lium job` has not started AND we have a record in `Redis`, it is an exception case. Restore the limit there, before the check.
2. **In the validator connector**, before container creation: if a record exists in `Redis` for this `gpu_uuid`, it is an expected case, and we should restore the limit before creating the container for the customer.

With these two safety nets, we can guarantee that we will never face penalties due to reduced `PL`, and never have reduced `PL` for customer containers.

### Logging

Every `PL` change should be strictly logged: `executor_uuid`, `gpu_uuid`, `PL` before, `PL` after, status of change.

taiberium:
Tested the whole flow on a real H100 executor. The chain is down (DAH-2397), so the miner leg was stubbed; everything else is the real code path. The cap goes 350W -> 217W on create, holds while the filler runs, and returns to 350W on delete. The watts came from a local run of the real backend (Datura-ai/lium-io-backend#742), not hand-picked.

Found a real bug along the way: with persistence mode off, nvidia-smi -pl exits 0 but the limit quietly reverts once the GPU goes idle. Fixed in 0ed664ca: each set now enables -pm 1 first and reads the limit back to confirm it stuck.

I also ran the safety nets live: a failed cap refuses to start the filler, a customer container clears a leftover cap before starting, and the check restores an orphaned cap after 30 min while still zeroing a below-floor GPU with no record. 1036 unit tests pass.

arhangel66 (approved):
Hey @taiberium, this is a risky change, so let's walk through a few edge cases before shipping:

1. **What is our recovery plan if the validator's Redis state is lost?** The physical cap is made durable on the miner via `-pm 1`, but the original limit is only stored in a Redis record. If these diverge, the miner's cap gets stuck. Could you confirm with Muhammed/Pixel how we ensure these Redis keys aren't lost? Specifically:
   - Is persistence (`RDB`/`AOF`) enabled, or is it a pure cache?
   - What is the `maxmemory-policy`? Any `*-lru` or `*-random` eviction policy could drop our keys under memory pressure.
   - Do the keys survive a validator redeploy?

2. **What happens if Redis is unreachable during a check?** While the filler is alive (backend reports `owner="lium"`), the check skips before touching Redis, so that path is safe. But for an orphaned cap (owner is no longer `lium`), the rescue path reads from Redis — a read error makes the code skip the record, treating the below-floor GPU as a miner violation and penalizing an innocent miner. **Should a Redis read error cause us to penalize?** That's a fail-closed choice; I'd like to confirm we're aligned on it rather than failing open (skipping the check) on a transient error.

3. **Don't forget to announce this change to miners.** As we are going to reactivate `PEARL` and will change their PL.

4. **The power-limit skip seems too broad.** The validator skips the power limit check for any `owner="lium"` job, but the cap is only applied to `PEARL`. So, miners can manually reduse PL for dolphin and we will not penalize it.
It's low-level, but I think will be needed to adj later.

taiberium:
Checked the Redis image we ship (daturaai/redis:7.4.2): appendonly is on, data sits on the redis_data volume, no maxmemory so nothing gets evicted. Keys survive restarts and redeploys. What it doesn't survive is a volume wipe or host move. For that case every cap logs the pre-cap watts to Loki, and a miner reboot resets the limit anyway.

On Redis being unreachable: you read it right, and penalizing on our own outage was wrong. Fixed in 3d92f8c7. A failed read now passes the check with GPU_POWER_LIMIT_STATE_UNAVAILABLE and retries next cycle. A clean read with no record still penalizes. Same commit adds a net at rental start: any GPU below 90% of default gets raised back to default straight from nvidia-smi, so a customer never inherits a cap even with Redis gone.

Announcement: agreed, we'll post it before reactivating PEARL.

Skip scope: agreed, taking it as a follow-up. The exemption should only cover GPUs that carry our cap record.

taiberium:
Superseded by #1136 — same tree, flattened to one linear commit because the repo only allows rebase-merge and this branch's merge commit from main can't be rebased. All review points addressed here carry over; staging evidence is in the comments above.

PR #1116, opened 2026-07-07
DAH-2360: retry connection-level errors in BackendClient + bounded keep-alive

arhangel66:
## Problem

Validator's `RentalVerificationCheck` intermittently fails with `RENTAL_CHECK_API_ERROR` ("API returned None") even though the backend is healthy: ~129 events per 6h across 27+ distinct executors on prod (2026-07-07), clustered at validation-wave boundaries.

Root cause: `BackendClient` holds a long-lived aiohttp session with a default keep-alive pool. When the backend/ingress closes a pooled connection, the validator's next write on it fails with `[Errno 32] Broken pipe` → `post()` returns `None` → the check fails the executor. Backend logs confirm the request never reached the app.

Notion: DAH-2360

## Fix

- Retry loop (2 retries, exponential backoff 1s → 5s) in `BackendClient.get()`/`post()` for connection-level errors only: `ServerDisconnectedError`, `ClientConnectorError`, `ClientOSError`. The short first pause heals stale-connection EPIPE instantly; the longer second one rides out a restarting backend pod. Timeouts and 5xx are deliberately NOT retried — by then the request may have reached the backend and run a long server-side operation twice.
- `TCPConnector(keepalive_timeout=30)` so stale pooled connections are dropped before the server closes them.
- Each retry logs a distinct `GET/POST connection error, retrying` warning for post-deploy measurement.

## Tests

- retry on `ClientOSError` (broken pipe) and `ServerDisconnectedError` succeeds on 2nd attempt
- returns `None` after retry budget exhausted (call count + backoff sequence asserted)
- timeout and generic `ClientError` are NOT retried (single call asserted)
- session keepalive bound asserted

`pdm run pytest tests/` in `neurons/validators`: 926 passed, 8 skipped (9 pre-existing greenlet env errors in `test_port_mapping_dao.py`, present on clean main).

### DIFF
--- a/neurons/validators/src/clients/backend_client.py
+++ b/neurons/validators/src/clients/backend_client.py
@@ -32,6 +32,20 @@ class BackendClient:
     _session: ClassVar[aiohttp.ClientSession | None] = None
     _session_lock: ClassVar[asyncio.Lock] = asyncio.Lock()
 
+    # DAH-2360: writing to a stale pooled keep-alive connection fails with
+    # EPIPE before the request reaches the backend, so these errors are safe
+    # to retry even for non-idempotent POSTs. Timeouts and 5xx are NOT retried:
+    # by then the request may have reached the backend and run server-side.
+    CONNECTION_ERRORS: ClassVar[tuple[type[Exception], ...]] = (
+        aiohttp.ServerDisconnectedError,
+        aiohttp.ClientConnectorError,
+        aiohttp.ClientOSError,
+    )
+    CONNECTION_RETRY_ATTEMPTS: ClassVar[int] = 2
+    CONNECTION_RETRY_BACKOFF_SECONDS: ClassVar[float] = 0.5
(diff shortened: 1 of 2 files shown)


neurons/validators/src/clients/backend_client.py
```diff
@@ -138,62 +110,102 @@ async def post(
...
+                            return None
+
+                        try:
+                            return response_model.model_validate(data)
+                        except ValidationError as e:
+                            logger.error(
+                                _m("Validation failed", extra=get_extra_info({**context, "error": str(e)}))
+                            )
+                            return None
+                except self.CONNECTION_ERRORS as e:
```
taiberium, line 185: **NIT:** connection error while reading the body retries this non-idempotent POST server-side. Scope the retry to the send phase.

taiberium (approved):

PR #1121, opened 2026-07-08
DAH-2364, graceful docker stop before forced container removal

arhangel66:
## Describe your changes

Adds a graceful `docker stop` (SIGTERM + 30s grace window, `CONTAINER_STOP_GRACE_SECONDS`) via the Docker SDK before the existing forced `remove_container(force=True)` in `DockerService.delete_container`.

Context: on wedged sysbox/containerd nodes (ticket-0258, 8xB200 at 23.153.44.21), the SIGKILL-only removal path fails with `could not kill container: ... did not receive an exit event`, the backend retries 4x, then deletes the pod row and the running container becomes an orphan holding all GPUs. A SIGTERM-first stop lets the workload (e.g. sglang) shut down cleanly and avoids triggering the containerd desync.

Behavior guarantees:
- **FILLER workloads skip the graceful stop entirely** and keep the pre-PR SIGKILL-only removal: the backend preempts a filler before a customer rent with a 30s total budget (`FILLER_STOP_WAIT_TIMEOUT_SECONDS`) and starts the rent anyway on timeout, so a grace window equal to that budget would let a SIGTERM-ignoring filler hold the GPUs into the customer rent.
- Stop failures are swallowed and logged (missing container -> info, anything else -> warning) and never change `delete_container`'s result.
- Forced removal path, absent-container tolerance for every workload kind (DAH-2345) and post-teardown steps are unchanged.
- docker-py extends the HTTP read timeout by the stop timeout, so the 30s grace cannot cause a read-timeout.

[DAH-2364 Graceful container teardown + escalation and alerting for failed pod deletions](https://app.notion.com/p/Graceful-container-teardown-escalation-and-alerting-for-failed-pod-deletions-397b8bfdbde9813eb8f6e27119b95912)

- [x] I wrote tests.
- [x] Need to take care of performance? (adds up to 30s to pod deletion only when the workload ignores SIGTERM; deletion is async fire-and-forget on the WS path)

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -3971,17 +3971,28 @@ class DockerService:
                 # DAH-2364: graceful stop first (SIGTERM + grace window) so the workload
                 # can exit cleanly before the forced removal below. Applies to customer
                 # rentals and fillers alike. Any stop failure is swallowed — the forced
-                # removal stays the single source of truth for deletion success.
+                # removal stays the single source of truth for deletion success. The stop
+                # is called directly (not via run_logged_rental_docker_sdk_operation) so an
+                # expected failure — chiefly an already-absent container — is not logged at
+                # ERROR and does not raise a false failed-deletion alert (DAH-2364).
+                stop_started = time.monotonic()
                 try:
-                    await run_logged_rental_docker_sdk_operation(
-                        operation="stop_container",
-                        log_extra=default_extra,
-                        call=lambda: docker_client.stop(
-                            container_name=payload.container_name,
-                            timeout=CONTAINER_STOP_GRACE_SECONDS,
-                        ),
+                    await docker_client.stop(
                         container_name=payload.container_name,
-                        stop_grace_seconds=CONTAINER_STOP_GRACE_SECONDS,
+                        timeout=CONTAINER_STOP_GRACE_SECONDS,
+                    )
+                    logger.info(
+                        _m(
+                            "Graceful container stop completed",
+                            extra=get_extra_info(
+                                {
+                                    **default_extra,
+                                    "container_name": payload.container_name,
+                                    "stop_grace_seconds": CONTAINER_STOP_GRACE_SECONDS,
+                                    "duration_ms": int((time.monotonic() - stop_started) * 1000),
+                                }
+                            ),
+                        ),
                     )
                 except Exception as exc:
                     if _is_missing_docker_container_error(exc):
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -112,6 +112,12 @@ LOG_STREAM_INTERVAL = 0.5  # 500ms — keeps build-log p95 emit→publish under 
 # would violate the SSE latency requirement.
 IN_CONTAINER_SSH_BOOTSTRAP_PATH = "/tmp/lium-ssh-bootstrap.sh"
 
+# DAH-2364: SIGTERM grace window (docker stop -t semantics) given to a container before
+# forced removal. SIGKILL-only removal of GPU-heavy workloads repeatedly wedged
+# containerd/sysbox ("could not kill container ... did not receive an exit event"),
+# leaving orphaned containers that hold GPUs and brick the executor.
+CONTAINER_STOP_GRACE_SECONDS = 30
+
 DOCKER_VOLUME_PLUGINS = {
     "s3fs": "mochoa/s3fs-volume-plugin"
 }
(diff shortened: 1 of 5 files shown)


neurons/validators/src/services/docker_service.py
```diff
@@ -3860,6 +3866,48 @@ async def delete_container(
                     private_key=private_key,
                 ) as docker_client,
             ):
+                # DAH-2364: graceful stop first (SIGTERM + grace window) so the workload
+                # can exit cleanly before the forced removal below. Applies to customer
+                # rentals and fillers alike. Any stop failure is swallowed — the forced
+                # removal stays the single source of truth for deletion success.
+                try:
+                    await run_logged_rental_docker_sdk_operation(
```
taiberium: NIT: 
whole method where you added this functionality is a horribly long one-function method.
Do you wanna decompose the method "delete_container"?

taiberium: This wrapper logs the stop failure at ERROR and re-raises, then the `except` below logs it again as info/warning and swallows it — so a stop you treat as non-fatal still emits an ERROR line plus a duplicate.

The already-gone-container case is the worst: stop -> NotFound -> ERROR `stop_container failed`, even though the next line correctly logs "already absent". DAH-2364 adds alerting for failed deletions, so these ERRORs will fire false alerts.

Suggest calling `docker_client.stop(...)` directly here (skip the ERROR-logging wrapper) and keep only the info/warning logging in the `except`.

arhangel66:
Thanks @taiberium — addressed the double-log / false-alert point: the graceful stop now calls `docker_client.stop()` directly instead of the ERROR-logging `run_logged_rental_docker_sdk_operation` wrapper, so an already-absent container no longer emits a false ERROR (and won't trip the failed-deletion alerting). Kept `info`(missing)/`warning`(other) in the `except`; forced removal stays the single source of truth. Left the `delete_container` decomposition out of scope for this PR. PTAL 🙏

neurons/validators/src/services/docker_service.py
```diff
@@ -103,6 +104,12 @@
 # would violate the SSE latency requirement.
 IN_CONTAINER_SSH_BOOTSTRAP_PATH = "/tmp/lium-ssh-bootstrap.sh"
 
+# DAH-2364: SIGTERM grace window (docker stop -t semantics) given to a container before
+# forced removal. SIGKILL-only removal of GPU-heavy workloads repeatedly wedged
+# containerd/sysbox ("could not kill container ... did not receive an exit event"),
+# leaving orphaned containers that hold GPUs and brick the executor.
+CONTAINER_STOP_GRACE_SECONDS = 30
```
taiberium, line 122: **Critical:** the graceful stop runs for `FILLER` too, and 30s is the backend's entire `FILLER_STOP_WAIT_TIMEOUT_SECONDS` budget. A filler that ignores SIGTERM burns it inside `docker stop`, so the backend gives up and starts the customer rent while the filler still holds the GPUs.

Skip the grace window for `FILLER`, or raise the backend timeout first.

neurons/validators/src/services/rental_docker_sdk.py
```diff
@@ -236,11 +236,14 @@ async def start(self, *, container_name: str) -> None:
             api_method=self._api_client.start,
         )
 
-    async def stop(self, *, container_name: str) -> None:
+    async def stop(self, *, container_name: str, timeout: int | None = None) -> None:
```
taiberium: NIT: `timeout` already means the HTTP client timeout on this class (see `create_volume`), here it means the SIGTERM grace. Rename to `stop_grace_seconds`.

neurons/validators/src/services/docker_service.py
```diff
@@ -151,6 +158,66 @@ def _missing_rental_docker_host_key_log_text(
     )
 
 
+class _BoundLog:
+    # structured logger bound to a default_extra, so a call site stays one line
+    def __init__(self, base_extra: dict):
```
taiberium: NIT: bare `dict` here and on `_delete_container_log_extra(...) -> dict`. `_message` has no return type and `**fields` is untyped.

neurons/validators/src/services/docker_service.py
```diff
@@ -151,6 +158,66 @@ def _missing_rental_docker_host_key_log_text(
...
+    # raise ValueError on an unsafe volume name; the quoted result is unused, these names
+    # reach the Docker SDK rather than a shell
+    if payload.local_volume:
+        _quote_safe_docker_volume_name(payload.local_volume, field_name="local_volume")
+    if payload.external_volume:
+        _quote_safe_docker_volume_name(payload.external_volume, field_name="external_volume")
+
+
+@contextlib.contextmanager
+def _best_effort(log: _BoundLog, step: str, **fields):
```
taiberium: NIT: generic helper, but the log line it writes says `delete_container post-teardown step failed`. Rename it `_best_effort_delete_step` or pass the message in. Return type is `Iterator[None]`.

neurons/validators/src/services/docker_service.py
```diff
@@ -151,6 +158,66 @@ def _missing_rental_docker_host_key_log_text(
     )
 
 
+class _BoundLog:
```
taiberium, line 172: NIT: this refactor is ~190 of the 200 changed lines and none of it is the graceful stop. It buries the two-line behavior change and makes a revert of DAH-2364 drag the logging rewrite with it.

neurons/validators/tests/test_docker_service.py
```diff
@@ -964,6 +968,181 @@ async def test_delete_customer_rental_treats_missing_container_as_deleted(
     )
 
 
+@pytest.mark.asyncio
+async def test_delete_container_stops_gracefully_before_forced_removal(
```
taiberium, line 977: NIT: the name says "before" but nothing asserts ordering, since `stopped_containers` and `removed_containers` are separate lists. Also no test for the already-absent container on the stop path, which is the branch `FILLER` hits.

taiberium (changes requested):
Request changes. 1 Critical (the 30s graceful stop eats the backend's 30s filler-stop wait), 5 NITs.

taiberium (approved):

PR #1122, opened 2026-07-08
DAH-2365, bound matrix/verifyx SSH runs and emit GPU_VERIFY_TIMEOUT

arhangel66:
## Describe your changes

Bounds the previously unbounded SSH runs in the validation pipeline and makes a hung GPU check produce an explicit, observable result instead of a silent executor drop:

- `matrix_validation_service`: matmul run is now bounded by `MATRIX_VERIFY_TIMEOUT_SECONDS = 120`. On timeout the validator emits a new reason code `GPU_VERIFY_TIMEOUT` (score 0) and best-effort kills the leaked remote `decrypt_challenge.py` via `pkill`.
- `verifyx_validation_service`: SSH commands bounded by `VERIFYX_COMMAND_TIMEOUT_SECONDS = 600` (timeout folds into the existing `SSH_TRANSPORT` classification).
- `miner_service` (WS + REST paths): `asyncio.gather` results are now filtered through `_filter_task_results`, which logs a warning with executor uuid/address when a task is dropped without a result (previously exceptions were filtered out silently).

Context: on the wedged 8xB200 node (ticket-0258) the matmul hung forever, the executor's pipeline died only at the outer 780s `wait_for`, produced no reason code, and the stale sweep later flipped the executor inactive with zero operator-visible signal.

Notes from review (accepted, not regressions):
- `pkill -f <script_path>` could in principle hit a concurrent legitimate matmul run on the same executor; runs are per-executor and serialized per validator cycle, so the overlap window is negligible.
- Worst case verifyx (600s) + matmul (120s) leaves ~60s of the 780s outer budget for other checks; both steps were previously unbounded, so this only makes failures more graceful.

[DAH-2365 Matrix multiplication GPU check: emit explicit timeout result and clean up remote process](https://app.notion.com/p/Matrix-multiplication-GPU-check-emit-explicit-timeout-result-and-clean-up-remote-process-397b8bfdbde981d88bd9ff8b47f02ba9)

- [x] I wrote tests.
- [x] Need to take care of performance? (no extra work on the happy path; timeout only bounds already-hung runs)

### DIFF
--- a/neurons/validators/src/services/matrix_validation_service.py
+++ b/neurons/validators/src/services/matrix_validation_service.py
@@ -1,14 +1,24 @@
+import asyncio
 import time
 import random
 import logging
 import json
 import os
+import shlex
 import uuid as uuid4
 from dataclasses import dataclass
+
+import asyncssh
+
 from core.utils import _m, get_extra_info
 from ctypes import CDLL, c_longlong, POINTER, c_void_p, c_char_p
 
 logger = logging.getLogger(__name__)
+
+# Hard cap on the remote matmul run. Real runs finish in 4-27s; without a cap a wedged
+# GPU (e.g. VRAM held by an orphaned container, DAH-2364) hangs the whole pipeline until
+# the outer JOB_TIME_OUT cancellation, which dies silently without a reason_code (DAH-2365).
+MATRIX_VERIFY_TIMEOUT_SECONDS = 120
 
 
 class DMCompVerifyWrapper:
@@ -135,6 +145,7 @@ class ValidationResult:
     stderr: str = ""
     error_message: str = ""
     metrics: dict | None = None
+    timed_out: bool = False
 
     def __bool__(self) -> bool:
         """Allow using ValidationResult in boolean context for backward compatibility."""
@@ -195,6 +206,18 @@ class ValidationService:
         max_dim_k = max_elements // (2 * dim_n) - dim_n
 
         return max_dim_k
+
+    async def _kill_remote_matmul(self, ssh_client, script_path: str, log_extra: dict) -> None:
+        # best-effort kill of the leaked matmul process so it stops holding GPU memory
+        try:
+            await ssh_client.run(f"pkill -f {shlex.quote(script_path)}", timeout=10)
+        except Exception as e:
+            logger.warning(
+                _m(
+                    f"Failed to kill remote matmul process: {e}",
+                    extra=get_extra_info(log_extra),
+                )
+            )
 
     async def validate_gpu_model_and_process_job(
         self,
@@ -283,7 +306,21 @@ class ValidationService:
 
             # Run the script
             try:
-                result = await ssh_client.run(command)
+                result = await ssh_client.run(command, timeout=MATRIX_VERIFY_TIMEOUT_SECONDS)
+            except (asyncssh.TimeoutError, asyncio.TimeoutError):
+                # asyncssh's timeout only abandons the local wait — the remote process keeps
+                # running and holding GPU memory, so it must be killed explicitly.
+                error_msg = (
+                    f"Matrix multiplication timed out after {MATRIX_VERIFY_TIMEOUT_SECONDS}s"
+                )
+                logger.error(_m(error_msg, extra=get_extra_info(log_extra)))
+                await self._kill_remote_matmul(ssh_client, script_path, log_extra)
+                return ValidationResult(
+                    success=False,
+                    expected_uuid=verifier_params.uuid,
+                    error_message=error_msg,
+                    timed_out=True,
+                )
             except Exception as e:
                 error_msg = f"Failed to execute SSH command: {str(e)}"
                 logger.error(_m(error_msg, extra=get_extra_info(log_extra)))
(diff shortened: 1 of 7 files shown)


neurons/validators/src/services/matrix_validation_service.py
```diff
@@ -1,15 +1,25 @@
...
+
 from core.utils import _m, get_extra_info
 from ctypes import CDLL, c_longlong, POINTER, c_void_p, c_char_p
 
 logger = logging.getLogger(__name__)
 
+# Hard cap on the remote matmul run. Real runs finish in 4-27s; without a cap a wedged
+# GPU (e.g. VRAM held by an orphaned container, DAH-2364) hangs the whole pipeline until
+# the outer JOB_TIME_OUT cancellation, which dies silently without a reason_code (DAH-2365).
+MATRIX_VERIFY_TIMEOUT_SECONDS = 120
```
taiberium, line 21: Are you sure the timeout values are set properly and do not conflict with other timeouts in the lifecycle?

neurons/validators/src/services/verifyx_validation_service.py
```diff
@@ -24,6 +24,11 @@
 # Cap on stderr bytes captured per failure. Last 2 KB is kept when longer.
 STDERR_TAIL_BYTES = 2048
 
+# Hard cap on the remote verifyx run. Memory/storage/network probes can legitimately take
+# minutes, so this is generous — it only cuts true hangs (same silent-hang class as the
+# matrix check, DAH-2365) instead of blocking until the outer JOB_TIME_OUT cancellation.
+VERIFYX_COMMAND_TIMEOUT_SECONDS = 600
```
taiberium, line 30: Are you sure the timeout values are set properly and do not conflict with other timeouts in the lifecycle?

neurons/validators/src/services/task/messages.py
```diff
@@ -877,6 +877,17 @@ class CapabilityMessages:
         impact="Score set to 0",
         remediation="Run Docker GPU diagnostics (nvidia-smi) and ensure containers can access GPUs.",
     )
+    VERIFY_TIMEOUT = MessageTemplate(
```
taiberium, line 880: Will this message be shown properly on the provider UI?

taiberium (approved):

neurons/validators/src/services/miner_service.py
```diff
@@ -428,6 +425,30 @@ async def request_job_to_miner(
                 str(e),
             )
 
+    def _filter_task_results(self, executors, raw_results, default_extra: dict) -> list:
```
taiberium: Nit: missing type hints. `executors: list[ExecutorSSHInfo]`, `raw_results: list`, and a concrete return type would match our types-everywhere rule.

neurons/validators/src/services/miner_service.py
```diff
@@ -428,6 +425,30 @@ async def request_job_to_miner(
...
+            logger.warning(
+                _m(
+                    "Executor task dropped without result",
+                    extra=get_extra_info(
+                        {
+                            **default_extra,
+                            "executor_uuid": str(executor_info.uuid),
+                            "executor_ip_address": executor_info.address,
+                            "error_type": type(result).__name__ if result is not None else None,
+                            "error": str(result) if result is not None else "empty result",
```
taiberium: Nit: `str(TimeoutError())` is `""`, so an outer wait_for timeout logs `error: ""`. `error_type` still shows `TimeoutError`, so cosmetic only.

neurons/validators/src/services/matrix_validation_service.py
```diff
@@ -196,6 +207,18 @@ def get_max_matrix_dimensions(self, gpu_memory, dim_n):
 
         return max_dim_k
 
+    async def _kill_remote_matmul(self, ssh_client, script_path: str, log_extra: dict) -> None:
+        # best-effort kill of the leaked matmul process so it stops holding GPU memory
+        try:
+            await ssh_client.run(f"pkill -f {shlex.quote(script_path)}", timeout=10)
```
taiberium: Nit: this is truly best-effort — default SIGTERM won't kill a CUDA call stuck in D (uninterruptible) state, which is the wedge class this targets. Fine to keep; the explicit reason code is the real win.

taiberium (commented):
Minor nits only — nothing blocking.

arhangel66:
Thanks @taiberium! Nits addressed:
- added type hints to `_filter_task_results` (`executors: list[ExecutorSSHInfo]`, `raw_results: list`, `-> list[JobResult]`)
- `str(TimeoutError()) == ""` now falls back to `error_type` so the log line isn't empty

On the no-code nits: verifyx (600s) + matmul (120s) fit inside the 780s outer budget; `GPU_VERIFY_TIMEOUT` renders fine on the provider UI (`reason_code` is a free-form string there, no allowlist). Left `pkill`/SIGTERM as-is per your note.

arhangel66:
Follow-up on the SIGTERM/D-state nit: switched the remote kill to `pkill -9 -f` (SIGKILL). Reasoning — for a pure D-state wedge no signal lands anyway, so `-9` is no worse there, but for every other hang (userspace loop, ignored/slow SIGTERM) it's strictly more reliable and can't be caught. VRAM is reclaimed on process death regardless of signal, and there's no cleanup worth preserving for a throwaway probe. Tests updated.

PR #1124, opened 2026-07-09
DAH-2375, remove anonymous volumes in failed-create cleanup and prune dangling orphans

arhangel66:
### Task Link

[DAH-2375](https://www.notion.so/398b8bfdbde981709269d4f2f5682f22)

---

## Problem

Executor nodes slowly fill their disk with orphaned anonymous docker volumes. Dind-based rental images (`daturaai/pytorch:*-dind`, `daturaai/dind`) declare `VOLUME /var/lib/docker`, so docker creates one anonymous volume per container. The happy-path delete removes containers with `-v` and reaps them, but the failed-create/retry cleanup paths remove containers WITHOUT `-v`, orphaning the anonymous volume on every failed create / port retry. Nothing ever prunes them (`volume_prune()` helper existed but was never called), so orphans accumulate forever — observed 1345 volumes / ~32GB on the staging A4000, pushing the disk to 84% and silently blocking the PEARL fit check.

## Solution

Two-part fix:

1. **Stop the leak** — all container-removal paths now also remove the container's anonymous volumes (`docker rm -fv` / `remove_volumes=True`). This is safe for retry data: the `-v` flag (Docker API `v=1`) removes only anonymous volumes and never touches named volumes such as `volume_<pod_id>`.
2. **Garbage-collect the backlog** — `ContainerCleanup.prune_dangling_anonymous_volumes()` runs on every executor each validation cycle (via `StaleContainerCleanupCheck`) and removes volumes that are dangling AND have a 64-hex anonymous name. Named volumes are excluded by the name filter; in-use volumes are excluded twice over — the `dangling=true` filter never lists a volume referenced by any container (running or stopped), and `docker volume rm` itself refuses to remove a volume that raced back into use. It removes at most `VOLUME_RM_MAX_PER_PASS = 200` per pass so the `docker volume rm` command stays small; since GC runs every cycle, any larger backlog drains over a few passes. Best-effort: never raises, never changes the check verdict.

The dangerous `DockerCommand.volume_prune()` (`prune -af` would delete unused NAMED volumes) was removed entirely.

## Changes

- `docker_service.py`: `_remove_failed_rental_container_for_retry` → `remove_volumes=True`; `_remove_failed_container_for_retry` → `docker rm -fv` (+ fixed misleading comment); custom-build dind teardown → `-fv`; stale doc comments updated
- `custom_build_orphan_sweep.py`: orphaned build-dind removal → `-fv`
- `container_cleanup.py`: new `prune_dangling_anonymous_volumes()` + `get_dangling_anonymous_volumes()` wired into `cleanup()` (including the idle-executor path, whose early return was dropped so the sweep still runs); constants `ANON_VOLUME_NAME_PATTERN`, `VOLUME_RM_MAX_PER_PASS = 200`
- `docker_utils.py`: new command builders `volume_ls_dangling()` and variadic `volume_remove(*names)`; removed the now-unused `remove()` and dangerous `volume_prune()`
- Tests: updated exact-command assertions to `-fv`; new test for the SDK retry path; new tests for the volume GC (filter keeps named volumes, removes anonymous, caps removals per pass, swallows SSH failures, respects `dry_run`, and is invoked by `cleanup()`)

## Deployment Steps

Use GitHub Action.

## Review Request

- [x] Just code review

## Other PRs

None.

## Risks

- Volume GC runs a couple extra SSH commands per executor per validation cycle (cheap: `volume ls` + one capped `rm`).
- A dangling volume is by definition referenced by no container (running or stopped), so no rental can be using it; if one races back into use between the `ls` and the `rm`, `docker volume rm` refuses it and it is silently skipped.
- Named volumes (`volume_<pod_id>`) are protected twice: `-v`/`v=1` never removes named volumes, and the GC only matches bare 64-hex anonymous names.

### Test Case 1: full validator test suite

**Actions**

`cd neurons/validators && pdm run pytest`

**Expected Output**

Green, apart from the 3 pre-existing failures in `test_sshd_bootstrap_script.py` that reproduce on clean `origin/main` and are unrelated.

### Test Case 2: volume GC unit coverage

**Actions**

`pdm run pytest tests/test_container_cleanup.py -q`

**Expected Output**

Removes bare 64-hex dangling volumes; keeps named ones (`volume_*`, `hc_*`, `pod_*`, `executor_db_data`) even when dangling; caps removals at `VOLUME_RM_MAX_PER_PASS` per pass; returns 0 without raising on SSH failure; respects `dry_run`; `cleanup()` return contract unchanged.

## Checklist

- [x] Risks are documented
- [x] Tests are described

### DIFF
--- a/neurons/validators/src/core/docker_utils.py
+++ b/neurons/validators/src/core/docker_utils.py
@@ -69,7 +69,6 @@ class DockerCommand:
         """Build docker volume ls command listing dangling volume names."""
         return "/usr/bin/docker volume ls -qf dangling=true"
 
-
     @staticmethod
     def inspect_created_timestamp(container_id: str) -> str:
         """Build docker inspect command to get creation timestamp in seconds."""
--- a/neurons/validators/src/core/docker_utils.py
+++ b/neurons/validators/src/core/docker_utils.py
@@ -33,11 +33,6 @@ class DockerCommand:
         )
 
     @staticmethod
-    def remove(name: str) -> str:
-        """Build docker rm command."""
-        return f"/usr/bin/docker rm -f {name}"
-
-    @staticmethod
     def remove_with_volumes(name: str) -> str:
         """Build docker rm command that also removes anonymous volumes."""
         return f"/usr/bin/docker rm -fv {name}"
@@ -64,34 +59,16 @@ class DockerCommand:
         return f"/usr/bin/docker inspect {container_id} --format '{{{{.State.ExitCode}}}}' 2>&1"
 
     @staticmethod
-    def volume_prune() -> str:
-        """Build docker volume prune command."""
-        return "/usr/bin/docker volume prune -af"
-
-    @staticmethod
-    def volume_remove(volume_name: str) -> str:
-        """Build docker volume rm command."""
-        return f"/usr/bin/docker volume rm {volume_name} 2>/dev/null || true"
+    def volume_remove(*volume_names: str) -> str:
+        """Build docker volume rm command, tolerating per-volume failures."""
+        names = " ".join(shlex.quote(name) for name in volume_names)
+        return f"/usr/bin/docker volume rm {names} 2>/dev/null || true"
 
     @staticmethod
     def volume_ls_dangling() -> str:
         """Build docker volume ls command listing dangling volume names."""
         return "/usr/bin/docker volume ls -qf dangling=true"
 
-    @staticmethod
-    def volume_inspect_created(volume_names: list[str]) -> str:
-        """Build docker volume inspect command printing name and creation time per line."""
-        names = " ".join(shlex.quote(name) for name in volume_names)
-        return (
-            "/usr/bin/docker volume inspect "
-            f"--format '{{{{.Name}}}} {{{{.CreatedAt}}}}' {names} 2>/dev/null"
-        )
-
-    @staticmethod
-    def volume_remove_batch(volume_names: list[str]) -> str:
-        """Build docker volume rm command for multiple volumes, tolerating per-volume failures."""
-        names = " ".join(shlex.quote(name) for name in volume_names)
-        return f"/usr/bin/docker volume rm {names} 2>/dev/null || true"
 
     @staticmethod
     def inspect_created_timestamp(container_id: str) -> str:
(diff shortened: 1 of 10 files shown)


neurons/validators/src/services/container_cleanup.py
```diff
@@ -104,8 +109,65 @@ async def cleanup(
...
 
+    async def prune_dangling_anonymous_volumes(
+        self,
+        ssh_client,
+        executor_uuid: str,
+    ) -> int:
+        # reap orphaned anonymous volumes left behind by container removals without -v
+        extra = {"executor_uuid": executor_uuid}
+        try:
+            dangling_anonymous_volumes: list[str] = await self.get_dangling_anonymous_volumes(ssh_client)
```
taiberium: **NIT:** `dangling_anonymous_volumes` is typed `list[str]`, but the helper can return `None` and the next line checks `is None`.

```suggestion
            dangling_anonymous_volumes: Optional[list[str]] = await self.get_dangling_anonymous_volumes(ssh_client)
```

neurons/validators/src/core/docker_utils.py
```diff
@@ -64,14 +59,16 @@ def inspect_exit_code(container_id: str) -> str:
...
+        return f"/usr/bin/docker volume rm {names} 2>/dev/null || true"
 
     @staticmethod
-    def volume_remove(volume_name: str) -> str:
-        """Build docker volume rm command."""
-        return f"/usr/bin/docker volume rm {volume_name} 2>/dev/null || true"
+    def volume_ls_dangling() -> str:
+        """Build docker volume ls command listing dangling volume names."""
+        return "/usr/bin/docker volume ls -qf dangling=true"
+
```
taiberium, line 71: **NIT:** Stray extra blank line here, two blanks before the next method. Drop one.

taiberium (approved):
Approve. 2 NITs, none blocking.

neurons/validators/src/services/container_cleanup.py
arhangel66: Fixed in e935fab4 — annotation is now `Optional[list[str]]`.

neurons/validators/src/core/docker_utils.py
arhangel66: Fixed in e935fab4 — dropped the extra blank line.

PR #1126, opened 2026-07-09
DAH-2380 - Resolve default template digests from Docker Hub in validator

fortunelucky777:
## Describe your changes

- Add a stateless `fetch_default_image_digests()` that builds a fresh `repo:tag → sha256` snapshot from Docker Hub at each job-cycle start (token GET + manifest **HEAD** with the manifest-list accept header, so the digest matches `docker pull` RepoDigests on amd64 and the requests are exempt from Docker Hub pull rate limits).
- The image list comes from the backend via shared config: `SharedConfig.default_docker_images` carries the backend's `DOCKER_IMAGES` entries verbatim, and the validator **derives the `repo:tag` refs locally** from the `image`/`tag` keys (no hardcoded copy, no backend-derived refs field). Malformed entries are skipped, fail-open.
- The snapshot is threaded through the existing `rented_data`-style parameter rails into `ContextConfig`; `CachedTemplateVerificationCheck` compares the executor's local `RepoDigest` against it instead of `docker_image_digest` on the backend response.
- Fail-open by design: a ref whose fetch fails this cycle is absent from the snapshot, so the check skips rather than zeroing an honest miner on a re-pushed tag or a Docker Hub outage.
- Unit tests for ref derivation, digest fetch, snapshot fail-open, and updated cached-template verification tests.

[DAH-2380 - Validator fetches default template digests from Docker Hub](https://app.notion.com/p/Validator-fetches-default-template-digests-from-Docker-Hub-398b8bfdbde98165a783e641bdc0f227)

- [x] I wrote tests.
- [x] Need to take care of performance? (per-cycle: one HEAD per default image, ~2 refs)

## Other PRs (rollout ordering)

1. **lium-core#5** — adds `SharedConfig.default_docker_images` (v0.1.6, must be released to PyPI first).
2. **lium-io-backend#754** — publishes `default_docker_images` on `/v1/shared-config` and stops serving per-template digests.
3. This PR — after 0.1.6 is on PyPI, regen the lockfile here to pin it; deploy backend + validator together.

## Test plan

- [x] Full validator suite: 999 passed, 8 skipped
- [x] `uvx ruff@0.5.1` clean on changed files
- [ ] Deploy validator + backend together; confirm `recommended_image_digest_match` still populates on a node with the recommended image pre-pulled

### DIFF
--- a/neurons/validators/src/core/validator.py
+++ b/neurons/validators/src/core/validator.py
@@ -74,10 +74,6 @@ class Validator:
             keypair=keypair,
         )
         self.default_docker_image_digest_service = get_default_docker_image_digest_service()
-        # Retain the task ref so the background refresh loop isn't garbage-collected mid-run.
-        self._digest_refresh_task = asyncio.create_task(
-            self.default_docker_image_digest_service.run_refresh_loop()
-        )
 
         port_tester = PortTester()
         runner = ContainerRunner()
@@ -220,6 +216,19 @@ class Validator:
             if current_block - self.last_job_run_blocks >= settings.BLOCKS_FOR_JOB:
                 job_block = (current_block // settings.BLOCKS_FOR_JOB) * settings.BLOCKS_FOR_JOB
                 job_batch_id = await self.subtensor_client.get_time_from_block(job_block)
+
+                # DAH-2380: refresh default cache-template digests from Docker Hub at the
+                # start of each job cycle, so verification compares executors against the
+                # current tags. Fail-open — a Docker Hub hiccup must never block the cycle.
+                try:
+                    await self.default_docker_image_digest_service.refresh()
+                except Exception as exc:
+                    logger.error(
+                        _m(
+                            "[sync] default docker image digest refresh failed; using last cache",
+                            extra=get_extra_info({**self.default_extra, "error": str(exc)}),
+                        ),
+                    )
 
                 # Fetch all rented executors from backend API
                 rented_executors = await self.backend_client.get_all_rented_executors()
--- a/neurons/validators/src/core/validator.py
+++ b/neurons/validators/src/core/validator.py
@@ -3,35 +3,34 @@ import json
 import os
 import time
 
-from payload_models.payloads import MinerJobRequestPayload
 from clients.backend_client import BackendClient
-
-from core.config import settings
+from clients.subtensor_client import SubtensorClient
 from incentive.eligibility import is_missing_discord_after_cutoff
 from incentive.factory import IncentiveFactory
 from incentive.rental_price import precompute_all_estimates
-from core.utils import _m, get_extra_info, get_logger
-from clients.subtensor_client import SubtensorClient
+from payload_models.payloads import MinerJobRequestPayload
+from services.attestation_service import AttestationService
+from services.collateral_contract_service import CollateralContractService
+from services.default_docker_image_digest_service import get_default_docker_image_digest_service
 from services.docker_service import DockerService
 from services.executor_connectivity.container_runner import ContainerRunner
 from services.executor_connectivity.dind_probe import DindProbe, DindVerifier
+from services.executor_connectivity.orchestrator import ConnectivityOrchestrator
 from services.executor_connectivity.port_probe import PortProbe
 from services.executor_connectivity.port_selector import PortSelector
 from services.executor_connectivity.port_tester import PortTester
 from services.executor_connectivity.port_verifiers import BatchVerifier, FallbackVerifier
-from services.executor_connectivity.orchestrator import ConnectivityOrchestrator
 from services.executor_connectivity_service import ExecutorConnectivityService
 from services.file_encrypt_service import FileEncryptService
+from services.matrix_validation_service import ValidationService
 from services.miner_service import MinerService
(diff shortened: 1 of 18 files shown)


neurons/validators/src/services/const.py
```diff
@@ -272,3 +272,10 @@
         },
     }
 }
+
+# Default cache-template images whose digests the validator pre-fetches from Docker Hub
+# (DAH-2380). Keep in sync with lium-io-backend `DOCKER_IMAGES`.
+DEFAULT_DOCKER_IMAGE_REFS: tuple[str, ...] = (
```
arhangel66: You will have to keep it synced. Once a new template is created in the backend, you'll need to update it here as well, right?

We already have a shared constant to share values like this between projects.

Please use it, or at least create a backlog task to do so.

arhangel66 (approved):
Approved, but please read the comment

neurons/validators/src/services/default_docker_image_digest_service.py
```diff
@@ -0,0 +1,125 @@
...
+        updated: dict[str, str] = {}
+        timeout = aiohttp.ClientTimeout(total=30)
+        async with aiohttp.ClientSession(timeout=timeout) as session:
+            for image_ref in self._image_refs:
+                digest = await fetch_registry_digest(session, image_ref)
+                if digest:
+                    updated[image_ref] = digest
+
+        async with self._lock:
+            self._digests.update(updated)
```
taiberium: **Critical:** partial refresh keeps the last digest when a fetch fails (`update` merges only successes). After the cutoff, a stale entry vs a re-pushed tag gives a false `DIGEST_MISMATCH` and score 0. Clear failed refs instead.

neurons/validators/src/core/validator.py
```diff
@@ -74,6 +73,8 @@ async def initiate_services(self):
             base_url=settings.COMPUTE_REST_API_URL or "",
             keypair=keypair,
         )
+        self.default_docker_image_digest_service = get_default_docker_image_digest_service()
+        asyncio.create_task(self.default_docker_image_digest_service.run_refresh_loop())
```
taiberium: **NIT:** the refresh task isn't retained, so it can be garbage-collected mid-run.

```suggestion
        self._digest_refresh_task = asyncio.create_task(self.default_docker_image_digest_service.run_refresh_loop())
```

neurons/validators/src/services/const.py
taiberium: **NIT:** this must stay byte-identical to the backend `DOCKER_IMAGES`; on drift `get_digest` returns `None` and the digest check silently no-ops.

taiberium (changes requested):
Request changes. 1 Critical (stale digest cache can false-fail after cutoff), 2 NITs.

taiberium (approved):

taiberium (changes requested):
Critical comments exists

neurons/validators/src/services/const.py
fortunelucky777: I am not sure about adding more templates in the future but Fish even wants to have 1 template. But having them in shared config makes more sense.

fortunelucky777:
Pushed two follow-up commits addressing review + the shared-constant feedback:

- `0789ef53` — **fix (Critical, @taiberium):** `refresh()` now replaces the digest map wholesale so a failed Docker Hub fetch **drops** the ref (check fails open) instead of keeping a stale digest that would false-`DIGEST_MISMATCH` a re-pushed tag and zero an honest miner. Also retains the refresh task ref (NIT). + regression test.
- `c271e1d4` — **refactor (@arhangel66 / @taiberium):** the default image list is no longer a hardcoded copy in `const.py`. It's now read from `SharedConfig` (served by the backend from its `DOCKER_IMAGES`), so a new template auto-propagates without editing/restarting the validator. Bumps `lium-core>=0.1.6`.

Cross-repo (supersedes the `feat/2380-remove-backend-digest-v2` note above):
- Backend: Datura-ai/lium-io-backend#758 (draft) — serves `default_docker_image_refs` via `/v1/shared-config`.
- lium-core: `feat/validator-docker-hub-digest` adds the field (v0.1.6) — **must be released to PyPI first**, then regen lockfiles here + in backend. (Branch not yet pushed — needs write access to Datura-ai/lium-core.)

fortunelucky777:
Follow-up `a5be94c4`: moved digest fetching to **the start of each job cycle** (per reviewer discussion) instead of a time-based background loop.

- `refresh()` is now called at the top of the job-cycle branch in `sync()` (before executors are verified), so verification always compares against current Docker Hub tags for that cycle. The call is **fail-open** — wrapped so a Docker Hub hiccup logs and proceeds on the last good cache, never blocking the cycle.
- Removed `run_refresh_loop`, the retained `_digest_refresh_task`, `refresh_seconds`, and the now-unused `DEFAULT_DOCKER_IMAGE_DIGEST_REFRESH_SECONDS` setting. This makes the earlier task-retention NIT (@taiberium) moot — there's no background task anymore.
- The shared-config list resolution is unchanged (still read per refresh, i.e. per cycle now).

taiberium (approved):

PR #1127, opened 2026-07-09
DAH-2379: surface Docker deletion in progress

pixel29913:
## Describe your changes

Handle Docker `409 removal already in progress` during container deletion as a soft/non-terminal delete state.

When Docker reports that removal is already in progress, the validator now returns `FailedContainerRequest` with `error_code=DeletionInProgress` and passes the exact raw Docker exception string through `msg`. The backend companion PR uses that code to avoid treating the retry collision as a provider-impacting delete failure.

This keeps normal delete failures unchanged, still treats missing containers as successful idempotent deletes, and preserves the existing graceful-stop-before-force-remove delete flow from `main`.

Deploy order: deploy the backend companion PR first, because older backend code strictly parses the enum and needs to understand `DeletionInProgress` before validators can send it.

Tests run:

- `pdm run pytest tests/test_docker_service.py::test_docker_container_removal_in_progress_detection_covers_docker_py_api_error tests/test_docker_service.py::test_delete_container_removal_in_progress_returns_soft_failure tests/test_docker_service.py::test_delete_container_failure_msg_includes_underlying_error tests/test_docker_service.py::test_delete_customer_rental_treats_missing_container_as_deleted tests/test_docker_service.py::test_delete_filler_container_treats_missing_container_as_deleted tests/test_docker_service.py::test_delete_container_stops_gracefully_before_forced_removal tests/test_docker_service.py::test_delete_container_stop_failure_still_removes`
- `python -m py_compile neurons/validators/src/services/docker_service.py neurons/validators/tests/test_docker_service.py`
- `git diff --check -- neurons/validators/src/services/docker_service.py neurons/validators/tests/test_docker_service.py neurons/validators/src/payload_models/payloads.py`

[DAH-2379](https://app.notion.com/p/398b8bfdbde9816298f7fb8607610971)

Backend companion PR: https://github.com/Datura-ai/lium-io-backend/pull/753

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/payload_models/payloads.py
+++ b/neurons/validators/src/payload_models/payloads.py
@@ -596,6 +596,7 @@ class FailedContainerErrorCodes(enum.Enum):
     UnknownError = "UnknownError"
     NoSshKeys = "NoSshKeys"
     ContainerNotRunning = "ContainerNotRunning"
+    DeletionInProgress = "DeletionInProgress"
     NoPortMappings = "NoPortMappings"
     InvalidExecutorId = "InvalidExecutorId"
     ExceptionError = "ExceptionError"
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -194,6 +194,7 @@ _LOCAL_VOLUME_TIMEOUT_GB_PER_SEC = 10
 _LOCAL_VOLUME_TIMEOUT_MAX_SEC = 180
 _FILLER_EXTERNAL_PORT_OFFSET = 20
 _DOCKER_NO_SUCH_CONTAINER_PHRASE = "No such container"
+_DOCKER_REMOVAL_IN_PROGRESS_PHRASES = ("409", "removal", "already in progress")
 HOST_KEY_REQUIRED_EXTRA = {
     "ssh_host_key_missing": True,
     "docker_sdk_host_key_required": True,
@@ -321,16 +322,24 @@ def _parse_volume_size_to_bytes(value: str | None) -> int | None:
     return int(float(number) * _VOLUME_SIZE_SUFFIX_MULTIPLIERS[suffix.lower()])
 
 
-def _is_missing_docker_container_error(exc: Exception) -> bool:
-    if _DOCKER_NO_SUCH_CONTAINER_PHRASE in str(exc):
-        return True
+def _exception_texts(exc: Exception) -> list[str]:
+    texts = [str(exc)]
     if isinstance(exc, RetryError):
         last_exception = exc.last_attempt.exception()
-        return (
-            last_exception is not None
-            and _DOCKER_NO_SUCH_CONTAINER_PHRASE in str(last_exception)
-        )
-    return False
+        if last_exception is not None:
+            texts.append(str(last_exception))
+    return texts
+
+
+def _is_missing_docker_container_error(exc: Exception) -> bool:
+    return any(_DOCKER_NO_SUCH_CONTAINER_PHRASE in text for text in _exception_texts(exc))
+
+
+def _is_docker_container_removal_in_progress_error(exc: Exception) -> bool:
+    return any(
+        all(phrase in text.lower() for phrase in _DOCKER_REMOVAL_IN_PROGRESS_PHRASES)
+        for text in _exception_texts(exc)
+    )
 
 
 def _is_stale_vloopback_mountpoint_error(exc: Exception) -> bool:
@@ -4580,10 +4589,7 @@ class DockerService:
         docker_client: RentalDockerSdkClient,
         payload: ContainerDeleteRequest,
         log: _BoundLog,
-    ) -> None:
-        # DAH-2345: deletion is idempotent for every workload kind — a container that is already
-        # gone (e.g. removed by failed-create cleanup) must not fail the delete, or the backend
-        # retries a doomed request until force-removal and penalizes the miner.
+    ) -> FailedContainerRequest | None:
         try:
             await run_logged_rental_docker_sdk_operation(
                 operation="remove_container",
@@ -4598,6 +4604,29 @@ class DockerService:
                 remove_volumes=True,
             )
         except Exception as exc:
+            if _is_docker_container_removal_in_progress_error(exc):
+                # Docker is still processing the original force-remove request;
+                # let the backend keep polling without treating this as terminal.
+                error_msg = str(exc)
+                log.info(
+                    "Container deletion is still in progress",
+                    container_name=payload.container_name,
+                    error=error_msg,
+                )
+                return FailedContainerRequest(
+                    miner_hotkey=payload.miner_hotkey,
+                    executor_id=payload.executor_id,
+                    pod_id=payload.pod_id,
+                    workload_kind=payload.workload_kind,
+                    msg=error_msg,
+                    error_type=FailedContainerErrorTypes.ContainerDeletionFailed,
+                    error_code=FailedContainerErrorCodes.DeletionInProgress,
+                )
+
+            # DAH-2345: deletion is idempotent for every workload kind — a container
+            # that is already gone (e.g. removed by failed-create cleanup) must not
+            # fail the delete, or the backend retries a doomed request until
+            # force-removal and penalizes the miner.
             if not _is_missing_docker_container_error(exc):
                 raise
             log.info(
@@ -4605,6 +4634,7 @@ class DockerService:
                 container_name=payload.container_name,
                 error=str(exc),
             )
+        return None
 
     async def delete_container(
         self,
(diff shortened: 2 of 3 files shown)


neurons/validators/src/services/docker_service.py
```diff
@@ -3874,7 +3879,31 @@ async def delete_container(
                         remove_volumes=True,
                     )
                 except Exception as exc:
-                    # DAH-2345: deletion is idempotent for every workload kind — a container
+                    if _is_docker_container_removal_in_progress_error(exc):
```
taiberium: NIT: Can we hide this complex logic behind an internal method to make the code easier to read?

taiberium (approved):

neurons/validators/src/services/docker_service.py
```diff
@@ -208,6 +208,11 @@ def _is_missing_docker_container_error(exc: Exception) -> bool:
     return False
 
 
+def _is_docker_container_removal_in_progress_error(exc: Exception) -> bool:
```
taiberium, line 338: **NIT:** `_is_missing_docker_container_error` unwraps tenacity `RetryError` before matching, and this one doesn't.

Nothing retries `remove_container` today so there's no bug, but if a retry wrapper is added later, `str(RetryError)` won't contain the Docker text and this predicate will quietly stop matching. Worth making the two symmetric.

neurons/validators/src/services/docker_service.py
```diff
@@ -208,6 +208,11 @@ def _is_missing_docker_container_error(exc: Exception) -> bool:
...
+    text = str(exc).lower()
+    return "409" in text and "removal" in text and "already in progress" in text
```
taiberium: **NIT:** The match phrases are inline literals here.

The file keeps this kind of thing in module constants (`_DOCKER_NO_SUCH_CONTAINER_PHRASE` at line 137, `_VLOOPBACK_MOUNT_ERROR_PHRASES` at line 123), so these could move up too.

neurons/validators/src/payload_models/payloads.py
```diff
@@ -511,6 +511,7 @@ class FailedContainerErrorCodes(enum.Enum):
     UnknownError = "UnknownError"
     NoSshKeys = "NoSshKeys"
     ContainerNotRunning = "ContainerNotRunning"
+    DeletionInProgress = "DeletionInProgress"
```
taiberium, line 599: **NIT:** Deploy order.

The old backend parses `error_code` strictly, so an unknown `"DeletionInProgress"` raises a pydantic `ValidationError` and drops the whole message. Backend #753 has to deploy first. That's noted in the backend PR, worth repeating in this description.

neurons/validators/src/services/docker_service.py
```diff
@@ -3874,7 +3879,31 @@ async def delete_container(
...
+                        return FailedContainerRequest(
+                            miner_hotkey=payload.miner_hotkey,
+                            executor_id=payload.executor_id,
+                            pod_id=payload.pod_id,
+                            workload_kind=payload.workload_kind,
+                            msg=error_msg,
+                            error_type=FailedContainerErrorTypes.ContainerDeletionFailed,
+                            error_code=FailedContainerErrorCodes.DeletionInProgress,
+                        )
+                    # DAH-2345: deletion is idempotent for every workload kind - a container
```
taiberium: **NIT:** Unrelated change: the em dash in this DAH-2345 comment became a plain hyphen. Revert it to keep the diff clean.

```suggestion
                    # DAH-2345: deletion is idempotent for every workload kind — a container
```

neurons/validators/tests/test_docker_service.py
```diff
@@ -1015,6 +1016,55 @@ async def test_delete_container_failure_msg_includes_underlying_error(
     assert "500 Server Error: daemon exploded" in result.msg
 
 
+@pytest.mark.asyncio
+async def test_delete_container_removal_in_progress_returns_soft_failure(
```
taiberium, line 1421: **NIT:** This is the only new test and the error string is hand written.

Nothing exercises the real docker-py `APIError` format, and nothing proves a plain 409 (`You cannot remove a running container`) still falls through to the normal failure path. A direct unit test on `_is_docker_container_removal_in_progress_error` would cover both.

taiberium (approved):
Approve. 5 NITs, none blocking.

neurons/validators/src/services/docker_service.py
pixel29913: Addressed in the rebased shape: the removal-in-progress handling now lives inside _force_remove_container, so delete_container keeps the high-level flow: graceful stop, force remove, best-effort cleanup.

neurons/validators/src/services/docker_service.py
pixel29913: Addressed: missing-container and removal-in-progress matching now both go through _exception_texts, which unwraps Tenacity RetryError via last_attempt.exception().

neurons/validators/src/services/docker_service.py
pixel29913: Addressed: moved the match phrases to _DOCKER_REMOVAL_IN_PROGRESS_PHRASES next to the existing Docker error constants.

neurons/validators/src/payload_models/payloads.py
pixel29913: Addressed: repeated the backend-first deploy order in the validator PR description.

neurons/validators/src/services/docker_service.py
pixel29913: Addressed: the DAH-2345 comment keeps the em dash again.

neurons/validators/tests/test_docker_service.py
pixel29913: Addressed: added test_docker_container_removal_in_progress_detection_covers_docker_py_api_error covering docker-py APIError, RetryError wrapping, and unrelated 409 fallthrough.

fortunelucky777 (approved):
Approving the change itself — the mechanism holds up. Verified against the checkouts:

- `FailedContainerErrorCodes` mirrors exactly across the only two copies that exist (validator `payload_models/payloads.py` + backend `protocol/validator_requests.py`) — no third consumer left behind.
- The phrase matcher fires on the real production string, not just the test's fake client: `_call_api` raises `RentalDockerOperationError(_wrap_error_message(...))`, and `_wrap_error_message` appends the full `str(exc)`, so `409` / `removal` / `already in progress` all survive the wrapping.
- Returning a `FailedContainerRequest` is contract-compatible — `_failed_delete` already returns one, and both `miner_service` call sites pass `result` straight back without inspecting it.

**Blocking on one thing: this branch is 16 commits behind `main` and conflicts in `docker_service.py`.**

The conflict is not cosmetic. `b38ea5e0` (DAH-2427) landed after this base and wrapped the exact call this PR rewrites:

```python
                try:
                    await self._force_remove_container(docker_client, payload, log)
                except Exception:
                    # DAH-2427: a failed force-remove (backend FAILED / STOP_FAILED) is the
                    # classic wedge path — sweep before propagating so a wedged card does not
                    # outlive the failed delete. No-op while a live compute app still exists.
                    if payload.workload_kind == WorkloadKind.FILLER:
                        with _best_effort_delete_step(log, "sweep_wedged_gpus_after_failed_remove"):
                            await _sweep_wedged_gpus_after_teardown(ssh_client, log)
                    raise
```

Please rebase and nest the new early return **inside** that `try` rather than replacing it:

```python
                try:
                    removal_failure = await self._force_remove_container(docker_client, payload, log)
                    if removal_failure is not None:
                        return removal_failure
                except Exception:
                    # DAH-2427 sweep unchanged
                    ...
                    raise
```

`return` never reaches `except`, so `DeletionInProgress` still short-circuits while a *genuine* force-remove failure keeps sweeping the wedged GPU before propagating. Resolving by taking this branch's side wholesale would drop that sweep and re-open the DAH-2427 failure — a card stays pinned into the next rental.

Non-blocking nit: neither test exercises the real wrapping — the predicate test uses a bare `APIError`, and the integration test injects into the fake client. Asserting the predicate against `RentalDockerOperationError(_wrap_error_message("Docker SDK remove container failed", APIError(...)))` would close that seam; if the wrapper ever changes, this feature degrades silently back to the old behavior with no alarm.

pixel29913:
Rebased onto latest `main` and resolved the `docker_service.py` conflict from DAH-2427.

Specific review follow-ups addressed:
- kept the new `DeletionInProgress` early return inside the force-remove `try`
- preserved the DAH-2427 wedged-GPU sweep for genuine force-remove failures
- kept missing-container delete idempotency unchanged
- added coverage for the real wrapped Docker path: `RentalDockerOperationError(_wrap_error_message(... APIError ...))`, plus retry wrapping
- preserved the exact Docker error in `FailedContainerRequest.msg`

Staging smoke after deploy:
- verified validator/connector image `9815010` and backend image `8c8f387`
- rented pod `be8e27ba-5796-4213-988a-140d156233a2` on A6000 executor `185.216.20.168`
- SSH, `nvidia-smi`, logs endpoint, reboot, post-reboot GPU check, and delete all passed
- final pod list was empty
- connector/backend logs had no pod-specific `ERROR`, no `409`, and no Docker `500`

Note: the staging smoke covered the normal lifecycle path. The actual Docker `409 removal already in progress` branch is covered by the unit tests rather than a live staging reproduction.

PR #1129, opened 2026-07-10
DAH-2392 - Disable SSH connect debug logging by default

fortunelucky777:
## Describe your changes

Follow-up to DAH-2272. The SSH connect phase-timing instrumentation added to investigate the slow-connect / "Started in subnet" stall shipped **enabled by default**. That investigation is now closed, so this turns the debug logging off by default while keeping the tooling in the tree (re-enablable via env for any future investigation).

**What was noisy**
- **Layer 1 — per-connection phase timing:** `neurons/validators/src/services/ssh_connect_timing.py` emitted a `ssh_connect_phase_timing` info line on **every** `asyncssh.connect`. This was a hard `logger.info` **not gated by any flag** — so `SSH_DEBUG_LOGGING=false` alone did *not* silence it.
- **Layer 2 — asyncssh handshake DEBUG:** `neurons/validators/src/core/utils.py` raised the `asyncssh` logger to DEBUG (debug level 2) per connection when `SSH_DEBUG_LOGGING` was on — and the flag defaulted to `True`.

**How it's fixed**
- `core/config.py` — `SSH_DEBUG_LOGGING` default flipped `True → False` (off everywhere by default; set `SSH_DEBUG_LOGGING=true` per env to re-enable).
- `services/ssh_connect_timing.py` — the per-connect `ssh_connect_phase_timing` block is now gated behind `settings.SSH_DEBUG_LOGGING`, so the flag actually controls both layers.
- The abnormal `ssh_connection_lost` warning is intentionally left **ungated** — it only fires on abnormal drops (keepalive timeout / mid-deploy reset), so it's a low-volume signal, not per-connection noise.
- The two `docker_service.py` callsites are unchanged; they use `connect_with_phase_timing` purely as a context manager, so connects behave identically — just without the per-connection log when off.

**Tests**
- `tests/test_ssh_connect_timing.py`: an autouse fixture enables the flag for the existing emission tests, plus a new `test_phase_timing_suppressed_when_flag_off` proving the line is silent when off while the connection is still established/closed and abnormal drops still warn.
- `pytest tests/test_ssh_connect_timing.py` → **10 passed**.
- `ruff@0.5.1` clean on the changed lines (two pre-existing findings in `ssh_connect_timing.py`, `I001`/`UP035`, predate this branch and are left untouched per repo convention).

[DAH-2392 — Disable DAH-2272 SSH connect debug logging by default](https://app.notion.com/p/Disable-DAH-2272-SSH-connect-debug-logging-by-default-399b8bfdbde981f8aeecfd2c1036b00f)

### Verification (from the spec)

**Pre-deploy**
- [x] `pytest neurons/validators/tests/test_ssh_connect_timing.py` passes (10 passed).
- [x] With `SSH_DEBUG_LOGGING` unset (new default False): `connect_with_phase_timing` does **not** emit `ssh_connect_phase_timing`, and the asyncssh logger stays at WARNING (covered by the suppression test + `test_apply_asyncssh_log_level_off`).
- [x] With `SSH_DEBUG_LOGGING=true`: the phase-timing line and asyncssh DEBUG still appear (tooling re-enablable) — covered by the emission tests + `test_apply_asyncssh_log_level_on`.
- [x] `ruff` clean on changed lines.

**Post-deploy** (whoever deploys the validator re-runs):
- Query validator logs (Loki) for `ssh_connect_phase_timing` and confirm the per-connection line has stopped appearing.
- Confirm no `asyncssh` DEBUG handshake spam.
- Confirm the validator still verifies executors normally (SSH connects still succeed).

> No read-only baseline probe applies here — the "bug" is log noise from deployed validators, not a reproducible functional defect. Coverage is via unit tests; the post-deploy Loki check above is the confirmation.

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/core/config.py
+++ b/neurons/validators/src/core/config.py
@@ -175,17 +175,16 @@ class Settings(BaseSettings):
     DRY_RUN: bool = Field(env="DRY_RUN", default=False, description="Run validation without publishing scores/weights")
     CONTAINER_CLEANUP_DRY_RUN: bool = Field(env="CONTAINER_CLEANUP_DRY_RUN", default=False, description="Dry run mode for stale container cleanup")
 
-    # DAH-2272: raise the asyncssh logger to DEBUG (debug level 2) so the SSH
-    # handshake (banner / key exchange / auth) is logged per connection. Enabled
-    # by default while DAH-2272 is under investigation, so the handshake trace is
-    # captured whenever the rare slow-connect transient hits; set
-    # SSH_DEBUG_LOGGING=false (env) to silence it without a deploy. Lives on the
-    # main settings (not DebugSettings, which is local-dev only) so it is
-    # configurable in staging/prod where the real slow connects happen.
-    # TODO(DAH-2272): flip default back to False once the slow-connect
-    # investigation is closed — debug-level handshake logging shouldn't ship on
-    # by default long-term.
-    SSH_DEBUG_LOGGING: bool = Field(env="SSH_DEBUG_LOGGING", default=True, description="Enable verbose asyncssh SSH handshake debug logging")
+    # DAH-2272: when on, raise the asyncssh logger to DEBUG (debug level 2) so the
+    # SSH handshake (banner / key exchange / auth) is logged per connection, and
+    # emit the per-connect ``ssh_connect_phase_timing`` breakdown (see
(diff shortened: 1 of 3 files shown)


arhangel66 (approved):
looks good

PR #1130, opened 2026-07-10
DAH-2391: isolate metrics collection from executor event loop

arhangel66:
## Problem

The executor agent is a single-process, single-event-loop FastAPI app. `/hardware_utilization` and `/containers/{name}` call sync psutil / pynvml / docker-py — including `exec_run("df -k ...")` into the renter's container — directly on the event loop with no timeouts. Any hang (dead FUSE mount: 61s block reproduced on staging) or a burst of ordinary renter polling (10 concurrent calls serialize to 19.2s) blocks `/ping` and `/upload_ssh_key`, producing false-offline flips, `EXECUTOR_INACTIVE_MID_RENTAL` penalties, and failed rental starts while the node is actually healthy.

Investigation: see DAH-2391.

## Changes

- **Dedicated `ThreadPoolExecutor(4)` for metrics** — blocking collection never runs on the event loop; `wait_for(8s)` bounds every call. A leaked thread (stuck docker daemon) degrades metrics only, never `/ping`.
- **`Semaphore(3)` with immediate 503** — excess concurrent metrics requests are rejected, not queued, so polling bursts cannot pile up.
- **`df` via killable subprocess** — `docker exec ... df -k` with a 5s timeout replaces docker-py `exec_run`: a `df` stuck in D-state on a dead FUSE mount kills the subprocess instead of permanently leaking a pool thread; the response carries `"stale": true` zeros.
- **Single module-level docker client** with `timeout=5` instead of `docker.from_env()` per request.
- **Container lookup in `/containers/{name}/logs`** offloaded to the pool with a 5s bound (was a raw docker call on the loop).
- **CI**: executor test suite now runs on PRs (`test.yml` had no executor job at all).

`sr25519` verification deliberately stays sync on the loop (~1ms; offloading it to a shared pool would queue pings behind slow metrics calls).

## Rollout note

Do not roll out to the fleet (runner image rebuild) before backend PR #755 (3-consecutive-miss offline debounce) is live: a fleet-wide watchtower restart can exceed the single 10s ping window.

## Tests

5 new tests in `tests/test_metrics_loop_isolation.py`: ping stays <1s while a metrics call blocks its worker; over-limit request gets 503 while in-flight ones finish; timeout returns 503; df timeout returns stale zeros; df success parses and passes the timeout.
`pdm run pytest` — 45 passed.

### DIFF
--- a/neurons/executor/src/routes/apis.py
+++ b/neurons/executor/src/routes/apis.py
@@ -5,9 +5,10 @@ import os
 import threading
 import time
 import tomllib
+from collections.abc import Callable
 from concurrent.futures import ThreadPoolExecutor, TimeoutError as FutureTimeoutError
 from pathlib import Path
-from typing import Annotated, Optional
+from typing import Annotated, Any, Optional
 
 import bittensor
 import docker
@@ -54,7 +55,7 @@ _metrics_executor = ThreadPoolExecutor(max_workers=4, thread_name_prefix="metric
 _metrics_semaphore = asyncio.Semaphore(METRICS_MAX_CONCURRENT)
 
 
-async def _run_in_metrics_pool(label: str, func, timeout: float | None = None):
+async def _run_in_metrics_pool(label: str, func: Callable[[], Any], timeout: float | None = None) -> Any:
     # run one blocking call in the dedicated pool, bounded by a timeout read at call time
     if timeout is None:
         timeout = METRICS_TIMEOUT_SECONDS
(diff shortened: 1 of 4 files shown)


neurons/executor/src/routes/apis.py
```diff
@@ -314,9 +363,16 @@ async def stream_container_logs(
         if not reserved_follow_stream:
             raise HTTPException(status_code=429, detail="Too many active live log streams")
 
-    try:
+    def _lookup_container():
         client = docker.from_env()
```
taiberium, line 367: **NIT:** `_lookup_container` opens a fresh `docker.from_env()` with no timeout; reuse the shared 5s client so a wedged daemon can't leak the thread for docker-py's 60s default.

taiberium (approved):

PR #1131, opened 2026-07-10
DAH-2391: remove per-host postgres, store pod logs in a file

arhangel66:
## Problem

Every GPU host ran a full Postgres container for exactly ONE diagnostic table (`pod_log`: container lifecycle events written by `monitor.py`, read by `/pod_logs`). In production nothing calls `/pod_logs` except the validator CLI — yet this Postgres was a node-killing single point of failure:

- `db` had **no restart policy** → once dead it stayed dead (h100micro incident: autoheal restarted it ~60× to SIGKILL during a tenant disk-IO storm).
- `run.sh` gated the whole agent on `alembic upgrade head` under `set -e` → dead DB = **eternal restart loop**, `/ping` and `/upload_ssh_key` never boot, node penalized while the host, GPU and the renter's container are alive.
- `core/db.py` ran `create_all(engine)` at import time → the app could not boot without a reachable DB even without alembic.

## Changes

- **Postgres removed entirely.** `PodLog` events go to a JSONL file on a shared named volume (`pod_logs:/var/lib/executor-pod-logs`), single writer (`monitor.py`), read by `/pod_logs`. Append is best-effort and never raises — a **full or failing disk degrades pod-log collection only, never the agent**. Corrupt/torn lines are skipped on read; a 50 MB cap stops unbounded growth. HTTP response shape is unchanged (verified against the miner's `PodLog.parse_obj`).
- Deleted: `db` service (all compose variants), `core/db.py`, `daos/`, `migrations/`, `alembic.ini`, `docker-compose.local.yml`, `DB_URI` setting; deps `alembic`, `sqlmodel`, `sqlalchemy`, `psycopg2-binary` dropped from `pyproject`/`pdm.lock`. `run.sh` no longer runs alembic.
- **Healthcheck now probes the app, not the GPU**: `curl http://127.0.0.1:${INTERNAL_PORT}/version` on the loopback replaces `nvidia-smi` — a wedged agent goes unhealthy and autoheal restarts it (self-heal), while GPU health stays the validators' job. Conservative thresholds (30s × 5 retries + 120s start_period): ~2.5 min of continuous failure needed, so a host-IO storm can't trigger false restarts.
- **autoheal scoped to the executor only** (default `autoheal=true` label instead of `AUTOHEAL_CONTAINER_LABEL=all`) — it can never again restart other containers to death.

The CVM/dstacktee runner is not affected separately: `Dockerfile.runner` bakes this same `docker-compose.app.yml`, so the topology change ships with the runner image.

## Rollout note

Ship after PR #1130 (event-loop isolation) has baked, and only once backend PR #755 (3-consecutive-miss debounce) is live — a fleet-wide watchtower stack recreation can exceed the single 10s ping window. Historical pod logs in the old `db_data` volume are abandoned in place (ephemeral diagnostics).

## Tests

6 new tests in `tests/test_pod_log_store.py`: SQL-semantics roundtrip (container rows + null-container rows, oldest first), corrupt-line tolerance, missing file, never-raise on ENOSPC, size-cap drop, service integration. `pdm run pytest` — 46 passed. All three compose variants validated with `docker compose config`.


neurons/executor/src/services/pod_log_store.py
```diff
@@ -0,0 +1,63 @@
...
+
+POD_LOG_FILE = Path(os.getenv("POD_LOG_FILE", "/var/lib/executor-pod-logs/pod_logs.jsonl"))
+MAX_FILE_SIZE_BYTES = 50 * 1024 * 1024
+
+
+def append_pod_log(log: PodLog) -> None:
+    # best-effort append that never raises, so a full disk cannot break the caller
+    try:
+        POD_LOG_FILE.parent.mkdir(parents=True, exist_ok=True)
+        if POD_LOG_FILE.exists() and POD_LOG_FILE.stat().st_size > MAX_FILE_SIZE_BYTES:
```
taiberium, line 27: **NIT:** once the file passes 50MB every new event is dropped for good with no rotation, so you lose the freshest logs; truncate to keep the tail instead.

taiberium (approved):

fortunelucky777 (approved):

neurons/executor/src/services/pod_log_store.py
arhangel66: Good catch — fixed in 48748f4d. Past the 50 MB cap the file is now truncated to its newest 25 MB (`_truncate_to_tail`, atomic `os.replace`) and the append proceeds, so the freshest events always survive and it is the oldest ones that age out. The seek starts one byte early and reads to the next newline, so the line the cut lands in is discarded and nothing extra is when it lands on a break. Test `test_append_over_cap_drops_oldest_events_and_keeps_the_newest` covers it.

PR #1132, opened 2026-07-13
DAH-2395, capture container death diagnostics before rm -fv on failed create_container

taiberium:
### Task Link

[DAH-2395](https://app.notion.com/p/Validator-capture-container-death-diagnostics-OOMKilled-ExitCode-logs-before-rm-fv-on-failed-cre-39cb8bfdbde981928277db22b2785e05)

## Why

This morning four Dolphin default jobs died on RTX PRO 6000 nodes about 15 seconds after a successful `docker run`. The only trace left is `failure_reason = "UnknownError: Failed create_container"` in the DB. We still don't know whether the worker got OOM-killed or crashed on startup, and we can't find out: the validator force-removes the dead container right after the failure, and the host belongs to the miner, so there is nowhere to go back and look. Identical hardware runs Dolphin fine on another node. The answer was inside the container we deleted.

While digging into this we also found that the executor-side container monitor has been blind for a long time. It records docker events (die, oom, exit codes) into the local `podlog` table, but still filters names by the legacy `container_` prefix, while the validator names real workloads `pod_*` and `filler_*`. On the staging executor `podlog` has 32,737 rows since March. Every one of them is a port-check probe. Not a single real pod or filler.

## What changed

- Validator: on a failed container creation it now reads `docker inspect .State` (OOMKilled, ExitCode) and the last 100 log lines before running `docker rm -fv`, then writes one structured WARNING to Loki: `Failed container diagnostics before cleanup`. Best effort: if the capture itself fails, cleanup proceeds exactly as before. The success path is untouched.
- Executor monitor: watches `pod_*` and `filler_*` containers instead of `container_*`. Real workload deaths get recorded again, probe noise stops filling the table.

Next time a filler dies like this, the answer is one Loki query away instead of a dead end.

## Tests

TDD: 6 new validator tests and 4 new executor tests. Full suites: validators 1009 passed (the 3 failures in `test_sshd_bootstrap_script.py` are pre-existing on `origin/main` and environment-specific), executor 44 passed. Heads up: CI currently runs validator tests only.

The monitor fix reaches miners with the next executor image release.

### DIFF
--- a/neurons/executor/src/monitor.py
+++ b/neurons/executor/src/monitor.py
@@ -3,7 +3,6 @@ import logging
 import json
 import time
 from datetime import datetime
-from core.config import settings
 from core.db import get_session
 from daos.pod_log import PodLog, PodLogDao
 
@@ -11,7 +10,11 @@ from daos.pod_log import PodLog, PodLogDao
 logging.basicConfig(filename="container_monitor.log", level=logging.INFO,
                     format='%(message)s')  # we will log pre-formatted JSON strings
 
-monitor_prefix = "container_"  # only monitor containers with this prefix
+# Container-name prefixes the validator creates for real workloads: rental pods
+# (pod_*) and idle-node default jobs (filler_*). The legacy "container_" prefix
+# matches only the validator's throwaway port-check probes today, so watching it
+# recorded pure probe noise while every pod/filler death went unlogged (DAH-2395).
+MONITORED_CONTAINER_PREFIXES: tuple[str, ...] = ("pod_", "filler_")
 
 
 # Helper: Determine stop reason classification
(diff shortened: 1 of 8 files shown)


arhangel66:
Really nice catch — this adds exactly the kind of evidence we've been missing for container creation failures.

This logic looks very similar to `_collect_pod_diagnostics` in `checks/rented_machine.py` (from DAH-2193). It captures the same `docker inspect .State` (`ExitCode`, `OOMKilled`, `Error`, `FinishedAt`) and `docker logs --tail` for pods that fail while running (`POD_NOT_RUNNING`, `STALE_POD_NOT_RUNNING`).

**Before merging, could we refactor this into a shared helper function called from both places?** This would give us a consistent field shape in Loki, so create-time and run-time failures are queryable together. This path would also get the `host_context` check for free, which helps distinguish a container failure from a host reboot.

A neutral home like `core/docker_utils.py` (next to `inspect_exit_code`) would be a good fit, as `DockerService` shouldn't import from the `checks` layer. When combining them, we'd just need to reconcile a couple of small things: the log tail length (`50` vs. `100`) and the data shape. The flatter shape from this PR (top-level `OOMKilled`/`ExitCode` instead of nested under `state`) is the better one to keep.

Happy to pair on it if that's easier.

arhangel66 (changes requested):
Please see the comment above.

arhangel66 (approved):
approved, but need to do the followup task

PR #1135, opened 2026-07-13
DAH-2397: bump bittensor 9.10.1 → 10.5.0 (validators + miners)

arhangel66:
## Why

Staging subnet is down: testnet chain runtime (spec **429**, metadata V15) now encodes `netuid` as a `NetUid` newtype (`Composite{u16}`). `bittensor 9.10.1` + `async-substrate-interface 1.6.3` encode it as a bare int → every `metagraph()` / `neurons_lite()` call raises `ValueError: Invalid type for data: 37`. The validator cannot sync the metagraph → no scraping/validation/scoring on staging.

Empirically verified (live probes against both chains): **bittensor 10.5.0 works with BOTH** current mainnet (spec 424, n=256 on netuid 37) and upgraded testnet (spec 429, n=94) — safe to roll to prod ahead of the mainnet runtime upgrade. The 9.x line is unfixable: the encode fix lives only in async-substrate-interface 2.x (cyscale), which conflicts with 9.x's scalecodec.

## What

**Dependencies** (validators — first commit; miners — this PR):
- `bittensor==9.10.1` → `==10.5.0`, `async-substrate-interface>=2.0.4`, `bittensor-cli>=9.22.0`, `bittensor-commit-reveal>=0.4.0`
- `celium-collateral==1.1.6` → `==1.2.0` (Datura-ai/celium-collateral-contracts#8 — drops `scalecodec`, which blocks cyscale import)

**Code migration (all verified against installed 10.5.0 + live testnet):**
- Removed lowercase factories → classes: `bittensor.subtensor`→`Subtensor`, `async_subtensor`→`AsyncSubtensor`, `wallet`→`Wallet`, `axon`→`Axon`, `config`→`Config` (call sites + evaluated type annotations)
- `sync_evm_address_maps`: asi 2.x `query_map` yields decoded records `("0x…", block)` instead of ScaleObj-wrapped nested bytes — probed live on testnet
- `cli_service.transfer`: `dest=` → `destination_ss58=` (renamed in 10.x)

**Verified unchanged (no code changes needed):**
- `query(...).value` — 10.5.0 returns ScaleType with `.value` (probed live; 6 sites)
- `set_weights` return — `ExtrinsicResponse.__iter__` yields `(success, message)`, tuple unpacking compatible
- `process_weights_for_netuid` — same signature
- `serve_axon`, `is_hotkey_registered`, `get_neuron_for_pubkey_and_subnet`, `Balance.from_tao` — kwargs-compatible
- `AsyncSubtensor.initialize()` returns self — miners' pattern survives

## Blocked on

- [ ] Datura-ai/celium-collateral-contracts#8 merged + `1.2.0` published to PyPI
- [ ] `pdm lock` regen in `neurons/validators` and `neurons/miners` (resolves only after 1.2.0 is on PyPI)
- [ ] Unit tests + import smoke with the final env

## Out of scope

`neurons/executor` and `watchtower` — only use `bittensor.Keypair` for offline signature verification, never connect to the chain. Not affected.

## Deploy

Staging first (already down — zero risk), verify `Invalid type for data` errors stop and `Machine specs scraped` resumes, then prod canary before the mainnet runtime upgrade (~1-2 weeks window).


neurons/validators/src/clients/subtensor_client.py
```diff
@@ -150,7 +150,7 @@ def initialize_subtensor(self):
                     extra=get_extra_info(self.default_extra),
                 ),
             )
-            subtensor = bittensor.subtensor(config=self.config)
+            subtensor = bittensor.Subtensor(config=self.config)
```
taiberium, line 153: **NIT:** the old subtensor is never closed after the swap, same leak that Datura-ai/lium-miner-portal#153 fixes.

neurons/validators/src/core/config.py
```diff
@@ -347,46 +347,18 @@ def get_redis_connection_url(self) -> str:
...
 
-    def get_bittensor_config(self) -> bittensor.config:
-        parser = argparse.ArgumentParser()
-        # bittensor.wallet.add_args(parser)
-        # bittensor.subtensor.add_args(parser)
-        # bittensor.axon.add_args(parser)
+    def get_bittensor_config(self) -> bittensor.Config:
+        # bittensor >=10.3.2 ignores argparse defaults (BT_NO_PARSE_CLI_ARGS is on by
+        # default), so assign values directly or Subtensor silently falls back to finney
+        config = bittensor.Config(argparse.ArgumentParser())
```
taiberium: **NIT:** the empty parser is ignored since CLI parsing is off by default, so `bittensor.Config()` is enough.

neurons/miners/src/core/config.py
```diff
@@ -79,54 +79,26 @@ class Settings(BaseSettings):
...
 
-    def get_bittensor_config(self) -> bittensor.config:
-        parser = argparse.ArgumentParser()
-        # bittensor.wallet.add_args(parser)
-        # bittensor.subtensor.add_args(parser)
-        # bittensor.axon.add_args(parser)
+    def get_bittensor_config(self) -> bittensor.Config:
+        # bittensor >=10.3.2 ignores argparse defaults (BT_NO_PARSE_CLI_ARGS is on by
+        # default), so assign values directly or Subtensor silently falls back to finney
+        config = bittensor.Config(argparse.ArgumentParser())
```
taiberium: **NIT:** same as the validators config, `bittensor.Config()` works without the throwaway parser.

taiberium (approved):

fortunelucky777 (approved):

PR #1136, opened 2026-07-14
DAH-2356: Lium PEARL GPU power cap (70%, min-clamped) — apply, validator exemption, pre-cap restore (executor)

taiberium:
Replaces #1115. The repo allows rebase-merge only, and the old branch carried a merge commit from main that GitHub refuses to rebase, so this is the same change flattened into one linear commit. The tree is byte-identical to the reviewed and staging-tested head of #1115 (e80c7f37), verified with an empty diff.

What's inside (validator side of DAH-2356, pairs with Datura-ai/lium-io-backend#742):
- apply the backend-computed 70% PEARL cap host-side with nvidia-smi, persistence mode + readback verify, fail-closed for the filler
- GpuPowerLimitCheck exemption while a Lium filler runs, stale-cap rescue via frozen Redis records
- restore the exact pre-cap limit on filler teardown
- Redis read failure passes the check for the cycle (GPU_POWER_LIMIT_STATE_UNAVAILABLE) instead of zero-scoring an innocent miner
- state-free net at rental start: any GPU below 90% of default is raised back to default, so a customer never inherits a cap even with all records lost

Review discussion and staging test evidence (cap 140->100, restore, skip-check, raise 217->310, penalty) live in #1115.


taiberium:
Closing: work continues on the original approved PR #1115.

PR #1138, opened 2026-07-15
DAH-2420, include failing step's error in create_container failure message

taiberium:
### Task Link

[DAH-2420](https://app.notion.com/p/Default-Job-filler-create-failures-hide-the-real-cause-capture-failure_step-error-text-39eb8bfdbde9817d85a6f09cb564f7a3)

---

## Problem

`create_container` failures returned only the message `"Failed create_container"`, so the backend stored a bare `"UnknownError: Failed create_container"` with no cause. The failing step's real error (e.g. `set_environment` → `409 container is not running`, i.e. the workload exited at boot) was written only to the exc_info log line and never made it into the stored reason / alerts.

## Solution

Append the failing step's (capped) exception text to the returned `FailedContainerRequest.msg`. `failure_step` already rides the response.

## Changes

- `create_container` except block: `failure_msg = f"{message}: {error_text[:500]}"` for every failure (subsumes the old `docker_sdk_ssh_host_key`-only special case).
- Test that a create failure carries `failure_step` + the underlying error in `msg`.

## Deployment Steps

Use GitHub Action.

## Review Request

- [x] Just code review

## Other PRs

Backend counterpart: [lium-io-backend#766](https://github.com/Datura-ai/lium-io-backend/pull/766) — surfaces `failure_step` in the filler's stored reason + alert.

## Risks

- Low. Only the text of the failure message changes (an error suffix is appended). No flow/behaviour change; secrets aren't in exception strings for these paths. Full docker_service suite green (144).

### Test Case 1
**Actions**: create_container fails at the `set_environment` step.
**Expected Output**: `FailedContainerRequest.failure_step == "set_environment"` and `msg` contains the underlying error text.

## Checklist

- [x] Risks are documented
- [x] Tests are described

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -3593,12 +3593,11 @@ class DockerService:
             await self.redis_service.remove_pending_pod(payload.miner_hotkey, payload.executor_id, payload.pod_id)
 
             # Port release now handled by backend
-            failure_msg = str(log_text)
-            if (
-                current_step == "docker_sdk_ssh_host_key"
-                and isinstance(e, RentalDockerConnectionError)
-            ):
-                failure_msg = f"{failure_msg}: {e}"
+            # Carry the failing step's exception text into the returned message so the backend stores
+            # the real cause instead of a bare "Failed create_container" (capped to keep it small; the
+            # full traceback stays in the exc_info log above). failure_step already rides the response.
+            error_text = (str(e) or e.__class__.__name__)[:500]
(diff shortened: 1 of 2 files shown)


taiberium:
Closing. Digging into the real failures showed the container gets force-removed on the host seconds after it starts (exit 137, not OOM), and that death is already captured by the DAH-2395 diagnostics log. failure_step here just points at the symptom (set_environment), so it adds little for this case. Can revisit if we want the same detail for the early pre-container steps (pull/gpu/ssh).

PR #1139, opened 2026-07-15
DAH-2364: reduced graceful stop window for fillers instead of skipping

arhangel66:
## Describe your changes

Follow-up to #1121. Instead of **skipping** the graceful `docker stop` for FILLER workloads entirely (the SIGKILL-only path), fillers now get a graceful stop too, but with a **shorter grace window** — `FILLER_CONTAINER_STOP_GRACE_SECONDS = 15`, half of the customer `CONTAINER_STOP_GRACE_SECONDS = 30`.

### Why

The reason fillers were excluded was the preemption race: the backend preempts a filler before a customer rent with a total budget of `FILLER_STOP_WAIT_TIMEOUT_SECONDS = 30` (`compute-app`), polls the filler's stopped-status every 0.25s, and **starts the rent anyway on timeout**. A grace window equal to that whole budget would let a SIGTERM-ignoring filler burn all 30s inside `docker stop`, so the customer pod could land on GPUs the filler still holds.

A grace window **strictly below** that budget removes the race while still giving well-behaved fillers a clean SIGTERM exit — which is the whole point of #1121: a clean exit avoids the containerd/sysbox wedge that leaves orphaned containers holding GPUs.

### Timeline with the 15s window

| Case | Timeline | Result |
|------|----------|--------|
| Filler **respects** SIGTERM (typical sglang/PEARL/miner) | exits cleanly in 1–3s → forced removal is instant → `STOPPED` ~t=3, backend sees it within 0.25s | GPU free well before 30s; **no containerd/sysbox wedge** |
| Filler **ignores** SIGTERM (rare) | SIGKILL at t≈15 → forced removal ~t=16 → `STOPPED` ~t=17 | GPU free by ~17s, ~13s of headroom inside the 30s budget, no collision |

The one double-rare case — a wedged node **and** a SIGTERM-ignoring filler — is no worse than today (the pre-#1121 SIGKILL-only path already failed there), and the common SIGTERM-respecting filler on a wedged node is now actively helped.

`docker-py` already extends the HTTP read timeout by the stop timeout (from #1121), so the 15s grace cannot cause a read-timeout.

### Cross-repo invariant

`FILLER_CONTAINER_STOP_GRACE_SECONDS` (validator) must stay strictly below `FILLER_STOP_WAIT_TIMEOUT_SECONDS` (compute-app backend) with room for the forced removal and the stopped-callback. Documented in a comment next to the constant.

[DAH-2364 Graceful container teardown + escalation and alerting for failed pod deletions](https://app.notion.com/p/Graceful-container-teardown-escalation-and-alerting-for-failed-pod-deletions-397b8bfdbde9813eb8f6e27119b95912)

- [x] I wrote tests (rewrote `test_delete_filler_skips_graceful_stop` → `test_delete_filler_stops_with_reduced_grace`; full `test_docker_service.py` + `test_rental_docker_sdk.py` = 170 passed)
- [x] Need to take care of performance? (adds up to 15s to filler teardown only when the filler ignores SIGTERM; deletion is async fire-and-forget on the WS path)

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -121,6 +121,15 @@ IN_CONTAINER_SSH_BOOTSTRAP_PATH = "/tmp/lium-ssh-bootstrap.sh"
 # that hold GPUs and brick the executor.
 CONTAINER_STOP_GRACE_SECONDS = 30
 
+# Fillers get a shorter grace window than customer workloads. The backend preempts a filler
+# before a customer rent with a total budget of FILLER_STOP_WAIT_TIMEOUT_SECONDS (30s in
+# compute-app) and starts the rent anyway on timeout. This grace must stay strictly below that
+# budget with room for the forced removal and the stopped-callback, so a SIGTERM-ignoring filler
+# can never burn the whole budget inside docker stop and hold the GPUs into the customer rent.
+# Half the budget leaves ~15s of headroom while still letting a well-behaved filler exit cleanly
+# and avoid the containerd/sysbox wedge. Keep in sync with compute-app FILLER_STOP_WAIT_TIMEOUT_SECONDS.
+FILLER_CONTAINER_STOP_GRACE_SECONDS = 15
+
 DOCKER_VOLUME_PLUGINS = {
     "s3fs": "mochoa/s3fs-volume-plugin"
 }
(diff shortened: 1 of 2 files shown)


taiberium (approved):

taiberium:
NIT: `filler_no_grace` (test_docker_service.py:1250) is a stale name now that fillers get a grace window. Rename it, e.g. `filler_reduced_grace`.

PR #1140, opened 2026-07-15
DAH-2427, ghost GPUs: signature-based wedge detection + targeted reset in GpuUsageCheck; verifyx timeout + context release; teardown wedge sweep

taiberium:
## Describe your changes

Fixes the "ghost GPU" blind spot: a hard-killed GPU process leaves an orphaned CUDA kernel that pins the card at 100% utilization with 0 MiB and no process — and today the validator neither notices nor heals it, so the card keeps being offered to renters for days (postmortem, Jul 15: GPU 1 on 4db213a4 / 204.9.206.20 sat pinned for 9 days; GPUs 4-5 on ee0cdd5f / 204.9.206.18 still pinned; plus a 6-card wave on 6e062ec9 after a DPHN teardown).

Root cause chain (three links, this PR closes the first two):

- **(a) Source** — an abruptly killed GPU workload (failed PEARL create force-removed via `docker rm -fv`, a rental-stop force-remove, a DPHN teardown, or a hung verifyx run) leaves its kernel behind. Confirmed via `filler_run` 00ded466-af1a-4648-84af-80ebb782c87b + 15-min GPU snapshots + connector logs.
- **(b) Survival** — the driver keeps the kernel resident forever; `verifyx_executor.py` releases its native CUDA context only in `__del__`, and `gpus_utility.py` holds an NVML handle that makes `nvidia-smi -r` refuse with "in use by another client".
- **(c) Blindness** — `GpuUsageCheck._find_violation` returns `None` on an empty `gpu_processes` list, so a ghost's 100% utilization is recorded every cycle and never compared against the 5% limit.

### Layer A — detection + recovery (`gpu_usage.py`, `gpu_wedge.py`)

- `GpuUsageCheck` no longer requires processes to flag a card. Stateless (review feedback): the signature (utilization >= 95%, memory <= 1%, empty process list) triggers the cure immediately — a clean CUDA context create/destroy cycle per wedged UUID (shared in `services/gpu_wedge.py`) — and the verdict comes from re-sampling the live card: clean -> the check passes and the node stays online (`GPU_WEDGE_CURED`); still latched -> fatal fail, node leaves the pool (`GPU_WEDGED`). No Redis, no timers, no cross-cycle state. The context cycle replaced the original `pkill` + `nvidia-smi -r` after review evidence showed `-r` is refused on ~45% of the Blackwell fleet and the pkill only worked against a leaked NVML fd; the cure needs no root, tolerates concurrent NVML clients, and cleared 11/11 latched cards in live tests. Source-agnostic — catches filler, rental and verifyx orphans alike.
- The dry-run pipeline builds the check with `recover_wedged_gpus=False`: it detects and reports but never mutates the executor.

### Layer B — stop leaving kernels behind

- **verifyx**: the validator-side hard timeout landed on main first via DAH-2365 (`VERIFYX_COMMAND_TIMEOUT_SECONDS = 600`, native asyncssh `timeout=`) — this branch merged it and dropped its own duplicate. What this PR adds on top: `verifyx_executor.py` releases the native context in `try/finally` and on SIGTERM/SIGINT/SIGHUP (channel close on timeout delivers SIGHUP), instead of relying on `__del__`.
- **filler teardown**: `delete_container` runs a best-effort wedged-GPU sweep after the forced removal — on the success path and on the failed-remove path (what backend records as FAILED / STOP_FAILED). The sweep no-ops while any live compute app exists on the host.

Layer C from the postmortem (hardening `gpus_utility.py` timeouts/watchdog + hung-monitor restart in `start_gpu_monitor.py`) is follow-up scope tracked in the ticket, not in this PR.

Recovery runbook change: no reboot, no miner, no root needed — a CUDA context open/close on the latched card clears the counter (`CUDA_VISIBLE_DEVICES=<uuid> python3` + cuInit/cuCtxCreate/cuCtxDestroy). Also hardened: all executor-side NVML init/shutdown blocks are now try/finally, so a scrape exception can no longer leak the fd that blocks per-GPU maintenance.

[DAH-2427 Ghost GPUs: detect and auto-recover orphaned CUDA kernels](https://app.notion.com/p/39eb8bfdbde98129b854d47d079d0ce9)

- [x] I wrote tests.
- [x] Need to take care of performance? (detection adds one Redis hash read/write per idle cycle; the teardown sweep adds ~5s settle + two parallel nvidia-smi queries per filler delete)

### DIFF
--- a/neurons/validators/src/services/task/checks/gpu_usage.py
+++ b/neurons/validators/src/services/task/checks/gpu_usage.py
@@ -1,6 +1,31 @@
-from ..messages import GpuUsageMessages as Msg, render_message
+import logging
+from dataclasses import dataclass
+
+from services.const import (
+    GPU_MEMORY_UTILIZATION_LIMIT,
+    GPU_UTILIZATION_LIMIT,
+    GPU_WEDGE_MEMORY_MAX,
+    POD_CONTAINER_PREFIX,
+)
+from services.gpu_wedge import (
+    GpuCureOutcome,
+    cure_wedged_gpus,
+    matches_wedge_utilization,
+    query_wedged_gpu_uuids,
+)
+
+from ..messages import GpuUsageMessages as Msg
+from ..messages import render_message
 from ..pipeline import CheckResult, Context
-from services.const import GPU_MEMORY_UTILIZATION_LIMIT, GPU_UTILIZATION_LIMIT, POD_CONTAINER_PREFIX
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class WedgedGpu:
+    gpu_uuid: str
+    gpu_utilization: float | None
+    memory_utilization: float | None
 
 
 class GpuUsageCheck:
@@ -9,10 +34,24 @@ class GpuUsageCheck:
     check_id = "gpu.validate.usage"
     fatal = True
 
+    def __init__(self, dry_run: bool = False):
+        # A dry run reports a ghost GPU but must not touch the executor (no CUDA-context cure).
+        self.dry_run = dry_run
+
     async def run(self, ctx: Context) -> CheckResult:
         gpu_details = ctx.state.gpu_details
         gpu_processes = ctx.state.gpu_processes
 
+        ghost_result = await self._detect_and_cure_ghost_gpus(ctx, gpu_details, gpu_processes)
+        if ghost_result is not None:
+            return ghost_result
+
+        return self._judge_process_based_usage(ctx, gpu_details, gpu_processes)
+
+    def _judge_process_based_usage(
+        self, ctx: Context, gpu_details: list[dict], gpu_processes: list[dict]
+    ) -> CheckResult:
+        """The legacy guard: flag GPU load owned by live processes outside our workloads."""
         violation = _find_violation(gpu_details, gpu_processes)
 
         if violation is None:
@@ -67,6 +106,72 @@ class GpuUsageCheck:
         )
         return CheckResult(passed=False, event=event)
 
+    async def _detect_and_cure_ghost_gpus(
+        self, ctx: Context, gpu_details: list[dict], gpu_processes: list[dict]
+    ) -> CheckResult | None:
+        """Cure ghost GPUs in place; fail the executor only when a card stays latched.
+
+        DAH-2427: a hard-killed GPU process can latch the busy counter — 100% utilization
+        with no memory and no process — which the process-based guard cannot see. Stateless
+        by design: the cure (a CUDA context cycle) is harmless, so it runs on first sight,
+        and the verdict comes from re-sampling the live card afterwards. Fail-open: this
+        detection is an addition to the legacy guard, so an unexpected error here must never
+        break validation for the whole executor. Returns None when there is no ghost.
+        """
+        try:
+            if gpu_processes:
+                return None
+
+            wedged_gpus: list[WedgedGpu] = [
+                WedgedGpu(
+                    gpu_uuid=detail["uuid"],
+                    gpu_utilization=detail.get("gpu_utilization"),
+                    memory_utilization=detail.get("memory_utilization"),
+                )
+                for detail in gpu_details
+                if _is_wedge_candidate(detail)
+            ]
+            if not wedged_gpus:
+                return None
+
+            if self.dry_run:
+                event = render_message(
+                    Msg.GPU_WEDGED,
+                    ctx=ctx,
+                    check_id=self.check_id,
+                    what={
+                        "wedged_gpus": wedged_gpus,
+                        "cure_attempted": False,
+                        "cure_results": [],
+                    },
+                )
+                return CheckResult(passed=False, event=event)
+
+            wedged_uuids: list[str] = [gpu.gpu_uuid for gpu in wedged_gpus]
+            cure_outcomes: list[GpuCureOutcome] = await cure_wedged_gpus(ctx.runner, wedged_uuids)
+
+            # The verdict comes from the card itself, not from the cure's exit code.
+            still_wedged: list[str] = sorted(
+                set(await query_wedged_gpu_uuids(ctx.runner)) & set(wedged_uuids)
+            )
+            what = {
+                "wedged_gpus": wedged_gpus,
+                "cure_attempted": True,
+                "cure_results": cure_outcomes,
+                "still_wedged": still_wedged,
+            }
+            if not still_wedged:
+                event = render_message(
+                    Msg.GPU_WEDGE_CURED, ctx=ctx, check_id=self.check_id, what=what
+                )
+                return CheckResult(passed=True, event=event)
+
+            event = render_message(Msg.GPU_WEDGED, ctx=ctx, check_id=self.check_id, what=what)
+            return CheckResult(passed=False, event=event)
+        except Exception as exc:  # noqa: BLE001
+            logger.warning("Ghost GPU detection failed, falling back to the usage guard: %s", exc)
+            return None
+
 
 def _find_violation(gpu_details: list[dict], gpu_processes: list[dict]) -> dict | None:
     if not gpu_processes:
@@ -85,3 +190,14 @@ def _find_violation(gpu_details: list[dict], gpu_processes: list[dict]) -> dict 
             }
 
     return None
+
+
+def _is_wedge_candidate(detail: dict) -> bool:
+    gpu_utilization = detail.get("gpu_utilization")
+    memory_utilization: float = detail.get("memory_utilization", 0) or 0
+    has_uuid: bool = bool(detail.get("uuid"))
+    return (
+        has_uuid
+        and matches_wedge_utilization(gpu_utilization)
+        and memory_utilization <= GPU_WEDGE_MEMORY_MAX
+    )
--- a/neurons/executor/src/gpus_utility.py
+++ b/neurons/executor/src/gpus_utility.py
@@ -165,12 +165,16 @@ async def manage_docker_images(
     driver_version = "unknown"
     try:
         pynvml.nvmlInit()
-        handle = pynvml.nvmlDeviceGetHandleByIndex(0)  # Get first GPU
-        gpu_name = pynvml.nvmlDeviceGetName(handle)
-        driver_version = pynvml.nvmlSystemGetDriverVersion()
-        if isinstance(gpu_name, bytes):
-            gpu_name = gpu_name.decode("utf-8")
-        pynvml.nvmlShutdown()
+        # finally: a leaked NVML handle makes this process the "in use by another client"
+        # blocker for any later per-GPU maintenance (DAH-2427 review finding).
+        try:
+            handle = pynvml.nvmlDeviceGetHandleByIndex(0)  # Get first GPU
+            gpu_name = pynvml.nvmlDeviceGetName(handle)
+            driver_version = pynvml.nvmlSystemGetDriverVersion()
+            if isinstance(gpu_name, bytes):
+                gpu_name = gpu_name.decode("utf-8")
+        finally:
+            pynvml.nvmlShutdown()
     except Exception as e:
         logger.error(f"Failed to get GPU name: {e}")
 
--- a/neurons/executor/src/services/hardware_service.py
+++ b/neurons/executor/src/services/hardware_service.py
@@ -142,17 +142,21 @@ def get_system_metrics():
     gpus = []
     try:
         pynvml.nvmlInit()
-        gpu_count = pynvml.nvmlDeviceGetCount()
-        
-        for i in range(gpu_count):
-            handle = pynvml.nvmlDeviceGetHandleByIndex(i)
-            util = pynvml.nvmlDeviceGetUtilizationRates(handle)
-            mem = pynvml.nvmlDeviceGetMemoryInfo(handle)
-            gpus.append({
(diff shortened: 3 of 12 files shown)


arhangel66:
# Review: The `nvidia-smi -r` approach is fragile; proposing a more robust context-based fix

Thanks for this PR. Let's start with a shared understanding of the bug, based on our research in `untracked/ghost_gpu/RESEARCH_JOURNAL.md`. The "ghost GPU" state is a **Blackwell GSP "engine busy" counter latch**. It's triggered when a process is killed while a CUDA kernel is running (e.g., `docker rm -f` on a working pod, which causes this state in ~86% of attempts). This is a telemetry-only issue: the card still functions at full speed, produces no Xid errors, and holds no memory or processes. The state persists only because persistence mode keeps the device initialized.

## Analysis of the `nvidia-smi -r` approach

This PR attempts to recover wedged GPUs by running these two commands:

1. `pkill -f gpus_utility.py`
2. `nvidia-smi -i <GPU-UUID> -r`

This sequence was observed to work once during a manual session (on 2026-07-15, host 204.9.206.20, driver 580.159.03). However, our investigation shows its success depends on several conditions that are not reliably met across our fleet.

### Condition 1: The driver must support the reset command

On driver version **580.126.09**, the reset command is unavailable even on an idle card with no open file descriptors (`fuser -v /dev/nvidia*` was empty). We confirmed this across multiple configurations on two Blackwell test machines (RTX PRO 6000 SE):

| config | card state | `nvidia-smi -i <uuid> -r` |
|---|---|---|
| pm1 + persistenced daemon | latched | `In use by another client` |
| pm1 flag, no daemon (= prod shape) | latched | `In use by another client` |
| pm1 flag, no daemon | clean | `In use by another client` |
| pm0, no daemon | clean | `In use by another client` |

Based on a 24-hour sample, ~96 Blackwell executors run driver 580.159.03, while ~77 run 580.126.09. This means the `nvidia-smi -r` command will fail on **~45% of the Blackwell fleet** due to the driver version alone.

### Condition 2: `pkill` must target the process holding the NVML handle

The `pkill` command is only effective if `gpus_utility.py` is holding a leaked NVML handle. In its current steady state, it does not: the scrape loop is commented out (`gpus_utility.py:304-310`) and `manage_docker_images` performs a clean `nvmlInit` / `nvmlShutdown` cycle at startup (`gpus_utility.py:166-175`).

The reason the `pkill` was effective during the manual recovery session was likely a **file descriptor leak**. The `nvmlInit`/`nvmlShutdown` block lacks a `finally` clause. If an exception occurs between these calls, the process will permanently hold file descriptors for all GPUs on the host, becoming the "another client" that blocks the reset.

This same leak pattern exists in `hardware_service.py:144-156` and `:261-289`, which are part of the main `executor.py` process. We cannot kill the executor, so if the leak occurs there, this PR's recovery method is blocked.

You can check any live executor for a leaked holder with:
```bash
P=$(pgrep -f gpus_utility.py); ls -l /proc/$P/fd | grep -c nvidia   # >0 = leaked
```

### Condition 3: Nothing else may hold or touch the device

Two separate blockers here:

- **Transient clients racing the reset**: `machine_scrape` and the hardware-utilization polls open NVML handles for seconds at a time. If the reset runs at the same moment, it fails with the same `In use by another client`.
- **Miner-run `nvidia-persistenced` daemons**: miners control their hosts, and a host that runs the persistenced daemon (instead of the `-pm 1` flag) has a *permanent* fd holder — the reset is blocked there regardless of timing.

Furthermore, even when successful, killing `gpus_utility.py` has the side effect of stopping the image pre-pull daemon until the next validation cycle restarts it.

## Recommended Solution: Clean CUDA Context Cycle

A more reliable and less disruptive solution is to trigger a clean CUDA context initialization and destruction on the affected GPU. This method clears the latched counter without requiring root, killing processes, or issuing a device reset — and it does not care what other clients hold the device.

This approach was effective in **11/11 tests** across two Blackwell machines, including on driver 580.126.09 where `nvidia-smi -r` is impossible. The recovery takes 2–5 seconds per GPU and its dependencies (`python3`, `libcuda.so.1`) are already present in the executor container (`NVIDIA_DRIVER_CAPABILITIES=all`).

The recovery logic would execute the following for each wedged GPU UUID, instead of the `pkill` and `nvidia-smi` commands:

```bash
CUDA_VISIBLE_DEVICES=<GPU-UUID> python3 -c "
import ctypes
lib = ctypes.CDLL('libcuda.so.1')
assert lib.cuInit(0) == 0
dev = ctypes.c_int(); assert lib.cuDeviceGet(ctypes.byref(dev), 0) == 0
ctx = ctypes.c_void_p(); assert lib.cuCtxCreate_v2(ctypes.byref(ctx), 0, dev) == 0
assert lib.cuCtxDestroy_v2(ctx) == 0
print('ctx open/close OK')
"
```

For manual ops on an unrented box, this loop sweeps every GPU on a host (the PR integration should keep targeting only the wedged UUIDs):

```bash
for uuid in $(nvidia-smi --query-gpu=uuid --format=csv,noheader); do
  echo "== $uuid"
  CUDA_VISIBLE_DEVICES=$uuid python3 -c "
import ctypes
lib = ctypes.CDLL('libcuda.so.1')
assert lib.cuInit(0) == 0
dev = ctypes.c_int(); assert lib.cuDeviceGet(ctypes.byref(dev), 0) == 0
ctx = ctypes.c_void_p(); assert lib.cuCtxCreate_v2(ctypes.byref(ctx), 0, dev) == 0
assert lib.cuCtxDestroy_v2(ctx) == 0
print('ctx open/close OK')
" || echo "FAILED: $uuid"
done
```

## Additional Recommendation: Prevent NVML Handle Leaks

Regardless of which recovery method we implement, we should make our NVML initialization calls exception-safe. The file descriptor leak is the likely root cause of the "another client" error that made `pkill` necessary in the manual recovery session.

Please wrap the NVML blocks in `try...finally` to ensure `nvmlShutdown()` is always called. This will prevent future `In use by another client` errors caused by this leak. The affected locations are:

- `neurons/executor/src/gpus_utility.py:166-175`
- `neurons/executor/src/services/hardware_service.py:144-156`
- `neurons/executor/src/services/hardware_service.py:261-289`

---

<details>
<summary><b>Full verification runbook</b> — every step reproduced live on a Blackwell box (RTX PRO 6000 SE, 580.126.09), real outputs quoted</summary>

# Ghost GPU: hands-on verification on shadeform@216.243.220.75

Machine: 1× RTX PRO 6000 Blackwell SE, driver 580.126.09, persistence mode ON —
exactly the configuration where the latch sticks.

Scripts used: `untracked/ghost_gpu/ghost_burn.py` (endless ~0.4 s CUDA kernel
launches, 512 MiB) and `untracked/ghost_gpu/gpu_ghost_cure.py` (self-contained
detector + cure, stdlib + ctypes, PTX JIT — no nvcc, no pynvml).

## Part 1 — create a ghost GPU

1. Baseline: `nvidia-smi --query-gpu=utilization.gpu,memory.used,power.draw,pstate --format=csv,noheader` → `0 %, 0 MiB, ~30 W, P8`
2. Start the workload: `cd ~/ghost_test && nohup python3 ghost_burn.py > burn.log 2>&1 &`
3. Confirm it burns: `100 %, 1071 MiB, ~400 W` — a kernel is genuinely in flight.
4. Kill mid-kernel (the trigger; same thing `docker rm -f` does): `pkill -9 -f "ghost_burn[.]py"`
5. Ghost signature observed: `100 %, 0 MiB, ~97 W, P0` and **zero** compute apps.
   Hit rate ~86% per kill; if the card reads 0%, the kill landed between kernels — repeat 2–4.

## Part 2 — the PR's recovery, replayed verbatim

1. Who holds the device: `sudo fuser -v /dev/nvidia*` → only `nvidia-persistenced` (this box runs the daemon; prod uses the pm flag).
2. `pkill -f "gpus_utility[.]py"` → `exit=1` (nothing to kill; steady-state it holds no NVML handle anyway).
3. `sudo nvidia-smi -i GPU-dffc6d14-31fb-774c-0f43-1355257782fa -r` →

```
The following GPUs could not be reset:
  GPU 00000000:01:00.0: In use by another client
reset exit=255
```

4. Still `100 %, 0 MiB` — ghost survived.
5. Reset matrix on this driver build (all with zero fd holders): pm1+daemon / pm1 flag / pm0, card latched / clean → **always** `In use by another client`. On 580.126.09 the reset is never available. (On 580.159.03 the manual prod session showed it working once the leaked fd holder was killed.)

## Part 3 — the context cure

1. `cd ~/ghost_test && time python3 gpu_ghost_cure.py` →

```
INFO ghost GPU detected: GPU-dffc6d14-31fb-774c-0f43-1355257782fa — running cure (mode=kernel)
INFO cure result for GPU-dffc6d14-31fb-774c-0f43-1355257782fa: CLEARED
real  0m5.3s
```

2. State after: `0 %, 0 MiB, 30 W, P8` — utilization cleared AND the card dropped to true idle (a latched card sits at ~97 W / P0).
3. The bare one-liner (ctx open/close only, no kernel) was verified separately on a fresh ghost: `ctx open/close OK` → `0 %, 0 MiB, 31 W, P8` three seconds later.
4. Cumulative cure reliability across all experiments: **11/11**.

</details>

neurons/validators/src/services/task/checks/gpu_usage.py
```diff
@@ -9,10 +41,25 @@ class GpuUsageCheck:
...
+        # reset) nor the shared wedge timers the production pipeline relies on.
+        self.dry_run = dry_run
+
     async def run(self, ctx: Context) -> CheckResult:
         gpu_details = ctx.state.gpu_details
         gpu_processes = ctx.state.gpu_processes
 
+        ghost_event = await self._detect_and_recover_ghost_gpus(ctx, gpu_details, gpu_processes)
+        if ghost_event is not None:
+            return CheckResult(passed=False, event=ghost_event)
```
arhangel66: can we skip fatal error in case the recovery was successfull?

arhangel66: Also, if we can skip and don't use redis - lets don't use it. 

Just a direct, stupid, and stateless check:
noticed bad state -> try to fix
failed to fix? -> exclude node.

arhangel66 (changes requested):
look to comments https://github.com/Datura-ai/lium-io/pull/1140#issuecomment-4992531432

taiberium:
Great research — you're right on all three counts. Our prod success with `pkill` + `nvidia-smi -r` was luck standing on a leaked fd and one specific driver build, not a mechanism.

Adopted in 26e1c5ba:
- recovery is now the CUDA context cycle you proposed, built per wedged UUID (`CUDA_VISIBLE_DEVICES=<uuid> python3 -c "cuInit/cuCtxCreate/cuCtxDestroy"`), success judged by the `ctx open/close OK` marker. `pkill` and `nvidia-smi -r` are gone from the recovery path entirely, so the pre-pull daemon side effect is gone too.
- all three NVML blocks (`gpus_utility.py`, both spots in `hardware_service.py`) now shut down in `try/finally`, so a scrape exception can't leave the process as the "another client" blocker.
- the GPU_WEDGED remediation text now points operators at the context cycle instead of `-r`.

Verified the exact command our code builds end-to-end on a live box over the validator SSH path (quoting, libcuda resolution, exit 0 + marker). Your 11/11 on latched Blackwell cards stands as the efficacy evidence — the two still-latched cards on 204.9.206.18 will be the live prod confirmation after release.

taiberium:
Both taken, in ef44bb52. You're right that the sustain window guarded a destructive reset that no longer exists — with the harmless context cycle the whole timer lost its reason.

New flow, exactly your shape: signature seen (util >= 95, ~0 mem, no processes) -> cure immediately -> re-sample the card with live nvidia-smi -> clean means the check PASSES and the node stays online (`GPU_WEDGE_CURED` event); still latched means fail and the node leaves the pool (`GPU_WEDGED`). The verdict comes from the card, not from the cure's exit code. Redis is gone from the check entirely — no keys, no timers, no cross-cycle state.

One side effect worth naming: a workload caught mid-exit now gets a pointless-but-harmless 2-5s context cycle instead of a 10-minute grace. Seemed like the right trade.

taiberium:
Reproduced a real ghost and cured it with the shipped code — closing the last untested gap.

Box: fresh RTX PRO 6000 Blackwell SE on Shadeform (216.243.220.86), driver **580.126.09** (the branch where `nvidia-smi -r` never works), persistence ON, full executor stack + a live Dolphin filler.

- Repro #1: `docker rm -f` on the working filler — landed between kernels, card released cleanly (the ~86% is per mid-kernel kill, not per attempt).
- Repro #2: burn loop of ~0.4s CUDA kernels (ctypes + PTX, global-store so ptxas can't elide it) + `kill -9` mid-kernel → latched on the first try: **100% / 0 MiB / ~99W / P0 / zero processes**, stable.
- Then the production path end to end, over `SSHCommandRunner`: `query_wedged_gpu_uuids` detected the latch → `cure_wedged_gpus` ran the context cycle (`ctx open/close OK`) → re-query clean → card back to 0% / 0 MiB.
- Re-latched and ran the full `GpuUsageCheck.run` against the live box: `passed=True`, `GPU_WEDGE_CURED`, `still_wedged=[]` — the node stays online, exactly the reviewed behavior.

Repro scripts live on the box in `~/ghost_test/` (burn.py + repro_loop.sh) if anyone wants to replay it.

arhangel66 (approved):
looks good

PR #1141, opened 2026-07-15
DAH-2419: capture worker log when a Lium filler is up but not mining (non-fatal diagnostic)

taiberium:
## Problem

A Lium filler can be RUNNING while the GPU shows it isn't mining — a dead worker, or a firmware-wedged card the worker no longer holds (100% util, 0 MiB, no process). The **why** lives in the worker's own logs, which run in a container on the miner's host and never reach our Loki. So today we can't see the in-container failure (OOM / Triton-JIT / bad model download).

## What this adds

A new **non-fatal** pipeline check `FillerMiningHealthCheck`: when a node has a Lium filler but no GPU process belongs to that filler, it snaps the worker's `docker logs` over the already-open SSH session and logs them to Loki (`Filler running but not mining — worker diagnostics`, flat fields via `to_log_fields()`).

Reuses `collect_container_death_diagnostics` (DAH-2395) — it reads `docker inspect` + logs tail; on a **live** wedge container that tail is the worker's recent output.

## Safety (strictly diagnostic)

- `fatal = False`, `passed` **always True** — never fails a check, never changes a miner's score.
- SSH log capture is **rate-limited to once per hour per executor** (a wedge persists for hours; we must not SSH every cycle).
- Any capture/redis error folds to "skip", never raises into the pipeline.
- No-ops when there's no filler (rented nodes) — returns before any SSH.

## Scope / limitation

Detects the strategy-agnostic cases: **dead worker** and **firmware wedge** (filler present but not on the GPU). It does **not** catch a partial-load DPHN (vLLM present but only ~14 GB of the ~70 GB model loaded) — that needs the filler's strategy, which the validator doesn't receive today (`filler_containers_by_executor` carries only the container name). Follow-up: add strategy to that map to enable the per-strategy VRAM threshold.

## Relation

Same ticket as the backend self-heal (lium-io-backend#767): the backend classifies + acts on not-mining fillers; this validator side captures the *cause* from the worker log. Complements the RTX PRO 6000 firmware-wedge work (v1.238 driver gate) by surfacing the in-container failures the gate doesn't explain.

## Tests

`test_filler_mining_health_check.py` — non-fatal, no-filler no-op, filler-on-GPU OK (no capture), not-on-GPU captures, rate-limit suppresses the second capture, diagnostics failure still passes. 6 pass (+ neighbouring pipeline tests green).

### DIFF
--- a/neurons/validators/src/services/task/checks/__init__.py
+++ b/neurons/validators/src/services/task/checks/__init__.py
@@ -4,6 +4,7 @@ from .capability import CapabilityCheck
 from .collateral import CollateralCheck
 from .custom_build_orphan_sweep import CustomBuildOrphanSweepCheck
 from .duplicate_executor import DuplicateExecutorCheck
+from .filler_mining_health import FillerMiningHealthCheck
 from .finalize import FinalizeCheck
 from .gpu_count import GpuCountCheck
 from .gpu_fingerprint import GpuFingerprintCheck
@@ -34,6 +35,7 @@ __all__ = [
     "CollateralCheck",
     "CustomBuildOrphanSweepCheck",
     "DuplicateExecutorCheck",
+    "FillerMiningHealthCheck",
     "FinalizeCheck",
     "GpuCountCheck",
     "GpuFingerprintCheck",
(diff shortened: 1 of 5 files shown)


taiberium:
Closing this one. #767 (lium-io-backend) already detects and restarts not-mining fillers from GPU telemetry, and lium-io-backend#775 gives us SSH access into live fillers, so we can read the worker output directly. This check only covered the dead-worker case anyway — the validator doesn't know the filler strategy, so partial-load DPHN was out of reach. If we want automatic fleet-scale capture later, the right way is to carry the strategy in filler_containers_by_executor first.

PR #1143, opened 2026-07-17
DAH-2439: verify Lium filler container liveness in RentalVerificationCheck (shadow/enforcement FTs)

taiberium:
Notion: DAH-2439. ISSUE-050 follow-up: some hosts kill Lium filler containers while the run stays RUNNING and keep earning unrented incentive — the fatal check trusted the backend's word and never looked at the host.

- RentalVerificationCheck: instead of blindly passing on "active filler runtime", SSH `docker ps` the filler container each cycle (mirrors the customer-pod flow in RentedMachineCheck).
- Container missing → live re-check via new backend endpoint `GET /internal/filler-runs/{id}/active` (mirrors `/internal/pods/{pod_id}/rental-active`, backend PR: Datura-ai/lium-io-backend DAH-2439 branch).
- Run not RUNNING (stopped / mid-transition) → benign pass. Run RUNNING past a 10-min grace → confirmed kill: customer-facing `FILLER_CONTAINER_MISSING` event with remediation + container death diagnostics (DAH-2395 collector).
- Fail open on re-check API errors, docker-daemon errors, SSH transport loss.
- Two FTs: `FILLER_LIVENESS_CHECK_ENABLED` (master switch, default on = shadow: log only, scoring untouched) and `FILLER_LIVENESS_ENFORCEMENT_ENABLED` (default off; flips the confirmed kill into a fatal fail → no unrented incentive).

Tests: 28 cases in test_rental_verification_check.py; full validator suite green (3 pre-existing test_sshd_bootstrap_script failures, unrelated).

### DIFF
--- a/neurons/validators/src/clients/backend_client.py
+++ b/neurons/validators/src/clients/backend_client.py
@@ -17,6 +17,7 @@ from protocol.vc_protocol.compute_requests import (
     DefaultDockerImage,
     DefaultDockerImagesResponse,
     ExecutorHealthCheckResponse,
+    FillerRunActiveResponse,
     PodRentalActiveResponse,
     RentedExecutorsResponse,
 )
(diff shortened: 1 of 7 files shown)


arhangel66 (approved):
The PR body says "Fail open on … SSH transport loss", but in enforcement mode the SSH branch intentionally fails the cycle (passed=not enforce), consistent with RentedMachineCheck's handling for rented machines — and that's the right call, since always-pass would let a host evade the check by dropping SSH before the probe. Please fix the PR description (and the code comment) to state the fail-closed policy explicitly, so whoever flips FILLER_LIVENESS_ENFORCEMENT_ENABLED later isn't misled. Optional hardening: normalize started_at to naive UTC before subtracting, in case the backend contract ever goes tz-aware (today it's pinned naive by the backend test).

PR #1144, opened 2026-07-17
DAH-2440: enable inspector collector for filler workloads

jam6099:


neurons/validators/src/services/docker_service.py
```diff
@@ -4275,11 +4271,10 @@ async def delete_container(
                         executor_info, payload.container_name
                     )
 
-                # Stop inspector only after the last customer pod leaves this executor.
+                # Stop inspector only after the last rented pod leaves this executor.
```
taiberium, line 4274: NIT: the guard now covers fillers too, so `_has_rented_customer_containers` no longer describes what it checks. `_has_rented_containers` would match.

taiberium (approved):

PR #1148, opened 2026-07-20
DAH-2448, executor disk reserve so validator writes survive a full disk

arhangel66:
### Task Link

[DAH-2448](https://app.notion.com/p/Executor-disk-reserve-validator-writes-survive-100-full-disk-no-XFS-false-penalties-3a3b8bfdbde98157a28debbca81b75f6)

---

## Problem

On executor machines without XFS/vloopback, a renter can fill the host disk to 100%. The validator's `UploadFilesCheck` (fatal) then fails with `UPLOAD_FAILED`, the executor is marked inactive, and the miner collects a false `EXECUTOR_INACTIVE_MID_RENTAL` penalty — even though the GPUs are healthy and rented. This has been recurring in waves (~30 penalties over 3 weeks), tied to disk-heavy customers.

For a rented executor the validation pipeline short-circuits at `TenantEnforcementCheck` (`halt=True`), so only two disk-dependent fatal steps actually run: `UploadFilesCheck` and `MachineSpecScrapeCheck`. Both just need a few writable paths to survive a full disk.

## Solution

The executor container (privileged) bootstraps a small pre-allocated disk reserve at start and mounts it over exactly the paths the validation cycle writes to, so those writes keep succeeding at 100% host disk. Executor-only change — no protocol change, no miner action, rolls out with the normal executor release train.

The reserve is a 5 GB loop-backed ext4 image inside a named Docker volume:
- `fallocate` + `mkfs.ext4 -E nodiscard` — `nodiscard` is mandatory, otherwise mkfs discards the blocks and silently destroys the pre-allocation.
- backing file lives in the `reserve_data` named volume (host fs, survives `--force-recreate`), so the reservation exists before the disk ever fills.

Mounted into the executor:
- `overlay` over `/root/app` (upper/work in the reserve) — validator sftp uploads land in `{root_dir}/{uuid}`;
- `bind` `/tmp` — `machine_scrape` onefile self-extraction and tempfiles;
- `bind` `/root/.ssh` — per-cycle `authorized_keys` appends.

Everything is fail-open: any failure logs a warning and the executor continues unprotected (today's behavior). A janitor loop (every 15 min) removes upload dirs / `/tmp` entries older than 90 min and evicts oldest under ≥80% reserve pressure — required because the current validator pipeline never deletes its own upload dirs.

`/run` is deliberately NOT covered: Debian's `/var/run` → `/run` symlink means a tmpfs there would shadow the bind-mounted `docker.sock`/`dstack.sock`. Consequence: an executor restart while the disk is already full may still fail (same as today, where alembic/postgres also fail); the reserve protects the running executor, which is the penalty scenario.

## Changes

- `neurons/executor/setup_disk_reserve.sh` (new) — reserve bootstrap + mounts + janitor, fully fail-open.
- `neurons/executor/run.sh` — invoke bootstrap (fail-open), start janitor, re-`cd /root/app` through the overlay.
- `neurons/executor/Dockerfile` — add `e2fsprogs` (`python:3.11-slim` has no `mkfs.ext4`).
- `neurons/executor/docker-compose.app.yml` / `.app.dev.yml` / `.app.local.yml` — `reserve_data` named volume, mounted only on the `executor` service.

## Deployment Steps

Use GitHub Action.

Not urgent for this PR, but before fleet rollout: canary on a kernel-5.15 box (overlay nesting was verified on 6.8) and a CVM smoke check (dstack.sock path not exercised on the plain staging box). Then the normal `executor-v*` release with image backup.

## Review Request

- [x] Just code review

## Other PRs

Complementary backend-side work (renter-facing disk caps) tracked separately in the `reserve-system-disk-space` proposal; not required for this PR.

## Risks

- **overlay over `/root/app`**: verified to mount on the container's own overlay2 rootfs on kernel 6.8; kernel 5.15 not yet canaried (arithmetic is version-independent, but worth a canary).
- **CVM machines**: same compose/Dockerfile/run.sh change; TDX/GPU attestation calls in `upload_ssh_key` already fail soft, but the dstack.sock path was not exercised on staging.
- **`/root/.ssh` semantics**: keys now survive container recreation (previously lost on recreate). Strictly an improvement; validator re-uploads keys per cycle anyway.
- **Fail-open**: on any bootstrap failure the executor runs exactly as today (unprotected) — no new hard-fail path.

## Test Cases

Verified on a staging box (ext4, no XFS, kernel 6.8, RTX A4000) with a locally-built `compute-subnet-executor:local` image (nothing pushed to Docker Hub, no `executor-v*` tag). Full battery: 24/26 passing (2 "failures" were defects in the probes themselves, re-verified by hand).

### Test Case 1 — validator writes at 100% host disk

**Actions**: fill the host disk to 100%, then from the validator side push an sftp upload into `/root/app/<uuid>`, write a 200 MB `NamedTemporaryFile` in `/tmp`, and append to `/root/.ssh/authorized_keys`.

**Expected Output**: all three succeed and physically land in the reserve; the executor API and `nvidia-smi` stay alive. Paired proof at 100% disk: a host-side `dd` writes 0 bytes (ENOSPC) while the container writes 50 MB to `/tmp` in the same window.

### Test Case 2 — restart reuse and spec integrity

**Actions**: restart the executor container after freeing the disk; check mounts and reported disk specs.

**Expected Output**: the reserve is re-attached without re-`mkfs`, overlay/binds restored, ssh keys persisted; `df /` inside the container still equals the host disk (reported `hard_disk` specs unchanged — only `free` drops by 5 GB, `total` unchanged).

### Test Case 3 — fail-open

**Actions**: start with the volume missing, and start with a reserve dir too small for `fallocate`.

**Expected Output**: each logs a warning and the executor continues unprotected; no hard fail.

## Checklist

- [x] Risks are documented
- [x] Tests are described

### DIFF
--- a/neurons/executor/setup_disk_reserve.sh
+++ b/neurons/executor/setup_disk_reserve.sh
@@ -54,9 +54,14 @@ setup_backing_fs() {
     fi
 
     # nodiscard is mandatory: default mkfs.ext4 discards blocks and silently
-    # destroys the fallocate reservation
+    # destroys the fallocate reservation. lazy init must be off too: the
+    # kernel's post-mount ext4lazyinit zeroes inode tables via WRITE_ZEROES,
+    # which loop turns into hole-punching in the backing file (~80M of the
+    # reservation lost; a write into a hole on a full disk aborts the journal
+    # and flips the reserve read-only)
+    local mkfs_opts="nodiscard,lazy_itable_init=0,lazy_journal_init=0"
     if [ "$(blkid -o value -s TYPE "$loop" 2>/dev/null)" != "ext4" ]; then
-        mkfs.ext4 -q -F -E nodiscard -m 0 -L lium-reserve "$loop" || { log "mkfs.ext4 $loop failed"; return 1; }
+        mkfs.ext4 -q -F -E "$mkfs_opts" -m 0 -L lium-reserve "$loop" || { log "mkfs.ext4 $loop failed"; return 1; }
     fi
 
     mkdir -p "$MNT" || return 1
@@ -64,9 +69,18 @@ setup_backing_fs() {
         # reserve content is disposable — rebuild a corrupted fs instead of
         # staying unprotected forever
         log "mount failed — re-creating reserve fs on $loop"
-        mkfs.ext4 -q -F -E nodiscard -m 0 -L lium-reserve "$loop" || { log "mkfs.ext4 $loop failed"; return 1; }
+        mkfs.ext4 -q -F -E "$mkfs_opts" -m 0 -L lium-reserve "$loop" || { log "mkfs.ext4 $loop failed"; return 1; }
         mount "$loop" "$MNT" || { log "mount $loop failed"; return 1; }
     fi
+    # an fs with a previously aborted journal can still mount, but read-only —
+    # same rebuild rule as a failed mount
+    if ! touch "$MNT/.rw_probe" 2>/dev/null; then
+        log "reserve mounted read-only — re-creating fs on $loop"
+        umount "$MNT" || return 1
+        mkfs.ext4 -q -F -E "$mkfs_opts" -m 0 -L lium-reserve "$loop" || { log "mkfs.ext4 $loop failed"; return 1; }
+        mount "$loop" "$MNT" || { log "mount $loop failed"; return 1; }
+    fi
+    rm -f "$MNT/.rw_probe"
     touch "$MARKER" || true
 }
 
@@ -122,6 +136,9 @@ janitor_loop() {
     # pipeline never deletes its upload dirs, and a scrape session killed
     # mid-run leaks its /tmp files
     while :; do
+        # re-extend the backing file: any holes punched at runtime are
+        # re-filled while the host disk still has room
+        fallocate -l "${SIZE_MB}M" "$IMG" 2>/dev/null
         find /root/app -maxdepth 1 -type d -regextype posix-extended \
             -regex '.*/[0-9a-f]{32}' -mmin +90 -exec rm -rf {} + 2>/dev/null
         find /tmp -mindepth 1 -maxdepth 1 -mmin +90 -exec rm -rf {} + 2>/dev/null
--- a/neurons/executor/setup_disk_reserve.sh
+++ b/neurons/executor/setup_disk_reserve.sh
@@ -48,7 +48,14 @@ setup_backing_fs() {
         ensure_loop_node "$loop" || return 1
     else
         [ -f "$MARKER" ] && log "WARNING: reserve initialized earlier but no loop attached — previous device may have leaked"
-        loop=$(losetup -f 2>&1) || { log "losetup -f failed: $loop"; return 1; }
+        # stderr dropped so a stray warning cannot corrupt the path; the shape
+        # check below is the validation ([ -b ] would be wrong here: the node
+        # may legally not exist yet — ensure_loop_node creates it)
+        loop=$(losetup -f 2>/dev/null)
+        case "$loop" in
+            /dev/loop[0-9]*) ;;
+            *) log "losetup -f failed (no free loop device?)"; return 1 ;;
+        esac
         ensure_loop_node "$loop" || return 1
         losetup "$loop" "$IMG" || { log "losetup $loop $IMG failed"; return 1; }
     fi
--- a/neurons/executor/setup_disk_reserve.sh
+++ b/neurons/executor/setup_disk_reserve.sh
@@ -1,0 +1,146 @@
+#!/bin/bash
+# Disk reserve for validator writes (DAH-2448).
+#
+# On machines without XFS/vloopback a renter can fill the host disk to 100%;
+# the validator's upload then fails (UPLOAD_FAILED) and the miner collects a
+# false penalty. This script pre-allocates a 5G loop-backed ext4 inside the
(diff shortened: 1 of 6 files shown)


arhangel66:
## Staging verification log: before/after cycle against the validator's real check

Full experiment run on a staging box (ext4, kernel 6.8, RTX A4000) on 2026-07-20, reproducing the production failure first and then proving the fix on the same machine. Nothing was pushed to Docker Hub and no `executor-v*` tags were created; the branch image was built locally on the box.

### Method

- The probe executes the **actual `UploadFilesCheck`** from this repo verbatim (together with the real `messages.py`/`models.py`; only the `pipeline.CheckResult`/`Context` containers are stubbed, because the real `pipeline.py` imports the whole validator service stack). It connects over the network via asyncssh to the executor's sshd, the same path a validator takes. Payload: 2×1 MiB script-like files + one 64 MiB binary per cycle.
- Baseline = prod image `daturaai/compute-subnet-executor@sha256:d9c6218a3908ee123b1b8ee1f93802992a45965e175656b03f8dadb3831caa96` with the `origin/main` compose. The update was applied watchtower-style: build on the box + compose stop→start of the running stack.
- "Full disk" = 0 bytes free including the root reserve (`sudo fallocate` + `dd` top-off). Control on the host at that state: `dd ... conv=fsync` as root → `dd: error writing ...: No space left on device`, rc=1.

### Results

| State | Host disk | Real check result | Notes |
|---|---|---|---|
| prod image | free | `UPLOAD_OK` ×2 | sanity baseline |
| prod image | 100% | **`UPLOAD_FAILED` ×2** (sftp `Failure`) | executor API still served `{"version":"3.1.0"}` and the container reported healthy — the false-penalty scenario reproduced literally |
| this branch | free | `UPLOAD_OK` ×2 | after the in-place update; 5120 MiB reserve pre-allocated |
| this branch | 100% | **`UPLOAD_OK` ×2** | uploads landed in the reserve (`/root/app/afc329c06523488885e956876a826877`, `/root/app/569b87a976eb4433bd9c80666109e518`); after a forced `sync`, every uploaded file was md5-identical to its source (`b561f87202d04959e37588ee05cf5b10` for the 1 MiB files, `84b0d1834a4eace74de575cc82776fe8` for the 64 MiB blob); dmesg clean; a parallel 50 MiB `dd conv=fsync` inside the container succeeded while the host could not write a byte |

### What the cycle caught (fixed by dd062ed7)

The first 100%-disk run on this branch initially "passed", but the reserve fs died minutes later:

```
Buffer I/O error on dev loop7, logical block 524416, lost async page write
Aborting journal on device loop7-8.
EXT4-fs (loop7): Delayed block allocation failed for inode 132618 ... error 5
EXT4-fs (loop7): This should not happen!! Data will be lost
EXT4-fs (loop7): Remounting filesystem read-only
```

Root cause: the kernel's post-mount `ext4lazyinit` zeroes inode tables via WRITE_ZEROES, which the loop driver turns into hole-punching in the backing file — exactly 80 MiB on a 5 GiB fs (327680 inodes × 256 B), undoing that part of the `fallocate` reservation. A writeback into a hole at 100% host disk gets ENOSPC → journal abort → reserve goes read-only with data loss. `nodiscard` alone does not prevent this.

Spike confirmation on the box: with the old mkfs flags a fully allocated 1024 MiB image drops to 1008 MiB within ~45 s of the first mount; with `lazy_itable_init=0,lazy_journal_init=0` it stays at 1024 MiB. After the fix the 5120 MiB reserve allocation is stable (checked right after boot and 105 s later), and the full-disk cycle above was re-run end-to-end with the integrity checks.

Hardening added alongside the flag change: the janitor re-`fallocate`s the backing file every cycle (re-fills any holes punched at runtime while the host still has room), and the bootstrap write-probes the mounted fs and rebuilds it if it came up read-only (an fs with a previously aborted journal still mounts, but ro).

### Pre-existing behavior observed at 100% disk (unchanged by this PR)

- `docker exec` fails on the host (runc cannot write its /tmp bootstrap file), so container healthchecks fail too → containers get flagged unhealthy and autoheal restarts them.
- postgres died during the full-disk window and needed a restart afterwards. The reserve's scope is the validator write paths, not the DB — matching the "restart/DB at full disk" limitation already documented in the description.

neurons/executor/setup_disk_reserve.sh
```diff
@@ -0,0 +1,163 @@
...
+    # idempotent: extends/repairs the allocation, no-op when already full-size
+    fallocate -l "${SIZE_MB}M" "$IMG" || { log "fallocate ${SIZE_MB}M failed (host disk too full?)"; return 1; }
+
+    local loop
+    loop=$(losetup -j "$IMG" 2>/dev/null | head -1 | cut -d: -f1)
+    if [ -n "$loop" ]; then
+        ensure_loop_node "$loop" || return 1
+    else
+        [ -f "$MARKER" ] && log "WARNING: reserve initialized earlier but no loop attached — previous device may have leaked"
+        loop=$(losetup -f 2>&1) || { log "losetup -f failed: $loop"; return 1; }
```
taiberium: **NIT:** `losetup -f 2>&1` folds stderr into `$loop`, so a warning on success corrupts the device path. Use `2>/dev/null` and check `[ -b "$loop" ]`.

taiberium (approved):

fortunelucky777 (approved):
Really good idea!

PR #1149, opened 2026-07-20
feat: DAH-2458 — seed deploy profile with backend pre-dispatch spans

fortunelucky777:
## Describe your changes

DAH-2458. The deploy profiler in `create_container` starts counting at `REQUESTED_FROM_BACKEND`, so it structurally can't see the backend-side time *before* the container-create request reaches the subnet — most of which is the blocking filler-preemption wait (routinely ~30s). This lets the backend hand its own measured pre-dispatch spans to the subnet so the persisted profile is **one ordered `backend → subnet → backend` timeline** instead of subnet-only.

**Changes (`neurons/validators`):**
- `ProfilerStepName`: add `FILLER_PREEMPTION`, `BACKEND_PREP`, `BACKEND_FINALIZE`.
- `ProfilerStep.from_wire(dict)`: rebuild a typed step from a wire dict; returns `None` for an unknown/malformed name so profiler drift can never break a deploy.
- `ContainerCreateRequest.pre_dispatch_profilers: list[dict] = []`: new optional field carrying the backend's measured spans.
- `create_container`: **seed** the profiler list from `pre_dispatch_profilers` (ahead of the anchor) before appending the subnet steps.
- `create_container`: the final `FINISHED_IN_SUBNET` step now also **stamps the subnet's wall-clock finish time** (`timestamp`, alongside its `duration`), so the backend derives its `Backend finalize` span with one subtraction (`pending_finished_at − timestamp`) instead of re-summing every step's duration. The `duration` is unchanged, so the deploy-summary total and the lium-stats ETL pivot are unaffected.

**Compatibility:** the field is optional/defaulted — an older backend that omits it degrades gracefully to today's subnet-only profile, and a newer backend hitting an older validator just has its spans dropped. No hard deploy-ordering with the backend.

[DAH-2458](https://www.notion.so/DAH-2458)

- [x] I wrote tests. (`from_wire` round-trip + unknown/malformed drop; `FINISHED_IN_SUBNET` carries a wall-clock timestamp; full `test_deploy_optimizations.py` green — 32 passed)
- [ ] Need to take care of performance? (no — seeding is a bounded list copy)

## Other PRs
- **lium-io-backend #790** — measures the spans and sends `pre_dispatch_profilers`.
- **lium-stats #72** — pivots the three new steps into columns + charts.


arhangel66 (approved):
LGTM with #790. Non-blocking: move ProfilerStep.from_wire's cls(...) inside the try and catch ValidationError — right now a known step name with a bad-typed duration/timestamp escapes the guard and can abort the deploy, contrary to the "never fatal" comment. Add a negative test for it.

PR #1152, opened 2026-07-21
DAH-2439: classify filler death — punish only external kills (strike counter for ambiguous stops)

taiberium:
Notion: DAH-2439 follow-up. Prod shadow data (20h) showed the liveness check flags two different things: 515 containers REMOVED by the host (real ISSUE-050 kills) and 5 that died by SIGTERM while retrying pool connection. Punishing must hit only external kills, never a filler that crashed or failed to start on its own.

- `classify_container_death()` in docker_utils: REMOVED / STOPPED (137/143, not OOM) / SELF_CRASHED / OOM_KILLED / NEVER_STARTED / CLEAN_EXIT / UNKNOWN, + `container_uptime_seconds` (kill_timing=at_start/after_running in the event).
- RentalVerificationCheck: REMOVED → punishable outright (nothing legitimate deletes a RUNNING run's container). STOPPED → strike system: a lone SIGTERM could be the worker exiting 143 itself, so incident #1 per executor = `FILLER_KILL_SUSPECTED` (logged, no penalty even in enforcement); >= `FILLER_KILL_STRIKE_THRESHOLD` (default 2) within 7 days = punished. Everything else → `FILLER_CRASHED`, never punished, never a strike (self-heal territory, DAH-2419).
- Strikes: one per filler_run_id (Redis SETNX dedup — the same corpse seen every cycle counts once), accumulated per executor, 7-day TTL. Redis unavailable → strict verdict (no free pass), shadow default still means no money impact.
- Shadow/enforcement semantics unchanged; all new verdicts pass in shadow.

Tests: 13 classifier + 9 verdict cases; full validator suite 1149 green (3 pre-existing sshd failures, unrelated).

### DIFF
--- a/neurons/validators/src/core/docker_utils.py
+++ b/neurons/validators/src/core/docker_utils.py
@@ -158,10 +158,14 @@ class ContainerDeathKind(str, Enum):
 _ZERO_DOCKER_TIMESTAMP_PREFIX = "0001-01-01"
 _KILL_SIGNAL_EXIT_CODES = (137, 143)  # 128+SIGKILL, 128+SIGTERM
 _REMOVED_MARKERS = ("no such object", "no such container")  # docker casing varies; match lowercased
-# A reboot/compose-restart SIGTERMs every container at once, so the executor stack restarts around
-# the same time. If it (re)started no earlier than this many seconds before the filler died, the
-# SIGTERM was collateral of that restart, not a filler-targeted `docker stop`.
+# A reboot SIGTERMs every container at shutdown and brings the executor stack back up on boot, so
+# the executor's StartedAt lands at or shortly AFTER the filler's death. The restart must bracket
+# the death to count as collateral: up to this long after (slow boot), with only a small skew
+# tolerance before (clocks). A filler-targeted `docker stop` leaves the executor started hours/days
+# earlier (far-negative delta); an executor that restarts long after the kill (far-positive delta)
+# does not excuse a kill that already happened.
 _HOST_RESTART_WINDOW_SECONDS = 600
+_HOST_RESTART_CLOCK_SKEW_SECONDS = 120
 
 
 def _never_started(diagnostics: ContainerDeathDiagnostics) -> bool:
@@ -183,13 +187,17 @@ def _stop_coincided_with_host_restart(diagnostics: ContainerDeathDiagnostics) ->
     finished = _parse_docker_timestamp(diagnostics.finished_at)
     if executor_started is None or finished is None:
         return False
-    return (executor_started - finished).total_seconds() > -_HOST_RESTART_WINDOW_SECONDS
+    delta_seconds = (executor_started - finished).total_seconds()
+    return -_HOST_RESTART_CLOCK_SKEW_SECONDS <= delta_seconds <= _HOST_RESTART_WINDOW_SECONDS
 
 
 def classify_container_death(diagnostics: ContainerDeathDiagnostics) -> ContainerDeathKind:
+    # Match the "no such object/container" marker only in OUR inspect error, never in the
+    # container's own logs_tail: a genuinely removed container fails inspect and lands the
+    # marker in capture_error, while a still-existing stopped container whose workload merely
+    # printed that string would otherwise be misread as removed and punished outright.
     capture_error = (diagnostics.capture_error or "").lower()
-    logs_tail = (diagnostics.logs_tail or "").lower()
-    if any(marker in capture_error or marker in logs_tail for marker in _REMOVED_MARKERS):
+    if any(marker in capture_error for marker in _REMOVED_MARKERS):
         return ContainerDeathKind.REMOVED
     if diagnostics.oom_killed:
         return ContainerDeathKind.OOM_KILLED
--- a/neurons/validators/src/core/docker_utils.py
+++ b/neurons/validators/src/core/docker_utils.py
@@ -2,6 +2,8 @@ import asyncio
 import json
 import shlex
 from dataclasses import dataclass
+from datetime import datetime
+from enum import Enum
 
 import asyncssh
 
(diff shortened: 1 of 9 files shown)


arhangel66 (changes requested):
**Requesting changes** — two logic gaps in the money path, both confirmed by running the head code. Enforcement is off by default, so these are pre-flip blockers rather than a live incident, but both should land before enforcement goes on.

**1. (blocking) One filler run becomes a second strike over time** — `redis_service.py` `register_filler_kill_strike`
The executor counter `filler_kill_strikes:{executor_uuid}` gets its TTL refreshed on *every* observation (both branches), but the per-run dedup key `filler_kill_seen:{filler_run_id}` is set once and never refreshed. If the same stopped corpse keeps being observed while the run stays RUNNING, after the 7-day TTL the dedup key expires, the counter is still alive, and the next observation re-`INCR`s the same run → strike 2 → punished at threshold 2 **without a second real kill**. Please store timestamped unique incidents (e.g. a sorted set, prune entries older than the window) so each incident expires independently, and add a real Redis TTL regression test — the current `FakeStrikeRedis` returns a fixed count and never exercises SETNX/INCR/EXPIRE.

**2. (blocking) Late executor restart hides a targeted stop** — `docker_utils.py` `_stop_coincided_with_host_restart`
The check has only a lower bound: `(executor_started - finished) > -600`. With no upper bound, an executor that (re)started *any* time after the filler died still reads as a coincident reboot → `HOST_REBOOT` → never punished. Repro: filler SIGTERM'd, executor container restarted a day later → classified `host_reboot` instead of `stopped`. Please bound the delta on both sides (symmetric window in the intended direction) and add a test with a much later executor start.

**3. (non-blocking) Redis-down punishes the first stop with no grace**
When Redis is unavailable `_register_kill_strike` returns `None`, which the caller treats as “threshold reached” → under enforcement the *first* ambiguous stop is punished outright, skipping the grace incident. It's documented as intentional, but it contradicts the one-grace-incident design and can misfire during a Redis blip — worth reconsidering or at least logging loudly.

Classifier split (`docker_utils` diagnoses, Redis counts, `RentalVerificationCheck` decides) looks right; only the TTL model and the one-sided restart window need fixing.

surcyf123:
Triage: @taiberium — changes requested 49 days ago (two money-path gaps), conflicting since (DAH-2439 kill classifier). Still planned? It scores as P2 (idle-earnings fairness) if you pick it up.

neurons/validators/src/core/docker_utils.py
```diff
@@ -135,6 +137,105 @@ def to_log_fields(self) -> dict[str, object]:
...
+    avoid the timezone ambiguity of `uptime -s`. A filler-targeted `docker stop` leaves the executor
+    stack untouched, so its start time stays hours/days before the death.
+    """
+    context = diagnostics.host_context or {}
+    executor_started = _parse_docker_timestamp(context.get("executor_container_started_at"))
+    finished = _parse_docker_timestamp(diagnostics.finished_at)
+    if executor_started is None or finished is None:
+        return False
+    delta_seconds = (executor_started - finished).total_seconds()
+    return -_HOST_RESTART_CLOCK_SKEW_SECONDS <= delta_seconds <= _HOST_RESTART_WINDOW_SECONDS
```
surcyf123, line 191: An executor container restart within 600 s after the death is not a host reboot, and the provider controls it: `docker stop filler_x && docker restart <executor>` gets `host_reboot` every time. `host_context["host_boot_time"]` is already collected; require the host boot to bracket the death (better: `btime` from `/proc/stat`) and treat an executor-only restart as STOPPED.

neurons/validators/src/services/redis_service.py
```diff
@@ -70,6 +70,40 @@ async def acquire_executor_lock(
...
+                    pipe.incr(strikes_key)
+                    pipe.expire(strikes_key, ttl_seconds)
+                    strikes, _ = await pipe.execute()
+                return int(strikes)
+            # Already-counted incident (same run seen again this window). Refresh BOTH keys so a
+            # persistent killer keeps its strikes alive AND this one dead run never re-counts as a
+            # fresh incident: without refreshing the seen key, a run stuck RUNNING past the TTL would
+            # let its seen marker expire mid-observation and be double-charged as a second strike.
+            current_strikes = await self.redis.get(strikes_key)
+            if current_strikes is None:
```
surcyf123, line 99: Returning 1 without recreating the counter means the next distinct run also INCRs from nothing to 1, so a real second incident is undercounted. `SET strikes_key 1 EX ttl` here keeps the count honest.

surcyf123 (commented):
Comment only: this needs a rebase on main before a verdict, and it changes penalty logic, so the decision is Mikhail's, not the loop's.

Mikhail's three points from 22 Jul look addressed at head `bad5704a`: the seen key is refreshed on repeat observation (`redis_service.py:100`, pinned by `test_repeat_observation_refreshes_seen_key_ttl` on fakeredis, which is in the validator test deps), the restart window is bounded on both sides (`docker_utils.py:191`, with tests for 5 min before and 20 min after), and Redis down now fails open. One thing I would still fix before enforcement goes on:

1. **[medium]** `neurons/validators/src/core/docker_utils.py:191`: `HOST_REBOOT` is granted whenever the executor container (re)started up to 600 s after the filler died. An executor container restart is not a host reboot, and the provider controls it: `docker stop filler_x && docker restart <executor>` classifies as `host_reboot` every time, no strike, no penalty. `_collect_host_context` already gathers `host_boot_time`, but the classifier never reads it. Fix: require the host boot to bracket the death (read `btime` from `/proc/stat`, a UTC epoch, instead of `uptime -s`, which removes the timezone problem the docstring mentions), and treat an executor-only restart as `STOPPED`, or at least as a strike.
2. **[low]** `neurons/validators/src/services/redis_service.py:99`: when the seen key is alive but the counter is gone, this returns 1 without recreating the counter, so the next distinct run also lands on 1 and a real second incident is undercounted. Fail-open direction, but cheap to fix: `SET strikes_key 1 EX ttl` in that branch.
3. **[low]** `neurons/validators/src/services/task/messages.py:287`: `GpuPowerLimitMessages.RECAPPED_DRIFTED_FILLER` is not on main and nothing in this PR reads it. Drop it from this branch, or move it to the PR that uses it.
4. **[low]** PR body vs code: the body says REMOVED is punishable outright; the code strike-gates REMOVED the same as STOPPED (the `_filler_killed_result` docstring explains why from the shadow data). Update the body so whoever reviews the enforcement flip reads the right rule.

Hold for Mikhail (arhangel66) on the penalty math; Rustam owns the check itself.

neurons/validators/src/services/task/messages.py
```diff
@@ -281,6 +281,14 @@ class GpuPowerLimitMessages:
         category="policy",
         impact="Proceed; Lium lowered the limit for its own idle filler, node stays rentable",
     )
+    RECAPPED_DRIFTED_FILLER = MessageTemplate(
+        event="Re-applied Lium filler GPU power cap that had drifted back up",
+        reason="GPU_POWER_LIMIT_RECAPPED_DRIFTED_FILLER",
+        severity="warning",
```
surcyf123, line 287: `RECAPPED_DRIFTED_FILLER` is not on main and nothing in this PR reads it. Drop it here or move it to the PR that uses it.

PR #1153, opened 2026-07-21
DAH-2467, mixed scoring for partially rented GPU-split nodes

taiberium:
### Task Link

[DAH-2467](https://app.notion.com/p/GPU-split-partially-rented-node-loses-idle-incentive-on-remaining-GPUs-run-fillers-per-GPU-and-pa-3a4b8bfdbde98108b455c5f5fe741959)

---

## Problem

On a GPU-split node, renting even 1 of 8 GPUs moves the whole executor into the mining pool: rented state is per-executor, so the 7 free GPUs stop earning the unrented (idle) incentive for the entire rental. Providers report this kills GPU-splitting economics and are disabling splitting because of it (Discord #lium-io, 2026-07-21).

## Solution

Score a partially rented split-opted-in executor in BOTH pools: the rented GPUs stay in the mining pool, the free GPUs earn in the unrented (rental-share) pool.

Implementation: the incentive engine expands such an executor into two virtual `JobResult` portions before scoring — a rented portion (`gpu_count` = rented GPUs) and a free portion (`is_rented=False`, `gpu_count` = free GPUs) — so every existing eligibility and formula rule (model program, bucket caps at `gpu_splitting_min_count`, soft price limit, sysbox/driver gates, miner-default-job exclusion) applies to each portion unchanged. After scoring, the portions are merged back into the single executor result (full `gpu_count` restored, incentives summed, both pools' calculation logs kept), so machine-spec publish and backend accounting still see exactly one message per executor.

The rented-GPU count comes from the backend's rented-executors response: each `RentedPod` now carries `gpu_count` (Datura-ai/lium-io-backend#793). When the backend doesn't report it (older backend, no assigned-GPU rows), the validator falls back to today's whole-box rented scoring — fail-safe, nothing changes.

## Changes

- `RentedPod.gpu_count` (protocol): GPUs held by each rented pod, `None` = unknown.
- `JobResult.rented_gpu_count` + `TaskResultHandler._get_rented_gpu_count`: summed per-pod counts; `None` when any pod lacks a count.
- `RentalPriceIncentive.calculate_mining_scores`: expands partially rented split executors into rented/free portions, scores as usual, merges back (`_expand_partially_rented_split_results` / `_merge_partially_rented_split_results`).
- Miner weight aggregation is untouched — `miner_incentives` accumulates both portions during post-processing.

## Deployment Steps

Use GitHub Action. Deploy the backend PR (Datura-ai/lium-io-backend#793) first — without it the new field is absent and the validator keeps current behavior.

## Review Request

- [x] Just code review

## Other PRs

- Datura-ai/lium-io-backend#793 — backend half: runs Lium fillers on the free GPUs of partially rented split nodes and reports per-pod `gpu_count` to the validator.

## Risks

- Money-affecting change in the incentive engine, but gated three times: the executor must be rented, split-opted-in (`supports_gpu_splitting`), and the backend must report per-pod GPU counts. Any missing piece → exact current behavior (verified by test).
- The free portion inherits the rented executor's validation score — no separate synthetic validation runs on the free GPUs (same trust level as the rented box itself). The DAH-2439 filler-liveness enforcement will add a work-verification layer on top.
- Unrented capacity buckets now also absorb free portions of split nodes (counted at the `gpu_splitting_min_count` bucket), which slightly increases bucket competition — same rule as fully idle split nodes today.

## Test Cases

`pytest` (neurons/validators): 1143 passed, 8 skipped; the 3 `test_sshd_bootstrap_script.py` failures are environment-related and reproduce on `main` unchanged.

### Test Case 1

**Actions** 8-GPU H200 split node, 1 GPU rented, engine run with a pinned rental share.

**Expected Output** One merged result: `mining_score` computed on 1/8 GPUs, total incentive = mining incentive + full rental share, 7 free GPUs counted in the (`H200`, 1) unrented bucket, `miner_incentives` matches the merged sum.

### Test Case 2

**Actions** Same node but `rented_gpu_count=None` (old backend) / fully rented / no split opt-in / idle.

**Expected Output** No expansion — behavior identical to today (whole-box mining for rented, normal idle path otherwise).

## Checklist

- [x] Risks are documented
- [x] Tests are described


arhangel66 (changes requested):
Вот финальный текст ревью для #1153. Готов вставлять как есть — **пока не отправлял**.

---

**Requesting changes.** The scoring idea is right and the mechanics are clean, but three things about the merged result block this: one silently costs miners money, two make the payout unexplainable to the provider.

I replayed this against prod. `f3016420-761d-4637-b83c-e837a8432585` (4×H200, `min_gpu_count_for_rental = 1`, one RUNNING pod holding 2 GPUs, sysbox on) is exactly this feature's target. Cycle `2026-08-25 11:26:12`:

| | today | with this PR |
|---|---|---|
| mining | 0.001653237068822 | 0.000862660167212 |
| unrented | — | 0.002431478928505 |
| **total** | **0.001653237068822** | **0.003294139095717** |

~2× for a half-idle box, and the extra comes out of burn rather than out of other miners (`rental_share` is linear in `total_rental_cost`, so the per-card idle payout is invariant). That part works.


### 1. Blocking — the provider earnings split misattributes the payout

The merge sums both portions' `incentive` but keeps the rented portion's classification: `is_rented` stays `True`, so `incentive_source` is `rented_emission` and `node_state_at_cycle` is `rented`.

The Miner Portal splits provider income by exactly that field — `provider_emissions_service.py:42-46`, again at `:182-186`:

```sql
SUM(c.attributed_alpha_rao) FILTER (WHERE c.incentive_source = 'rented_emission') AS rented_alpha_rao,
SUM(c.attributed_alpha_rao) FILTER (WHERE c.incentive_source = 'idle_incentive')  AS idle_alpha_rao,
```

For `f3016420` above, 0.002431478928505 of 0.003294139095717 — 74% — is idle income reported to the provider as rental income, and `idle_alpha_rao` for that node comes out zero. The one thing GPU splitting is sold on is the one thing the earnings page won't show.

Totals are safe: `compute-app/apps/server/src/services/provider_emissions.py` weights TMC alpha by `incentive` per `(cycle_started_at, validator_hotkey, executor_id, cycle_id)`, and the merge keeps one row per executor with the correct sum. Only the breakdown breaks.

### 3. Blocking — `incentive_formula_inputs` no longer reproduces `incentive`

`eligible_for_rental_share` stays `False` on the merged row, so `incentive_formula_version` is `mining_v1` and the inputs are the mining dict, while every unrented-pool field (`hourly_rate`, `effective_rate`, `count_bucket`, `max_cap`, `unrented_cap_multiplier`, `total_unrented_by_gpu_type`, `cap_dilution_applied`) is `None`. Recomputing from those inputs yields 0.000862660167212 against a stored `incentive` of 0.003294139095717 — 26%. And `incentive_formula_inputs.gpu_count` reports 4 (restored by the merge) while `mining_score` was computed on 2, so the formula does not close against its own inputs.

This is not internal — `NODE_CYCLES_SQL` returns `incentive_formula_version` and `incentive_formula_inputs` straight to the provider via `GET /{hotkey}/emissions/nodes/{executor_id}/daily/{date}`.

### 4. Non-blocking — no `ENABLE_*` flag

Every other money gate in this module ships behind `ENABLE_UNRENTED_SOFT_PRICE_LIMIT` / `ENABLE_UNRENTED_VRAM_OVER_DISK_LIMIT` / `ENABLE_UNRENTED_FLAGSHIP_CAPABILITY_LIMIT` / `ENABLE_UNRENTED_POWER_CAP_LIMIT`, with a shadow log first. Here the only switch is whether the backend sends `RentedPod.gpu_count`, which is one global lever with no per-validator rollback and no shadow mode. Given item 1, I'd like an `ENABLE_SPLIT_PARTIAL_RENTAL_SCORING` next to the others — two lines, and it matches the file's own convention.

### 5. Minor

- Nothing in `full_log_text` states the total. It ends with two separate `incentive` values (0.000862660167212 and 0.002431478928505) and the miner has to add them himself to get the number that is actually accounted.
- `rented_machine.py:156` builds the event context as `[{"name": ..., "pod_id": ...}]`, so the `log_text` header gives no hint that only 2 of 4 GPUs are rented.

### Suggested shape for the payload

Carry the free portion's rental fields onto the merged result; emit `incentive_formula_version = "mixed_v1"` with both blocks of inputs plus `rented_gpu_count` / `free_gpu_count`; add a third value for `incentive_source` / `node_state_at_cycle` (`mixed`) and split amounts (`incentive_rented` / `incentive_idle`) on the wire, so the portal can filter on numbers instead of one label. That last part needs coordination across `incentive_per_validator_cycle` (compute-app) and `provider_emissions_service` (miner portal) — the portal queries have to move off `incentive_source` before the first split node lands in a report. Happy to take that thread if you'd rather keep this PR scoped to the validator.

No rush on any of it: prod's `RentedPod` has no `gpu_count` yet (`compute-app/apps/server/src/protocol/compute_app_requests.py`), so `_get_rented_gpu_count` returns `None` and this PR is a no-op on prod until the backend side ships. That's the window to get the payload right.

I also have prod-snapshot golden-gate coverage for partially rented split nodes locally — a ladder of 1/7 … 7/1 rented on real fleet data, with assertions on merged `gpu_count` and per-executor `incentive`. I verified it catches a reverted expansion, a merge that drops the free portion's incentive, and a merge that forgets to restore `gpu_count`. Happy to push it onto this branch.

taiberium:
All four points are in, plus one your replay would have hit next.

- Earnings split: every cycle now carries `incentive_rented` and `incentive_idle`, always summing to `incentive`. The portal splits by those amounts and falls back to the label only for rows written before this. Datura-ai/lium-miner-portal#170.
- Formula inputs: a merged row reports `mixed_v1` with both blocks and the two GPU counts, snapshotted before the merge restores the whole-box `gpu_count`.
- Flag: `ENABLE_SPLIT_PARTIAL_RENTAL_SCORING`, next to the other four.
- Log: the total line is there, and the rented-pods context now carries `rented_gpu_count`.

The one you did not hit: migration 010 put CHECK constraints on `incentive_source` and `node_state_at_cycle`, so the first `mixed` row would have failed the insert and lost that cycle. Migration 011 replaces both. Verified on a timescaledb container.

The portal frontend also indexed a colour map by the label, so it would have thrown on the first mixed cycle: Datura-ai/lium-miner-portal-frontend#170.

Deploy order is portal, then backend, then validator. Yes please to the golden-gate tests, push them onto this branch.

arhangel66 (approved):
Approving, and the golden-gate tests are on the branch (`dbee9c7`): a `snapshot_partial_split_ladder` fixture derived from `snapshot_2026-02-13` with five real rented 8x A100 nodes opted into splitting at 1, 2, 4, 6 and 7 GPUs rented (two on sysbox, three not) and five whole-box controls, plus per-executor `incentive` and `gpu_count` assertions on every existing snapshot. Flipping `ENABLE_SPLIT_PARTIAL_RENTAL_SCORING` off makes three of them fail, so the gate is real.

All four review points are addressed properly — the split amounts go over the wire instead of a label, the mixed formula snapshot closes against its own incentive, the flag matches the file's convention. I re-checked the deploy order (portal, backend, validator) and the validator is a no-op without the backend, so the ordering protects itself.

One thing to fix before this ships, and it is the same class of problem as the earnings split: a case the tier table cannot express.

A remainder is always rated at `gpu_splitting_min_count` (`rental_price.py:508-516`), but `GPU_COUNT_CUSTOM_PRICES` only prices `"1"` and `"8"` (`"*": 0` for everything else) and `MAX_UNRENTED_GPUS_BY_TYPE` only caps buckets 1 and 8. `max(hourly_rate, rate_for_min)` does not rescue it — both lookups resolve to the same count. So on a node with `min_gpu_count_for_rental` = 2 or 4 the remainder gets `hourly_rate = 0` and `max_cap = 0`, while mining has already shrunk to the rented GPUs, so the provider loses its share of the mining pool for the free GPUs with nothing on the other side. The same node fully idle sits in bucket 8 and is paid normally.

Prod has five such machines today: one H200 8x and one RTX A4000 4x at `min = 2`, three RTX 5090 8x at `min = 4`. The RTX 5090 owner who rents 4 of 8 GPUs comes out behind where he'd be renting nothing at all. That is the opposite of what DAH-2467 is trying to fix.

Cheapest fail-safe: skip the expansion when the `min_count` tier has no configured price or cap, so those nodes keep today's whole-box behaviour. The ladder fixture only covers `min = 1`, so whichever way you fix it, a `min = 2` or `min = 4` node in the fixture would be the regression case. Happy with a follow-up PR if you'd rather keep this one scoped — it just has to land before the validator side goes out.

Two notes, neither blocking:

- The comment in `_resolve_bucket` calls the remainder tier "a pricing decision". In the current config it isn't one — tiers 1 and 8 both resolve to `DEFAULT_PRICE`, i.e. the same `MACHINE_PRICES` per card, so what actually changes is which cap bucket the remainder competes in. The config could price counts differently in principle, but today it doesn't, and that distinction is the whole reason the `min = 2` / `min = 4` case above is broken.
- `DefaultIncentive._pre_process_job_result` counts the virtual free portion as another executor, so `total_executors` / `successful_executors` drift by the number of partially rented split nodes. Observability only, no effect on weights.

PR #1154, opened 2026-07-21
DAH-2469: central miner serves executors from bulk portal snapshot

arhangel66:
## Summary
Reworks `MinerPortalAPI.fetch_executors` (central mode) to serve from an in-memory snapshot `{miner_hotkey: [executors]}` refreshed via the portal's new bulk endpoint, instead of one portal request per hotkey.

Root cause being fixed (DAH-2469): the validator wave fans out ~40 concurrent per-hotkey portal GETs; portal pool exhausts 3-4x/day; `fetch_executors` returned `[]` on ANY error, indistinguishable from a miner with no machines, so the validator zeroed the miner's whole fleet for the cycle (placeholder `11111111-1111-1111-1111-111111111111`).

## Behavior
- One bulk request per 5 min (`SNAPSHOT_TTL_SECONDS=300`) instead of ~40 per-hotkey GETs per cycle; snapshot dict is fully replaced on each successful refresh.
- Single-flight: concurrent callers past TTL await one in-flight refresh (`asyncio.Lock`).
- Stale-on-error: a failed refresh keeps serving the previous snapshot with a loud error log — a portal outage no longer looks like "miner has zero executors". 30s backoff between failed refresh attempts.
- `executor_id` filtering now local; call site (`ExecutorService.get_executors_for_validator`) unchanged.
- Residual `[]`-on-error case: cold start (miner restart) while the portal is down — same as today's behavior, logged.

## Deploy ordering
Deploy AFTER lium-miner-portal PR https://github.com/Datura-ai/lium-miner-portal/pull/155 — the bulk endpoint must exist first, otherwise every refresh 404s and (with no snapshot yet) all opted-in miners get empty lists.

## Testing
`neurons/miners/tests/test_miner_portal_api.py` (6 tests): one bulk call serves all hotkeys; single-flight under 10 concurrent calls; TTL expiry fully replaces the snapshot; stale served on refresh failure; failed-refresh backoff then recovery; local executor_id filter. Full miners suite: 19 passed. Ruff clean.

Notion: DAH-2469

### DIFF
--- a/neurons/miners/src/clients/miner_portal_api.py
+++ b/neurons/miners/src/clients/miner_portal_api.py
@@ -29,7 +29,7 @@ class MinerPortalAPI:
 
     _snapshot: dict[str, list[dict[str, Any]]] = {}
     _snapshot_fetched_at: float | None = None
-    _last_refresh_attempt_at: float = 0.0
+    _last_refresh_attempt_at: float | None = None
     _refresh_lock: asyncio.Lock = asyncio.Lock()
 
     @classmethod
@@ -61,8 +61,13 @@ class MinerPortalAPI:
         async with cls._refresh_lock:
             if cls._is_fresh():
                 return cls._snapshot
-            # after a failed refresh, don't hammer the portal on every call
-            if time.monotonic() - cls._last_refresh_attempt_at < FAILED_REFRESH_RETRY_SECONDS:
+            # after a failed refresh, don't hammer the portal on every call; None means
+            # never attempted - monotonic counts from host boot, so a 0.0 sentinel would
+            # silently swallow the very first refresh on a freshly booted host
+            if (
+                cls._last_refresh_attempt_at is not None
+                and time.monotonic() - cls._last_refresh_attempt_at < FAILED_REFRESH_RETRY_SECONDS
+            ):
                 return cls._snapshot
             cls._last_refresh_attempt_at = time.monotonic()
 
--- a/neurons/miners/src/clients/miner_portal_api.py
+++ b/neurons/miners/src/clients/miner_portal_api.py
@@ -28,7 +28,7 @@ class MinerPortalAPI:
     """
 
     _snapshot: dict[str, list[dict[str, Any]]] = {}
-    _snapshot_fetched_at: float = 0.0
+    _snapshot_fetched_at: float | None = None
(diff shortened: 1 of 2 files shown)


fortunelucky777 (approved):
Approving. The design holds up: replacing ~40 per-hotkey GETs per wave with one bulk snapshot, and — the part that actually fixes DAH-2469 — serving the previous snapshot on a failed refresh instead of `[]`, so a portal outage stops reading as "this miner has no machines".

Checked the one contract change that could itself have zeroed fleets: the bulk query adds `Miner.opt_in_status IS TRUE`, which the old per-hotkey query did not have. It's correct — the validator only routes a miner to the central miner when that miner appears in the portal's opted-in list (`subtensor_client.py:287-303`), which uses the same gate, so the two sets match.

Four non-blocking notes:

1. `miner_portal_api.py:32` — `_last_refresh_attempt_at` starts at `0.0`, but the backoff at line 65 compares it against `time.monotonic()`, which on Linux counts from host boot. If the host has been up less than 30s when the first call lands, the backoff swallows the very first refresh and returns an empty snapshot for every miner, with no log line at all. Suggest initialising it to `None` and skipping the backoff check when it has never been set.

2. `miner_portal_api.py:123` — the "portal returned `{}`" guard keeps the old snapshot *and* stamps it fresh. A portal that persistently answers `200 {}` therefore freezes the snapshot forever behind a 5-minute warning; deleted nodes and changed IPs never propagate. Worth a staleness ceiling: after N consecutive empty/failed refreshes, either accept the empty result or escalate the log to error.

3. `price_per_gpu` rides the snapshot into `ExecutorSSHInfo` (`executor_service.py:303`) and feeds the validator's hard price cap (`score_calculator.py:44`, score → 0) and the soft unrented cap (`rental_price.py:188`). A provider who lowers their price to escape a cap now stays capped for up to 5 more minutes. Either call this out in the ticket, or bypass the TTL when `executor_id` is set.

4. Companion portal PR #155: `signature_auth_miner` has its subnet-registration check commented out, so any self-generated keypair can sign and pull the full IP:port inventory of every opted-in provider in one request. Raised on that PR.

Not reviewed: the portal DAO's SQL performance at real table sizes, and `debug_routes.py:17`, which calls `get_executors_for_validator` with one argument against a two-argument signature — pre-existing, untouched here.

PR #1155, opened 2026-07-21
DAH-2465: validator protects/tolerates all filler containers on GPU-split nodes

taiberium:
Part of **DAH-2465** (GPU-split Dolphin fillers). Validator half of the fix; backend half is in lium-io-backend PR #792. Must ship together.

## Problem (found live on staging, 2026-07-21)

DAH-2465 runs one Dolphin filler container per GPU VRAM bundle. But the validator only ever knew about ONE filler per executor: `RentedExecutorsResponse.filler_containers_by_executor` was `dict[str, str]`. So on a split node:

- **`StaleContainerCleanupCheck` reaped the sibling.** `container_cleanup` protects containers in `rented_data`; it added only `get_filler_container()` (one). The other bundle's container was "stale" and force-removed (`docker rm -fv`) mid-cycle — we watched a live 2xL40S bundle get SIGKILLed ~58 min in.
- **`GpuUsageCheck` zeroed the score.** It passed only if ALL GPU processes belonged to the single mapped filler; a second bundle's processes tripped "GPU busy outside validator" → score 0.

## Fix

`filler_containers_by_executor` becomes `dict[str, list[str]]` — every active filler container on the node. An older backend that sends a single string is still accepted (field validator normalizes str -> [str], mode="before").

- `container_cleanup`: protect EVERY filler container.
- `gpu_usage`: GPU is clean when every process belongs to ANY of the node's fillers.
- `get_filler_containers()` (list) added; `get_filler_container()` kept (returns the first) for the skip-check / logging callers (`capability`, `verifyx`, `rented_machine`, `rental_verification`) — unchanged.

## Tests

352 passed in the filler/rental/cleanup/gpu-usage/backend-client scope. New: `gpu_usage` allows processes from two bundles; `container_cleanup` preserves both bundle fillers.


arhangel66 (approved):
looks good

PR #1156, opened 2026-07-21
chore: DAH-2470 — publish cache-prefetch state for the cached-template check

fortunelucky777:
## Describe your changes

When `executor.validate.cached_template` zeroes a node for a stale or missing default image, the log line does not say **why**. The executor's cache-prefetch loop already narrates every branch it takes — with digests and exception text — but `neurons/executor/src/core/logger.py` attaches only a `StreamHandler`, so that diagnosis is written on the provider's machine and never leaves it. This PR is transport for it, not new logging.

**Executor** — `run_cache_template_prefetch()` records a named outcome for every exit path and publishes `/var/lib/lium/cache_prefetch_state.json` after each sweep. Atomic write, capped at 4 KB, every failure swallowed. No pull decision, interval or retry changes; a failed pull is recorded and then **re-raised**, so the sweep still aborts and backs off exactly as before.

**Validator** — `CachedTemplateVerificationCheck` reads that file and attaches it to the event as `prefetch_state`, so it reaches Loki through the existing `LoggerSink`. Only when `should_fail` is true (roughly 2–6 nodes per cycle against 86–88 checked), so healthy nodes pay neither an SSH round trip nor log volume. The read is wrapped and can never change `passed`.

Worth knowing: **the validator's shell lands inside the executor container**, not on the host — `run.sh` starts sshd there and `docker-compose.app.yml` maps `${SSH_PORT}:22` to it. So this is a plain `cat`. No bind-mount, no compose change, no runner-image change.

### Reading the output

```
last_outcome            = sweep_ok                    <- the loop itself is healthy
images[…].last_outcome  = remote_digest_unreadable    <- this image is stuck
images[…].last_error    = 429 Client Error: Too Many Requests …
images[…].outcome_counts.remote_digest_unreadable = 208
```

That is a provider-side answer. `pull_ok` climbing while `local_digests` never moves would be ours.

A missing or unreadable file reports a distinguishable value — `missing`, `empty`, `unparseable`, `oversized`, `ssh_read_failed` — because absence is itself a finding (executor image too old, loop never started, `COMPUTE_REST_API_URL` unset).

### Notes

- The file is deliberately **not** durable. Watchtower recreates the container and resets the counters, so `started_at` and `sweep_count` are always published and a fresh reset can never be mistaken for a healthy node.
- One outcome beyond the twelve exit paths: `sweep_ok`. Without it a single transient error would pin `last_outcome` to `loop_error` forever, and a reader would chase a problem that had already passed.
- DAH-2461 means the suspected branch (registry digest unreadable → keep the stale copy, never retry) is now reached only when the backend sends no `docker_image_digest`. Still instrumented, and the counters can refute it as well as confirm it.

[DAH-2470 — Get every SN51 node to a healthy result on the validator cached-template check](https://app.notion.com/p/Get-every-SN51-node-to-a-healthy-result-on-the-validator-cached-template-check-3a4b8bfdbde9818eae41f7bcb544acfb)

- [x] I wrote tests.

Executor 50 passed, validator 1138 passed / 4 skipped. `ruff check` and `ruff format` add no new findings (four remain on untouched lines, each verified byte-identical at `main`). Four executor test modules fail to *collect* in my environment on a stale installed `datura` — reproduced on `main`, unrelated to this branch.

Performance: one extra SSH read on the failure path only, and up to ~6 KB added to those events. Healthy nodes are unchanged — asserted by a test.

Also proven end to end locally: the real prefetch loop with the registry lookup forced to 429 wrote the document, and the real check read it back through real shell commands (`docker image inspect` against a local image at a genuinely different digest, then `head -c 8192`), producing `passed=False` with the executor's state in the event.

### DIFF
--- a/neurons/executor/src/services/cache_prefetch_state.py
+++ b/neurons/executor/src/services/cache_prefetch_state.py
@@ -24,6 +24,7 @@ import copy
 import functools
 import json
 import os
+from dataclasses import asdict, dataclass, field
 from datetime import UTC, datetime
 from pathlib import Path
 
@@ -167,6 +168,36 @@ def _drop_images(doc: dict) -> None:
     doc["images"] = {}
 
 
+@dataclass(slots=True)
+class _ImageRecord:
+    """Everything known about one template, as a type rather than a bare dict.
+
+    ``slots=True`` is the whole point: a mistyped field name raises instead of
+    quietly adding an attribute nobody publishes. That matters here more than
+    usual, because every writer below runs under ``@_never_raises`` — with a dict,
+    the typo would be swallowed and the document would ship missing the field.
+    """
+
+    last_outcome: str | None = None
+    last_outcome_at: str | None = None
+    last_error: str | None = None
+    local_digests: list[str] = field(default_factory=list)
+    local_digest_first_seen_at: str | None = None
+    digest_changed_at: str | None = None
+    last_local_error: str | None = None
+    expected_digest: str | None = None
+    remote_digest: str | None = None
+    last_remote_read_ok_at: str | None = None
+    last_remote_error: str | None = None
+    last_pull_attempt_at: str | None = None
+    last_pull_ok_at: str | None = None
+    last_pull_error: str | None = None
+    last_cleanup_error: str | None = None
+    last_disk_required_bytes: int | None = None
+    last_disk_available_bytes: int | None = None
+    outcome_counts: dict[str, int] = field(default_factory=dict)
+
+
 class CachePrefetchState:
     """Accumulates what the prefetch loop knows, and publishes it as JSON.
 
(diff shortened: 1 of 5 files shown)


neurons/executor/src/services/cache_prefetch_state.py
```diff
@@ -0,0 +1,404 @@
...
+
+    # -- per-image facts --------------------------------------------------------
+
+    def _record(self, image_ref: str) -> dict:
+        return self._images.setdefault(
+            image_ref,
+            {
+                "last_outcome": None,
+                "last_outcome_at": None,
+                "last_error": None,
```
taiberium: NIT: the per-image record is a bare dict mutated by string key across ~10 methods, so a mistyped key silently adds a field. A dataclass would type-check it.

taiberium (approved):

arhangel66 (approved):
looks good

PR #1157, opened 2026-07-22
DAH-2471, report missing filler container on the active re-check

taiberium:
### Task Link

[DAH-2471](https://www.notion.so/3a5b8bfdbde981c4ab52f83101a85f52)

---

## Problem

The backend cannot see filler runs whose container vanished while the executor stays alive: they sit `RUNNING` forever, and the validator's kill re-check then treats the zombie as a live filler — the blocker for enabling DAH-2439 enforcement.

The validator already observes exactly the missing signal: it calls `GET /internal/filler-runs/{id}/active` **only** when the filler container is absent on the host. The backend just never learns that this is why it was asked.

## Solution

Make the observation explicit: the filler-liveness re-check now sends `container_missing=true` on the active re-check. The backend (Datura-ai/lium-io-backend#799) accumulates these reports and closes a run the validator keeps seeing containerless, only on sustained evidence (≥3 reports over ≥30 min).

Older backends ignore the unknown query param, so deploy order is backend first, validator second; either order is safe.

## Changes

- `clients/backend_client.py`: `get_filler_run_active(..., container_missing: bool = False)` appends `?container_missing=true`.
- `checks/rental_verification.py`: the missing-container re-check passes `container_missing=True`.
- Tests: query-string assertions in `test_backend_client.py`; flag recorded by the fake backend in `test_rental_verification_check.py`.

## Deployment Steps

Use GitHub Action. Deploy after (or independently of) the backend PR — the flag is ignored by older backends.

## Review Request

- [x] Just code review

## Other PRs

- Datura-ai/lium-io-backend#799 — DAH-2471 backend reconciliation phase consuming this flag.

## Risks

- Minimal: one boolean query param on an existing internal call; no behavior change in the check itself (same call sites, same pass/fail logic).

### Test Case 1 — unit

**Actions**: `pdm run pytest tests/test_backend_client.py tests/test_rental_verification_check.py`

**Expected Output**: green — 51 passed. (Full `tests/` run: 1135 passed; the 3 pre-existing `test_sshd_bootstrap_script.py` failures reproduce on an untouched checkout and are unrelated.)

## Checklist

- [x] Risks are documented
- [x] Tests are described


arhangel66 (approved):
looks good

PR #1158, opened 2026-07-22
Propagate pipeline_id / log context through port connectivity probes

arhangel66:
### Task Link

No Notion task yet (logging-only improvement; branch `fix/connectivity-log-ctx`).

---

## Problem

Logs emitted during executor port-connectivity verification carried no correlation identifiers. Lines like `testing 40 ports (attempt 1/2, timeout=60s)`, `batch verification failed, trying fallback`, or `fallback: port 8003 ok (2/10)` had no `pipeline_id`, `job_batch_id`, `executor_uuid`, `executor_ip`, or `miner_hotkey`, so there was no way to tell which validation pipeline / executor / miner a given line belonged to. Only `dind_probe` received a partial context (`executor_uuid`, `executor_ip`).

## Solution

Thread a `log_ctx` dict from `PortConnectivityCheck` down through the whole probing chain and attach it to every log line via the existing structured-logging helpers (`_m(...)` + `get_extra_info(...)`). The message text stays a compact one-liner; identifiers go into the JSON formatter's structured `extra` field, so log lines are not widened.

## Changes

- `checks/port_connectivity.py`: pass `log_ctx` (`pipeline_id`, `job_batch_id`, `miner_hotkey`, `executor_uuid`, `executor_ip`) into `verify_ports`.
- `executor_connectivity/service.py`: accept `log_ctx`, forward to the orchestrator, use it in the sysbox-fallback log and convert the error log to structured form.
- `executor_connectivity/orchestrator.py`: accept `log_ctx`, enrich once with executor identity, forward to both `port_probe` and `dind_probe` (removes the duplicated inline dict).
- `executor_connectivity/port_probe.py`: accept/forward `log_ctx`; structured fallback warning.
- `executor_connectivity/port_verifiers.py`: `BatchVerifier` / `FallbackVerifier` (and `_attempt`) accept `log_ctx`; all `%`-style logs converted to `_m(..., extra=get_extra_info(...))` with the message text preserved.
- `tests/test_port_connectivity_check.py`: `DummyConnectivityService.verify_ports` gains the `log_ctx` kwarg to match the new signature.

## Deployment Steps

Use GitHub Action.

## Review Request

- [x] Just code review

## Other PRs

None.

## Risks

- Logging-only change; no behavior or control-flow change. All new `log_ctx` parameters are optional (default `None`), so existing callers and tests are unaffected.
- Log messages moved from `%`-style args to f-strings inside `_m(...)`; the visible text is unchanged, but records now carry a structured `extra` object.

[...]

PR #1159, opened 2026-07-22
fix(validator): SemiBatchVerifier tier so firewalled nodes verify up to 50 ports

arhangel66:
## Problem

The validator's port-connectivity check under-counts verified ports on any executor host running `ufw` with a default-deny inbound policy (only SSH allowed) — standard host hardening. `verified_ports` directly caps how many renter/filler containers the platform places on the node, so these hosts get throttled far below their hardware (and DPHN filler splitting suffers).

Mechanism (reproduced live on a prod executor, subnet `13567ffb-c410-4edc-b09f-0e47c0363bac`):
- The **batch** verifier uses `--network=host`; its netcat listeners bind directly on host ports and inbound traffic to them traverses the INPUT chain → **ufw drops it** → `0/N` verified.
- After 2 attempts it fell back to a **sequential verifier that publishes with `-p` (bypasses ufw via nat/DOCKER DNAT, so it works) but tested only the first 10 ports**.
- Net: firewalled nodes cap at ≤10 verified ports regardless of how many they actually opened.

Repro on such a host: an external TCP connect to a `--network=host`-bound port times out (dropped by ufw INPUT), while a `-p`-published port connects fine. Default docker runtime is plain `runc` (not sysbox) — ufw is the blocker.

## Fix

Add a new **`SemiBatchVerifier`** tier between `BatchVerifier` (`--network=host`) and the existing `FallbackVerifier` (sequential, one container per port). `PortProbe` now cascades **batch → semi-batch(50) → fallback(10)**, each tier falling through to the next when it verifies nothing.

`SemiBatchVerifier` publishes up to 50 ports in ONE container via `-p external:internal` and reuses the existing batch netcat script + concurrent `test_many`. Published ports reach their listeners through the nat/DOCKER DNAT chain — the same path renter pods use — so they bypass ufw's default-deny INPUT. One container + concurrent probes keeps it fast. If the publish container cannot start (e.g. a candidate port is already held by a live container — rare, since `PortSelector` already excludes rented ports), it verifies nothing and the cascade falls through to the existing sequential fallback.

Existing `BatchVerifier` and `FallbackVerifier` are left untouched; the change is additive.

- `port_verifiers.py`: new `SemiBatchVerifier` class
- `port_probe.py`: middle cascade tier + constructor param
- `ioc.py`, `core/validator.py`: wire the new tier into both DI paths
- `test_port_verifiers.py` (SemiBatchVerifier: single-container publish, external→internal mapping, cap at 50, empty-on-start-failure), `test_port_probe.py` (3-tier cascade)

## Testing

`pytest tests/executor_connectivity/ tests/test_port_connectivity_check.py` → 30 passed. Full validator suite otherwise green except pre-existing/unrelated failures (`test_sshd_bootstrap_script` timing, `test_port_mapping_dao` DB-fixture errors) that do not touch these modules.

**Verified end-to-end on staging** (full detail in the comment below). Stand: staging executor `02fdbaf8-2598-40c5-9597-23843ec7bab7` with 200 declared ports and `ufw` default-deny — a copy of the prod node above. Deployed this branch to the staging validator and compared against the previous build:

| metric | before | after |
| --- | --- | --- |
| `verified_ports` | **10** | **50** |
| verification wall clock | 55.2s | 24.4s |
| control nodes (no ufw) | 299 / 9 | 299 / 9 — unchanged |

Healthy nodes never reach the new tier. The firewalled node also gets *faster*: 50 ports probed concurrently in 4.7s vs ~20s for 10 sequential.

## Notes

- #1158 (log-context threading over the same files) is merged; this branch was rebased onto `main` and retargeted.
- Not addressed here (kept the existing class untouched): `FallbackVerifier` still publishes `-p internal:internal`, which is wrong for `port_mappings` miners where external != internal. It is now only the last-resort 3rd tier; `SemiBatchVerifier` uses the correct `-p external:internal`. Worth a separate follow-up.

### DIFF
--- a/neurons/validators/src/services/executor_connectivity/port_probe.py
+++ b/neurons/validators/src/services/executor_connectivity/port_probe.py
@@ -2,7 +2,7 @@ import logging
 
 from core.utils import _m, get_extra_info
 from services.executor_connectivity.models import PortPair, PortProbeResult
-from services.executor_connectivity.port_verifiers import BatchVerifier, FallbackVerifier
+from services.executor_connectivity.port_verifiers import BatchVerifier, FallbackVerifier, SemiBatchVerifier
 
 logger = logging.getLogger(__name__)
 
@@ -13,9 +13,11 @@ class PortProbe:
     def __init__(
         self,
         batch_verifier: BatchVerifier,
+        semi_batch_verifier: SemiBatchVerifier,
         fallback_verifier: FallbackVerifier,
     ):
         self.batch_verifier = batch_verifier
+        self.semi_batch_verifier = semi_batch_verifier
         self.fallback_verifier = fallback_verifier
 
     async def probe(
@@ -36,7 +38,19 @@ class PortProbe:
 
         if not successful:
             logger.warning(
-                _m("batch verification failed, trying fallback", extra=get_extra_info(log_ctx))
+                _m("batch verification failed, trying semi-batch", extra=get_extra_info(log_ctx))
+            )
+            successful, failed = await self.semi_batch_verifier.verify(
+                ports,
+                ssh_client=ssh_client,
+                host=host,
+                max_ports=50,
+                log_ctx=log_ctx,
+            )
+
+        if not successful:
+            logger.warning(
+                _m("semi-batch verification failed, trying fallback", extra=get_extra_info(log_ctx))
             )
             successful, failed = await self.fallback_verifier.verify(
                 ports,
--- a/neurons/validators/src/services/executor_connectivity/port_verifiers.py
+++ b/neurons/validators/src/services/executor_connectivity/port_verifiers.py
@@ -3,6 +3,7 @@ import logging
 import uuid
 
 import aiohttp
+from asyncssh import SSHClientConnection
 
 from core.utils import _m, get_extra_info
 from services.executor_connectivity.container_runner import ContainerRunner
@@ -219,7 +220,7 @@ class SemiBatchVerifier:
         self,
         ports: list[PortPair],
         *,
-        ssh_client,
+        ssh_client: SSHClientConnection,
         host: str,
         max_ports: int = 50,
         log_ctx: dict | None = None,
@@ -240,23 +241,41 @@ class SemiBatchVerifier:
             )
         )
 
-        # short self-kill window: a missed cleanup would otherwise hold every published
-        # port (and its docker-proxy) against the next cycle's verification
-        start_result = await self.runner.run(ssh_client, container_name, script, publish_flags, 20)
-        if not start_result.ok:
-            logger.warning(
-                _m(
-                    f"semi-batch start failed: status={start_result.status} logs={start_result.logs}",
-                    extra=get_extra_info(log_ctx),
+        # any error below must not escape: an exception here would zero the whole
+        # verification (fatal PortCountCheck) instead of falling through to the
+        # sequential fallback tier
+        started = False
+        try:
+            # short self-kill window: a missed cleanup would otherwise hold every published
+            # port (and its docker-proxy) against the next cycle's verification
+            start_result = await self.runner.run(ssh_client, container_name, script, publish_flags, 20)
+            if not start_result.ok:
+                logger.warning(
+                    _m(
+                        f"semi-batch start failed: status={start_result.status} logs={start_result.logs}",
+                        extra=get_extra_info(log_ctx),
+                    )
                 )
+                return [], ports_to_test
+
+            started = True
+            async with aiohttp.ClientSession() as session:
+                successful, failed = await self.port_tester.test_many(session, host, ports_to_test, token)
+
+        except Exception as e:
+            logger.error(
+                _m(f"semi-batch: error: {str(e)[:100]}", extra=get_extra_info(log_ctx))
             )
             return [], ports_to_test
 
-        try:
-            async with aiohttp.ClientSession() as session:
-                successful, failed = await self.port_tester.test_many(session, host, ports_to_test, token)
         finally:
-            await self.runner.cleanup(ssh_client, container_name)
+            if started:
+                try:
+                    await self.runner.cleanup(ssh_client, container_name)
+                except Exception as e:
+                    logger.warning(
+                        _m(f"semi-batch: cleanup failed: {str(e)[:50]}", extra=get_extra_info(log_ctx))
+                    )
 
         logger.info(
             _m(f"semi-batch: {len(successful)}/{len(ports_to_test)} verified", extra=get_extra_info(log_ctx))
--- a/neurons/validators/src/services/executor_connectivity/port_verifiers.py
+++ b/neurons/validators/src/services/executor_connectivity/port_verifiers.py
@@ -200,3 +200,65 @@ class FallbackVerifier:
             _m(f"fallback: {len(successful)}/{len(ports_to_test)} verified", extra=get_extra_info(log_ctx))
         )
         return successful, failed
+
+
+class SemiBatchVerifier:
+    """Verifies ports by publishing a whole batch in ONE container via -p.
+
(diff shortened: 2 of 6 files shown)


arhangel66:
## Verified end-to-end on staging (2026-07-23)

Ran the full validator cycle against a stand built to match the prod node this PR was written for:
staging executor `02fdbaf8-2598-40c5-9597-23843ec7bab7` (64.247.196.173, L40S), `RENTING_PORT_RANGE`
widened to `40000-40199` (200 ports), `ufw` with `Default: deny (incoming)` + `allow 22`.

**Before** — validator image `654752d` (staging's previous build):

```
attempt 1 failed, retrying in 2s
progress: 0/200 verified                     <- --network=host listeners dropped by ufw
batch verification failed, trying fallback
fallback: testing 10 ports sequentially
fallback: 10/10 verified
verification complete total_time=55.24s ok=10 port_range=40000-40199
```

**After** — validator image `6c4244f` (this PR's head, deployed via `staging_branch_deployment.yml`):

```
progress: 0/200 verified                     <- same first tier, still dropped
batch verification failed, trying semi-batch
semi-batch: publishing 50 ports in one container
semi-batch: 50/50 verified                   <- 4.7s
verification complete total_time=24.38s ok=50 port_range=40000-40199
```

| metric | before | after |
| --- | --- | --- |
| `verified_ports` (backend DB) | **10** | **50** |
| verification wall clock for that node | 55.2s | 24.4s |
| control node 185.216.20.168 (no ufw, 300 ports) | 299 | 299 |
| control node 149.36.1.151 (no ufw, 10 ports) | 9 | 9 |

Two things the numbers show beyond the port count:

- **No change for healthy nodes.** They still succeed on the first tier, so the new tier is never
  reached — both controls came back byte-identical across the deploy.
- **The firewalled node gets faster, not slower.** semi-batch probes 50 ports concurrently in 4.7s,
  where the sequential fallback spent ~20s on 10 ports; total verification dropped 55.2s -> 24.4s.

Also worth noting for whoever reads the logs: the summary line reports `batch=ok` and
`100% available` in *both* runs — it is derived from the final success count, not from which tier
actually produced it. That label is why this was invisible in prod for so long. Untouched here
(out of scope for this PR).

Staging was restored afterwards: ufw disabled, port range back to `40000-40009`, validator
redeployed to the branch it was running before.

neurons/validators/src/services/executor_connectivity/port_verifiers.py
```diff
@@ -200,3 +200,65 @@ async def verify(
...
+                    f"semi-batch start failed: status={start_result.status} logs={start_result.logs}",
+                    extra=get_extra_info(log_ctx),
+                )
+            )
+            return [], ports_to_test
+
+        try:
+            async with aiohttp.ClientSession() as session:
+                successful, failed = await self.port_tester.test_many(session, host, ports_to_test, token)
+        finally:
```
taiberium, line 271: **Critical:** an SSH error in `cleanup` or `runner.run` throws past the fallback tier and zeroes the executor at fatal `PortCountCheck`. `FallbackVerifier` guards its cleanup, this one doesn't.

neurons/validators/src/services/executor_connectivity/port_verifiers.py
```diff
@@ -200,3 +200,65 @@ async def verify(
...
+        ports: list[PortPair],
+        *,
+        ssh_client,
+        host: str,
+        max_ports: int = 50,
+        log_ctx: dict | None = None,
+    ) -> tuple[list[PortPair], list[PortPair]]:
+        """Publish up to max_ports in one container, then probe them concurrently."""
+        log_ctx = log_ctx or {}
+        ports_to_test = ports[:max_ports]
```
taiberium, line 230: NIT: always the same first 50 ports, so one occupied port kills the tier for that node every cycle. Worth trying the next slice when the container won't start.

neurons/validators/src/services/executor_connectivity/port_probe.py
```diff
@@ -36,7 +38,19 @@ async def probe(
...
         if not successful:
             logger.warning(
-                _m("batch verification failed, trying fallback", extra=get_extra_info(log_ctx))
+                _m("batch verification failed, trying semi-batch", extra=get_extra_info(log_ctx))
+            )
+            successful, failed = await self.semi_batch_verifier.verify(
+                ports,
+                ssh_client=ssh_client,
+                host=host,
+                max_ports=50,
```
taiberium, line 47: NIT: `max_ports=50` here repeats the verifier's own default. Pick one place.

neurons/validators/src/services/executor_connectivity/port_verifiers.py
```diff
@@ -200,3 +200,65 @@ async def verify(
...
+
+    def __init__(self, port_tester: PortTester, runner: ContainerRunner):
+        self.port_tester = port_tester
+        self.runner = runner
+
+    async def verify(
+        self,
+        ports: list[PortPair],
+        *,
+        ssh_client,
```
taiberium: NIT: `ssh_client` has no type here, and `ContainerRunner.run` already types it as `SSHClientConnection`.

taiberium (changes requested):
Requested changes.

neurons/validators/src/services/executor_connectivity/port_verifiers.py
arhangel66: Fixed in 99edd7e3: the whole body is now guarded — any exception logs and returns `([], ports)` so the cascade falls through to the sequential fallback instead of zeroing the executor at fatal PortCountCheck. Cleanup runs only if the container started and is guarded separately (same pattern as FallbackVerifier). Added 3 tests.

neurons/validators/src/services/executor_connectivity/port_verifiers.py
arhangel66: By design. Everything we know to be occupied is already excluded upstream: PortSelector drops rented ports (incl. filler) and StaleContainerCleanupCheck reaps orphaned containers before the port checks. What's left is only ports held by things outside our knowledge (miner's host services / foreign containers) — that residual case is exactly what the 3rd sequential tier is for, so the node still gets its 10 ports instead of 0.

neurons/validators/src/services/executor_connectivity/port_probe.py
arhangel66: Keeping the explicit param: the call site mirrors the fallback line right below it (`max_ports=10`), and the cascade's sizing reads in one place in `probe()`. The class default just makes the verifier usable standalone.

neurons/validators/src/services/executor_connectivity/port_verifiers.py
arhangel66: Done in 99edd7e3, typed as `SSHClientConnection`.

fortunelucky777 (approved):
Approve. Additive middle tier — SemiBatchVerifier publishes up to 50 ports via -p (DNAT path) so ufw default-deny nodes verify 50 instead of capping at 10.

Verified: no missed PortProbe construction site (both DI paths + test updated), downstream verified_ports/PortCountCheck contract unchanged, and every failure path falls through to the sequential fallback instead of zeroing the probe.

Two non-blocking notes:
- All-or-nothing publish: one already-allocated port among the 50 fails the whole tier and drops the node to the 10-port fallback for that cycle (overlaps the DAH-2272 stale rented_ports window). Graceful, never worse than pre-PR; worth a follow-up to chunk or drop-and-retry.
- The port_test_pub_* container holds up to 50 published host ports for ~20s and isn't matched by wait_for_port_check_containers' rental sweep; relies on the 20s self-kill + 90s create-retry, same as existing port_test_* containers.

taiberium (approved):

PR #1161, opened 2026-07-23
DAH-2475, filler cache volumes + real failure reasons + Redis pool hardening

taiberium:
> [!IMPORTANT]
> **Updated 2026-07-24 — review before the sections below.** After a high-effort adversarial review and a live staging run:
>
> 1. **The cache affordability decision now lives HERE** (`select_affordable_cache_volumes`, option E): volumes that already exist mount free of charge (no disk check — mounting costs nothing); only a missing volume is judged, against live `df` on the real docker root. The sweep runs BEFORE the decision, keyed on the *requested* names, so a renamed cache can no longer strand a tight node with a dead 37 GB set.
> 2. **`msg`/`detail` protocol split** in `FailedContainerRequest`: `msg` is the renter-safe headline, `detail` the full structured diagnosis (exception, host, failure step). Old backends simply see the headline — the leak window is closed structurally, not by string parsing.
> 3. **The rental reclaim fires on non-storage-limit nodes too** (`volume_limit_gb=None` means "the renter may use the whole disk", not "no disk needed"), the disk-tight reclaim measures the real docker root through the daemon and reports only what actually got removed, and the Redis retry now actually retries (`redis.asyncio.retry.Retry` — the sync class silently performed zero retries; verified empirically).
>
> Live-verified on staging 2026-07-24: new DPHN container mounted both cache volumes and warm-started straight to mining (weights volume unchanged at 22.8G, `connected to lighthouse`, 744/744 ok); a 13h orphaned filler container was reaped on the next create.

### Task Link

[DAH-2475](https://app.notion.com/p/Filler-launch-resilience-TP-valid-bundles-soft-backoff-in-flight-cap-admin-restart-API-real-fai-3a5b8bfdbde981cab994c12bb9190173)

---

## Problem

Three findings from the 2026-07-22 08:35 UTC incident, where 17 DPHN bundle creates failed across 3 executors and put healthy nodes into a 12h launch backoff:

1. **The real error never left the validator.** Every failure reached the backend as a bare `"ExceptionError: Resulted in an exception"`. `_StructuredMessage.__str__` returns only the headline, so the `extra` dict holding the actual exception was dropped before the message was sent — the true cause was in neither `filler_run.failure_reason` nor the log line, and every incident cost a manual investigation.

2. **A Redis blip destroyed a finished container.** At `failure_step=finalize`, `add_rented_pod` raised `Timeout connecting to server` and the exception fell into `cleanup_failed_container_creation` — tearing down a container that was already built and running (~40 min of DPHN model download thrown away), marking the run FAILED and backing the node off. That hash is a cache the backend rebuilds wholesale every ~10 min, and the backend already learns the container from the `ContainerCreated` callback.

3. **Root cause of the blip: the Redis client had no resilience config at all.** `aioredis.from_url(url)` with no options means an unbounded pool, no retries, no timeouts and no health checks — so a wave of concurrent creates opened a thundering herd of fresh connections (51 `Timeout connecting to server`/min at 08:35, again at 08:41 with the next wave, zero outside the waves), and with no retry every operation caught in that window failed outright.

Separately, a DPHN cold start re-downloads ~37 GB (23 GB HF weights + 12 GB worker runtime) on every container start, because the cache lives in the ephemeral `volume_{run_id}` that stop deletes — a ~40 min cold start before the node earns anything.

## Solution

- **Real failure reasons**: `_handle_container_error` now ships `to_full_string()` — headline plus the `extra` dict — so one change covers all its call sites; the create path in `docker_service` does the same. Logging still passes the structured object, so `extra` keeps landing in Loki as fields.
- **Finalize is best-effort**: the rented-pod cache write is wrapped so a Redis failure logs and keeps the container instead of destroying it.
- **Redis pool hardening**: `BlockingConnectionPool` (bounded at 64, callers *queue* rather than get `MaxConnectionsError`, which the default pool raises before the retry wrapper is even reached), explicit socket/connect timeouts, keepalive, 30s health checks, and 3 retries with exponential backoff on `ConnectionError`/`TimeoutError`.
- **Persistent DPHN cache volumes**: FILLER-only named volumes mounted from the request's `cache_volumes`, outside the `volume_` prefix so no GC path and no delete request touches them. Absorbed from #1160 (which this supersedes), with its validation kept (name regex, no `volume_` prefix, no host-path / `:` / `..` targets).
- **Create-time sweep** (new, not in #1160): removes every `dphn_cache_*` volume the current request does NOT name. The backend now bakes the model+runtime version into the volume NAME, so a Dolphin model update renames the cache and the previous set becomes stale by name and is swept — the invariant is at most ONE model's cache per host. Without this, #1160's `_vN` kill switch would orphan the old volume and leak another ~37 GB per bump.

- **Reclaim, two levels** (new): the sweep only removes STALE versions, never reclaims — a node's free disk is low precisely BECAUSE the cache is downloaded, so reclaiming there would raise free disk, the backend would grant the cache again next cycle, and the node would re-download ~37 GB forever. Reclaim therefore lives where the decision is made on real free space: (1) at customer-rental create, if the requested volume does not fit beside the cache (+20 GB headroom), the cache goes before volume sizing — the disk belongs to the renter; (2) a periodic backstop removes `dphn_cache_*` when free disk is under the rental listing floor (100 GB) AND no filler container is RUNNING, rescuing a node that would otherwise be delisted and unreachable by both renters and fillers. It runs from `StaleContainerCleanupCheck`, which is the only caller — the create-time sweep never reclaims on purpose.

Two deliberate non-actions in the sweep, both to avoid thrash:
- no cache requested → sweep does nothing (a node denied the cache has low free disk *because* the cache is downloaded; deleting would raise free disk, the next cycle would grant it again, and the node would re-download ~37 GB forever — reclaim belongs to the rental/disk-pressure path);
- a live filler sibling present → skip the listing entirely (docker cannot remove an in-use volume, so it is provably a no-op).

## Changes

- `payload_models/payloads.py`: `CacheVolume {name, target}` + optional FILLER-only `ContainerCreateRequest.cache_volumes`.
- `services/docker_service.py`: `_validate_cache_volume`, `_build_cache_volume_mounts`, mounts appended in `_build_rental_container_run_spec`; `sweep_stale_cache_volumes` + `_find_cache_volumes_to_sweep`; `_cache_rented_pod_best_effort`; full-text failure message on create failure.
- `services/miner_service.py`: `_handle_container_error` accepts the structured message and converts once; the sites that construct response models directly convert explicitly.
- `services/redis_service.py`: blocking pool + timeouts + retry + health check.
- `services/container_cleanup.py`: `reclaim_dphn_cache_when_disk_is_tight` (disk-pressure backstop); `services/const.py`: single home for `DPHN_CACHE_VOLUME_PREFIX`.
- `tests/test_cache_volumes.py` (22) + `tests/test_dphn_cache_reclaim.py` (5): mounts, validation, sweep semantics, sibling skip, best-effort finalize, and both reclaim levels.

## Deployment Steps

Use GitHub Action. **Deploy this BEFORE the backend PR** — until the validator ships, the backend's `cache_volumes` field is simply dropped and DPHN cold-starts exactly as today, so either order is safe, but validator-first makes the cache take effect on the backend release.

The image PR should ship first of all three (see Other PRs).

## Review Request

- [x] Just code review

## Other PRs

- computenet-docker-images (ship first): flock guard so siblings sharing the cache volume cannot corrupt the worker binary
- lium-io-backend (ship last): versioned cache names, in-flight create cap, count-based backoff, admin roll-strategy endpoint

## Risks

- **Supersedes #1160** (same mechanism, absorbed here with the sweep added) — close it rather than merging both.
- Redis pool cap of 64: callers now queue up to 10s instead of opening unbounded connections. If a real workload needs more concurrency the symptom is latency, not errors; raise `REDIS_MAX_CONNECTIONS`.
- The cache volumes are shared by all filler containers on a node — the worker binary is protected by the image's flock, and HF weights rely on `huggingface_hub`'s own file locking.

## Verification

- Validator suite: **1170 passed, 8 skipped**. (3 `test_sshd_bootstrap_script` failures are pre-existing and reproduce on a clean `main`.) The 3 `test_sshd_bootstrap_script` failures are pre-existing and reproduce on a clean checkout of `main`.

## Verified on staging (2026-07-23)

Deployed this branch to staging and forced a DPHN launch on a 4xL40S node. The validator's own run-spec log shows the mechanism working end to end:

```
volume_47389aac-…                                    -> /root          (ephemeral, deleted on stop)
dphn_cache_hf_nvidia-qwen3-6-35b-a3b-nvfp4-20e94d55  -> /root/.cache
dphn_cache_dp_0-0-3-544c8fc9                         -> /opt/dolphinpod
```

**A bug the unit tests missed, found only by running this on a node:** the sweep never fired. Its "skip while a live filler sibling holds the cache" guard counted the run's OWN container — the backend creates the filler_run row (STARTING) before it builds the create request, so `filler_<own_id>` is already in `active_container_names` and the guard was always true. Stale cache sets would have accumulated forever, which is the exact leak the versioned names exist to prevent. Fixed by excluding the container being created, with a regression test; after the fix the sweep does the right thing on the node:

```
stale = ["dphn_cache_dp_0-0-3-544c8fc9"]                     <- previous image tag, removed
kept  = ["dphn_cache_dp_0-0-6-370292b2",                      <- current runtime
         "dphn_cache_hf_nvidia-qwen3-6-35b-a3b-nvfp4-…"]      <- weights, model unchanged
```

Note only the runtime set is swept on an image bump: the weights volume is keyed on the model, so the 23GB of weights survive a version change.

Both bundles on the node received the SAME two cache volumes (their `volume_<run_id>` differ), so the sharing is real; the nested `/root` + `/root/.cache` mounts are accepted by dockerd; and the names carry the model and the image tag, which is what makes a version bump sweep the old set.

## Review round — three defects found in this PR and fixed

Review (thanks) caught three things my own tests had not, all of them mine:

1. **The Redis hardening broke the validator's main inbound channel.** `socket_timeout` leaked into
   `pubsub.listen()`, which is a blocking read on channels that are idle most of the time, so the
   subscription died after 10 s of silence and reconnected in a loop — and the loop never called
   `aclose()`, so with the now-bounded pool it also leaked a connection per round. Measured against a
   live Redis: before the hardening `listen()` survived 14 s of silence, after it died at exactly
   10.0 s. Fixed by giving subscriptions their own small pool with no `socket_timeout` (commands keep
   theirs) and closing the pubsub in a `finally`. Re-measured: 14 s of silence survived, command
   timeout still enforced.
2. **The disk-pressure reclaim was never called.** It existed only in its own definition and its
   tests — zero call sites in `src/` — while this PR's description claimed the level was done. Now
   wired into `StaleContainerCleanupCheck` and covered by a test that asserts the call happens.
3. **Its "no live filler" guard used `docker ps -a`,** which lists stopped containers, so one
   long-exited `filler_*` would have wedged the guard shut forever. Now a RUNNING-only check.


arhangel66 (approved):
looks good

PR #1162, opened 2026-07-23
feat: DAH-2251 — pay referral incentive from residual burn, split across referrers by EMA

fortunelucky777:
## Describe your changes

Pay referral rewards through **subnet mining emission** instead of a backend alpha payout. The validator draws a **Referral Pool** from the residual burn and splits it across the miners who referred paying customers, in proportion to each referrer's EMA of referred revenue (read from the backend feed).

This replaces the original **DAH-2481** carve-out, which is superseded. That version took the cut **from every miner** (diluting them) and paid **one funding wallet**. This version takes the cut **from residual burn** (miners are never diluted) and splits it **across many referrer hotkeys by EMA**. The work now sits under DAH-2251 alongside the backend rail.

### How it works

`incentive/default.py::_apply_referral_pool`, called from `calculate_final_weights`. `RentalPriceIncentive` extends `DefaultIncentive` and does not override `calculate_final_weights`, so the algorithm running in production inherits this unchanged.

- `share = clamp(REFERRAL_EMISSION_SHARE, 0.0, 1.0)` — default **0.0 → inert**, byte-identical to today. The clamp preserves NaN, and the `share > 0` guard rejects it.
- `referral_pool = min(share * total_score, burn_total)`, where `burn_total` is the burn left **after** the rental-share (idle) pool takes its first claim. A short burn therefore shrinks referral, never rental-share and never miners.
- The pool is subtracted from burn hotkeys proportionally, then added to referrer hotkeys by EMA share. Value only moves burn → referrers, so the total is conserved and every non-referrer miner keeps its exact normalized weight.
- **Eligibility:** a referrer must be registered on the subnet this cycle, have positive EMA, and **not** be a burn slot — a burner cannot also collect referral emission.
- **Fail closed:** unset/zero/NaN share, an unreachable/stale/malformed/empty feed, no residual burn, or no eligible referrer each leave the weight vector untouched.

### Epoch numbering (cross-repo, load-bearing)

`core/validator.py` passes the current subnet epoch into `calculate_final_weights` for the feed's staleness guard, computed as:

```
epoch_index = (block + netuid + 1) // (tempo + 1)
```

This **must** match the backend's `services/subtensor.py::get_current_epoch_index` exactly. A plain `block // tempo` diverges from it and the gap grows without bound, so the feed would read as permanently stale the moment `REFERRAL_EMISSION_SHARE > 0`. Any failure computing it just disables the staleness check, never the cycle; the client still fails closed on transport, parse, and empty.

### Supporting changes

- `clients/referral_feed_client.py` — new, fail-closed client for `<COMPUTE_REST_API_URL>/v1/referral-weights`. Transport error, malformed JSON, missing key, empty, or stale all return `{}`. Deliberately **not** built on the fail-**open** `SharedConfigClient`.
- `core/config.py` — adds `REFERRAL_EMISSION_SHARE` (float, default `0.0`), `REFERRAL_FEED_URL`, `REFERRAL_FEED_MAX_STALENESS_EPOCHS` (default `3`). Purely additive: **+24/−0**.
- `core/validator.py` — computes and passes `current_epoch`.
- `incentive/base.py` — ABC signature gains `current_epoch`.
- `tests/test_referral_emission_pool.py` — 9 tests: no dilution, total conserved, pool equals `min(share×total, burn)` with gains matching losses, cap at burn total, fail-closed on empty feed and on non-positive/NaN share, both eligibility filters, and no-op at zero burn.
- `tests/test_referral_feed_client.py` — 7 tests: happy path, drops non-positive EMA, and fail-closed on HTTP error, malformed JSON, missing key, empty, and stale epoch.

Diff is 7 files, **+564/−6**. CI green: Validators Tests, Executor Tests, CodeQL, Analyze (python/actions).

[DAH-2251](https://www.notion.so/DAH-2251)

Supersedes [DAH-2481](https://www.notion.so/DAH-2481) (single-wallet carve-out — abandoned).

Backend feed + measurement: Datura-ai/lium-io-backend#774. **That PR must merge first** — it serves `/v1/referral-weights`, which this reads. Until it does, the client fails closed and this branch is inert regardless of the share setting.

Provider-facing UI for the same rail: Datura-ai/lium-miner-portal#157 and Datura-ai/lium-miner-portal-frontend#155.

- [ ] Default-inert verified (`REFERRAL_EMISSION_SHARE=0.0` → prod-snapshot incentive unchanged)
- [ ] Miners never diluted (pool drawn only from residual burn)
- [ ] Fail-closed on feed error / stale / empty
- [ ] Epoch numbering matches the backend feed (`(block + netuid + 1) // (tempo + 1)`) against a live feed
- [ ] Testnet: `set_weights` lands the configured slice on referrer UIDs from burn (before enabling in prod)
- [ ] Share value for launch agreed with ops (currently unset; 5% discussed)


arhangel66 (approved):
The mechanism looks right — single merge point, fail-closed coldkey guard, conserved
totals, clean tests. Two things before merge, one blocking.

**Blocking — the funding wallet never reaches the weight path.**
`SubtensorClient.fetch_miners` keeps only `neuron.axon_info.is_serving or neuron.uid in
(settings.BURNERS + settings.NEW_BURNERS)`. The rollout in the description only asks ops
to *register* the referral wallet on SN51, and a registered hotkey that doesn't run an
axon has `is_serving == False`, so it is dropped before `calculate_final_weights` ever
sees it. `_apply_referral_carveout` then finds no UID and no-ops; `set_weights` builds
its `uids`/`weights` arrays from the same filtered list, so even a score under that
hotkey would never reach chain. The new tests pass the funding neuron straight into
`calculate_final_weights`, so they don't cover this.

Please include the configured `REFERRAL_EMISSION_UID` in the weight-bearing neuron path,
and add a test that starts from a metagraph containing a referral neuron with
`axon_info.is_serving == False` and asserts it appears in the final uids/weights.

**Non-blocking — silent failure has no log.** When `share > 0` but the funding UID isn't
in the miner list, `_apply_referral_carveout` returns without logging anything (only the
coldkey mismatch warns). That's exactly the failure mode above, and it leaves no trace.
Worth a warning.

**Non-blocking — reported earnings will drift by the share.** `result.incentive`
published on `MACHINE_SPEC_CHANNEL` (`miner_service.py`) and `usd_per_epoch` in
`RentalPriceEstimate` (`rental_price.py`) are both computed before the carve-out. Once
the share is non-zero, provider-facing expected earnings will overstate actual emission
by exactly that share. Either scale them by `1 - share` or state explicitly that the
backend reconciles it.

Full validator suite on this HEAD: 1135 passed, 8 skipped; the 3 failures + 9 errors
(`test_sshd_bootstrap_script`, `test_port_mapping_dao`) reproduce on the base commit and
are unrelated to this PR. `tests/test_referral_emission_carveout.py`: 8 passed.

neurons/validators/src/core/config.py
```diff
@@ -165,6 +165,30 @@ class Settings(BaseSettings):
     }
     ENABLE_NEW_BURN_LOGIC: bool = True
 
+    # DAH-2481 — referral incentive paid via emission. A fixed fraction of the miner
```
taiberium: **NIT:** this comment says the share comes from miner (non-burn) emission, but the pool is drawn from residual burn and miners are never diluted. Fix the wording.

neurons/validators/src/core/config.py
```diff
@@ -165,6 +165,30 @@ class Settings(BaseSettings):
...
+    # redirected until ops sets a non-zero share at launch (coordinated with the backend
+    # referral rail).
+    REFERRAL_EMISSION_SHARE: float = Field(env="REFERRAL_EMISSION_SHARE", default=0.0)
+
+    # DAH-2481 — fail-closed feed of epoch-stable referral EMA weights, served by the
+    # backend at `<COMPUTE_REST_API_URL>/v1/referral-weights`. Defaults track the same
+    # base URL as COMPUTE_REST_API_URL; override independently via REFERRAL_FEED_URL if
+    # the feed is ever hosted elsewhere.
+    REFERRAL_FEED_URL: str = Field(
+        env="REFERRAL_FEED_URL", default="https://lium.io/api/v1/referral-weights"
```
taiberium: **NIT:** the default does not track `COMPUTE_REST_API_URL`, it is a hardcoded prod URL. On staging the feed silently points at prod unless `REFERRAL_FEED_URL` is also set.

neurons/validators/src/clients/referral_feed_client.py
```diff
@@ -0,0 +1,91 @@
+"""Fail-closed client for the backend's epoch-stable referral-weights feed (DAH-2481).
```
taiberium: **NIT:** the new files still reference DAH-2481, which this PR supersedes. Update the ticket refs to DAH-2251.

taiberium (approved):

PR #1163, opened 2026-07-23
DAH-2419: scrape real resident VRAM (nvmlMemory.used) as gpu.memory_used_mb

taiberium:
Ticket: [DAH-2419](https://app.notion.com/p/39eb8bfdbde981688042efca7f0a8d65)

**Deploy this BEFORE the backend counterpart** ([lium-io-backend#801](https://github.com/Datura-ai/lium-io-backend/pull/801)) — that PR reads the field this one starts sending.

## Problem

`gpu.memory_utilization` is `nvmlUtilization_t.memory` — the percent of time the memory **bus** was being read or written, not how much VRAM is occupied. The backend was multiplying it by card capacity as if it were an occupancy fraction, so a healthy Dolphin worker holding its weights but idle between requests reported ~0 "VRAM".

Prod check (187 RUNNING DPHN fillers): the single largest bucket is **171 GPUs at exactly `memory_utilization = 0`**, the rest spread over 23-69% — a bus duty cycle, not a model footprint.

## Change

`nvmlDeviceGetMemoryInfo(handle)` was already being called for `.total` (`gpu.capacity`); this adds the `.used` field next to it as **`gpu.memory_used_mb`**, plus its entries in `ORIGINAL_KEYS` and `generate_key_mappings()`.

`gpu.memory_utilization` is left untouched — the metric is correct for what it measures and is used elsewhere (`gpu_usage` checks, wedge detection).

Deliberately **not** added to `KEYS_FOR_ENCRYPTION_KEY_GENERATION`: that list derives the encryption key, and the new field does not need to participate.

## Tests

295 passed, 2 skipped (validator specs/scrape/gpu suites). Key-mapping consistency verified: every `gpu.*` key emitted by the scrape is present in both `all_keys` and `ORIGINAL_KEYS`.


arhangel66 (approved):
approved

PR #1166, opened 2026-07-24
DAH-2451: read NVML allowlist from shared config + report unknown drivers

taiberium:
## What

Read the NVML allowlist from shared config and report unknown drivers for reactive verification (DAH-2451).

- `pipeline_factory` sources `nvml_digest_map` from `shared_client.config.nvml_ml_digests`, falling back to the packaged `LIB_NVIDIA_ML_DIGESTS` constant when the backend is unreachable (shared config empty). Also threads `nvml_invalid_drivers`.
- On a genuinely unknown driver (not in the map, not a known spoof), `NvmlDigestCheck` calls `BackendClient.report_unknown_driver` — fire-and-forget; the backend resolves it against NVIDIA's official installer and the verdict returns via shared config on a later cycle. Known spoofs are rejected but never re-reported.

Net effect: a new NVIDIA driver stops requiring a validator code change + redeploy — it's a backend DB row that reaches the validator through shared config.

## Depends on

`lium-core 0.1.8` (`nvml_ml_digests` / `nvml_invalid_drivers`): Datura-ai/lium-core#7 — **merge + release first**, then bump `lium-core>=0.1.8` here. Backend: Datura-ai/lium-io-backend#811.

Includes the already-merged `595.84` seed via rebase onto `main`.

## Tests

Updated `test_nvml_digest_check`: unknown driver is reported once; a known spoof is not. Existing digest cases unchanged. (Validator suite runs in CI — the local pdm/bittensor env is impractical to stand up here.)

### DIFF
--- a/neurons/validators/pdm.lock
+++ b/neurons/validators/pdm.lock
@@ -5,7 +5,7 @@
 groups = ["default", "dev"]
 strategy = ["inherit_metadata"]
 lock_version = "4.5.0"
-content_hash = "sha256:af83b9b4f8ade6ca68d0eaaadf1a6ebc678e704f24465a15774a7f71352ff13c"
+content_hash = "sha256:8ee9e62b9269e48f4d7b901bd7f92c568d55b76297e33c521776f1019c3e9c5f"
 
 [[metadata.targets]]
 requires_python = ">=3.11,<3.13"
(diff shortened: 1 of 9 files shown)


arhangel66 (approved):
Looks good

PR #1168, opened 2026-07-24
DAH-2490: reject kernels without ID-mapped mounts in the sysbox installer

arhangel66:
## Problem

Providers on kernel 5.15 cannot pass the sysbox GPU check, and the installer gives them no way to find out why. It runs all the way to step 6/7 and prints diagnostics that contain everything except the one fact that matters:

```
nvidia-container-cli: mount error: /var/lib/docker/overlay2/<id>/merged/proc/driver/nvidia: no such file or directory
```

The most recent case (provider LOY, node `jssz-4090-52-105`, 8x RTX 4090, driver 550.54.14, Docker 27.5.1) cost the provider an hour of driver, toolkit and CDI config changes — none of them relevant — before it reached human support.

## Cause

Sysbox needs ID-mapped mounts, which the kernel only supports for overlayfs from 5.19 onward. Below that, `sysbox-mgr` falls back to shiftfs and relocates the container rootfs to `/var/lib/sysbox/shiftfs/<uuid>`, while the NVIDIA legacy hook still mounts `/proc/driver/nvidia` at the original overlay2 path — which no longer exists.

Confirmed by direct experiment on 2026-07-17: identical failure on kernel 5.15, works on 6.8. Toolkit version (1.14.3 vs 1.19.1), Docker version (24/28/29), hook mode (legacy vs CDI) and driver version (535 vs 580) were each ruled out by test — upgrading the driver alone did not help.

## Changes

- `nvidia_docker_sysbox_setup.sh` refuses to install on a kernel that cannot support GPU passthrough, and names the fix (`linux-generic-hwe-22.04` + reboot). The gate reads `Overlayfs on ID-mapped mounts supported by kernel: no` from the sysbox-mgr journal when sysbox is already installed, and falls back to a `5.19` version comparison when it is not — so a host where sysbox-mgr reports `yes` on an odd kernel is never rejected on version alone.
- The gate sits **after** the "already working, nothing to do" exit, so a host that somehow works today cannot be broken by it.
- Diagnostics now print `Kernel:` and `ID-mapped mounts:`.
- `machine_scrape` reports the host kernel as `specs.kernel` (`uname -r` returns the host kernel even from inside the executor container, unlike `lsb_release`). Today the fleet has no kernel data at all, so the scope of this failure class is invisible — the closest proxy is driver major version, where the sysbox failure rate is 42-46% on 535-575 versus 14% on 580 and 4% on 595.

## Verification

- Version gate exercised against real kernel strings: `5.4.0-42-generic` and `5.15.0-91-generic` rejected; `5.19.0-45-generic`, `6.8.0-134-generic`, `6.14.0-1012-aws` accepted.
- `bash -n` clean.
- `tests/test_scrape_encryption_key_order.py` passes — the new keys are outside `KEYS_FOR_ENCRYPTION_KEY_GENERATION`, which is built from `gpu.*` only, so the encryption key is unchanged.

Docs side: Datura-ai/lium-docs#90

Task: DAH-2490

## Review round 2

External review (model panel + a second code pass) found three things worth fixing:

- `IDMAPPED="no"` is a live probe result, not a version verdict — it also comes back `no` when Docker's data-root sits on a filesystem without ID-map support (ZFS) or when sysbox-mgr runs with the feature disabled. Telling that host to install an HWE kernel is useless advice. The gate now picks its message: below 5.19 → kernel fix, 5.19+ → `docker info --format '{{.Driver}}'` and the sysbox-mgr journal.
- The "already working" early exit required the verify image to be present locally, so a working host that had pruned the image fell through to the gate and could be rejected on kernel version alone. It now pulls the image (needed for step 7 anyway) and lets the real GPU test decide.
- The gate is a `case` over `yes` / `no` / anything-else, so a future change in the reported value cannot silently let an old kernel through.

The reboot advice now warns about active rentals, non-DKMS drivers (a `.run`-installed driver will not load on the new kernel), and that Ubuntu 20.04 has no HWE path to 5.19.

Truth table exercised over 10 (report, kernel) combinations — all as expected.


neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -48,13 +87,34 @@ fi
...
 
+# Reached only when the real GPU test above did not pass, so a working host is never rejected here.
+IDMAPPED=$(sysbox_idmapped_report)
+case "$IDMAPPED" in
+    yes) ;;
+    no)
+        if kernel_supports_idmapped; then fail_no_idmapped; else fail_old_kernel; fi
+        exit 1
+        ;;
+    *)  # sysbox-mgr reported nothing this boot — fall back to the kernel version
```
taiberium, line 110: NIT: When sysbox is not installed there is no report, so this rejects on `uname -r` alone with no way to override.

neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -48,13 +87,34 @@ fi
...
 
-if command -v sysbox-runc &>/dev/null && docker info 2>/dev/null | grep -q sysbox-runc \
-   && docker image inspect "$VERIFY_IMAGE" &>/dev/null \
-   && docker run --rm --runtime=sysbox-runc --gpus all "$VERIFY_IMAGE" nvidia-smi &>/dev/null; then
-    ok "Sysbox is already working. Nothing to do."
-    exit 0
+if command -v sysbox-runc &>/dev/null && docker info 2>/dev/null | grep -q sysbox-runc; then
+    # pull first: without the image the real test cannot run and the host would be judged on kernel version alone
+    if ! docker image inspect "$VERIFY_IMAGE" &>/dev/null; then
+        echo "  Pulling $VERIFY_IMAGE to test the current setup..."
```
taiberium, line 93: NIT: This pulls the image before the active-rental check, so a run that will abort anyway still downloads it onto a live node.

neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -22,12 +22,51 @@ cleanup() { [ -n "$DOWNLOADED_DEB" ] && rm -f "$DOWNLOADED_DEB"; }
...
 docker_version_ge() {
+    # the DAEMON version — it writes the OCI spec sysbox has to accept, and it can differ from the client
     local ver major minor
-    ver=$(docker --version | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1)
+    ver=$(docker version --format '{{.Server.Version}}' 2>/dev/null | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1)
     major=${ver%%.*} minor=${ver#*.} minor=${minor%%.*}
     [ "$major" -gt "$1" ] 2>/dev/null || { [ "$major" -eq "$1" ] && [ "$minor" -ge "$2" ]; } 2>/dev/null
 }
 
+kernel_supports_idmapped() {
```
taiberium, line 32: NIT: This repeats the major/minor parsing already in `docker_version_ge`.

neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -227,13 +295,21 @@ else
...
-    echo "    Docker:              $(docker --version 2>/dev/null || echo unknown)"
+    echo "    Docker daemon:       $(docker version --format '{{.Server.Version}}' 2>/dev/null || echo unknown)"
+    echo "    Kernel:              $(uname -r)"
+    echo "    ID-mapped mounts:    $(sysbox_idmapped_report | grep . || echo 'not reported by sysbox-mgr')"
     echo "    NVIDIA driver:       $(nvidia-smi --query-gpu=driver_version --format=csv,noheader 2>/dev/null || echo FAILED)"
     echo "    /proc/driver/nvidia: $(ls /proc/driver/nvidia &>/dev/null && echo exists || echo MISSING)"
     echo "    sysbox-runc:         $(sysbox-runc --version 2>/dev/null | head -1 || echo 'not found')"
     echo "    CDI specs:           $(ls /var/run/cdi/nvidia.yaml /etc/cdi/nvidia.yaml 2>/dev/null || echo none)"
     echo "    daemon.json cdi:     $(jq -r '.features.cdi // "not set"' /etc/docker/daemon.json 2>/dev/null)"
+    echo "    daemon.json time-ns: $(jq -r '.features["time-namespaces"]' /etc/docker/daemon.json 2>/dev/null)"
```
taiberium, line 306: NIT: This prints `null` when the key is unset, while the `cdi` line above falls back to `not set`.

taiberium (approved):

neurons/validators/src/services/file_encrypt_service.py
```diff
@@ -93,6 +93,8 @@
     'hard_disk_utilization': "utilization",
     'hard_disk_scrape_error': "hard_disk_scrape_error",
     'data_os': "os",
+    'data_kernel': "kernel",
```
arhangel66, line 100: compute-app's `MachineSpecs` declares neither `kernel` nor `kernel_scrape_error`, and executor upserts normalize specs through `MachineSpecs.model_validate(...).model_dump()`, so Pydantic drops both keys. Non-blocking here, but the backend field needs to ship before the kernel shows up in fleet data.

arhangel66 (commented):
See the inline comments.

neurons/validators/src/services/file_encrypt_service.py
arhangel66: Closed by compute-app eecd104e on `DAH-2490-machine-specs-kernel` (lium-io-backend#910): `MachineSpecs` now declares both `kernel` and `kernel_scrape_error`, so neither key is stripped on upsert.

taiberium (approved):

PR #1169, opened 2026-07-27
DAH-2496: keep sysbox for S3FS volume pods via uid-aligned mount

taiberium:
## What

Keep `sysbox-runc` on pods that mount an S3FS external volume, instead of silently downgrading them to plain `runc`.

The s3fs mount is now created with `uid`/`gid` set to the host's sysbox subuid base, so the pod's root owns the objects it sees.

## Why

Attaching a volume set `payload.is_sysbox = False` with no flag, log line or warning — the renter lost user-namespace isolation and Docker-in-Docker without knowing.

The downgrade is not necessary. sysbox cannot ID-shift a FUSE mount (`mount_setattr(MOUNT_ATTR_IDMAP)` returns EINVAL on s3fs, OK on ext4) and does not treat that as fatal, so the container starts either way — it just sees the bucket as `nobody` and cannot write objects it did not create.

Verified on staging against a real S3 bucket, on a node running host Docker 29.1.4 with sysbox subuid base 231072:

| operation under sysbox | `allow_other` (today) | `+ umask=0000` | `+ uid=<base>` |
| --- | --- | --- | --- |
| create file, mkdir, 20 MB write | OK | OK | OK |
| append to a file created outside the pod | EACCES | EACCES | OK |
| chown | EPERM | EPERM | OK |
| owner seen inside the pod | nobody | nobody | root |

`umask=0000` cannot work: it only masks bits away, and s3fs keeps owner and mode in object metadata. Owner alignment is what DAH-2201 recommended.

## Fallback

`resolve_sysbox_uid_base()` returns `None` when `/etc/subuid` has no sysbox entry or reserves more than one 65536-wide slice — a wider range means sysbox picks a different base per container, so no single uid matches. In that case the pod still drops to `runc`, but now it says so in the deploy log instead of doing it silently.

## Tests

`tests/test_docker_service_rental_security.py` — sysbox is kept with a volume attached, the fallback path, both mount-option shapes, and `/etc/subuid` parsing.

Follow-ups in separate PRs: the Create Pod form and the public docs still tell users the two cannot be combined.

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -150,6 +150,10 @@ DOCKER_VOLUME_PLUGINS = {
     "s3fs": "mochoa/s3fs-volume-plugin"
 }
 
+# DAH-2496: sysbox installers reserve one 65536-wide subuid slice per host and map
+# every container's root to its base. A wider range is handed out per container.
+SYSBOX_SUBUID_SLICE_SIZE = 65536
+
 # DAH-1991: tolerate concurrent health_check_* / container_* on the executor.
 # Probe TTL is short (~30s); same-command retry within a 90s budget covers the
 # documented race without regenerating port mappings.
@@ -2195,12 +2199,49 @@ class DockerService:
 
         return True
 
+    async def resolve_sysbox_uid_base(
+        self,
+        ssh_client: asyncssh.SSHClientConnection,
+        log_extra: dict,
+    ) -> int | None:
+        """Host uid that sysbox maps container root to, or None when unusable.
+
+        Only a single slice is usable: a wider range means sysbox hands out a
+        different base per container, and one static s3fs uid cannot match them.
+        """
+        result = await ssh_client.run("grep '^sysbox:' /etc/subuid")
+        entries: list[str] = (result.stdout or "").strip().splitlines()
+        if not entries:
+            return None
+
+        fields: list[str] = entries[0].strip().split(":")
+        if len(fields) != 3:
+            return None
+
+        try:
+            uid_base: int = int(fields[1])
+            slice_size: int = int(fields[2])
+        except ValueError:
+            return None
+
+        if slice_size != SYSBOX_SUBUID_SLICE_SIZE:
+            logger.warning(
+                _m(
+                    "sysbox subuid range is not a single slice; cannot align s3fs owner",
+                    extra=get_extra_info({**log_extra, "slice_size": slice_size}),
+                )
+            )
+            return None
+
+        return uid_base
+
     async def create_s3fs_volume(
         self,
         ssh_client: asyncssh.SSHClientConnection,
         log_extra: dict,
         volume_info: ExternalVolumeInfo,
         log_tag: str,
+        sysbox_uid_base: int | None = None,
     ):
         responses = []
         # install docker volume plugin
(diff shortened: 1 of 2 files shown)


taiberium:
## Verified on staging

Deployed to staging and rented a pod with a volume attached, on a node with sysbox subuid base 231072:

```
/proc/self/uid_map   0  231072  65536     <- sysbox is on, with the volume mounted
append to a file created outside the pod  OK   (was EACCES)
chown                                     OK   (was EPERM)
create / mkdir / 20 MB write / delete     OK
write as an unprivileged user             OK
docker run hello-world  ->  "Hello from Docker!"   <- nested Docker with the volume attached
```

## Second commit

The first attempt on staging kept falling back to runc. The SSH session the validator uses lands **inside the executor container**, whose `/etc/subuid` is its own (`liumuser:100000:65536`) and has no sysbox entry — so the base never resolved. The host file is now read through dockerd, which resolves `-v` paths on the host:

```
docker run --rm -v /etc/subuid:/etc/subuid:ro --entrypoint cat <executor image> /etc/subuid
```

taiberium:
## Polish pass + re-verified on staging

Re-deployed to staging at commit `16c9d8e1` and re-ran the full check on a fresh pod with the volume attached:

```
/proc/self/uid_map   0  231072  65536     <- sysbox on, volume mounted
docker run hello-world  ->  "Hello from Docker!"
append to a file created outside the pod  OK
chown / 20 MB write / delete              OK
```

Three changes came out of the polish pass:

- **Subuid is read through the host PID namespace** (`cat /proc/1/root/etc/subuid`) instead of starting a throwaway container. The executor shares the host PID namespace, so this is one ssh round trip instead of a container create/start/rm on the deploy path, and it drops the dependency on an image tag that prod does not pin (`daturaai/compute-subnet-executor:latest` — prod compose pins by digest, so the tag may not exist locally).
- **The runtime decision moved after the volume actually attaches.** Previously the sysbox downgrade happened before `create_s3fs_volume`, so a failed volume creation left the pod on runc with no volume at all — isolation traded for nothing.
- Naming aligned with `/etc/subuid` vocabulary (`resolve_sysbox_subuid_base`, `sysbox_subuid_base`), parsing collapsed to tuple unpacking, and the two mount-option tests merged into one parametrized case.

Tests: 1222 passed, 3 pre-existing failures in `tests/test_sshd_bootstrap_script.py` (they fail on a stashed tree too).

taiberium:
## The "plugin is broken on Docker 29" claim does not hold

DAH-2201 reported that `mochoa/s3fs-volume-plugin` fails on Docker 29.1.4 with
`failed to chmod … propagated-mount …: input/output error`, independent of sysbox.

Swept nine Docker versions on a throwaway box against a **real S3 bucket** (a platform-created volume with its own scoped IAM user), same plugin version 2.0.9 and the same option string `create_s3fs_volume` uses. Plugin reinstalled per version, volume created, container started and wrote a file:

```
27.5.1  OK      29.2.1  OK
28.5.2  OK      29.3.0  OK
29.0.4  OK      29.4.2  OK
29.1.4  OK      29.5.3  OK
29.6.2  OK
```

The same sweep against **MinIO** fails on all nine versions. Docker chmods the volume mount root at container create; s3fs stores that as object metadata, which real S3 accepts and MinIO rejects on the bucket root. So the failure is an artifact of the test backend, not a Docker version issue.

**Do not verify s3fs behaviour against MinIO** — it silently disagrees with S3 on both chmod and stored ownership, which is also what made the earlier `umask=0000` proposal look like it worked.

taiberium:
## Edit-pod path verified on staging

The pod-edit flow (`POST /api/pods/{id}/reboot` with a volume, logged as `edit_pod: true`) runs the same `create_container`, so it resolves the subuid base and reconfigures the plugin a second time for a pod whose volume is already mounted. Checked on a node where the host is reachable.

Before the edit: `uid_map 0 231072 65536`, `/mnt` mounted, marker file written.
After the edit:

```
uid_map              0  231072  65536      <- sysbox kept
mount                s3fs on /mnt
files                marker written before the edit is still there
append / chown       OK
docker run hello-world  ->  "Hello from Docker!"
plugin options       allow_other,uid=231072,gid=231072
```

No orphans: one container, one volume during the run, and both gone after deletion.

That closes the last untested path in this change. Pods without a volume never enter the new block, so they are unaffected by construction.

arhangel66 (approved):
The uid alignment logic and the runc fallback look right. One question before I approve: DEFAULT_S3FSOPTS is global plugin state on the host, but its value now depends on the individual request, and the plugin disable / set credentials / set options / enable / volume create sequence is a series of separate awaits with no lock around it. DockerService.lock only guards the logs queue.

Can two volume creations run concurrently against the same host? If so, a runc request can reset the options to allow_other between our set and our volume create, and we keep sysbox-runc with an unwritable mount — the exact failure this PR fixes, but now silent. If concurrency there is impossible, say so and I'll approve as is. Otherwise an asyncio.Lock around the whole plugin-config + volume-create section (plus a test with interleaved sysbox/runc requests) should cover it.

PR #1170, opened 2026-07-27
DAH-2499: carry the real cause into the set_environment failure

arhangel66:
## Problem

`create_container` re-raises a bare `RuntimeError("Failed to set environment variables")`. The actual cause is logged separately by `rental_docker_observability`, so the ERROR record lium-stats classifies carries no diagnostic content at all.

Over the last 7 days (12–19 Aug) that is 10 of 579 `create_container` failures — 1.7%. Every one of them takes the exception branch, and the causes are already known and specific:

- `Docker container is not ready for exec: status='exited' running=False … exit_code=143`
- `Docker SDK exec failed: 409 … container … is not running`
- `Docker SDK inspect container failed: 404 … No such container: filler_…`

The `exit_status != 0` branch produced nothing in that window.

> Opened on 27 Jul, when this signature was 41 of 256 — 16% and the largest unclassifiable one. The mix has shifted since; see Follow-up.

## Change

`add_environment_variables_with_rental_docker` returns the failure cause (`str | None`) instead of a bool, and the caller puts it in the `RuntimeError` message. The cause therefore reaches `extra.error` on the `Failed create_container` record, which is what Datura-ai/lium-stats#83 reads, and `filler_run.failure_reason` on the backend.

Same shape as the neighbouring `Failed to add SSH public keys: exit_status=...; stderr=...; stdout=...`, which classifies fine today.

## What this buys

The three texts above are matched by rules that already exist in lium-stats#83 and sit above the generic `failed to set environment variables` rule, so the records reclassify with no classifier change:

| now | after |
|---|---|
| `environment.set_failed` (ambiguous) | `container.exited_before_exec` |
| `environment.set_failed` (ambiguous) | `container.disappeared` |

Which also settles the question: there is no distinct "environment" failure class — the container is already dead by the time we exec (`exit_code=143`, i.e. SIGTERM, and `exit_code=0`). `set_environment` is just where we first notice.

Nothing changes for the renter: `msg` is built from `str(log_text)` and `_StructuredMessage.__str__` returns the headline only. Environment values cannot leak either — they travel over stdin (`sh -c "cat >> /etc/environment"`), and only the command's own `exit_status`/`stderr`/`stdout` reach the message.

## Follow-up (not in this PR)

`gpu_power_cap` is now the largest unclassifiable signature: **125 of 579** failures in the last 7 days, all carrying the same bare text `GPU power cap could not be applied; refusing to start PEARL filler uncapped`. `apply_filler_gpu_power_limits` has five distinct `return False` paths collapsing into that one string — same disease, same fix, separate PR.

Full breakdown over the same window:

| failure_step | count | share |
|---|---|---|
| docker_run | 258 | 45% |
| gpu_power_cap | 125 | 22% |
| add_public_keys | 90 | 16% |
| docker_pull | 80 | 14% |
| set_environment | 10 | 1.7% |
| other | 16 | 3% |

For contrast, `docker_run` — the largest step — already carries full detail (`Docker SDK run container failed: 400 Client Error … /containers/create?name=pod_…`) and needs nothing.

## Verification

Rebased onto current `main` (no conflicts). `pdm run pytest tests/test_docker_service.py tests/test_docker_service_rental_security.py -q` in `neurons/validators`: 208 passed, 1 skipped.

The test on `create_container` asserts the cause reaches the logged `extra.error`, not just that the step is reported.

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -2693,13 +2693,14 @@ class DockerService:
         environment: dict[str, str] | None,
         log_tag: str,
         log_extra: dict,
-    ) -> bool:
+    ) -> str | None:
+        # returns the failure cause, or None when the environment was applied
         exec_spec = build_environment_exec_spec(
             container_name=container_name,
             environment=environment,
         )
         if exec_spec is None:
-            return True
+            return None
 
         try:
             result = await exec_logged_rental_docker_sdk_operation(
@@ -2721,7 +2722,7 @@ class DockerService:
                 ),
                 exc_info=True,
             )
-            return False
+            return str(exc)
 
         if result.exit_status != 0:
             await self.stream_log(
@@ -2741,9 +2742,9 @@ class DockerService:
                     }),
                 )
             )
-            return False
+            return f"exit_status={result.exit_status}; stderr={result.stderr}; stdout={result.stdout}"
 
-        return True
+        return None
 
     async def resolve_sysbox_subuid_base(
         self,
@@ -4739,15 +4740,15 @@ class DockerService:
 
                     # add environment variables
                     current_step = "set_environment"
-                    environment_ok = await self.add_environment_variables_with_rental_docker(
+                    environment_error = await self.add_environment_variables_with_rental_docker(
                         docker_client=docker_client,
                         container_name=container_name,
                         environment=custom_options.environment if custom_options else None,
                         log_tag=log_tag,
                         log_extra=default_extra,
                     )
-                    if not environment_ok:
-                        raise RuntimeError("Failed to set environment variables")
+                    if environment_error:
+                        raise RuntimeError(f"Failed to set environment variables: {environment_error}")
 
                     # Historical name — key injection moved before the bootstrap
                     # (DAH-2341), so this step now times the environment setup.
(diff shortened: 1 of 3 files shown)


neurons/validators/src/services/docker_service.py
```diff
@@ -2741,9 +2742,9 @@ async def add_environment_variables_with_rental_docker(
                     }),
                 )
             )
-            return False
+            return f"exit_status={result.exit_status}; stderr={result.stderr}; stdout={result.stdout}"
```
taiberium, line 2745: NIT: `_format_exec_result_failure` in `rental_docker_sdk.py` already builds this exact string.

neurons/validators/src/services/docker_service.py
```diff
@@ -4739,15 +4740,15 @@ async def create_container(
...
+                    environment_error = await self.add_environment_variables_with_rental_docker(
                         docker_client=docker_client,
                         container_name=container_name,
                         environment=custom_options.environment if custom_options else None,
                         log_tag=log_tag,
                         log_extra=default_extra,
                     )
-                    if not environment_ok:
-                        raise RuntimeError("Failed to set environment variables")
+                    if environment_error:
```
taiberium, line 4750: NIT: The function returns `str | None`, so `if environment_error is not None` matches that contract better than a truthiness check.

taiberium (approved):

PR #1171, opened 2026-07-27
DAH-2512: one s3fs plugin instance per volume

taiberium:
## What

Give every S3FS volume its own plugin instance (`s3fs-<volume name>`) instead of reconfiguring one shared `s3fs` instance per host. Pod deletion now removes that instance only.

## Why

The shared instance holds a single credential pair and dies on `plugin disable`, so attaching or detaching one pod's volume breaks every other pod mounting a volume on the same host. Reproduced on a node with a running container:

```
ls /mnt                          -> 6 files
docker plugin disable s3fs -f    # what create_s3fs_volume does for the next volume
ls /mnt                          -> ls: /mnt: Socket not connected
docker plugin enable s3fs
ls /mnt                          -> ls: /mnt: Socket not connected   (does not recover)
```

The first pod's mount stays dead until its container is recreated, with no error surfaced to the renter. The delete path has the same problem: it disabled the shared instance for everyone.

Skipping the reconfiguration when nothing changed would not fix it — each volume has its own IAM user and the plugin stores one credential pair per instance, so a second volume overwrites the first one's credentials regardless. The plugin's own documentation says credentials are global per instance and recommends a separate alias per key set.

## Scope

Pre-existing, not a regression from DAH-2496 — but while volumes silently disabled sysbox and Docker-in-Docker there was little reason to use them, and DAH-2496 makes them usable. Only nodes rented in parts (GPU splitting) can host two volume pods at once.

## Leftovers on already-provisioned hosts

Hosts that served a volume before this change keep the old shared `s3fs` instance installed and unused; volumes created with it keep working until their pod goes away. `remove_s3fs_volume_plugin` is a no-op for them, which is harmless.

## Tests

`tests/test_docker_service_rental_security.py` — the per-volume alias is used for install / credentials / enable / volume create and the shared instance is never touched; removal drops only its own instance; an unsafe volume name issues no commands.

Based on #1169 — same functions. Merge that first.

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -150,6 +150,17 @@ DOCKER_VOLUME_PLUGINS = {
     "s3fs": "mochoa/s3fs-volume-plugin"
 }
 
+S3FS_PLUGIN_IMAGE = DOCKER_VOLUME_PLUGINS["s3fs"]
+
+
+def s3fs_plugin_alias(volume_name: str) -> str:
+    """DAH-2512: one plugin instance per volume.
+
+    A shared instance holds a single credential pair and dies on `plugin disable`,
+    so attaching or detaching one pod's volume breaks every other pod on the host.
+    """
+    return f"s3fs-{volume_name}"
+
 # DAH-2496: sysbox installers reserve one 65536-wide subuid slice per host and map
 # every container's root to its base. A wider range is handed out per container.
 SYSBOX_SUBUID_SLICE_SIZE = 65536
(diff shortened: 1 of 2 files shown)


arhangel66 (approved):
Re-read the pilot journal before finalizing — it reorders my earlier review. Only the first
item blocks.

**1. The automatic miner restart loop is the real risk.** Each relaunch mints a new
WORKER_ID, and the pilot measured worse than "back of the qualification queue": on
2026-07-24 one restart got the worker flagged `failed` on the provider's board, then 7+ hours
of zero billable traffic with the websocket alive, legs ADMITTED and 290 audits at 100%.
Nothing looks broken in that state, so an automatic 60s loop can quietly burn an earning day
with no signal we'd catch. Please drop the loop, or cap it at one attempt and make it loud.
Related: the plan's rule is "restart only on 'serve is dead', never on 'no traffic'".

**2. Cheap and worth doing: a completion marker for the download.** Readiness is
`[[ ! -f "${CKPT_DIR}/config.json" ]]`, and config.json lands long before the ~35GB of
weights. Since the refill scheduler creates a *new container* per tick, a half-written cache
greets every later container, not just the next one. Write a marker after `hf download`
returns and check that.

**3. Three lines: store the sidecar PID and `wait` on specific PIDs.** With METRICS_TOKEN
set, the sidecar's PID is never captured, so `shutdown` never signals it and the
argument-less `wait` blocks forever — reproduced, still alive after 10s, only SIGKILL ended
it. (There's a second path that needs no token: `hf download` runs in the foreground and bash
defers the trap until it returns. Item 2 covers the consequence for the POC.)

**4. Two doc updates.** The plan asks for SIGTERM as a drain and lists drain-vs-latency as
undecided; this PR decides it as a drop. I think that's right — please put it in the plan
rather than only in a code comment. Same for the base image: the plan still says
`12.8.0-runtime-ubuntu22.04`.

Also worth deciding, though probably out of scope here: nothing in the image corresponds to
the pilot's `storm_guard.py`, and the 2026-07-24 storm cost 95 failures on 811 requests —
8,689 clean requests to undo against the 99% gate. A 24h run can fail its gate for that
reason with nothing wrong in the container.

**Please disregard my HELP/TYPE finding** — I had the impact wrong. The duplicates don't
break the scrape path (`textparse.PromParser` and `prometheus_client` both accept it; only
`promtool`'s parser errors), and our actual consumer skips every `#` line and sums across
label sets.

Everything else held up: the `engy_miner.py:278` WORKER_ID reference is accurate, and
`--checkpoint`, `--serve-url` and `MAX_INFLIGHT` match the vendored miner.

PR #1173, opened 2026-07-28
DAH-2514: report docker disk usage breakdown in the machine scrape

arhangel66:
## What

The scrape reports one `shutil.disk_usage("/")` and nothing else, so `used` is an opaque number — there is no way to tell a genuinely full node from one whose disk is held by image layers or by volumes that come back on teardown. This adds a breakdown of what filled it.

Three new keys under `hard_disk`, in kB like the existing fields:

| key | source |
|---|---|
| `images` | `/system/df` → `LayersSize` |
| `containers` | `/system/df` → Σ `Containers[].SizeRw` |
| `volumes` | `/system/df` → Σ `Volumes[].UsageData.Size` (local driver) **+** Σ vloopback backing-file blocks |

`total`, `used`, `free` and `utilization` are untouched, so TSDB history stays comparable. Reporting only — no gate, filter or sizing decision changes here.

## Notes

- **The volume figure is assembled from two sources on purpose.** `/system/df` only accounts for the `local` driver and is blind to every vloopback volume — exactly the ones fillers and pods run on. The backing files are read directly: the loopback plugin names each one after its volume (`ashald/docker-volume-loopback`, `manager.go:338`), under `<DockerRootDir>/plugins/*/rootfs<DATA_DIR>` reachable via `/proc/1/root` because the executor container shares the host PID namespace.
- **`st_blocks`, not `st_size`.** A preallocated volume occupies its whole declared size while holding nothing; a sparse one occupies only what it wrote. Blocks are correct for both.
- **The docker half is a separate `try`.** It folds into a new `hard_disk_docker_scrape_error` rather than the existing `hard_disk_scrape_error`, so a node that loses the socket keeps reporting total/used/free. `/system/df` walks the graph driver and is slow on image-heavy nodes, hence the explicit 20s socket timeout.
- **`verifyx.py` now merges `hard_disk` instead of replacing it.** VerifyX measures only its own df-derived fields, and `ENABLE_VERIFYX` defaults to True — replacing the dict wholesale would have dropped the new fields on every node where it runs. The merge falls back to the old replace when either side is not a dict, which the pipeline fixtures rely on.
- **Both obfuscation tables were updated** (`ORIGINAL_KEYS` and `generate_key_mappings()`); a new test parses the source and fails if a `hard_disk_*` key is registered in only one of them.
- Uses `http.client` over an `AF_UNIX` socket — the CLI returns human strings (`"45.08GB"`), the API returns exact integers.

## Testing

`tests/test_scrape_disk_breakdown.py` — 6 tests covering the `/system/df` parse, unknown (`-1`) sizes, sparse-vs-preallocated block accounting, a missing backing file, a non-200 response, and the two-table key registration. `machine_scrape.py` is a script (importing it runs the whole scrape), so the helpers are compiled out of the source and executed on their own, the same reason `test_scrape_encryption_key_order.py` parses rather than imports.

Full validator suite: 1236 passed. The 3 failures in `tests/test_sshd_bootstrap_script.py` reproduce on a clean `origin/main` and are unrelated.

## Rollout

Validator-only deploy — the scrape binary is rebuilt and pushed every job round, no executor release. Pairs with https://github.com/Datura-ai/lium-io-backend/pull/835, which declares the fields on `HardDiskSpec`; until that lands the new keys are silently stripped by Pydantic, which is harmless.


neurons/validators/src/miner_jobs/machine_scrape.py
```diff
@@ -751,6 +754,91 @@ def benchmark_network_speed():
...
+        if (volume.get("Driver") or "").startswith(VLOOPBACK_DRIVER_PREFIX)
+        and volume.get("Name")
+        and "/" not in volume.get("Name")
+    ]
+    if not names:
+        return 0
+
+    # DATA_DIR is set to <DockerRootDir>/loopback at install time, but the plugin writes it inside
+    # its own rootfs; reachable here because the executor container shares the host PID namespace.
+    data_dirs = glob.glob(
```
taiberium, line 807: **NIT:** this returns 0 when the glob matches nothing, so volumes under-report by terabytes with no error key, unlike the docker half.

neurons/validators/src/miner_jobs/machine_scrape.py
```diff
@@ -751,6 +754,91 @@ def benchmark_network_speed():
...
+                break
+            except OSError:
+                continue
+    return total
+
+
+def get_docker_disk_usage():
+    # what actually filled the disk, split by kind, in kB to match the other hard_disk fields
+    df = docker_api_get("/system/df")
+    containers = sum(max(int(each.get("SizeRw") or 0), 0) for each in df.get("Containers") or [])
```
taiberium: **NIT:** `each` says nothing about what it holds. Use `container` and `volume`.

taiberium (approved):

PR #1174, opened 2026-07-28
DAH-2520, zero the unrented incentive when total VRAM exceeds total disk

arhangel66:
### Task Link

[DAH-2520](https://app.notion.com/p/Validator-zero-unrented-incentive-when-total-VRAM-exceeds-total-disk-3abb8bfdbde9816faf49d2473bc497de)

---

## Problem

The unrented (idle) incentive pays idle machines so a renter can pick them up. A node without
enough disk to hold the data its own GPUs work on is not realistically rentable — yet it collects
idle emission today.

Prod right now (359 active + verified executors, all measurable): **1** node has less disk than
VRAM, **2** are below the 1.5x margin this PR enforces, 8 are below 2x.

- `a5c4c96f-d2cf-45d2-8d36-607cce56429b` — 8x H100 80GB HBM3, 637.2 GB VRAM / 483.3 GB disk (0.76x)
- `eaf781b7-e1f9-431d-ba2b-c486262a6e2e` — 8x RTX PRO 6000 Blackwell, 764.7 GB VRAM / 865.2 GB disk (1.13x)

Worth knowing before the flag is flipped: six H200 nodes of one hotkey
(`5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv`) sit at exactly 1.54x (1123.2 GB VRAM /
1731.4 GB disk) — just above the threshold.

## Solution

Gate the unrented pool on a disk/VRAM sanity check inside `RentalPriceIncentive`, mirroring the
DAH-2250 soft price limit: an otherwise-eligible idle executor whose total disk is below
`MIN_DISK_TO_VRAM_RATE` (1.5) times its summed GPU VRAM loses `eligible_for_rental_share`, stays
active, and is told why in its incentive log.

Both numbers already ride on `JobResult.spec` (`gpu.details[].capacity` in MB, `hard_disk.total`
in kB), so there is no new scrape and no protocol change.

Enforcement is off by default: with `ENABLE_UNRENTED_VRAM_OVER_DISK_LIMIT=false` the validator only
emits a shadow log line, so prod impact can be measured before any payout changes.

## Changes

- `incentive/rental_price.py` — `InsufficientDisk` model, `_insufficient_disk()` detector,
  `_log_insufficient_disk()` shadow log and `_log_insufficient_disk_unmeasured()` for scrapes the
  gate cannot read; the gate runs in `calculate_executor_score` right after the soft price limit.
- `core/config.py` — new `ENABLE_UNRENTED_VRAM_OVER_DISK_LIMIT` flag, default `False`.
- `incentive/miner_incentive_log.py` — `ZeroIncentiveReason.INSUFFICIENT_DISK_FOR_VRAM`
  (append-only enum) and `no_payout_because_insufficient_disk_for_vram()`, which takes the whole
  measurement and quotes both numbers so the miner knows how much disk to add.
- `tests/test_unrented_insufficient_disk.py` — 31 tests; `tests/test_miner_incentive_log.py` — pin
  the new reason code in the stable-contract test.

Deliberate choices:

- Compares against **total** disk, not free space: free/used are distorted by preallocated volumes
  (DAH-2514), total is not.
- A scrape missing either number is never flagged — fail open, a miner never loses incentive over
  missing telemetry; the unmeasurable case is logged separately so a shadow report cannot read it
  as "disk is fine".
- A malformed scrape can never raise out of the gate: `calculate_mining_scores` has no per-result
  guard, so one bad spec would cost every miner the cycle's weights.
- Rented executors are untouched: the gate sits inside the idle-and-eligible branch only.
- Disk exactly equal to 1.5x VRAM clears the gate; only strictly less is flagged.

## Deployment Steps

Use GitHub Action. Ships in shadow mode; flip `ENABLE_UNRENTED_VRAM_OVER_DISK_LIMIT=true` in a
separate step once the shadow log confirms the expected nodes.

## Review Request

- [x] Just code review

## Other PRs

None.

## Risks

- **Open review point (not addressed in this PR):** the backend picks filler nodes without reading
  `eligible_for_rental_share` (`default_job/runner.py` → `get_idle_executors_without_miner_default_job`),
  so a gated node keeps running Lium's Dolphin/PEARL fillers while earning nothing. The one node the
  gate zeroes today (`a5c4c96f-d2cf-45d2-8d36-607cce56429b`) is running 4 DPHN fillers right now.
  Needs a decision before the flag is flipped: exempt nodes with a Lium default job in the validator
  (`default_job_owner == "lium"` is already on `JobResult`), or mirror the gate in the backend's
  filler selection.
- With the flag on, a wrong or truncated `gpu.details` list would under-count VRAM and silently
  keep an executor in the pool (fail-open direction, not a payout loss).
- Six H200 nodes sit at 1.54x, so a change in which filesystem the scrape reports would pull them
  under the threshold in one cycle.
- No effect while the flag is off other than one extra log line per affected idle executor.

### Test Case 1 — unit

**Actions** `pdm run pytest tests/test_unrented_insufficient_disk.py` in `neurons/validators/`.

**Expected Output** 31 passed: detection at 1128 GB VRAM vs 500 GB disk, disk exactly at the 1.5x
margin not flagged, equal totals flagged, partial scrape (no disk / no gpu details / disk 0) not
flagged, malformed scrapes never raise out of scoring, flag default is `False`, shadow mode keeps
eligibility and logs nothing to the miner, enforcement drops eligibility and appends
`insufficient_disk_for_vram` with both numbers, enough-disk and rented executors untouched.

### Test Case 2 — full validator suite

**Actions** `pdm run pytest tests/` in `neurons/validators/`.

**Expected Output** 1253 passed. The 3 failures in `test_sshd_bootstrap_script.py` and 9 errors in
`test_port_mapping_dao.py` are pre-existing on `main` (bash sshd harness and a missing DB), untouched
by this branch.

### Test Case 3 — staging

**Actions** Deployed this branch to staging via `staging_branch_deployment.yml`
(run 30380313276, image `lium-staging-validator:92c024d`).

**Expected Output** Validator pod healthy; over 20 minutes it completed 122 rental-share
calculations and emission splits with no error from the incentive path. Live specs in the validator
flow carry both fields (`hard_disk.total = 101430960` kB, `capacity = 49140.0` MB), confirming the
gate reads real data rather than failing open. Staging has no machine short on disk, so enforcement
itself is covered by the unit tests only.

## Checklist

- [ ] Proper labels added — this repo has no `ready-review` / `require-qa` labels; `ready_for_deploy` goes on after approval
- [x] Risks are documented
- [x] Tests are described

### DIFF
--- a/neurons/validators/src/core/config.py
+++ b/neurons/validators/src/core/config.py
@@ -202,6 +202,12 @@ class Settings(BaseSettings):
     # (shadow mode) so prod impact can be observed before enforcing.
     ENABLE_UNRENTED_SOFT_PRICE_LIMIT: bool = Field(env="ENABLE_UNRENTED_SOFT_PRICE_LIMIT", default=False)
 
+    # DAH-2520 — unrented VRAM/disk sanity gate. When True, an unrented executor whose
+    # total GPU VRAM exceeds the machine's total disk loses the unrented rental incentive
+    # while staying active. When False, the breach is only logged (shadow mode) so prod
+    # impact can be observed before enforcing.
+    ENABLE_UNRENTED_VRAM_OVER_DISK_LIMIT: bool = Field(env="ENABLE_UNRENTED_VRAM_OVER_DISK_LIMIT", default=False)
+
     COLLATERAL_CONTRACT_ADDRESS: str = Field(
         env='COLLATERAL_CONTRACT_ADDRESS', default='0x8A4023FdD1eaA7b242F3723a7d096B6CC693c7C6'
     )
--- a/neurons/validators/src/incentive/miner_incentive_log.py
+++ b/neurons/validators/src/incentive/miner_incentive_log.py
@@ -8,7 +8,7 @@ WHERE A NODE EARNS (two "pools" of subnet emission; the rest is burned):
 
 HOW A NODE PICKS A POOL:
   rented?                                              -> mining pool (always earns)
-  idle AND model in program AND price ok AND disk >= VRAM
+  idle AND model in program AND price ok AND disk >= 1.5x VRAM
                                AND capacity?            -> unrented pool (earns)
   otherwise                                            -> 0 incentive (reason below)
 
@@ -22,7 +22,7 @@ WHAT THIS CATALOG HOLDS — every `MinerLogLine` the miner-facing log block
    Group B — idle but does not qualify for the unrented pool:
      GPU model not in the unrented program (earns only when rented),
      price above the market soft limit (lower the price to earn),
-     total GPU VRAM above total disk (add disk to earn),
+     total disk below 1.5x total GPU VRAM (add disk to earn),
      no unrented capacity for that GPU-count tier this cycle,
      NVIDIA driver below the minimum, sysbox runtime not enabled
 
@@ -46,6 +46,7 @@ from pydantic import BaseModel, Field
 from core.utils import _m, _StructuredMessage, get_extra_info
 
 if TYPE_CHECKING:
+    from incentive.rental_price import InsufficientDisk
     from services.task_service import JobResult
 
 
@@ -66,7 +67,7 @@ class ZeroIncentiveReason(StrEnum):
     PRICE_ABOVE_MARKET_P90_SOFT_LIMIT = "price_above_market_p90_soft_limit"
     NO_UNRENTED_CAPACITY_FOR_GPU_COUNT = "no_unrented_capacity_for_gpu_count"
     NVIDIA_DRIVER_BELOW_MINIMUM = "nvidia_driver_below_minimum"
-    VRAM_EXCEEDS_DISK = "vram_exceeds_disk"
+    INSUFFICIENT_DISK_FOR_VRAM = "insufficient_disk_for_vram"
     SYSBOX_NOT_ENABLED = "sysbox_not_enabled"
 
 
@@ -227,23 +228,25 @@ class MinerLogLine(BaseModel):
         )
 
     @staticmethod
-    def no_payout_because_vram_exceeds_disk(
-        result: JobResult, total_vram_gb: float, total_disk_gb: float, rate: float
+    def no_payout_because_insufficient_disk_for_vram(
+        result: JobResult, measured: InsufficientDisk
     ) -> MinerLogLine:
-        required_disk_gb: float = round(total_vram_gb * rate, 1)
+        # takes the measurement whole: the two totals are interchangeable floats, and swapping
+        # them would quietly tell the miner to add disk he already has
+        required_disk_gb: float = round(measured.vram_gb * measured.rate, 1)
         return MinerLogLine._no_payout(
             result,
-            reason=ZeroIncentiveReason.VRAM_EXCEEDS_DISK,
+            reason=ZeroIncentiveReason.INSUFFICIENT_DISK_FOR_VRAM,
             message=(
-                f"No unrented incentive: this executor has {total_disk_gb} GB of total disk but "
-                f"{total_vram_gb} GB of GPU VRAM. An idle executor must have at least "
-                f"{rate}x its GPU VRAM in disk to earn the unrented incentive. Give it at least "
-                f"{required_disk_gb} GB of total disk, or rent it out to earn."
+                f"No unrented incentive: this executor has {measured.disk_gb} GB of total disk but "
+                f"{measured.vram_gb} GB of GPU VRAM. An idle executor must have at least "
+                f"{measured.rate}x its GPU VRAM in disk to earn the unrented incentive. Give it at "
+                f"least {required_disk_gb} GB of total disk, or rent it out to earn."
             ),
             extra_fields={
-                "total_vram_gb": total_vram_gb,
-                "total_disk_gb": total_disk_gb,
-                "vram_over_disk_rate": rate,
+                "total_vram_gb": measured.vram_gb,
+                "total_disk_gb": measured.disk_gb,
+                "min_disk_to_vram_rate": measured.rate,
                 "required_disk_gb": required_disk_gb,
             },
         )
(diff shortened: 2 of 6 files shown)


neurons/validators/src/incentive/rental_price.py
```diff
@@ -487,6 +564,19 @@ async def calculate_executor_score(
                 )
                 job_result.incentive_logs.append(reason.to_log_line())
 
+        # DAH-2520 VRAM/disk gate: an idle machine with less disk than GPU VRAM is not
+        # realistically rentable, so it forfeits the unrented incentive (node stays active).
+        # While the flag is off we only log the would-be exclusion (shadow).
+        vram_over_disk = self._vram_over_disk(job_result) if eligible_for_rental_share else None
```
taiberium: **Critical:** the backend picks filler nodes without reading `eligible_for_rental_share`, so a gated node keeps running Dolphin/PEARL and earns nothing. Exempt nodes with a default job.

neurons/validators/src/incentive/rental_price.py
```diff
@@ -214,6 +228,69 @@ def _log_soft_price_limit(self, result: JobResult) -> None:
...
+        spec = result.spec
+        if not spec:
+            # no scrape at all: a synthetic or estimated job result, nothing to measure
+            return None
+        try:
+            gpu_details = (spec.get("gpu") or {}).get("details") or []
+            vram_gb = sum(float(gpu.get("capacity") or 0) for gpu in gpu_details) / 1024  # capacity is MB
+            # total size of the filesystem the scrape reports, not free space: free is
+            # distorted by preallocated volumes
+            disk_gb = float((spec.get("hard_disk") or {}).get("total") or 0) / 1024 ** 2  # total is kB
```
taiberium, line 244: **NIT:** this reads total disk while every other gate keys on free space, so a 95% full 1 TB node passes here after the marketplace already dropped it.

neurons/validators/src/incentive/rental_price.py
```diff
@@ -214,6 +228,69 @@ def _log_soft_price_limit(self, result: JobResult) -> None:
...
 
+    def _vram_over_disk(self, result: JobResult) -> VramOverDisk | None:
+        # machine carrying more GPU VRAM than it has disk; None when it fits or the scrape is unusable
+        spec = result.spec
+        if not spec:
+            # no scrape at all: a synthetic or estimated job result, nothing to measure
+            return None
+        try:
+            gpu_details = (spec.get("gpu") or {}).get("details") or []
+            vram_gb = sum(float(gpu.get("capacity") or 0) for gpu in gpu_details) / 1024  # capacity is MB
```
taiberium, line 241: **NIT:** this sums every card, but a split node only ever uses one bundle's share at a time, so split supply takes the hardest hit.

neurons/validators/src/incentive/miner_incentive_log.py
```diff
@@ -64,6 +66,7 @@ class ZeroIncentiveReason(StrEnum):
     PRICE_ABOVE_MARKET_P90_SOFT_LIMIT = "price_above_market_p90_soft_limit"
     NO_UNRENTED_CAPACITY_FOR_GPU_COUNT = "no_unrented_capacity_for_gpu_count"
     NVIDIA_DRIVER_BELOW_MINIMUM = "nvidia_driver_below_minimum"
+    VRAM_EXCEEDS_DISK = "vram_exceeds_disk"
```
taiberium: **NIT:** the value says `vram_exceeds_disk` while the rule is disk under 1.5x VRAM. The enum is append-only, so rename it before it ships.

neurons/validators/src/incentive/miner_incentive_log.py
```diff
@@ -8,7 +8,8 @@
 
 HOW A NODE PICKS A POOL:
   rented?                                              -> mining pool (always earns)
-  idle AND model in program AND price ok AND capacity? -> unrented pool (earns)
+  idle AND model in program AND price ok AND disk >= VRAM
```
taiberium: **NIT:** the catalog still describes the 1x rule while the code enforces 1.5x.

neurons/validators/src/incentive/miner_incentive_log.py
```diff
@@ -223,6 +226,28 @@ def no_payout_because_price_above_market_soft_limit(
             },
         )
 
+    @staticmethod
+    def no_payout_because_vram_exceeds_disk(
```
taiberium: **NIT:** three positional floats, and swapping two of them quietly tells the miner to add disk he already has. Pass the `VramOverDisk` instead.

neurons/validators/src/incentive/miner_incentive_log.py
```diff
@@ -223,6 +226,28 @@ def no_payout_because_price_above_market_soft_limit(
...
+        result: JobResult, total_vram_gb: float, total_disk_gb: float, rate: float
+    ) -> MinerLogLine:
+        required_disk_gb: float = round(total_vram_gb * rate, 1)
+        return MinerLogLine._no_payout(
+            result,
+            reason=ZeroIncentiveReason.VRAM_EXCEEDS_DISK,
+            message=(
+                f"No unrented incentive: this executor has {total_disk_gb} GB of total disk but "
+                f"{total_vram_gb} GB of GPU VRAM. An idle executor must have at least "
```
taiberium: **NIT:** adding disk is rarely an option on rented bare metal, and an H200 node needs 1692 GB, so the real outcome is the miner leaving.

neurons/validators/src/core/config.py
```diff
@@ -202,6 +202,12 @@ class Settings(BaseSettings):
     # (shadow mode) so prod impact can be observed before enforcing.
     ENABLE_UNRENTED_SOFT_PRICE_LIMIT: bool = Field(env="ENABLE_UNRENTED_SOFT_PRICE_LIMIT", default=False)
 
+    # DAH-2520 — unrented VRAM/disk sanity gate. When True, an unrented executor whose
+    # total GPU VRAM exceeds the machine's total disk loses the unrented rental incentive
+    # while staying active. When False, the breach is only logged (shadow mode) so prod
+    # impact can be observed before enforcing.
+    ENABLE_UNRENTED_VRAM_OVER_DISK_LIMIT: bool = Field(env="ENABLE_UNRENTED_VRAM_OVER_DISK_LIMIT", default=False)
```
taiberium, line 209: **NIT:** the 3 of 370 count was measured under the 1x rule. Re-read it off the shadow log before flipping this.

taiberium (changes requested):
Requested changes.

taiberium (approved):

PR #1175, opened 2026-07-29
DAH-2527, exclude idle filler ports from port verification and publish in chunks

arhangel66:
### Task Link

[DAH-2527](https://www.notion.so/Port-verification-collides-with-idle-fillers-exclude-filler-ports-and-publish-in-chunks-3acb8bfdbde981b89fdad1cb224fefa1)

---

## Problem

Three H200 executors (`152.236.142.238/240/241`) report 5–6 verified ports out of a ~1000-port
range, every validation cycle, for hours. Two independent causes stack:

1. **Idle fillers hold the ports we probe.** `PortSelector.select` sorts the port range ascending
   and takes the first N; DPHN fillers take the lowest free ports as well, so the collision is
   deterministic, not unlucky. The validator cannot see them: a filler creates no `pod` row, so
   `rented_ports` (built from `RentedPod.ports_mapping`) is empty for these executors. On `.240`
   the filler ports right now are 40001, 40003, 40005, 40008, 40009, 40025, 40050, 40057 — the
   first ones the selector picks.
2. **`SemiBatchVerifier` died atomically on the first busy port.** It published all 50 candidates
   in a single `docker run`, and docker refuses the whole run when one published port is already
   allocated (`Bind for 0.0.0.0:40001 failed: port is already allocated`). One busy port returned
   zero verified ports for the entire tier, and the cascade fell through to `FallbackVerifier`,
   which tests 10 ports one at a time — half of those busy too.

(The `--network=host` tier is separately dead on these hosts — `progress: 0/300 verified` twice per
cycle, a provider-side firewall dropping raw host listeners. That is out of scope here; it caps the
ceiling at 50 ports rather than 300.)

## Solution

Backend-supplied exclusions plus a smaller blast radius per busy port.

The backend now sends `filler_ports_by_executor` (see the compute-app PR below) — an exact list of
the external ports every active filler holds on an executor. It rides the same channel as
`all_filler_containers_by_executor` (DAH-2465) and is additive with an empty default, so backend
and validator can deploy in either order.

The validator carries those ports as a **separate** `filler_ports` argument rather than folding
them into `rented_ports`. That distinction is load-bearing: a non-empty `rented_ports` is read as
"this executor has a customer rental" by `ExecutorConnectivityService`'s sysbox fallback, and
`len(pods) > 0` gates `port_count`, `sysbox_required`, `rented_machine`, `rental_verification`,
`capability`, `verifyx`, `inspector` and `result_handler`. Filler ports must exclude candidates
without routing these machines down the rented branch.

The split therefore stops exactly where it stops meaning something. `ExecutorConnectivityService` is
the last layer that reads `rented_ports` as a rental flag, so it is also the layer that merges the
two lists; everything below it receives one `unavailable_ports` list and only needs to know "this
external port is taken, don't probe it".

Separately, `SemiBatchVerifier` now publishes in chunks of 10, one container per chunk, run
concurrently. A port that is busy anyway — a filler restarting between the snapshot and the probe,
a leaked `port_test_` container, a pod container that outlived its rental — now costs one chunk
instead of the whole tier.

Expected on the dirtiest machine (`.240`), chunks of 10 over the first 50 candidates, before the
exclusions even land:

```
40000–40009  fails   (40001, 40003, 40005, 40008, 40009 busy)
40010–40019  ok      → 10
40020–40029  fails   (40025 busy)
40030–40039  ok      → 10
40040–40049  ok      → 10
```

30 verified from chunking alone; with the exclusions in place those two chunks stop failing at all.

## Changes

- `protocol/vc_protocol/compute_requests.py` — mirror `filler_ports_by_executor` with a
  `get_filler_ports(executor_uuid)` accessor, defaulting to `{}` so an older backend contributes
  no exclusions.
- `services/task/checks/port_connectivity.py` — read the filler ports off `rented_data` (which
  works even though the executor has no `executors` entry at all) and pass them as `filler_ports`.
- `services/executor_connectivity/service.py` — new `filler_ports` argument, merged with
  `rented_ports` into the single `unavailable_ports` list handed to the orchestrator.
- `services/executor_connectivity/orchestrator.py` — `rented_ports` becomes `unavailable_ports`;
  the orchestrator no longer knows which of the two sources a port came from.
- `services/executor_connectivity/port_selector.py` — rename the third parameter `rented` →
  `unavailable_ports`: it now receives pod ports *and* filler ports, and the old name lied.
- `services/executor_connectivity/port_verifiers.py` — `SemiBatchVerifier` splits its candidates
  into `CHUNK_SIZE = 10` chunks and runs one container per chunk via `asyncio.gather`, each with
  its own token and its own cleanup. A chunk that fails to start contributes only its own ports to
  `failed`.

Deliberately **not** changed:

- `port_probe.py`. Its cascade already falls through only when a tier verifies *zero* ports, so a
  partial semi-batch result is carried forward as is.
- No single-port retry over failed chunks. It would run the ~15 s sequential tier in cases where it
  never ran before (partial success), for at most ten extra ports, while the exclusions remove the
  collisions that poison chunks in the first place.
- No randomized or high-end candidate selection — it would move the probed ports every cycle and
  destroy comparability between cycles.

## Deployment Steps

Use GitHub Action. Deploy order does not matter: the field is additive with an empty default on
both sides. A new validator against an old backend simply gets no exclusions and keeps the current
behaviour; an old validator against a new backend ignores the extra field.

## Review Request

- [x] Just code review

## Other PRs

- compute-app: https://github.com/Datura-ai/lium-io-backend/pull/840

## Risks

- **More containers per verification cycle.** The semi-batch tier now starts up to 5 `port_test_pub_*`
  containers instead of 1, concurrently over one asyncssh connection. Each still carries the same
  20 s self-kill window and its own cleanup in `finally`, so a missed cleanup self-heals as before;
  a crash mid-flight now strands 5 short-lived containers instead of 1.
- **`ConnectivityOrchestrator.verify` renames a required argument** (`rented_ports` →
  `unavailable_ports`). It has exactly one production caller, `ExecutorConnectivityService`, which
  is updated in the same commit; the orchestrator tests are updated too. `verify_ports` keeps
  `rented_ports` unchanged, so nothing outside `executor_connectivity/` sees the rename.
- **Wrong filler ports would shrink the candidate pool.** The exclusions come straight from
  `FillerRun.port_maps`; a stale row would remove a handful of ports from a ~1000-port range, which
  the selector simply skips over. Low impact.
- **A filler in `STARTING` contributes no exclusions.** The backend writes `port_maps` at the
  transition to `RUNNING`, while the rented-executors query also returns `STARTING` runs younger
  than 15 min. Such a filler already holds its ports but ships none, so the validator can still
  probe them — chunking is what keeps that from costing more than one chunk.
- The dead `--network=host` tier on these hosts is untouched — these executors stay capped at 50
  probed ports until that firewall story is handled separately.

### Test Case 1 — a busy port fails only its own chunk

**Actions** `test_semi_batch_busy_port_fails_only_its_own_chunk` — 25 candidate ports, the runner
refuses to start any container whose publish flags include `-p 40012:40012`.

**Expected Output** ports 0–9 and 20–24 verified, ports 10–19 failed. Previously all 25 failed.

### Test Case 2 — filler ports reach the exclusion set

**Actions** `test_verify_ports_merges_rented_and_filler_ports` calls `verify_ports` with
`rented_ports=[40100]` and `filler_ports=[40001, 40003]`;
`test_port_connectivity_passes_filler_ports_as_exclusions` builds a `RentedExecutorsResponse` with
an **empty** `executors` dict (the filler-only case) and
`filler_ports_by_executor={"executor-123": [40001, 40003]}`.

**Expected Output** the orchestrator receives `unavailable_ports` covering `{40001, 40003, 40100}`
and no `rented_ports` at all; the check calls `verify_ports` with `filler_ports=[40001, 40003]` and
`rented_ports=[]`.

### Test Case 3 — no regression in the tier's failure modes

**Actions** existing `test_semi_batch_returns_empty_when_run_raises` /
`_when_test_many_raises` / `_survives_cleanup_failure`, plus
`test_semi_batch_returns_empty_when_no_chunk_can_start` extended to 25 ports across 3 chunks, and
`test_port_probe_keeps_partial_semi_batch_result`.

**Expected Output** exceptions never escape the tier (an escaping one would zero the whole
verification and trip the fatal `PortCount` check); an all-chunks-fail run returns empty rather
than raising; a partial semi-batch result does not trigger the sequential tier.

### Test Case 4 — full suite

**Actions** `pdm run pytest` in `neurons/validators`.

**Expected Output** 1233 passed, 3 failed — the 3 failures are in
`tests/test_sshd_bootstrap_script.py` and reproduce identically on unmodified `main`.
`ruff check` over the touched trees reports the same 19 pre-existing findings as before the change.

## Checklist

- [x] Proper labels — this repo has no `ready-review`/`require-qa` labels defined; nothing to add
- [x] Risks are documented
- [x] Tests are described

### DIFF
--- a/neurons/validators/src/protocol/vc_protocol/compute_requests.py
+++ b/neurons/validators/src/protocol/vc_protocol/compute_requests.py
@@ -92,6 +92,10 @@ class RentedExecutorsResponse(BaseModel):
     # processes are tolerated. Additive over the legacy map above so backend/validator deploy in any
     # order — an older backend leaves this empty and get_filler_containers falls back to the one.
     all_filler_containers_by_executor: dict[str, list[str]] = {}
+    # executor_id -> external ports its active fillers hold (DAH-2527). A filler creates no pod, so
+    # these ports never show up in executors[].pods and port verification used to probe them and
+    # collide. Empty for a backend that predates the field — the collision simply stays.
+    filler_ports_by_executor: dict[str, list[int]] = {}
     banned_guids: list[str] = []
     gpu_splitting_config: dict[str, int] = {}  # executor_id → min_gpu_count_for_rental
     network_ema: dict[str, NetworkEMA] = {}  # executor_id → EMA network speeds, all active executors
@@ -139,6 +143,10 @@ class RentedExecutorsResponse(BaseModel):
         # (skip-checks, logging). Container protection and GPU-usage tolerance use get_filler_containers.
         containers = self.get_filler_containers(executor_uuid)
         return containers[0] if containers else None
+
+    def get_filler_ports(self, executor_uuid: str) -> list[int]:
+        """External ports held by this executor's fillers, to be excluded from port verification."""
+        return self.filler_ports_by_executor.get(str(executor_uuid), [])
 
     def get_default_job_owner(self, executor_uuid: str) -> str | None:
         return self.default_job_owner_by_executor.get(str(executor_uuid))
--- a/neurons/validators/src/protocol/vc_protocol/compute_requests.py
+++ b/neurons/validators/src/protocol/vc_protocol/compute_requests.py
@@ -145,7 +145,6 @@ class RentedExecutorsResponse(BaseModel):
         return containers[0] if containers else None
 
     def get_filler_ports(self, executor_uuid: str) -> list[int]:
-        """External ports held by this executor's fillers, to be excluded from port verification."""
         return self.filler_ports_by_executor.get(str(executor_uuid), [])
 
     def get_default_job_owner(self, executor_uuid: str) -> str | None:
--- a/neurons/validators/src/services/executor_connectivity/orchestrator.py
+++ b/neurons/validators/src/services/executor_connectivity/orchestrator.py
@@ -33,6 +33,7 @@ class ConnectivityOrchestrator:
         sysbox_runtime: bool,
         rented_ports: list[int] | None,
         ssh_client,
+        excluded_ports: list[int] | None = None,
         log_ctx: dict | None = None,
     ) -> PortVerificationResult:
         log_ctx = {
@@ -40,8 +41,10 @@ class ConnectivityOrchestrator:
             "executor_uuid": executor_info.uuid,
             "executor_ip": executor_info.address,
         }
-        rented = set(rented_ports) if rented_ports else set()
-        ports = self.port_selector.select(executor_info, BATCH_PORT_VERIFICATION_SIZE, rented)
+        # rented ports come from pods, excluded ones from idle fillers (DAH-2527) — different
+        # sources, same meaning here: the port is taken and probing it would only collide
+        unavailable = set(rented_ports or []) | set(excluded_ports or [])
+        ports = self.port_selector.select(executor_info, BATCH_PORT_VERIFICATION_SIZE, unavailable)
 
         if not ports:
             return PortVerificationResult(
--- a/neurons/validators/src/services/executor_connectivity/orchestrator.py
+++ b/neurons/validators/src/services/executor_connectivity/orchestrator.py
@@ -31,9 +31,8 @@ class ConnectivityOrchestrator:
         executor_info: ExecutorSSHInfo,
         miner_hotkey: str,
         sysbox_runtime: bool,
-        rented_ports: list[int] | None,
+        unavailable_ports: list[int] | None,
         ssh_client,
-        excluded_ports: list[int] | None = None,
         log_ctx: dict | None = None,
     ) -> PortVerificationResult:
         log_ctx = {
@@ -41,10 +40,9 @@ class ConnectivityOrchestrator:
             "executor_uuid": executor_info.uuid,
             "executor_ip": executor_info.address,
         }
-        # rented ports come from pods, excluded ones from idle fillers (DAH-2527) — different
-        # sources, same meaning here: the port is taken and probing it would only collide
-        unavailable = set(rented_ports or []) | set(excluded_ports or [])
-        ports = self.port_selector.select(executor_info, BATCH_PORT_VERIFICATION_SIZE, unavailable)
+        ports = self.port_selector.select(
+            executor_info, BATCH_PORT_VERIFICATION_SIZE, set(unavailable_ports or [])
+        )
 
         if not ports:
             return PortVerificationResult(
--- a/neurons/validators/src/services/executor_connectivity/port_selector.py
+++ b/neurons/validators/src/services/executor_connectivity/port_selector.py
@@ -11,11 +11,11 @@ from services.port_utils import get_all_ports
 class PortSelector:
     """Selects which ports to verify."""
 
-    def select(self, executor_info: ExecutorSSHInfo, size: int, unavailable: set[int]) -> list[PortPair]:
+    def select(self, executor_info: ExecutorSSHInfo, size: int, unavailable_ports: set[int]) -> list[PortPair]:
         """Select ports to check, skipping external ports already taken by pods or fillers."""
         all_ports = get_all_ports(executor_info.port_range, executor_info.port_mappings, executor_info.ssh_port)
         available_ports = [
-            PortPair(internal, external) for internal, external in all_ports if external not in unavailable
+            PortPair(internal, external) for internal, external in all_ports if external not in unavailable_ports
         ]
         return available_ports[:size]
 
--- a/neurons/validators/src/services/executor_connectivity/port_selector.py
+++ b/neurons/validators/src/services/executor_connectivity/port_selector.py
@@ -11,9 +11,11 @@ from services.port_utils import get_all_ports
 class PortSelector:
     """Selects which ports to verify."""
 
-    def select(self, executor_info: ExecutorSSHInfo, size: int, rented: set[int]) -> list[PortPair]:
-        """Select ports to check, based on executor params and info about rented external ports."""
+    def select(self, executor_info: ExecutorSSHInfo, size: int, unavailable: set[int]) -> list[PortPair]:
+        """Select ports to check, skipping external ports already taken by pods or fillers."""
         all_ports = get_all_ports(executor_info.port_range, executor_info.port_mappings, executor_info.ssh_port)
-        available_ports = [PortPair(internal, external) for internal, external in all_ports if external not in rented]
+        available_ports = [
+            PortPair(internal, external) for internal, external in all_ports if external not in unavailable
+        ]
         return available_ports[:size]
 
--- a/neurons/validators/src/services/executor_connectivity/port_verifiers.py
+++ b/neurons/validators/src/services/executor_connectivity/port_verifiers.py
@@ -204,13 +204,18 @@ class FallbackVerifier:
 
 
 class SemiBatchVerifier:
-    """Verifies ports by publishing a whole batch in ONE container via -p.
+    """Verifies ports by publishing them via -p, in one container per chunk.
 
     Middle tier of the cascade: published ports are reached through the nat/DOCKER
     DNAT chain, so they survive a ufw default-deny INPUT policy — the same policy
     that silently drops BatchVerifier's --network=host listeners. Returns nothing
-    verified if the container cannot start, letting the caller fall through.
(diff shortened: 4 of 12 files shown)


arhangel66:
## Staging experiment — chunked semi-batch port verification (DAH-2527)

**TL;DR** — Confirmed. Under a reproduction of the prod signature (default-deny firewall + two
already-allocated ports inside the probed range), the current validator verifies **9 ports**; the
DAH-2527 build verifies **30**, and the two busy ports cost only their own chunk of ten instead of
the entire tier. The filler-port exclusion half of the change was **not** exercised — staging's
backend was held by another branch, so only the validator half was deployed.

### Environment

| what | value |
|---|---|
| Executor (subnet uuid) | `66e9c110-d9db-4a1e-a2cd-6ea1647ae81b` |
| Host | `149.36.1.151` (staging A4000, Shadeform) |
| Miner hotkey | `5DfbmGgkgYoYKJk7rA9UFgeJFacfkBMhaeFZgJk7c4mCv7DT` |
| Validator hotkey | `5DHgdomxxHSQf7neJWcagvGhTMKpqvKWwuprEaFF8C653qty` |
| Baseline image | `lium-staging-validator:95e6b7c` (no DAH-2527 code) |
| Treatment image | `lium-staging-validator:1d042ad` (branch `DAH-2527-filler-port-exclusion`) |
| Backend | `lium-staging-backend:b116054` in **both** runs — the new `filler_ports_by_executor` field was never sent |

Stand preparation (all reverted afterwards):

- `RENTING_PORT_RANGE` widened `40000-40009` → `40000-40049`, executor container recreated, so the
  semi-batch tier had a full 50 candidates to work with.
- `ufw` set to default-deny incoming with only `22/tcp` allowed. This kills the host-network
  `BatchVerifier` tier exactly as the provider firewall does on the three prod H200s: docker
  published ports traverse the nat/DOCKER DNAT chain and survive the deny, raw host listeners do
  not. Verified externally before starting: `40009` (docker-published) open, a host-network
  listener on `40030` refused.
- Two ports inside the range deliberately held: `40009` by a **live PEARL filler**
  (`filler_97a2cbcd-9aa2-43ef-8835-0b27ea0d0a0e`, the real phenomenon under investigation) and
  `40025` by a dummy container `dah2527_busy_40025` (a second collision far from the first, to
  prove chunk isolation rather than a lucky boundary).

### Timeline (UTC, 2026-07-29)

| time | event |
|---|---|
| 07:40 | stand armed: port range widened, executor recreated, `40025` occupied |
| 07:48 | `ufw` enabled, default deny incoming, `22/tcp` allowed |
| 07:53:52 | **baseline** cycle on `95e6b7c` — `ok=9`, total 54.46 s |
| 07:58 | validator + backend staging deploys triggered for DAH-2527 |
| 08:02 | validator live on `1d042ad`; backend deploy **failed** (see side findings), stayed on `b116054` |
| 08:05:20 | **treatment** cycle — `ok=30`, total 24.41 s |
| 08:27 | stand restored (`ufw disable`, dummy removed, port range back to `40000-40009`); validator redeployed to `DAH-2243` |

### Mechanism

1. `BatchVerifier` starts 300 host-network listeners and the validator reaches none of them —
   `progress: 0/50 verified`, twice. This tier is dead behind the firewall and the cascade in
   `port_probe.py` falls through to semi-batch.
2. **Before:** `SemiBatchVerifier.verify` published all 50 candidates in a single `docker run`.
   Docker refuses the whole container when one published port is already allocated, so `40009`
   returned zero verified ports for the entire tier. The cascade then fell through again to
   `FallbackVerifier`, which tests only 10 ports one at a time — 9 of them passed, and that 9 is
   what the executor was scored on.
3. **After:** `verify()` splits the candidates into chunks of `CHUNK_SIZE = 10` and runs them
   concurrently, one container per chunk with its own token and its own cleanup
   (`port_verifiers.py:_verify_chunk`). A chunk that fails to start contributes its ports to
   `failed` and leaves the others untouched.
4. Chunks `40000-40009` (holding `40009`) and `40020-40029` (holding `40025`) failed; the other
   three returned all their ports. The tier produced a non-zero result, so `FallbackVerifier` never
   ran at all — which is also why the cycle got ~30 s faster.

### Key log evidence

Baseline, `95e6b7c` — one busy port zeroes the tier:

```
semi-batch: publishing 50 ports in one container
semi-batch start failed: ... Bind for 0.0.0.0:40009 failed: port is already allocated
semi-batch verification failed, trying fallback
fallback: 9/10 verified
verification complete total_time=54.46s 90% available, dind=ok batch=ok ok=9[40000,...] fail=1[40009]
```

Treatment, `1d042ad` — the blast radius is one chunk:

```
semi-batch: publishing 50 ports in 5 containers
semi-batch start failed for 10 ports: ... Bind for 0.0.0.0:40025 failed: port is already allocated
semi-batch start failed for 10 ports: ... Bind for 0.0.0.0:40009 failed: port is already allocated
semi-batch: 30/50 verified
verification complete total_time=24.41s 60% available, dind=ok batch=ok ok=30[40010, 40011, 40012, 40013, 40014] port_range=40000-40049 fail=20[40000, 40001, 40002, 40003, 40004]
```

`reason_code=PORT_VERIFY_OK`, `verified_port_count: 30` (baseline: 9).

### Blast radius

The three prod H200s (`152.236.142.238` / `.240` / `.241`) currently report 5–6 verified ports out
of a ~1000-port range every cycle, which is what they are scored on. The validator half alone lifts
them to roughly 30 — the fillers there hold 4–8 low ports, so two or three chunks of five are lost
and the rest survive. The backend half removes those collisions entirely and should give ~49–50.
No miner action is needed and nothing has to be restarted: the change is entirely validator-side and
takes effect on the first cycle after deploy.

### Side findings

1. **Staging's database carries a feature-branch migration.** `DAH-2243` deployed
   `e1b999164cda` (`add_pod_volume_encryption_preference`) to the staging DB at ~07:00. Any branch
   that does not contain that revision now fails its deploy: the `app-db-migration` Job dies with
   `FAILED: Can't locate revision identified by 'e1b999164cda'` and Pulumi aborts. Two independent
   deploys hit this within 20 minutes (`4434f26` and `bb06fa4`). Staging is effectively locked to
   whichever branch migrated it last, and the failure mode gives no hint of that.
2. **The failed deploy left staging in a mixed state.** Pulumi aborted after updating the `beats`
   deployment but before `compute-app` / `workers`, so `beats` ran `bb06fa4` while the API ran
   `b116054`. A failed backend deploy is not atomic across deployments.

### Recommendations

Not implemented here — separate tasks.

1. Make the migration Job's failure message name the offending revision's branch, or have the
   staging deploy detect "DB is ahead of this branch" and say so explicitly. The current error is a
   dead end for whoever hits it.
2. `CHUNK_SIZE = 10` is a guess anchored to the sequential tier's cap. Real filler density is 4–8
   ports per machine clustered at the bottom of the range, so the first one or two chunks eat almost
   all collisions. Worth revisiting to 5 once the backend exclusion half is live and we can measure
   what is actually left.
3. The dead host-network tier is the real ceiling here — it caps these machines at 50 candidates
   instead of 300, and no amount of chunking recovers that. Separate investigation.

fortunelucky777 (approved):
Approving.

Verified beyond the diff:
- `verify_ports` and `ConnectivityOrchestrator.verify` each have exactly one production caller (`port_connectivity.py` → `ExecutorConnectivityService`), both updated in this commit, so the `rented_ports` → `unavailable_ports` rename leaks nowhere.
- `rented_ports` still gates the sysbox fallback on its own, so a filler-only machine is not routed down the rented branch.
- Chunk results are concatenated in chunk order, so `successful` stays ascending for the orchestrator's `successful.pop(0)` DinD pick.
- `port_test_` is not in `RENTAL_CONTAINER_PREFIXES`, so the extra containers per cycle stay invisible to the stale-container reaper.
- Backend contract (#840) matches: `filler_ports_by_executor` is keyed by `Executor.executor_id` — the subnet UUID `ctx.executor.uuid` looks up — not the backend PK.

One nit, non-blocking:
- `tests/executor_connectivity/test_port_verifiers.py:72` — `assert runner.run.await_count == 50 / SemiBatchVerifier.CHUNK_SIZE` compares an int against `5.0`. `//` or a literal `5` reads clearer.

PR #1177, opened 2026-07-29
DAH-2528 - add occupancy-aware bucket fallback for split-capable idle nodes

fortunelucky777:
## Describe your changes

**Problem.** Bucket assignment for the unrented (idle) incentive ignores how full a bucket already is. `_resolve_bucket` pins a split-capable executor to its `gpu_count` bucket whenever that bucket has a configured cap — even when it is over cap and the node's split tier sits empty. Measured in prod (job cycle 2026-07-29 08:49:24 UTC): H200·8 at 128/64 GPUs (`cap_multiplier` 0.500, 16 executors each at incentive 0.007602590530147434) while H200·1 sat at 0/10.

**Fix.** A second, occupancy-aware pass (`_reassign_split_candidates`, run at the top of `_on_finish_pre_process`, before cap multipliers are computed):

- Candidates: idle, rental-eligible, split-capable executors initially placed in their `gpu_count` bucket.
- A node moves **whole** (all `gpu_count` GPUs + its weighted-rate contribution) into its `gpu_splitting_min_count` tier only when the source bucket is **strictly over cap**, the target tier has a positive cap, and the whole node **fits at or under the target cap** (strict admission — a move never over-fills the target, so incumbents are never diluted by a newcomer). Admission implies the mover always improves: target pays exactly 1.0, over-cap source pays < 1.0.
- Greedy in ascending executor-uuid order with fills recomputed after every accepted move, so an unchanged fleet reproduces byte-identical assignments every cycle.
- New miner-facing report line `unrented_bucket_reassigned` (fields: `event`, `bucket_reassigned_from`, `bucket_reassigned_from_multiplier`, `count_bucket`, `max_cap`, `unrented_cap_multiplier`) explains why the node was rated in another tier. Not a `ZeroIncentiveReason` — the node is paid, in a better bucket.
- `estimate_executor` / `precompute_all_estimates` inherit the rule automatically (same 3-stage pipeline, fresh instance per estimate), and `get_snapshot()` persists post-move fills — portal estimates stay equal to what the validator pays.
- Corrected the stale `_pre_process_job_result` / class docstrings that claimed split-capable executors land in their `gpu_splitting_min_count` bucket.

**Scope note (intended behavior, agreed with task owner).** Strict admission strands headroom smaller than a node (e.g. 8-GPU rigs leave 2 of 10 slots in the 1× tier unused), and with prod caps `{1: 10, 8: 64}` the fallback only fires while the 1× tier holds ≤ 2 GPUs and `gpu_splitting_min_count == 1`. A node's bucket may flip between cycles when other miners' 1× nodes arrive — the new log line makes each flip explainable. Non-splitting nodes, the reverse direction, and partial rentals (DAH-2467) are out of scope.

**Verification run:** full validator suite **1286 passed, 8 skipped**; 8 new tests + 1 updated (fallback happens / strict admission rejects / lands exactly at cap / greedy stops when source back at cap / deterministic rerun / source emptied to zero / unconfigured split tier / estimate parity); zero net-new `ruff@0.5.1` violations on touched files.

**Pre-deploy verification (from the task spec):**
1. Unit: construct a fleet where a model's 8 bucket is over cap and its 1 bucket has headroom. Assert a split-capable idle 8x node resolves to bucket 1 with all 8 GPUs counted there, and a non-splitting 8x node still resolves to bucket 8. ✅ `test_pcc_2528_split_node_falls_back_when_8_bucket_over_cap`
2. Unit: with the target bucket already at cap, assert the split-capable node stays in bucket 8 — no reassignment that makes things worse. ✅ `test_pcc_2528_no_move_when_whole_node_does_not_fit_target`
3. Staging: lower the 8 cap so the bucket is over-filled, register a split-capable idle node, force a job cycle. The incentive log shows the node in bucket 1, carries the reassignment reason, and its cap_multiplier is higher than the 8 bucket's.
4. Staging: run two consecutive cycles with an unchanged fleet and confirm byte-identical bucket assignments — no flapping.

**Post-deploy verification (from the task spec):**
- +1 d — validator incentive logs contain reassignment events (`event: unrented_bucket_reassigned`); for every reassigned executor the target-bucket cap_multiplier exceeds its source-bucket value, and no executor changes bucket between consecutive cycles while the fleet is unchanged.
- +7 d — mean cap_multiplier across active buckets rises against the 2026-07-29 08:49:24 UTC baseline (H200·8 = 0.500, H200·1 absent, RTX 5090·1 = 0.909), the count of unrented GPUs paid at full weight increases, and total_rental_cost still clears the burn-emission cap.

[DAH-2528 — Idle incentive: occupancy-aware bucket fallback for GPU-splitting nodes](https://app.notion.com/p/Idle-incentive-occupancy-aware-bucket-fallback-for-GPU-splitting-nodes-3acb8bfdbde9814e89a5ce806036b8af)

- [x] I wrote tests.

### DIFF
--- a/neurons/validators/src/incentive/miner_incentive_log.py
+++ b/neurons/validators/src/incentive/miner_incentive_log.py
@@ -28,7 +28,9 @@ WHAT THIS CATALOG HOLDS — every `MinerLogLine` the miner-facing log block
 
 2. CALCULATION REPORTS — the per-cycle score/incentive lines every scored node gets:
      mining_score_calculated, mining_incentive_calculated,
-     rental_incentive_calculated, mining_score_missing (internal-error case)
+     rental_incentive_calculated, mining_score_missing (internal-error case),
+     unrented_bucket_reassigned (DAH-2528: node rated against its split tier
+     because its own GPU-count tier was over capacity)
 
 The scoring code (rental_price.py / default.py) detects each condition where its
 data naturally lives (some per-executor upfront, some only after cohort aggregation),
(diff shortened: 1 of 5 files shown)


arhangel66 (approved):
looks good

PR #1178, opened 2026-07-30
DAH-2534: run rental container bootstrap execs as root so non-root images can start

taiberium:
## Problem

A renter whose Docker image declares a non-root `USER` can never start a pod.

`docker exec` inherits the image's `USER` when none is given. Every rental bootstrap
step we run through `_exec_in_container_sync` writes to `/root` or `/etc`, so on a
non-root image they all fail with:

```
mkdir: cannot create directory '/root': Permission denied
```

Prod, last 24h: **52 of 53 `rent_failed` rentals are this bug** — 30% of all 176 rental
attempts. One customer is fully blocked: 123 attempts on `ghcr.io/baseintelligence/prism-evaluator`,
120 `rent_failed`, rate climbing 20 → 29 → 64/day since 27 Jul. Their image is fine —
`useradd prism && USER prism` is Docker's own recommended practice. Docker never
guaranteed `USER` is root; assuming it was ours to get wrong.

Not a regression: `git log -S'exec_create'` shows the SDK path never passed `user=`,
and the legacy shell path never passed `-u root`. It only surfaces when a renter brings
a non-root image.

## Fix

Pin every exec that targets a **rental** container to root; leave DinD and helper
containers alone.

- `rental_docker_sdk.py` — `user="root"` on `exec_create`. Covers all five specs routed
  through it: key injection, key removal, `/etc/environment`, and the sshd bootstrap
  script create + run.
- `docker_service.py` — `-u root` on the 10 hand-built exec strings in
  `setup_encrypted_local_volume` and `run_jupyter`. `run_jupyter` had the identical
  `mkdir -p /root` defect, so without this a non-root image would have failed two steps
  later with the same error.
- `workspace_mount.py` — `-u root` in `VolumeAccess.docker_exec_args`, the third exec
  builder in the repo. It drives encrypted-workspace backup/restore (`du`/`find`/`mkdir`/`tar`);
  the restore-side writes would have hit the same failure.
- `docker_utils.py` — `-u root` in the `DockerCommand` builder.
- `rented_machine.py` — read the literal `/root/.ssh/authorized_keys` instead of `~`, so
  the verification read side matches the injection write side exactly.

The renter's own workload still runs as the image's `USER`; only our bootstrap execs are
pinned.

## What this can break

Enumerated every `docker exec` call site rather than trusting the first one:

- **Rental-container execs: 16 found, 16 now pinned.** 10 in `docker_service.py`
  (`setup_encrypted_local_volume`, `run_jupyter`), 5 SDK specs, 1 in `workspace_mount.py`.
- **DinD execs: 4, deliberately untouched** — `docker_service.py` 3161 / 3335 / 3425 / 3458
  target `lium-dind-build-*`, never a renter image. The dind image `daturaai/dind:0.0.1`
  already runs as root, so pinning would be a no-op anyway.
- **Helper-container execs untouched** — `backup_storage.py:170,173` exec into a throwaway
  `ubuntu` container, not a rental one.
- **The 99% case is unchanged.** For a root image `-u root` and `~` → `/root` are both no-ops.
- **Sysbox unaffected.** `-u root` means uid 0 inside the container's own userns;
  capabilities, `privileged` and the sysbox runtime are untouched.
- **Nothing relies on inheriting the image USER.** Every exec here is validator-side
  bootstrap writing `/root/.ssh`, `/etc/environment`, `/etc/ssh`, `/run`. The renter's
  workload starts via `docker run`, which still honors the image `USER`.
- **API support verified.** `docker[ssh]>=7.1.0` — `exec_create(..., user='')` is in the
  pinned signature; the value rides in a request body that was already being sent, so no
  extra round-trip to the daemon.
- **No new Redis touchpoints**, so no new failure mode when Redis is down.

Known and deliberately left alone: `install_open_ssh_server_and_start_ssh_service`
(`docker_service.py:2134`) has two unpinned rental execs, but it is **dead code** — both
production call sites (`:4439`, `:4865`) use the `_with_rental_docker` twin. Adding a flag
to dead code is pointless; deleting the method is a separate cleanup.

## Tests

TDD — the root-pinning assertion was red before the change. Coverage is folded into the
tests that already exercise these paths rather than added as parallel ones:
`test_rental_docker_sdk.py` asserts the full `exec_create` kwargs including `user`,
`test_rented_machine_check.py` asserts the built command carries `-u root`, and
`test_workspace_backup_restore.py` asserts the new argv.

Full validator suite: **1315 passed, 5 skipped**.

Three `test_sshd_bootstrap_script.py` failures are **pre-existing on `origin/main`** —
verified by stashing this diff and re-running; unrelated to this change.

## Deliberately not in this PR

Worth their own tickets:

- Template validation: reject or warn on a non-root image at save time.
  `docs/pod-users/templates.md:71` says user-template verification is currently disabled,
  so nothing catches this on save.
- Renter-facing error: surface "your image runs as non-root" instead of raw `mkdir` stderr.
- Routing guard: an executor that fails `add_public_keys` should not be silently re-picked
  120 times.

Note for the affected customer after this lands: the default volume mounts at `/root`,
which stays root-owned, so their `prism` process still cannot write there. Their image
already chowns `/workspace` and `/artifacts` — they likely want the volume path pointed
at one of those. That is a template setting, not a platform bug.


arhangel66 (approved):
Approving. I looked into the reviewer finding about `-u 0` on the final `run_jupyter.sh` exec — it's factually correct but not a blocker:

- `run_jupyter.sh` cannot run as a non-root user at all: it installs system packages (`apt-get`/`apk`), creates the venv in `/opt/jupyter-env`, writes `/jupyter.log`, and starts with `--allow-root --ServerApp.root_dir=/root`. Launching it as `.Config.User` would break Jupyter on exactly the non-root images this PR fixes.
- No regression: on root images `-u 0` is a no-op (Jupyter ran as root before); on non-root images the rental failed at bootstrap, so "Jupyter as image user" never existed as a working state.
- No security boundary crossed: renter SSH keys go into `/root/.ssh/authorized_keys`, so the renter is root inside the container by design.

Two small follow-ups, neither blocking merge:
1. On non-root images, notebooks will create root-owned files in a workspace we deliberately chown to the image user — worth a ticket for a privilege drop inside `run_jupyter.sh` after the install phase.
2. The PR description line "the renter's own workload still runs as the image's USER" deserves an explicit exception note: Jupyter is pinned to root, deliberately.

PR #1181, opened 2026-07-31
DAH-2306: heal reboot-stranded pods before penalising POD_NOT_RUNNING

arhangel66:
[Notion task](https://app.notion.com/p/Auto-recover-pods-stuck-after-host-reboot-stale-vloopback-mountpoint-POD_NOT_RUNNING-38eb8bfdbde981c7ab93c6740fb1be0d)

## What was broken

POD_NOT_RUNNING is ~23% of penalty failures (~83/350). Over a 7-day validator log window 22 of 31
POD_NOT_RUNNING pods failed with exit code 255 and
`VolumeDriver.Mount: cannot create mount point dir '/mnt/volume_<pod>': mkdir ...: file exists`,
and 20 of those coincide with a host reboot (e.g. pod `e0c58a78-532e-4283-87a0-c488a7bbe534`, host
boot 2026-06-24 16:12:28, pod finished 16:12:41).

Pods run with `--restart unless-stopped`. After a reboot dockerd auto-restarts the pod and calls the
vloopback plugin, but the mountpoint dir under `propagated-mount/volume_<pod>` survived the reboot
while the loopback mount did not, so the plugin's plain `mkdir` hits "file exists" and the container
stays down. `repair_stale_vloopback_mountpoint` (DAH-2133) already fixes exactly this, but only on
the create/rental path — nothing repaired it on the host-side auto-restart, so the pod stayed down
until the rental ended and the miner was penalised every cycle.

## How it is fixed

`TenantEnforcementCheck` now heals the pod before it penalises the miner
(`services/task/checks/rented_machine.py`). Recovery only runs when every one of these holds, and
they are what makes it safe:

- the check is not the dry-run instance (`TenantEnforcementCheck(recover_stale_pods=False)` in
  `build_dry_run_checks`) — the dry-run pipeline runs against live executors next to production
  validation and must not rmdir a host path or start a customer's container;
- the backend still reports the rental active (existing `get_pod_rental_active` call);
- the container's own `State.Error` carries all three phrases of the vloopback signature
  (`_is_stale_vloopback_mountpoint_error`, reusing `_VLOOPBACK_MOUNT_ERROR_PHRASES`);
- the pod has **no encrypted volume** — see the scope note below.

A pod a customer stopped through `ContainerStopRequest` exits cleanly with an empty `State.Error`,
so it can never match and is never resurrected.

On a match the check runs the existing `repair_stale_vloopback_mountpoint` and then restarts the
container through the **full** start path — `start_existing_container`, lifted out of
`start_container`: `docker start` plus the sshd bootstrap. A bare `docker start` was deliberately
rejected: it would report the pod "running" — and so silence the penalty — while the customer still
cannot get in, and the same method is what the create path uses to restore a gocryptfs plaintext
mount. After a successful start the pod is re-read,
so the recovered container's own SSH keys are what gets reported and a start that did not stick
still lands on POD_NOT_RUNNING.

Everything that is not the pod being down leaves the POD_NOT_RUNNING verdict alone rather than
turning it into a free pass. In particular a command that times out is reported as a failed
recovery, not as a dead transport: `asyncssh.TimeoutError` subclasses both `asyncssh.Error` and
`OSError`, so without an explicit catch it would reach the check's DAH-2055 handler and the miner
would dodge the penalty on any slow executor.

Everything else is wiring: `DockerService` and the decrypted executor SSH key are threaded into the
check context, and `DockerService` is now constructed before `TaskService` in both composition
roots. The `TYPE_CHECKING` imports in `services/task/` are load-bearing — `docker_service` imports
that package, so a runtime import back is a circular import.

## Scope note

**Encrypted volumes are out of scope.** For an encrypted rental docker mounts the ciphertext at
`/lium-cipher` and the plaintext path the customer actually uses comes from the create request —
it is recorded nowhere on the host and is not in `PodRentalActiveResponse` either. Remounting
gocryptfs at the `/root` default would leave a custom path (say `/data`) an ordinary container dir,
so everything the customer wrote there after recovery would land unencrypted on the miner's disk.
Recovery therefore inspects the container's mounts first and stands down when it sees
`/lium-cipher`, failing closed if the inspect itself does not answer. Re-enabling encrypted pods
needs `local_volume_path` on the backend's rental-active response — a separate compute-app change.

This PR also replaces the executor-side `entrypoint.sh` reconcile from the original spec. That plan
does not survive contact with the code: the runner entrypoint is not boot-only (the runner is
`restart: unless-stopped` and watchtower-managed on a 60s poll), it runs under `set -eu` after
`docker compose up --wait` so a failure there would brick the executor fleet-wide, a blanket
`docker start pod_*` would override a customer's explicit stop, and the executor host has neither
the rental ground truth nor the volume-encryption key needed to restart a pod correctly. The fix
therefore ships as a validator release, not an executor image release.

## Test plan

Unit (`cd neurons/validators && pdm run pytest tests/ -q`): 1364 passed. The 3 failures in
`tests/test_sshd_bootstrap_script.py` are a pre-existing baseline, unchanged on a clean tree.
`ruff check` on the touched files reports the same 8 errors before and after — no new lint.

- [x] service level: signature does not match -> no repair, no start
- [x] service level: no `State.Error` at all -> no repair
- [x] service level: encrypted volume (`/lium-cipher` mounted) -> no repair, no start
- [x] service level: container inspect fails -> treated as encrypted, no repair
- [x] service level: a repair command times out -> no start, POD_NOT_RUNNING preserved
- [x] service level: repair fails -> no start
- [x] service level: happy path -> repair on the volume the container actually mounts, start
      through the full path with no plaintext path
- [x] service level: start raises -> reported as not recovered, half-started container stopped
- [x] check level: recovered -> passes, no `clear_verified_job_info`, recovered pod's SSH keys reported
- [x] check level: recovery declines / pod stays down / recovery raises / no SSH key -> POD_NOT_RUNNING preserved
- [x] check level: SSH transport lost during recovery -> EXECUTOR_TRANSPORT_UNREACHABLE, not a penalty
- [x] check level: dry-run instance -> recovery never called, POD_NOT_RUNNING preserved
- [x] pipeline level: `build_dry_run_checks` carries `recover_stale_pods=False`

Pre-deploy verification on staging:

- [x] rent a pod, write a marker file into `/root`, reboot the executor host
- [x] confirm the failure signature: `docker inspect --format '{{json .State}}' pod_<id>` shows a
      non-empty `Error` with `cannot create mount point dir ... file exists`, and POD_NOT_RUNNING fires
- [x] deterministic fallback if the reboot does not reproduce it: stop the pod, `mkdir
      /var/lib/docker/plugins/<plugin_id>/propagated-mount/volume_<pod_id>`, then `docker start pod_<id>`

Post-deploy verification on staging:

- [x] on the next validation cycle the pod comes back up and no POD_NOT_RUNNING event fires
- [x] Loki carries `POD_STALE_MOUNT_RECOVERED` (failures land as
      `POD_STALE_MOUNT_RECOVERY_REPAIR_FAILED` / `_START_FAILED` / `_TIMED_OUT` /
      `POD_STALE_MOUNT_RECOVERY_FAILED`)
- [x] a healthy mounted pod on the same host is left untouched
- [ ] **re-run on an unencrypted pod.** The staging run in the comment below used an encrypted
      volume, which this revision now refuses; the shipped path needs its own run
- [ ] an encrypted pod hitting the same break logs
      `POD_STALE_MOUNT_RECOVERY_SKIPPED_ENCRYPTED_VOLUME` and still reports POD_NOT_RUNNING

Post-deploy on prod, +7 d:

- [ ] POD_NOT_RUNNING events with the `file exists` signature drop
- [ ] no rise in stale-volume or wrong-deletion errors

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -5018,13 +5018,13 @@ class DockerService:
         self,
         ssh_client: asyncssh.SSHClientConnection,
         container_name: str,
-        default_extra: dict,
+        log_extra: dict[str, Any],
     ) -> str | None:
         # the volume whose stale mountpoint kept the container down, read off the container itself.
         # `docker inspect` reports mounts of a stopped container, so the real name is taken from the
         # host instead of guessed from pod_id — a pod created through the edit path carries a
-        # backend-supplied volume name that no convention derives. repair_stale_vloopback_mountpoint
-        # is what decides: it accepts a candidate only if the driver is vloopback and the stale
+        # backend-supplied volume name that no convention derives. Every mount can be offered
+        # blindly: repair_stale_vloopback_mountpoint accepts only a vloopback volume whose stale
         # mountpoint dir is present and empty.
         inspect_result = await ssh_client.run(
             f"/usr/bin/docker inspect {shlex.quote(container_name)} "
@@ -5034,14 +5034,14 @@ class DockerService:
         if getattr(inspect_result, "exit_status", 0) != 0:
             return None
 
-        for candidate in (inspect_result.stdout or "").split():
+        for candidate_volume in (inspect_result.stdout or "").split():
             repaired = await self.repair_stale_vloopback_mountpoint(
                 ssh_client,
-                candidate,
-                {**default_extra, "local_volume": candidate},
+                candidate_volume,
+                {**log_extra, "local_volume": candidate_volume},
             )
             if repaired:
-                return candidate
+                return candidate_volume
         return None
 
     async def _stop_half_recovered_container(
@@ -5055,11 +5055,11 @@ class DockerService:
         # the POD_NOT_RUNNING verdict keeps matching what the customer sees. A no-op when the start
         # itself was what failed.
         try:
-            result = await ssh_client.run(
+            stop_result = await ssh_client.run(
                 f"/usr/bin/docker stop {shlex.quote(container_name)}",
                 timeout=_VLOOPBACK_REPAIR_COMMAND_TIMEOUT_SEC,
             )
-            return getattr(result, "exit_status", 1) == 0
+            return getattr(stop_result, "exit_status", 1) == 0
         except asyncio.CancelledError:
             raise
         except Exception:
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -186,12 +186,6 @@ _VLOOPBACK_DRIVER_PREFIX = "vloopback"
 # Shared with core.docker_utils so exactly one helper image lands on nodes.
 _VLOOPBACK_REPAIR_IMAGE = ALPINE_HELPER_IMAGE
 _VLOOPBACK_REPAIR_COMMAND_TIMEOUT_SEC = 30
-# DAH-2306: plaintext path an encrypted rental volume is remounted at when a pod is recovered after
-# a reboot. The backend's own path is recorded nowhere on the host — an encrypted volume is mounted
-# at /lium-cipher, not at its plaintext path — so recovery falls back to the same default the create
-# path uses. A rental that both encrypts its volume and asks for a non-default path comes back with
-# its data under /root.
-_RECOVERED_POD_LOCAL_VOLUME_PATH = "/root"
 _DOCKER_VOLUME_NAME_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9_.-]*$")
 # DAH-2475: slack kept free above a rental's requested volume before we decide the DPHN filler cache
 # has to go. Covers the image layers and scratch the pod needs beyond its own volume.
@@ -4875,7 +4869,7 @@ class DockerService:
         private_key: str,
         known_hosts_policy: asyncssh.SSHKnownHosts | None,
         container_name: str,
-        local_volume_path: str,
+        local_volume_path: str | None,
         pod_id: str,
         default_extra: dict[str, Any],
     ) -> None:
@@ -4883,6 +4877,9 @@ class DockerService:
         # drops: the gocryptfs plaintext mount of an encrypted rental volume, and the sshd the
         # customer connects through. A failed remount raises and the caller decides how to report
         # it; a failed sshd bootstrap only warns, because the pod itself is up either way.
+        # local_volume_path=None means the caller does not know the plaintext path, which is only
+        # safe for a pod without an encrypted volume: mounting gocryptfs at a guessed path would
+        # leave the customer's real path an ordinary container dir, writing plaintext to the host.
         pkey = asyncssh.import_private_key(private_key)
         async with self.rental_docker_client_factory.connect(
             executor_info=executor_info,
@@ -4906,6 +4903,11 @@ class DockerService:
                     container_name,
                 )
                 if encrypted_volume_name:
+                    if local_volume_path is None:
+                        raise RuntimeError(
+                            f"{container_name} holds an encrypted volume but no plaintext path "
+                            "was supplied; refusing to remount it at a guessed path"
+                        )
                     await self.setup_encrypted_local_volume(
                         ssh_client=ssh_client,
                         container_name=container_name,
@@ -4960,11 +4962,31 @@ class DockerService:
             return False
 
         log_extra = {**default_extra, "container_name": container_name}
-        repaired_volume = await self._repair_stale_mountpoint_of_container_volume(
-            ssh_client,
-            container_name,
-            log_extra,
-        )
+        try:
+            if await self._holds_encrypted_local_volume(ssh_client, container_name):
+                logger.warning(
+                    _m(
+                        "POD_STALE_MOUNT_RECOVERY_SKIPPED_ENCRYPTED_VOLUME",
+                        extra=get_extra_info(log_extra),
+                    )
+                )
+                return False
+            repaired_volume = await self._repair_stale_mountpoint_of_container_volume(
+                ssh_client,
+                container_name,
+                log_extra,
+            )
+        except TimeoutError:
+            # a command that did not finish in time is not a dead SSH transport: report it here so
+            # the check keeps its POD_NOT_RUNNING verdict instead of turning it into
+            # EXECUTOR_TRANSPORT_UNREACHABLE and dropping the penalty.
+            logger.warning(
+                _m(
+                    "POD_STALE_MOUNT_RECOVERY_TIMED_OUT",
+                    extra=get_extra_info(log_extra),
+                )
+            )
+            return False
         if not repaired_volume:
             logger.warning(
                 _m(
@@ -4987,7 +5009,7 @@ class DockerService:
                 private_key=private_key,
                 known_hosts_policy=known_hosts_policy,
                 container_name=container_name,
-                local_volume_path=_RECOVERED_POD_LOCAL_VOLUME_PATH,
+                local_volume_path=None,
                 pod_id=pod_id,
                 default_extra=log_extra,
             )
@@ -5013,6 +5035,27 @@ class DockerService:
             _m("POD_STALE_MOUNT_RECOVERED", extra=get_extra_info(log_extra))
         )
         return True
+
+    async def _holds_encrypted_local_volume(
+        self,
+        ssh_client: asyncssh.SSHClientConnection,
+        container_name: str,
+    ) -> bool:
+        # whether the pod's rental volume is gocryptfs-encrypted, which recovery has to refuse.
+        # An encrypted volume is mounted as the ciphertext at /lium-cipher and the plaintext path
+        # the customer actually uses comes from the create request, which is recorded nowhere on
+        # the host and is not in the backend's rental-active response either. Remounting at the
+        # /root default would silently turn a custom path into an ordinary container dir, so
+        # writes to it would land unencrypted on the miner's disk. Fails closed: an inspect that
+        # does not answer counts as encrypted, and the miner keeps the POD_NOT_RUNNING verdict.
+        inspect_result = await ssh_client.run(
+            f"/usr/bin/docker inspect {shlex.quote(container_name)} "
+            '--format \'{{range .Mounts}}{{println .Destination}}{{end}}\'',
+            timeout=_VLOOPBACK_REPAIR_COMMAND_TIMEOUT_SEC,
+        )
+        if getattr(inspect_result, "exit_status", 0) != 0:
+            return True
+        return _LIUM_CIPHER_MOUNT in (inspect_result.stdout or "").split()
 
     async def _repair_stale_mountpoint_of_container_volume(
         self,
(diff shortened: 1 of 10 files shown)


arhangel66:
## DAH-2306 — staging verification

- **PR:** https://github.com/Datura-ai/lium-io/pull/1181
- **Deployed:** `568c8995` (branch `DAH-2306-executor-boot-reconcile-stale-vloopback`), workflow run https://github.com/Datura-ai/lium-io-deployment/actions/runs/30776999434, 2026-08-03 01:35 UTC
- **Staging executor:** `185.216.20.168` (1×A6000), executor uuid `b9a7d204-fd8c-4451-a4d0-f5c33773d46a`, miner `5DfbmGgkgYoYKJk7rA9UFgeJFacfkBMhaeFZgJk7c4mCv7DT`
- **Pod under test:** `6b8bacab-8335-4c30-a713-27003840c8a0` (huid `cosmic-matrix-49`, name `dah2306c`), encrypted volume

### Pre-deploy baseline

Captured on 2026-07-31 against the same executor, before the fix was live. The same
break sequence left the pod down and the check reported `POD_NOT_RUNNING` for pod
`6b0394a0-b8bd-4c12-892a-aeb32d358911` at 12:20:46 UTC (pipeline
`38819baf-237b-451e-b2a8-8e5d3706d64d`) — miner scored 0 for the cycle and nothing
ever brought the pod back.

### Reproduction on staging (2026-08-03 02:19 UTC)

Marker written into the gocryptfs plaintext dir first, so a bare `docker start` can be
told apart from a real recovery:

```
$ lium exec cosmic-matrix-49 "echo '...' > /root/DAH-2306-marker.txt; mount | grep -E 'gocryptfs|lium-cipher'"
/dev/loop3 on /lium-cipher type xfs (rw,relatime,idmapped,nouuid,...)
/lium-cipher on /root type fuse.gocryptfs (rw,nosuid,nodev,relatime,...allow_other,...)
```

Then the reboot aftermath, reproduced on the host:

```
$ docker stop pod_6b8bacab-8335-4c30-a713-27003840c8a0
$ sudo mkdir -p /var/lib/docker/plugins/5050ec7eb81856dc0fb716edcc9d4daeab871695252663a87cd7796c63da0de3/propagated-mount/volume_6b8bacab-8335-4c30-a713-27003840c8a0
$ docker start pod_6b8bacab-8335-4c30-a713-27003840c8a0
Error response from daemon: error while mounting volume '': VolumeDriver.Mount: 01KZ2PM56EVWB3AEYRPTDPR3K1: cannot create mount point dir '/mnt/volume_6b8bacab-8335-4c30-a713-27003840c8a0': mkdir /mnt/volume_6b8bacab-8335-4c30-a713-27003840c8a0: file exists

$ docker inspect pod_6b8bacab-8335-4c30-a713-27003840c8a0 --format '{{.State.Status}} | {{.State.Error}}'
exited | error while mounting volume '': VolumeDriver.Mount: ... file exists
```

Restart policy on the container is `unless-stopped` — dockerd had already tried and given up.

### Post-deploy — validator log (pod `validator-7bd4c64cc6-mfq9m`, image `568c899`)

```
2026-08-03 02:21:53,969 INFO VLOOPBACK_STALE_MOUNTPOINT_REPAIRED
  {"executor_uuid": "b9a7d204-fd8c-4451-a4d0-f5c33773d46a",
   "container_name": "pod_6b8bacab-8335-4c30-a713-27003840c8a0",
   "local_volume": "volume_6b8bacab-8335-4c30-a713-27003840c8a0"}

2026-08-03 02:22:09,771 INFO POD_STALE_MOUNT_RECOVERED
  {"executor_uuid": "b9a7d204-fd8c-4451-a4d0-f5c33773d46a",
   "container_name": "pod_6b8bacab-8335-4c30-a713-27003840c8a0",
   "local_volume": "volume_6b8bacab-8335-4c30-a713-27003840c8a0"}

2026-08-03 02:22:09,977 INFO Executor already rented
  check_id "executor.validate.rented_state", reason_code "RENTED",
  pipeline_id "e6051a97-3817-4d21-9141-eca717c188ea",
  actual_score 1.0, job_score 1.0,
  recovered_pods ["pod_6b8bacab-8335-4c30-a713-27003840c8a0"]
```

`POD_NOT_RUNNING` count over the whole run: **0**.

### Post-deploy — pod state

```
$ docker inspect pod_6b8bacab-8335-4c30-a713-27003840c8a0 --format '{{.State.Status}} | started={{.State.StartedAt}} | err={{.State.Error}}'
running | started=2026-08-03T02:21:56.494790436Z | err=

$ docker exec pod_6b8bacab-... sh -c 'mount | grep -E "gocryptfs|lium-cipher"; cat /root/DAH-2306-marker.txt'
/dev/loop3 on /lium-cipher type xfs (rw,relatime,idmapped,nouuid,...)
/lium-cipher on /root type fuse.gocryptfs (rw,nosuid,nodev,relatime,...allow_other,...)
DAH-2306 marker, written before the break

$ sudo ls /var/lib/docker/plugins/5050ec7eb818.../propagated-mount/
volume_5a9deeb1-818e-4261-a68f-98191e15c581
volume_6b0394a0-b8bd-4c12-892a-aeb32d358911
```

The stale dir for `volume_6b8bacab-...` is gone; the two unrelated leftovers from 2026-07-31
were not touched.

### Verdict

Expected behaviour observed end to end. The still-rented pod was healed inside the same
validation cycle that would otherwise have penalised the miner: the check reported
`RENTED` with score 1.0 instead of `POD_NOT_RUNNING` with 0.

Two things the marker proves that a plain `docker start` would not:

- the gocryptfs plaintext mount was restored (`/lium-cipher → /root`), so the customer's
  data is visible, not just the container running;
- `/root/DAH-2306-marker.txt` written before the break is readable after it.

This run also exercised the post-review hardening in `568c8995`: the pod carried **two**
named volumes (`volume_6b8bacab-...` plus an anonymous `local` one), and the volume to
repair was read off the container with `docker inspect` rather than derived from `pod_id`.

neurons/validators/src/services/docker_service.py
```diff
@@ -4907,6 +4868,203 @@ async def start_container(
...
+        # the volume whose stale mountpoint kept the container down, read off the container itself.
+        # `docker inspect` reports mounts of a stopped container, so the real name is taken from the
+        # host instead of guessed from pod_id — a pod created through the edit path carries a
+        # backend-supplied volume name that no convention derives. Every mount can be offered
+        # blindly: repair_stale_vloopback_mountpoint accepts only a vloopback volume whose stale
+        # mountpoint dir is present and empty.
+        inspect_result = await ssh_client.run(
+            f"/usr/bin/docker inspect {shlex.quote(container_name)} "
+            '--format \'{{range .Mounts}}{{if eq .Type "volume"}}{{println .Name}}{{end}}{{end}}\'',
+            timeout=_VLOOPBACK_REPAIR_COMMAND_TIMEOUT_SEC,
```
taiberium, line 5075: **Critical:** `asyncssh.TimeoutError` subclasses `Error` and `OSError`, so a repair timeout reaches the check as EXECUTOR_TRANSPORT_UNREACHABLE and the pod dodges POD_NOT_RUNNING. Catch `TimeoutError` here.

neurons/validators/src/services/task/checks/rented_machine.py
```diff
@@ -189,6 +181,30 @@ async def run(self, ctx: Context) -> CheckResult:
                         },
                     )
 
+                try:
+                    if await _recover_pod_after_stale_vloopback_mount(
```
taiberium: **Critical:** `TenantEnforcementCheck` also runs in `build_dry_run_checks`, so dry-run mode now rmdirs host paths and starts customer containers. Gate recovery on `settings.DRY_RUN`.

neurons/validators/src/services/docker_service.py
```diff
@@ -4907,6 +4868,203 @@ async def start_container(
...
+                miner_hotkey,
+                log_extra,
+            )
+            require_rental_docker_ssh_host_key(executor_info)
+            await self.start_existing_container(
+                executor_info=executor_info,
+                private_key=private_key,
+                known_hosts_policy=known_hosts_policy,
+                container_name=container_name,
+                local_volume_path=_RECOVERED_POD_LOCAL_VOLUME_PATH,
```
taiberium: **Critical:** An encrypted pod with a custom volume path comes back with that path as a plain container dir, so writes after recovery land unencrypted on the miner's disk.

neurons/validators/src/services/task/checks/rented_machine.py
```diff
@@ -189,6 +181,30 @@ async def run(self, ctx: Context) -> CheckResult:
...
+                    # losing the transport here means pod state is unknown, not "pod down".
+                    return _executor_transport_unreachable_result(
+                        ctx=ctx,
+                        check_id=self.check_id,
+                        container_name=pod_container_name,
+                        pod_id=pod_id,
+                        transport_error=exc,
+                        extra=extra,
+                    )
+                if pod_running:
```
taiberium: NIT: `run` is 191 lines now. The recovery block is a clean seam to pull out.

taiberium (changes requested):
Requested changes.

neurons/validators/src/services/docker_service.py
taiberium: Narrow reach: both official templates with `supports_volume_encryption` leave `volumes` unset, so this needs a user template that sets a custom path and encryption.

neurons/validators/src/services/docker_service.py
arhangel66: Fixed in e8b2621d, one level up. `asyncssh.TimeoutError` also subclasses `asyncssh.Error`, and `repair_stale_vloopback_mountpoint` runs four more timed commands of its own, so catching it only at this call would leave the same hole. The whole repair phase in `recover_pod_after_stale_vloopback_mount` is now wrapped in `except TimeoutError` -> `POD_STALE_MOUNT_RECOVERY_TIMED_OUT` -> `return False`, so the check keeps its POD_NOT_RUNNING verdict. Test: `test_recover_pod_declines_when_a_repair_command_times_out`.

neurons/validators/src/services/task/checks/rented_machine.py
arhangel66: Fixed in e8b2621d. Gated with a constructor flag rather than `settings.DRY_RUN`, to match the two neighbours in the same list (`GpuUsageCheck(dry_run=True)`, `GpuPowerLimitCheck(restore_stale_caps=False)`): `TenantEnforcementCheck(recover_stale_pods=False)`. Tests: `test_tenant_enforcement_skips_recovery_in_dry_run` and `test_dry_run_pipeline_does_not_recover_pods`.

neurons/validators/src/services/docker_service.py
arhangel66: You are right, and it is worse than the known limitation I had written up: the custom path becomes an ordinary container dir, so writes after recovery land unencrypted on the miner's disk.

The path is not recoverable from the validator — docker mounts the ciphertext at `/lium-cipher`, the plaintext path lives only in the create request, and `PodRentalActiveResponse` carries just `active` + `rental_closed_at`. So in e8b2621d recovery inspects the container's mounts first and stands down when it sees `/lium-cipher`, failing closed if the inspect does not answer. Encrypted pods keep exactly the POD_NOT_RUNNING behaviour they have today.

Re-enabling them needs `local_volume_path` on the backend's rental-active response — a separate compute-app change, I will file it. Note this also invalidates the staging run in the comment above (it used an encrypted volume); re-verification on an unencrypted pod is now in the test plan.

neurons/validators/src/services/task/checks/rented_machine.py
arhangel66: Done in e8b2621d — `run` is 163 lines now, the block moved to `TenantEnforcementCheck._recover_downed_pod`.

taiberium (approved):

PR #1182, opened 2026-07-31
DAH-2535: add xet challenge in verifyx

jam6099:
- Bump validator/executor `libverifyx.so` with the Xet download challenge.
- Pass `enable_xet_challenge` via `VERIFYX_ENABLE_XET_CHALLENGE` (default on).
- When `xet_execution` is present, check pkg/size/hash and network min download speed, then log the outcome.
- Xet pass/fail does not change overall VerifyX success; failures are warnings only.


neurons/validators/src/services/verifyx_validation_service.py
```diff
@@ -415,24 +471,41 @@ def _perform_verification_checks(payload: dict) -> Dict[str, Any]:
...
-        "success": success,
+    result = {
+        "success": all(required_checks),
         "network": network_stats,
         "hard_disk": storage_stats,
         "ram": memory_stats,
         "errors": all_errors,
     }
+
+    if "xet_execution" in response_data:
```
taiberium, line 492: **Critical:** An exception here fails the whole VerifyX check, not just the xet warning. Wrap this block in try/except.

neurons/validators/src/services/verifyx_validation_service.py
```diff
@@ -406,6 +407,61 @@ def _verify_storage_test(challenge_data: dict, response_data: dict) -> Tuple[dic
     return stats, errors
 
 
+def _verify_xet_test(challenge_data: dict, response_data: dict) -> Tuple[dict, List[str]]:
```
taiberium, line 410: NIT: no test covers the six branches here. `neurons/validators/tests/` already has three verifyx test files.

taiberium (changes requested):
Requested changes.

neurons/validators/src/services/verifyx_validation_service.py
jam6099: What kind of exception? The check has no impact on the verifyx check result, see the line exactly above constructing the response and ignoring the xet execution results.

neurons/validators/src/services/verifyx_validation_service.py
```diff
@@ -406,6 +407,61 @@ def _verify_storage_test(challenge_data: dict, response_data: dict) -> Tuple[dic
...
+
+    status = xet_execution.get("status", "failed")
+    success = bool(xet_execution.get("success"))
+    speed_mbps = xet_execution.get("speed_mbps", 0.0)
+    errors: List[str] = []
+
+    if status == "failed":
+        errors.append(f"Xet execution failed: {xet_execution.get('error', 'Unknown error')}")
+        success = False
+    elif speed_mbps < settings.verifyx.NETWORK_MIN_DOWNLOAD_SPEED_MBPS:
```
jam6099, line 434: The same value is used in the existing verifyxx network test above. The whole point of the PR is that xet doesnt fail the verifyx test.

taiberium (approved):

arhangel66 (approved):

PR #1184, opened 2026-07-31
DAH-2521, add robust executor backup and restore runner

pixel29913:
## Background

The current backup path streams a live tar archive through Docker and the AWS CLI. For large or actively changing directories, that can produce misleading progress, excessive local disk/log growth, long-running operations, and failures that leave no independently restorable snapshot.

This PR introduces a production-shaped executor proof of concept for an incremental backup and exact-snapshot restore path. It intentionally leaves the current backup engine available while the new path is rolled out and monitored.

## What changed

- Install a checksum-verified, pinned backup binary plus explicit tar and namespace tooling in the executor image.
- Add a typed, operation-scoped runner for backup and exact-snapshot restore.
- Store each pod under an isolated repository prefix in the supplied per-user S3 bucket.
- Support configurable S3 concurrency, defaulting to the staging-tested value of 64 and bounded from 1 through 128.
- Stream normalized byte/file progress, restore byte checkpoints, heartbeats, bounded diagnostics, and terminal results.
- Treat a completed partial snapshot separately from a full success.
- Support long-lived and temporary AWS credentials without placing secret values in Docker arguments or copying them into the rental.
- Back up plain named volumes through a read-only sibling helper with Docker logging disabled, no local repository cache, and bounded temporary storage.
- Restore directly from the selected snapshot into a new or empty destination without writing a temporary archive to customer disk.
- Add a fail-closed encrypted-volume adapter that verifies the rental identity, PID/cgroup, ciphertext volume, and live gocryptfs mount, then enters only the rental user namespace from an operation-scoped helper.
- Add operation examples, executor documentation, and focused tests.

## User and developer impact

The new path avoids local archive staging and Docker-stdout data transfer, provides continuous liveness and useful progress, supports exact backup selection during restore, and allows later backups to upload only changed data. Existing backups and the legacy path are not migrated or removed by this PR.

## Validation

- Focused storage-runner tests: 18 passed.
- Complete executor suite with the updated runner: 105 passed.
- Python compilation and git diff checks pass.
- Real encrypted Sysbox/gocryptfs staging rental:
  - Backup completed successfully through the user-namespace helper.
  - Exact restore completed in 16.3 seconds for a 64 MiB incompressible model and emitted heartbeat/progress events.
  - Restored checksum, snapshot version, exclusion behavior, and symlink all matched.
  - A populated restore target was rejected before transfer.
- Isolated plain named-volume staging round trip:
  - Backup and exact restore completed successfully.
  - Source and restored 8 MiB model checksums matched; marker and symlink were preserved.
- Performance comparison on the same encrypted snapshot:
  - Native restore with 64 S3 connections: 118.7 seconds.
  - Streaming dump/extract restore with 64 S3 connections: 16–20 seconds.

The corrected runner has been exercised from a temporary path inside the existing staging executor. A new executor image still needs to be built and published from this commit before deployment.

### DIFF
--- a/neurons/executor/Dockerfile
+++ b/neurons/executor/Dockerfile
@@ -23,6 +23,8 @@ RUN apt-get update \
      pkg-config \
      speedtest-cli \
      ssh \
+     tar \
+     util-linux \
   && install -m 0755 -d /etc/apt/keyrings \
   && curl -fsSL https://download.docker.com/linux/debian/gpg | gpg --dearmor -o /etc/apt/keyrings/docker.gpg \
   && chmod a+r /etc/apt/keyrings/docker.gpg \
--- a/neurons/executor/Dockerfile
+++ b/neurons/executor/Dockerfile
@@ -9,6 +9,7 @@ WORKDIR /root/app
 RUN apt-get update \
   && apt-get install -y --no-install-recommends \
      build-essential \
+     bzip2 \
      ca-certificates \
      curl \
      e2fsprogs \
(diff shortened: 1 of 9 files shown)


pixel29913:
Superseded by the complete product implementation in #1188. Closing this PoC to avoid confusion.

PR #1188, opened 2026-08-03
DAH-2521: add resilient backup and restore engine

pixel29913:
## Summary

- Install checksum-verified backup and encrypted-volume tooling in the executor image.
- Run backup and restore as detached, operation-scoped executor jobs with mode-0600 specs, bounded logs, progress, heartbeats, cancellation, reporting leases, and durable terminal reporting.
- Stream new backups directly to object storage without staging a local archive.
- Support full and partial committed results, exact-snapshot restore, and legacy archive restore without local archive staging.
- Support plain volumes, running encrypted volumes, and new-pod encrypted restore before the customer workload starts.
- Have the validator invoke the installed executor runner while probing capabilities and preserving the existing legacy fallback during rollout.
- Clean operation artifacts after completion and retain cancellation/reporting fallbacks when a validator, executor, or backend connection is temporarily unavailable.

## Behavior

- Running-pod restore requires a new or empty destination.
- New-pod restore completes before workload startup.
- Empty source or destination paths resolve to the entire mounted volume.
- Backup tooling and object-storage credentials stay in Lium-controlled helpers, independent of the customer image.
- File contents, modes, ownership, timestamps, symlinks, and supported xattrs are preserved.

## Deployment

This requires the matching backend branch. Build and publish the executor image from this branch, deploy the matching validator, and leave the backend feature switch disabled until both are present.

The matching executor, validator, and backend have already been exercised together on staging. Production should still use the disabled-by-default backend rollout and begin with a small monitored cohort.

## Testing and staging validation

- Executor Tests, Validator Tests, and CodeQL pass in GitHub Actions.
- Plain and encrypted backup, online restore, and restore during new-pod creation pass with exact content and metadata verification.
- Active cancellation, completion races, reporting loss, runner cleanup, missing repository/snapshot behavior, old-executor fallback, and scheduled overlap were exercised on staging.
- Performance testing covered single-large-file, 100k-small-file, mixed, encrypted, and 100 GiB workloads and confirmed the selected restore path and S3 connection default.
- Final encrypted 500.85 GB real model/dataset workload completed in 18m36s for backup and 7m53s for restore. All 145 file SHA-256 checksums and metadata checks matched, API progress remained live, and temporary storage stayed bounded without a second local 500 GB copy.

Detailed staging report: https://github.com/Datura-ai/lium-io-backend/pull/852#issuecomment-5221475906


arhangel66 (approved):
Looks good. I traced the storage operation path end to end: compute-app mints a fresh `backup_log_id`/`restore_log_id` per request, `has_active_log_for_config` blocks a second backup while one is active, and the websocket retry only re-sends when `send_text` itself raised — so the `operation_id`-derived artifacts in `storage_operations.py` can't collide today.

PR #1190, opened 2026-08-04
DAH-2545, recover encrypted-volume pods after host reboot and report it

arhangel66:
DAH-2545, follow-up to DAH-2306.

A pod whose rental volume is gocryptfs-encrypted was skipped by stale-mount recovery, because
recovery could not know the plaintext path the customer mounts and remounting at a guessed one
would write their data out in the clear. The path is already stored on the backend as
`pod.local_volume_path`; Datura-ai/lium-io-backend#854 returns it in the rental-active response and
this PR consumes it. The passphrase is not stored anywhere and is not needed — it is derived from
`VOLUME_MASTER_SECRET` and `pod_id`.

**Deploy Datura-ai/lium-io-backend#854 first.** Against an older backend the field is simply absent
and encrypted pods keep being skipped, exactly as today.

## What is in here

- `local_volume_path` on `PodRentalActiveResponse`, passed from the rented-machine check into
  `start_existing_container` instead of the hardcoded `None`.
- `allow_init=False` on the recovery path. A pod being revived already had its volume initialised,
  so a missing `gocryptfs.conf` means the ciphertext is gone. Re-initialising there would hand the
  customer an empty volume, log the recovery as a success and spare the miner the penalty for the
  data they destroyed. It now fails instead.
- Three-state volume detection. An inspect that does not answer used to count as "encrypted", which
  was safe only while that always meant "stand down"; with a path in hand it would have meant
  starting a container whose volume we could not read. `UNKNOWN` now declines unconditionally.
- `VOLUME_MASTER_SECRET` and the plaintext path are both checked *before* `docker start`. The
  remount only happens after the start, so a missing one would leave the customer's pod flapping up
  and down once a cycle.
- Plaintext path validation: absolute, no `.` or `..` segment, not `/lium-cipher` and not inside it.
  The path is a customer-authored template field and the backend accepts anything with a leading
  slash.
- A cancellation between `docker start` and the remount now stops the container instead of leaving
  it up with an unmounted plaintext path. The whole check runs under an outer timeout, so this is
  an ordinary path, not an exotic one.
- Reports a successful recovery to the backend (`POST /internal/pods/{pod_id}/host-reboot-recovered`).
  This is **telemetry**, not a customer notification: it creates an internal event keyed on pod and
  downtime. Wiring it to an email is deliberately left out. Sent once, right after the recovery, and
  best-effort — a backend timeout drops it, and no later cycle retries, because by then the pod is
  running and recovery does not fire again. Acceptable for telemetry; it would need an outbox before
  anything customer-facing depends on it.

## Encrypted recovery ships switched off

`RECOVER_ENCRYPTED_VOLUMES` defaults to `false`, so this PR changes nothing in production until it
is deliberately turned on. The reason is a window this PR does not close:

A restarted container runs its own entrypoint and `startup_commands` the moment it starts, and
gocryptfs can only be mounted into it afterwards, over `docker exec`. Writes to the plaintext path
in that window land in the container's upper layer — the miner's disk — and gocryptfs then mounts
with `-nonempty` and *shadows* them: invisible to the customer, readable by the miner, permanently.

An earlier revision of this PR tried to close that window by `chmod 000`-ing the plaintext directory
in the container's upper layer before the start. That was wrong twice over and has been removed:

1. It does not work. The container's processes run as root and Docker grants `CAP_DAC_OVERRIDE` by
   default, which bypasses file permission checks. The write succeeds.
2. It made things worse. The path it built came from a customer-authored template, and the sanitizer
   only requires a leading slash, so a volume of `/a/../../../../../../etc` would have turned into
   `mkdir -p /etc && chmod 000 /etc` on the miner's host.

**The same window exists on the pod-creation path in production today** — `docker run` starts the
container with the customer's startup commands and the gocryptfs mount follows. It is a property of
the encryption feature, not of recovery. Closing it needs a real pre-start barrier for both paths
and is tracked separately; until then this switch stays off and encrypted pods keep the pre-DAH-2545
behaviour: skipped, with the miner keeping the `POD_NOT_RUNNING` verdict.

## Tests

`3 failed, 1418 passed, 9 skipped`. The three failures are in `test_sshd_bootstrap_script.py` and
reproduce on a clean `main`.

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -455,6 +455,13 @@ def _xor_wrap_passphrase(passphrase: str) -> tuple[str, str]:
     return pad.hex(), wrapped.hex()
 
 
+def _is_under_cipher_mount(path: str) -> bool:
+    # whether a plaintext path collides with the ciphertext mount. A descendant is as unusable as
+    # the mount itself: mounting the plaintext view inside the ciphertext it is decrypting is not
+    # something recovery should attempt on customer data.
+    return path == _LIUM_CIPHER_MOUNT or path.startswith(f"{_LIUM_CIPHER_MOUNT}/")
+
+
 def _gocryptfs_missing_config_branch(allow_init: bool) -> str:
     if allow_init:
         return f'  gocryptfs -init {_LIUM_CIPHER_MOUNT} -passfile "$_pf"'
@@ -5186,7 +5193,7 @@ class DockerService:
             settings.RECOVER_ENCRYPTED_VOLUMES
             and bool(local_volume_path)
             and bool(_PLAINTEXT_PATH_RE.fullmatch(local_volume_path or ""))
-            and local_volume_path != _LIUM_CIPHER_MOUNT
+            and not _is_under_cipher_mount(local_volume_path or "")
             and bool(settings.VOLUME_MASTER_SECRET)
         )
 
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -1,5 +1,6 @@
 import asyncio
 import contextlib
+import enum
 import ipaddress
 import logging
 import math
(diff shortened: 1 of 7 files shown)


neurons/validators/src/services/docker_service.py
```diff
@@ -5007,6 +5043,7 @@ async def start_existing_container(
                         pod_id=pod_id,
                         log_tag=f"start_container_{pod_id}",
                         log_extra=default_extra,
+                        allow_init=False,
```
taiberium, line 5061: **NIT:** `allow_init=False` sits on `start_existing_container`, so it also covers the backend start path at line 4925, which is not behind `RECOVER_ENCRYPTED_VOLUMES`.

taiberium (approved):

PR #1195, opened 2026-08-05
DAH-2588: poll for DinD SSH readiness instead of a fixed 5s wait

arhangel66:
[DAH-2588](https://app.notion.com/p/Stop-single-cycle-DinD-probe-failures-from-delisting-healthy-executors-3b3b8bfdbde981ebb922c29ac55e8866)

## What was broken

`DindVerifier.verify` waited a fixed 5s after `docker run`, then made a **single**
`asyncssh.connect` attempt. Across ~1000 prod probes the DinD container needs ~3s to accept a
login (p99 ~7s, max ~10s), so that wait sat on the median: when it lost, the probe returned
`success=False`, `ConnectivityOrchestrator` turned that into `sysbox_runtime = False`, and the
fatal `SysboxRequiredCheck` zeroed the executor's score, cleared its accumulated verification
and delisted it for the cycle.

Reference incident — executor `5f2cc740-3946-4663-9cc6-68a87d3ffa13`, wave `2026-08-05 05:23:48`,
unrented, 300/300 ports verified, scrape reported sysbox present:

```
05:30:04.925  DinD container created
05:30:11.251  DinD check failed  error "[Errno 111] Connect call failed ('192.222.55.74', 20000)"
05:31:50      SYSBOX_REQUIRED_MISSING  score 0, verification cleared
05:38:48      next wave: sysbox_runtime = TRUE, score 1.0, back on the market
```

Host uptime rose straight through the failing wave — the machine never went anywhere. The
provider saw a "machine went offline" notification and came to Discord.

## How it's fixed

`DindVerifier._connect_retrying_until_sshd_answers` polls `asyncssh.connect` every 1.5s under a
single 30s deadline. On success it logs `ssh_ready_sec` and `ssh_attempts`; on giving up it logs
`DinD SSH not ready before deadline` with the same fields, so the two cases stop looking
identical in Loki.

The deadline bounds the **whole** wait, not just the moment an attempt starts. Two guards do
that together: no attempt starts once the deadline is within one poll interval, and an attempt
that would outlive the deadline gets only the budget that is left
(`min(DIND_SSH_CONNECT_TIMEOUT_SECONDS, deadline - now)`). Without the second guard a blackholed
port stretches the probe to ~39s rather than 30s — an attempt greenlit at 27s still claimed a
fresh 12s — which is the DAH-2272 failure mode (a hung connect stalling the pipeline) coming
back stacked, and nothing upstream bounds it: `ConnectivityOrchestrator` awaits the probe with
no `asyncio.wait_for` around it.

The per-attempt `connect_timeout` / `login_timeout` otherwise stay at **12s** and are
deliberately not tightened toward the poll interval. asyncssh counts TCP connect, key exchange
*and* authentication against that budget and none of it carries over to the next attempt, so
tightening it would fail every attempt on a host that simply needs 6s to handshake — a new class
of false negative in place of the old one. The clamp above only shortens an attempt when less
than 12s of the deadline remains, which is time that attempt could not have used anyway.

This is the first of the three parts in the spec; the random probe port and the cross-wave
debounce are not in this PR.

Net effect on the probe budget is negative, not positive: a container ready at 3s is now caught
at ~3s instead of always waiting 5s.

## Test plan

- [x] `pdm run pytest tests/` in `neurons/validators` — 1470 passed, 9 skipped. (Three failures in
      `tests/test_sshd_bootstrap_script.py` are pre-existing on `main` — verified by running that
      file against a clean `origin/main` checkout — and excluded.)
- [x] Mutation check: forcing the retry loop to give up on the first attempt fails exactly
      `test_dind_verifier_retries_connect_until_sshd_is_up` and `test_dind_verifier_gives_up_after_deadline`.
      Removing the remaining-budget clamp fails exactly
      `test_dind_verifier_hanging_attempts_stay_inside_the_deadline`, which runs the production
      30/12/1.5 ratio scaled down 100x and asserts the probe still finishes inside the deadline.
- [x] `ruff check` on both changed files — 2 findings (`UP041`), both pre-existing on `main`.

### Pre-deploy verification

Run the probe ~20x against a staging executor and read `ssh_ready_sec` from
`DinD SSH connected`: a p99 above 5s confirms the race the fixed wait was losing. Then block the
DinD port on a healthy executor and confirm the probe now spends the full deadline and logs
`DinD SSH not ready before deadline` before failing — and that `ssh_waited_sec` lands at ~30s,
not ~39s.

### Post-deploy verification

- +1 d: `DinD check failed` rate falls below today's ~2.6%; no executor shows a single isolated
  `Sysbox required for unrented executor` wave with `ssh_ready_sec` absent.
- +7 d: one-to-three-wave `Sysbox required for unrented executor` hits collapse toward zero,
  while the ~17 persistently sysbox-less machines stay banned.

### DIFF
--- a/neurons/validators/src/services/executor_connectivity/dind_probe.py
+++ b/neurons/validators/src/services/executor_connectivity/dind_probe.py
@@ -1,8 +1,9 @@
 import asyncio
 import logging
+from typing import Any
 
 import asyncssh
-from asyncssh import SSHClientConnection
+from asyncssh import SSHClientConnection, SSHKey
 
 from core.docker_utils import DockerCommand
 from core.utils import _m, get_extra_info
@@ -17,12 +18,13 @@ logger = logging.getLogger(__name__)
 # "no sysbox", which delists an unrented executor for the whole cycle. Polling is also cheaper
 # than a longer sleep: a ready container is caught at ~3s instead of always waiting.
 #
-# The per-attempt timeout stays at 12s and is deliberately not tightened to the poll interval:
-# asyncssh counts TCP connect, key exchange AND authentication against it, none of which carry
-# over to the next attempt, so a host slow enough to need 6s would fail every single try.
-DIND_SSH_READY_TIMEOUT = 30
-DIND_SSH_CONNECT_TIMEOUT = 12
-DIND_SSH_POLL_INTERVAL = 1.5
+# The per-attempt timeout is deliberately not tightened to the poll interval: asyncssh counts TCP
+# connect, key exchange AND authentication against it, none of which carry over to the next
+# attempt, so a host slow enough to need 6s would fail every single try. An attempt is shortened
+# only when less than 12s of the deadline is left, which is time it could not have used anyway.
+DIND_SSH_READY_TIMEOUT_SECONDS = 30
+DIND_SSH_CONNECT_TIMEOUT_SECONDS = 12
+DIND_SSH_POLL_INTERVAL_SECONDS = 1.5
 
 
 class DindVerifier:
@@ -68,7 +70,9 @@ class DindVerifier:
 
             # Test SSH
             pkey = asyncssh.import_private_key(private_key)
-            async with await self._wait_for_ssh(host, port, pkey, log_ctx) as ssh:
+            async with await self._connect_retrying_until_sshd_answers(
+                host, port, pkey, log_ctx
+            ) as ssh:
                 # Test sysbox
                 if sysbox:
                     # daturaai/dind:0.0.1 bundles the hello-world image into the inner dockerd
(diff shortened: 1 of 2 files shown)


arhangel66:
## DAH-2588 — staging verification

- **PR:** https://github.com/Datura-ai/lium-io/pull/1195
- **Deployed:** `409e28ee`, workflow run https://github.com/Datura-ai/lium-io-deployment/actions/runs/30987595704, finished 2026-08-05T08:12:00Z
- **Validator restarted:** 2026-08-05 08:11:45 UTC (staging-subnet)
- Staging board: `validator` occupied as `arhangel66` for the whole run, released after.

Both staging executors belong to miner `5DfbmGgkgYoYKJk7rA9UFgeJFacfkBMhaeFZgJk7c4mCv7DT`.

### Pre-deploy baseline

```
query.sh '{namespace="staging-subnet", container="validator"} |= "DinD"' 90 40
```

Interval from `DinD container created` to `DinD SSH connected`, old code (`dind_probe.py:68`,
no `ssh_ready_sec` field):

| wave | executor | created → connected |
|---|---|---|
| 06:50 | `66e9c110-d9db-4a1e-a2cd-6ea1647ae81b` | 06:52:51.615 → 06:52:57.443 = 5.83s |
| 06:50 | `b9a7d204-fd8c-4451-a4d0-f5c33773d46a` | 06:52:54.019 → 06:52:59.197 = 5.18s |
| 07:05 | `66e9c110-d9db-4a1e-a2cd-6ea1647ae81b` | 07:07:56.692 → 07:08:03.025 = 6.33s |
| 08:07 | `b9a7d204-fd8c-4451-a4d0-f5c33773d46a` | 08:07:48.910 → 08:07:54.085 = 5.18s |
| 08:07 | `66e9c110-d9db-4a1e-a2cd-6ea1647ae81b` | 08:08:02.967 → 08:08:08.795 = 5.83s |

Never below 5s, because the 5s sleep ran before the single connect attempt regardless of
whether the container was already up.

### Post-deploy

```
query.sh '{namespace="staging-subnet", container="validator"} |= "ssh_ready_sec"' 30 30
```

New code (`dind_probe.py:173`), two consecutive waves:

```
2026-08-05 08:14:03,768  b9a7d204-fd8c-4451-a4d0-f5c33773d46a  ssh_ready_sec=3.21  ssh_attempts=3
2026-08-05 08:14:30,529  66e9c110-d9db-4a1e-a2cd-6ea1647ae81b  ssh_ready_sec=4.13  ssh_attempts=3
2026-08-05 08:28:42,000  b9a7d204-fd8c-4451-a4d0-f5c33773d46a  ssh_ready_sec=3.12  ssh_attempts=2
2026-08-05 08:28:55,244  66e9c110-d9db-4a1e-a2cd-6ea1647ae81b  ssh_ready_sec=4.30  ssh_attempts=3
```

Per-executor and reproducible: `b9a7d204` lands at 3.12–3.21s, `66e9c110` at 4.13–4.30s.

Full probe sequence for `b9a7d204-fd8c-4451-a4d0-f5c33773d46a`:

```
08:13:59,882  DinD start              line=49
08:14:00,553  DinD container created  line=67
08:14:03,768  DinD SSH connected      line=173  ssh_ready_sec=3.21 ssh_attempts=3
08:14:05,311  DinD check ok           line=102  sysbox_result=True
```

Failure counters over the same window:

```
query.sh '... |~ "DinD check failed|DinD creation failed|not ready before deadline"' 20 20   -> 0
query.sh '... |= "SYSBOX_REQUIRED_MISSING"' 20 10                                            -> 0
```

### Verdict

Expected behaviour observed. The probe now connects on readiness rather than on a timer:
5.18s → 3.12–3.21s and 5.83s → 4.13–4.30s on the same two executors, `sysbox_result=True`
throughout, zero probe failures and zero `SYSBOX_REQUIRED_MISSING` across two waves.

`ssh_attempts` is the load-bearing number: with a 1.5s poll interval, attempts land at t=0,
t≈1.5 and t≈3.0, so an `ssh_attempts` of 2 or 3 means sshd inside the container did not answer
at 0s — every probe here needed at least one retry. Real readiness on this hardware is
~1.5–3s plus ~1–1.6s of handshake, so the old fixed 5s wait carried roughly 2s of margin, not
the 4s a naive reading of the baseline suggests. Consistent with prod losing that race on
loaded hosts.

Not covered by this change and still open: `DinD creation failed` (18/h in prod, `docker run`
itself failing on "driver failed programming external connectivity") and the sticky probe port.
Neither appeared on staging during this run — staging has only two executors, so this run is a
"does it work" check, not a rate measurement.

neurons/validators/src/services/executor_connectivity/dind_probe.py
```diff
@@ -119,6 +125,74 @@ async def verify(
...
+        unreachable host are different diagnoses).
+        """
+        loop = asyncio.get_running_loop()
+        started_at = loop.time()
+        deadline = started_at + DIND_SSH_READY_TIMEOUT_SECONDS
+        attempts = 0
+
+        while True:
+            attempts += 1
+            attempt_timeout_seconds = min(
```
taiberium, line 154: NIT: this goes negative when the loop resumes late after the sleep, so the last attempt reports a timeout instead of the real connection error.

taiberium (approved):

neurons/validators/src/services/executor_connectivity/dind_probe.py
arhangel66: Fixed in d471225d. The loop now stops at the deadline itself (`while loop.time() < deadline`), so a poll that resumes late ends the wait instead of starting one more attempt with a negative budget; the last real error is stored and re-raised after the `not ready before deadline` warning, so a refused connection still reads as refused. New test `test_dind_verifier_late_poll_wakeup_starts_no_further_attempt` oversleeps the poll past the deadline and asserts a single attempt with a positive budget — it fails if the guard is reverted to `while True`.

PR #1196, opened 2026-08-06
DAH-2594: count a passed TDX attestation as a flagship capability

arhangel66:
## Problem

DAH-2546 gates the unrented (idle) incentive for 8x H200/B200/B300: the machine must have NCU
profiling counters open on the host, or real GPU splitting. Enforced in production since
2026-08-04.

A confidential-computing node can satisfy neither. NCU counters are opened on the host, and a CVM
provider has no access to the host GPU drivers — there is nothing for them to configure. Splitting
is likewise unavailable. So an attested CVM sits at zero idle incentive with no action available to
its owner.

The stated policy is broader than the code: idle incentive should require **CC or profiling or GPU
splitting**. `_missing_flagship_capability` only ever looked at the last two.

Live case: an 8x H200 CVM with `tdx_attestation_passed=true`, `ncu_profiling_access="restricted"`
and no splitting, excluded from the rental pool since 2026-08-05 16:47:05 UTC.

## Change

A third clause in the existing gate condition:

```python
if (
    ncu_profiling_access == NCU_PROFILING_UNRESTRICTED
    or has_real_splitting
    or result.tdx_attestation_passed
):
    return None
```

`JobResult.tdx_attestation_passed` is set by the validator itself from
`host_policy.attestation_passed` — `attestation_digest is not None AND gpu_attestation_passed is not
False`. A machine that submits no `tdx_quote` is never verified, so it can never reach `True`; the
signal is not miner-forgeable. The clause sits after the existing `spec is None` guard, so the
spec-less synthetic results from `precompute_all_estimates` keep failing open exactly as before.

The miner-facing message and the internal log now name all three paths.

`_log_flagship_capability_limit` gains one field, `tdx_quote_present`. It separates a self-declared
CVM whose attestation did not pass from an ordinary host — a distinction `tdx_attestation_passed`
cannot make, since it is `False` on every line this gate emits. Internal Loki log only; the wire
payload and the TimescaleDB column are untouched.

## Deliberately not changed

- **`ZeroIncentiveReason.FLAGSHIP_WITHOUT_NCU_OR_SPLIT` keeps its string value.** The enum is
  append-only — Loki dashboards and the DAH-2340 backend payload key off exact strings. The name is
  now a slight misnomer; that is the cheaper cost.
- **No `settings.ENABLE_TDX_ATTESTATION` in the condition.** With attestation off the field is
  already `False`.
- **No enforcement-flag change.** `ENABLE_UNRENTED_FLAGSHIP_CAPABILITY_LIMIT` stays as deployed.
- **Not extended to unattested CVMs.** A machine that presents a quote but fails attestation is
  already zeroed upstream by the Minimal-G5 gate in `score_calculator.py` and never reaches this
  code; where it would reach it, attestation is disabled and the quote is an unverified string.

## Verification

- `pdm run pytest tests/test_unrented_flagship_capability_limit.py tests/test_miner_incentive_log.py
  tests/test_rental_price_incentive_flow.py tests/test_rental_price_estimate.py
  tests/test_rental_price_helpers.py tests/prod_snapshot/ -q` → 136 passed, 4 skipped.
- Full package: `pdm run pytest tests/ -q` → 1483 passed, 9 skipped, 3 failed. The 3 are
  `tests/test_sshd_bootstrap_script.py`, pre-existing and macOS-only (BSD `sed`) — they fail
  identically with this diff stashed.
- RED confirmed before the fix: reverting only the two `src/incentive/` files leaves 6 failures —
  the two new parametrize rows, the enforced-CC row, the message assert and the two
  `tdx_quote_present` asserts.

## Effect once deployed

The gated CVM returns to the rental pool on the first scoring cycle after the validator rollout.
Nothing loses incentive: the change can only move an executor from excluded to included.

Provider documentation is a separate follow-up — `emission.mdx` does not describe this gate at all
today, and `gpu-splitting.md` still claims splitting unlocks no incentive.

### DIFF
--- a/neurons/validators/src/incentive/miner_incentive_log.py
+++ b/neurons/validators/src/incentive/miner_incentive_log.py
@@ -9,7 +9,7 @@ WHERE A NODE EARNS (two "pools" of subnet emission; the rest is burned):
 HOW A NODE PICKS A POOL:
   rented?                                              -> mining pool (always earns)
   idle AND model in program AND price ok AND disk >= 1.5x VRAM
-       AND (not flagship-8x OR NCU/split offered)
+       AND (not flagship-8x OR NCU/split/attested-CVM offered)
                                AND capacity?            -> unrented pool (earns)
   otherwise                                            -> 0 incentive (reason below)
 
@@ -24,7 +24,8 @@ WHAT THIS CATALOG HOLDS — every `MinerLogLine` the miner-facing log block
      GPU model not in the unrented program (earns only when rented),
      price above the market soft limit (lower the price to earn),
      total disk below 1.5x total GPU VRAM (add disk to earn),
-     8x H200/B200/B300 with neither NCU profiling nor GPU splitting (offer either to earn),
+     8x H200/B200/B300 with no NCU profiling, GPU splitting or passed TDX
+       attestation (offer any of the three to earn),
      no unrented capacity for that GPU-count tier this cycle,
      NVIDIA driver below the minimum, sysbox runtime not enabled
 
--- a/neurons/validators/src/incentive/miner_incentive_log.py
+++ b/neurons/validators/src/incentive/miner_incentive_log.py
@@ -346,11 +346,12 @@ class MinerLogLine(BaseModel):
             reason=ZeroIncentiveReason.FLAGSHIP_WITHOUT_NCU_OR_SPLIT,
             message=(
                 f"No unrented incentive: an idle {result.gpu_count}x {result.gpu_model} must offer "
-                "NCU profiling or GPU splitting. Either open the host profiling counters "
-                "(NVreg_RestrictProfilingToAdminUsers=0, then reboot), which also makes the node "
-                "rentable only as a whole host, or set a minimum GPU count below the full node in "
-                "the Miner Portal on a host that supports docker storage limits, "
-                "or rent it out to earn."
+                "NCU profiling, GPU splitting or confidential computing. Either open the host "
+                "profiling counters (NVreg_RestrictProfilingToAdminUsers=0, then reboot), which "
+                "also makes the node rentable only as a whole host, or set a minimum GPU count "
+                "below the full node in the Miner Portal on a host that supports docker storage "
+                "limits, or run the executor inside a confidential VM whose TDX attestation "
+                "passes, or rent it out to earn."
             ),
             extra_fields={
                 "ncu_profiling_access": missing.ncu_profiling_access,
--- a/neurons/validators/src/incentive/rental_price.py
+++ b/neurons/validators/src/incentive/rental_price.py
@@ -44,8 +44,8 @@ SOFT_LIMIT_PRICE_RATE = 1.1
 MIN_DISK_TO_VRAM_RATE = 1.5
 
 # DAH-2546 — flagship capability gate. An unrented 8x machine of these base models must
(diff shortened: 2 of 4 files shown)


neurons/validators/src/incentive/rental_price.py
```diff
@@ -334,7 +335,13 @@ def _missing_flagship_capability(
             and result.gpu_splitting_min_count
             and result.gpu_splitting_min_count < result.gpu_count
         )
-        if ncu_profiling_access == NCU_PROFILING_UNRESTRICTED or has_real_splitting:
+        # DAH-2594 — a CVM provider has no access to the host GPU drivers, so NCU counters are
+        # unreachable by construction; a passed attestation is that machine's capability path
+        if (
+            ncu_profiling_access == NCU_PROFILING_UNRESTRICTED
+            or has_real_splitting
+            or result.tdx_attestation_passed
```
taiberium: NIT: `attestation_passed` treats `gpu_attestation_passed=None` as passed, so a CVM sending no `nvidia_payload` earns here with no GPU attestation. Failing evidence gets nothing, so withholding it pays better.

neurons/validators/src/incentive/miner_incentive_log.py
```diff
@@ -346,11 +347,12 @@ def no_payout_because_flagship_without_ncu_or_split(
...
-                "NCU profiling or GPU splitting. Either open the host profiling counters "
-                "(NVreg_RestrictProfilingToAdminUsers=0, then reboot), which also makes the node "
-                "rentable only as a whole host, or set a minimum GPU count below the full node in "
-                "the Miner Portal on a host that supports docker storage limits, "
-                "or rent it out to earn."
+                "NCU profiling, GPU splitting or confidential computing. Either open the host "
+                "profiling counters (NVreg_RestrictProfilingToAdminUsers=0, then reboot), which "
+                "also makes the node rentable only as a whole host, or set a minimum GPU count "
+                "below the full node in the Miner Portal on a host that supports docker storage "
+                "limits, or run the executor inside a confidential VM whose TDX attestation "
```
taiberium, line 354: NIT: `gpu-profiling.md` still says profiling or splitting are the only two ways to keep the idle payout, so the docs need updating before providers see this.

taiberium (approved):

neurons/validators/src/incentive/rental_price.py
arhangel66: Fixed in 
13f96c0c: the gate now reads `result.attestation_digest` (the verified TDX quote) instead of `tdx_attestation_passed`, so the GPU-CC verdict no longer moves the payout in either direction — submitting failing evidence pays exactly what withholding it pays. Bad GPU evidence stays the GPU-attestation enforcement flag's job; once that flag is on, missing evidence fails attestation itself and no digest reaches the gate. Added a parametrized case for `gpu_attestation_passed=False` with a digest present.

neurons/validators/src/incentive/miner_incentive_log.py
arhangel66: Docs updated in https://github.com/Datura-ai/lium-docs/pull/104 — the two-ways wording in `providers/nodes/gpu-profiling.md`, `providers/nodes/gpu-splitting.md` and `providers/rewards/emission.mdx`, plus a tip in `providers/nodes/cvm.md`. It ships with this release.

PR #1197, opened 2026-08-06
DAH-2571, hand the RDMA verbs devices to the rental container

taiberium:
Step 2 of the InfiniBand epic (DAH-2568). Detection shipped and found the hardware: 16 hosts with live InfiniBand ports, 37 with live RoCE, and three H200s on one real 400 Gb/s NDR fabric. None of it is sellable, because a renter cannot reach the card — `/dev/infiniband/*` is not forwarded and RDMA cannot register memory under the stock 64 KB memlock.

## What it does

- `_query_shared_nodes` now also lists `/dev/infiniband/uverbs*` and `/dev/infiniband/rdma_cm`, so they land in the same `--device` block the NVIDIA nodes already use.
- A container that actually received one of those devices gets `memlock` unlimited. `ContainerRunSpec` grows a typed `ulimits` field; nothing else sets it.

## Three deliberate restrictions

**Only the verbs nodes, never the `/dev/infiniband` directory.** The directory form also hands over `issm*` — the subnet-manager interface — and `umad*`, raw MAD access. A renter holding `issm` can interfere with the fabric every other tenant on it depends on.

**Nothing on a partial-host rental.** The cards belong to the host, so the RDMA listing sits inside the same whole-host branch that already withholds `/dev/nvidia-caps/*` and the IMEX channels. On a split node, forwarding all 24 verbs devices would hand one tenant the cards another tenant is renting. Per-card selection is the follow-up and has to split along the same seam as the GPUs (DAH-2465).

**memlock is raised only where it buys something.** A pod whose `ram_total` is 0 carries no memory cgroup limit either (`mem_limit` is skipped for a falsy value), and unlimited memlock on top of that would let a tenant pin the host's RAM. Scoping it to containers that got an `/dev/infiniband/` device keeps the blast radius at the machines this feature is for.

The parameter that gates the whole-host set was named `include_caps`; it now says what it means, `is_whole_host_rental`, since it governs three different kinds of host-wide node.

## Proven on production hardware first

Read-only test on `204.9.206.243` (8xB300), 2026-08-06: under `--runtime=sysbox-runc` with the devices and `--ulimit memlock=-1:-1`, a throwaway container enumerated all 24 RDMA devices and the four ACTIVE InfiniBand ports, with `sm_lid` and LIDs matching `executor.specs.infiniband_ports` exactly, `memlock` unlimited inside, exit 0. sysbox hides nothing and blocks no ioctl — feasibility was checked before any code was written. `/dev/infiniband` on that host holds 24 `uverbs*`, 24 `umad*` and 4 `issm*` nodes, which is what motivated the narrow selection.

## What this does NOT protect against

**There is no per-tenant partitioning.** Two whole-host renters on two hosts of the same fabric share the default partition. Passive sniffing of another tenant's queue pairs is not reachable this way, but saturating the shared fabric is, and `rdma_cm` lets a renter resolve LIDs/GIDs and enumerate who else is on it. Closing that needs a `pkey` per rental (IB) or VLAN separation (RoCE), which is the next piece of work — worth knowing before the first pair is sold.

**Providers hand over RDMA at the plain per-GPU price.** Nothing in the incentive or pricing path reads the interconnect fields the scrape now collects, so an IB-equipped host earns exactly what it earned before.

## Also not in this PR

Per-card selection for split nodes, an address on the fabric (IPoIB or SR-IOV), node-to-node RDMA verification, and group rental.

## Testing

`tests/test_infiniband_passthrough.py` — the verbs nodes are forwarded and the directory form is not, the RDMA listing sits behind the whole-host guard, memlock is raised only for a container that got an RDMA device, it reaches the host config, a spec without ulimits still sends none, and a UUID repeated N times no longer passes a single-GPU rental off as whole-host. `tests/test_nvidia_devices.py` updated for the renamed parameter and now asserts the RDMA nodes in both directions.

Full validator suite: 1490 passed. The 3 failures in `test_sshd_bootstrap_script.py` reproduce on `main` untouched.

### DIFF
--- a/neurons/validators/src/core/config.py
+++ b/neurons/validators/src/core/config.py
@@ -230,6 +230,12 @@ class Settings(BaseSettings):
     # rental incentive while staying active. When False, the breach is only logged
     # (shadow mode) so prod impact can be observed before enforcing.
     ENABLE_UNRENTED_SOFT_PRICE_LIMIT: bool = Field(env="ENABLE_UNRENTED_SOFT_PRICE_LIMIT", default=False)
+    # DAH-2571. Off until a RoCEv2 queue pair from inside a container is shown to reach a neighbour
+    # on the same segment, or not to. RDMA bypasses the host network stack entirely — no iptables,
+    # no conntrack, no rate limit on a link the provider also uses for storage — so on the 21 prod
+    # hosts sharing an IPv4-mapped segment this would be a new, unfiltered path to a neighbour.
+    # Turning it on later costs nothing; taking it back after machines are sold does.
+    ENABLE_RDMA_DEVICE_PASSTHROUGH: bool = Field(env="ENABLE_RDMA_DEVICE_PASSTHROUGH", default=False)
 
     # DAH-2520 — unrented VRAM/disk sanity gate. When True, an unrented executor whose
     # total GPU VRAM exceeds the machine's total disk loses the unrented rental incentive
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -922,7 +922,7 @@ class DockerService:
             runtime="sysbox-runc" if payload.is_sysbox else None,
             cap_add=("NET_ADMIN",),
             sysctls={"net.ipv4.conf.all.src_valid_mark": "1"},
-            ulimits=self._memlock_ulimit_for(devices),
+            ulimits=self._memlock_ulimit_for(devices, payload.memory_gb),
             devices=devices,
             device_requests=gpu_devices.device_requests,
             cpu_count=cpu_count,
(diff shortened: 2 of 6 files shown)


arhangel66 (approved):
Nice narrow selection, and the UUID dedup is a good catch.

One thing I'd like settled before this ships. The PR frames the residual risk as a shared default partition, but on prod that part is small: of 18 hosts with active IB ports, 14 run their own SM with identical LID sets — each is its own subnet, so a whole-host renter there gets a fabric that is their own host. Only 4 sit behind an external SM.

The larger surface is RoCE — 42 hosts with an ACTIVE Ethernet-linked port, 21 with an IPv4-mapped GID in a shared segment (5 on 183.232.153.0/24, 4 on the public 38.255.28.0/24, 3 on 10.125.1.0/24). A renter can already reach those neighbours through the bridge NAT, so this isn't new reachability. What it adds is a path that skips the host network stack entirely: no ufw/iptables filtering, no conntrack, no visibility, no rate limit on a 100–400 Gb/s link the provider also uses for storage, plus direct access to the neighbours' RDMA stack.

The 2026-08-06 host test proved enumeration from inside a container but never sent RoCE traffic, so we don't actually know whether that path works. Could you check it on one host — a RoCEv2 QP from inside the container to a neighbour on the same /24? If the packets don't leave, this is moot. If they do, I'd put the RDMA listing behind a disabled-by-default flag, with a test that the default path forwards nothing: turning it on later is free, taking it back after the machines are sold is not.

Two smaller things:

_memlock_ulimit_for docstring says a tenant "can only pin their own allocation", but pod.ram_total is host RAM minus 2 GiB (2111321744 vs 2113418896 KB on a prod row), so on a whole-host rental that allocation is the machine — and mlocked pages don't reclaim. Worth rewording.
test_only_the_verbs_nodes_are_forwarded_never_the_whole_directory and test_rdma_devices_are_skipped_on_a_partial_host_rental assert on inspect.getsource text and substring offsets — they'd pass on a broken function with the right source text, and the behaviour is already covered in test_nvidia_devices.py.

taiberium:
All three taken, thanks — the RoCE framing is the part I had backwards.

**Flag added, default off.** `ENABLE_RDMA_DEVICE_PASSTHROUGH` now gates the listing on top of the whole-host branch, with your reasoning in the comment at the setting. A parametrised test asserts the probe command carries the RDMA globs only when it is on, so the default path forwards nothing.

**Docstring reworded.** You are right that `ram_total` is host RAM less ~2 GiB, so on a whole-host rental "their own allocation" is the machine. It now says what the cgroup actually buys — a ceiling the kernel enforces and the OOM killer can act on — rather than implying the tenant is confined to a share.

**Source-text tests replaced.** Both now drive `_query_shared_nodes` and assert on the command it builds: one that the directory form and `issm*`/`umad*` never appear, one that the flag decides. No more `inspect.getsource` or substring offsets.

On the measurement: I cannot rent or reach a prod host, so the RoCEv2 QP test is yours if you want it done soon. Minimal form on one of the `38.255.28.0/24` machines — inside a container with the devices, `ib_write_bw` bound to the RoCE device against a neighbour on the same segment, and `ethtool -S` on the host NIC before and after to see whether the packets actually left. If they do not, the flag can default on and I will send the flip; if they do, the flag stays off until we have per-tenant VLAN or pkey separation.

Either way nothing ships enabled, which was your point.

PR #1199, opened 2026-08-06
DAH-2596: add NVIDIA A800 80GB PCIe to the supported GPU catalog

taiberium:
A800 80GB PCIe is rentable on Vast (6 live offers) but was missing from every table on our side. It is the export-compliant A100: same GA100 silicon, same 80 GB HBM2e (offers report 81920 MB), only NVLink is cut to 400 GB/s. We already carry the H800 equivalents, so this was an oversight rather than a decision.

- `GPU_MODEL_RATES`: 0.0 — rentable, but no idle-emission weight.
- `GPU_VRAM_SIZES_MB`: 80 GB, same window as the A100.
- `BASE_GPU_MAP` / `MAX_UNRENTED_GPUS_BY_TYPE`: base model `A800`, no subsidy buckets — same treatment as H800.

Backend side: Datura-ai/lium-io-backend#870. Task: DAH-2596.

Tests: `pytest tests/test_gpu_precheck.py tests/test_gpu_vram_precheck_check.py tests/test_rental_price_incentive_flow.py` — 123 passed, 4 skipped.


taiberium:
Superseded by #1201 — all cards folded into one catalog pass.

PR #1202, opened 2026-08-06
DAH-2602: attach the rental container to the RDMA fabric with an SR-IOV function

taiberium:
Notion: https://app.notion.com/p/Give-the-rental-container-an-address-on-the-RDMA-fabric-3b4b8bfdbde981799dffc9de5fa0696d

## The problem

DAH-2571 forwards the RDMA verbs devices into the rental container, and that half works: `ibv_devinfo`
enumerates the cards, ports read `PORT_ACTIVE`, `memlock` is unlimited. RDMA still cannot cross a host.

Measured on prod on 2026-08-06 with two rented nodes on the same RoCE segment:

```
container:  inet 172.20.0.2/16  eth0        (the NAT bridge)
card GID:   0000:0000:0000:0000:0000:ffff:26ff:1c13  = 38.255.28.19  (the HOST)
result:     queue-pair handshake completes over the mapped TCP port, zero RDMA bytes move
```

RoCE sends with the source address carried in the card's GID. That address lives on the host's netdev,
not in the container's namespace, so the packets have nowhere to leave from. A colleague hit the same
wall one layer up — ranks advertise `172.20.0.3` and an ephemeral port during the collective's TCP
bootstrap. Same missing thing: the container has no routable identity of its own.

## The approach: give it one the tenant cannot forge

The tenant is root in its container and holds `NET_ADMIN` (it is there for the in-pod VPN path). So
anything enforced by the container's own kernel namespace is worth nothing. Two mechanisms put
enforcement out of its reach.

**SR-IOV virtual function.** One VF per rental, its identity programmed from the physical function:

```
ip link set <pf> vf <N> mac <derived> vlan <tenant> spoofchk on trust off
```

`spoofchk on` makes the NIC drop frames whose source does not match; `trust off` denies promiscuous
mode and MAC changes. Root inside the container can rename and reconfigure the interface all it likes —
frames that do not match die before the wire. This is the difference from macvlan, which trusts the
namespace owner with the interface's identity.

**Exclusive RDMA namespace mode.** The VF's RDMA device is moved into the container's namespace, where
it is the only device visible and its GID table holds only addresses that exist there. The
source-address problem dissolves instead of being patched, and a stolen verbs handle shows nothing
outside the namespace. Today's passthrough runs in the default *shared* mode, where any namespace
holding a verbs node sees every device on the box and every GID the host owns.

## Preconditions are read, never created

An unprepared host is the normal case, not an error — it keeps exactly today's behaviour, and a failure
anywhere in the attach is logged and swallowed rather than failing the rental.

- `rdma system set netns exclusive` — host-wide and disruptive to flip while RDMA is in use, so it
  belongs to provisioning.
- `/etc/lium/rdma-fabric.json` — `{"container_ip_range": "...", "vlan_id": N}`. Per host, because only
  whoever owns the fabric knows which addresses on the shared segment are free. A validator-wide
  setting cannot know that.
- A free virtual function. Discovery reads the PF's sysfs tree from the host namespace, where a VF
  already handed to a container no longer appears — so it lists exactly the unclaimed ones.

Inside the namespace the interface is renamed to **`fabric0`** before it gets its address. VF netdev
names vary per host and driver, and NCCL/gloo never pick a second interface on their own — a colleague's
cross-host run only worked with `NCCL_SOCKET_IFNAME`/`GLOO_SOCKET_IFNAME` pointed at the right device.
With one documented constant that setting is the same on every pod: `NCCL_SOCKET_IFNAME=fabric0`.

Addresses are derived from the VF index rather than allocated: indexes are unique per host, the range
belongs to that host alone, and a derived address needs no state to survive a validator restart or two
containers starting at once. MACs are derived from the container name, so a rebuilt container keeps the
address its neighbours' ARP caches and the switch's MAC table already learned.

## Not in this PR

- **Fabric-side partitioning.** The per-rental VLAN is programmed on the VF here, but the switch has to
  trunk it, and IB wants pkeys with limited membership instead. On the one real multi-node fabric in
  the fleet (`sm_lid=0xe`, the `69.63.236.160` trio) the subnet manager is the provider's, so IB
  partitioning is a conversation before it is code.
- **The macvlan fallback**, for hosts whose NICs expose no VFs. It is only safe together with dropping
  `NET_ADMIN` for those rentals — which costs the in-pod VPN — so half of it is worse than none.
- **A cross-host measurement.** No host on the fleet has exclusive mode and a declared range yet. That
  is what keeps this a draft.

## An ordinary rental cannot notice this

The attach is gated on a forwarded `/dev/infiniband/*` node, so a pod on a box with no card does not
merely skip the work — it issues no SSH command at all. Three tests hold that line: no RDMA devices →
`ssh.commands == []`; RDMA devices on an unprepared host → nothing is configured; a failing attach step
on a running pod → logged and swallowed, because a fabric the pod cannot reach is a degraded rental,
not a failed one.

Verified on staging. The branch is deployed on the staging validator, and a pod rented through it on a
host with no RDMA hardware came up completely untouched:

```
/dev/infiniband:        No such file or directory   -> attach returns before any SSH command
interfaces:             lo, eth0 172.20.0.2/16, docker0   (no fabric0)
ulimit -l:              8192   (default — memlock is only raised when RDMA devices are forwarded)
nvidia-smi:             NVIDIA RTX A6000
```

`Container creation step finished` in the staging validator log, and zero `RDMA_FABRIC` lines — an
unprepared host returns before logging, so nothing new is emitted on the 322-of-383 hosts with no card.

Still owed: the same run on a staging host that HAS a card, to exercise the branch with `forwards_rdma`
true. (An earlier attempt at that inspected a pod running released `main`, so it proved nothing here.)

Full validator suite: 1508 passed. The 3 failures in `test_sshd_bootstrap_script.py` are pre-existing —
verified by stashing this branch and re-running them on `main`.

Tests: `tests/test_rdma_fabric.py`, 16 cases — the three refusals, the hardware-enforced identity, the
ordering that makes the GID table non-empty, index-derived addresses that cannot collide, and a failed
step raising rather than leaving a half-attached function.
`pytest tests/test_rdma_fabric.py tests/test_nvidia_devices.py tests/test_infiniband_passthrough.py tests/test_scrape_infiniband.py` — 64 passed.

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -78,6 +78,7 @@ from services.gpu_power_limit import (
 )
 from services.gpu_wedge import cure_wedged_gpus, query_wedged_gpu_uuids
 from services.nvidia_devices import build_gpu_docker_config_for_executor
+from services.rdma_fabric import attach_container_to_rdma_fabric
 from services.redis_service import (
     STREAMING_LOG_CHANNEL,
     RedisService,
@@ -931,6 +932,55 @@ class DockerService:
             shm_size=custom_options.shm_size,
             entrypoint=custom_options.entrypoint,
         )
+
+    async def _attach_rdma_fabric_if_available(
+        self,
+        *,
+        ssh_client: asyncssh.SSHClientConnection,
+        container_name: str,
+        forwarded_devices: tuple[DeviceMount, ...],
+        default_extra: dict,
+    ) -> None:
+        """Give the container its own function on the RDMA fabric, when the host supports it.
+
+        Without this the container reaches its own cards but nothing across the wire: its only
+        address is the NAT bridge's, while the card's GID carries the host's (DAH-2602).
+
+        A failure here never fails the rental. Every precondition lives on the host — exclusive
+        RDMA namespace mode, a declared address range, a free virtual function — so an unprepared
+        host is the normal case, not an error, and the pod is still exactly what today's
+        passthrough delivers.
+        """
+        forwards_rdma = any(
+            device.path_on_host.startswith("/dev/infiniband/") for device in forwarded_devices
+        )
+        if not forwards_rdma:
+            return
+        try:
+            attachment = await attach_container_to_rdma_fabric(ssh_client, container_name)
+        except Exception:
+            logger.warning(
+                _m(
+                    "RDMA_FABRIC_ATTACH_FAILED",
+                    extra=get_extra_info({**default_extra, "container_name": container_name}),
+                ),
+                exc_info=True,
+            )
+            return
+        if attachment:
+            logger.info(
+                _m(
+                    "RDMA_FABRIC_ATTACHED",
+                    extra=get_extra_info({
+                        **default_extra,
+                        "container_name": container_name,
+                        "virtual_function_netdev": attachment.virtual_function.netdev,
+                        "virtual_function_rdma_device": attachment.virtual_function.rdma_device,
+                        "ipv4_cidr": attachment.ipv4_cidr,
+                        "vlan_id": attachment.vlan_id,
+                    }),
+                )
+            )
 
     @staticmethod
     def _memlock_ulimit_for(
@@ -4473,6 +4523,13 @@ class DockerService:
                     )
 
                     logger.info("Container creation step finished")
+
+                    await self._attach_rdma_fabric_if_available(
+                        ssh_client=ssh_client,
+                        container_name=container_name,
+                        forwarded_devices=run_spec.devices,
+                        default_extra=default_extra,
+                    )
 
                     # DAH-1524: isolate the bare `docker run` (dominated by the NVIDIA
                     # --gpus prestart hook, +sysbox/storage-opt) from the post-run
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -563,6 +563,11 @@ def build_startup_command_args(startup_commands: str | None) -> str:
     return " ".join(shlex.quote(token) for token in tokens)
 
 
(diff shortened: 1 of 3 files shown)


arhangel66 (approved):
I think it's ok to deploy now and do next as followup later if we will use it.
----

Two blockers before this leaves draft; both are cross-tenant, and partial rentals mean several tenants per host is the normal case (`is_partial_rental` in `nvidia_devices.py`).

**1. Concurrent rentals race for the same VF, and the loser breaks the winner.**
Discovery and the move are several SSH round-trips apart, and nothing serializes them per host. Two `create_container` calls running side by side list the same free function; the loser then runs `reserve_virtual_function` *after* the winner has already moved it, reprogramming the MAC of a live tenant's `fabric0` from the PF side. The winner silently drops off the fabric, the loser fails on the move and is swallowed. The docstring claims the derived address survives "a second container starting concurrently" — that holds for the address, not for discovery. `redis_service.acquire_executor_lock(executor_id)` already exists for exactly this shape of race in port allocation; holding it across discover → reserve → move closes the window.

**2. `virtual_function.index` is not host-unique, so the derived address is not either.**
Discovery walks every `/sys/class/infiniband/*` device and every PF under it, but `build_attachment` indexes the range by VF index alone. Once the first PF's functions are all claimed, the next rental gets `virtfn0` of the second PF and lands on the address the first rental is already using — same segment, ARP conflict, RDMA traffic going to the wrong place. Whether it fires depends on `sriov_numvfs` versus the number of concurrent rentals, so a well-provisioned box may never show it and the first host with `numvfs=2` will. Please key the slot off `(PF, VF index)` or give each PF a disjoint sub-range, and add a two-PF regression test — `..._two_containers_cannot_collide` only compares indices 0 and 1 of the same PF, so it can't catch this.

PR #1151, opened 2026-07-20
chore: DAH-2442 — install sysbox 0.7.0 in dstacktee CVM pre-launch

fortunelucky777:
## Describe your changes

Point the executor's dstacktee CVM pre-launch at our own **sysbox 0.7.0** installer image instead of the upstream `dstacktee/dstack-sysbox-installer:1.0.0` (sysbox 0.6.7) build.

- Image: `dstacktee/dstack-sysbox-installer:1.0.0@sha256:2f5dbea9…` → `daturaai/dstack-sysbox-installer:0.7.0@sha256:6e7e8e643a8c70467de5232b74cc1f6263ccfcc99d628425de6f1d5d59e0f116`
- Dropped the now-resolved `# TODO: pin digest` (the new image is Datura-owned and already digest-pinned).
- Comment now names the sysbox 0.7.0 build.

The installer image was built and published from [Datura-ai/dstack-sysbox-installer#1](https://github.com/Datura-ai/dstack-sysbox-installer/pull/1) (sysbox `v0.7.0`, commit `27cb713…`) via the repo's release workflow → `daturaai/dstack-sysbox-installer:0.7.0` (sigstore-attested). The digest above was verified identical across the GitHub release body and the DockerHub tag.

The `check_existing` stub-out workaround is unchanged and still matches the shipped `install-sysbox-complete.sh` (the `check_existing()` function is present), so the force-reinstall over 0.5.11's baked sysbox still applies.

## Issue ticket number and link

DAH-2442 — bump dstacktee CVM sysbox to 0.7.0

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [ ] I wrote tests. — N/A: shell pre-launch script executed inside the CVM. Validated with `bash -n` + digest cross-check; runtime validation is a fresh CVM rental exercising the sysbox/GPU compatibility probe.
- [ ] Need to take care of performance? — No.

PR #1134, opened 2026-07-13
DAH-2386: Route TDX attestation to a QEMU 10.x verifier by vm_config qemu_version

fortunelucky777:
## Describe your changes

**Context.** A `dstack-verifier` image embeds exactly **one** QEMU version's ACPI oracle, so one verifier can attest exactly one host-QEMU version. RTMR0 is a byte-exact SHA-384 of the QEMU-emitted ACPI tables, so a quote from a different QEMU version simply does not verify against the wrong oracle. The prod fleet runs **QEMU 9.2.1** (default verifier, digest-pinned, unchanged). This PR adds an optional **second** verifier plus version-based routing so a future mixed 9.x/10.x fleet stays verifiable, with today's behavior byte-for-byte unchanged.

**What changed**

- **Config** (`neurons/validators/src/core/config.py`): new optional `TDX_VERIFIER_QEMU10_URL` (default `None`).
- **Routing** (`neurons/validators/src/services/attestation_service.py`):
  - `_parse_qemu_version` reads `vm_config.qemu_version` from a quote, handling both wire forms — `vm_config` as a JSON string and as an already-decoded object.
  - `_select_verifier_url` routes a quote to the qemu10 verifier **only** when that URL is configured **and** the parsed major version is `>= 10`; everything else (no qemu10 URL, a 9.x host, or an absent/unparseable version) routes to the default verifier.
  - Ultra-defensive: the whole body is wrapped in `try/except` and any failure falls back to the default verifier. This runs on every attestation (the entire 9.2.1 fleet), so it can never break the existing path.
  - `_call_verifier` now selects the URL via the helper and logs which verifier was chosen (`verifier=qemu10|default`, plus the parsed `qemu_version`). The not-configured guard and `/verify` suffix logic are unchanged.
- **Compose** (`docker-compose.app.dstacktee.yml` + `.dev.yml`): new `dstack-verifier-qemu10` service pinned by digest `daturaai/dstack-verifier@sha256:3151aab51d618a0996cc353979c72b04f2137fd5be17be7487f1242faab3f83d` (tag `0.5.11-qemu10.2.1`), same restart policy and `DSTACK_VERIFIER_IMAGE_DOWNLOAD_URL` as the default verifier; validator gains `TDX_VERIFIER_QEMU10_URL` env and a `depends_on` entry. `docker-compose.local.yml` is intentionally untouched.
[...]

PR #1203, opened 2026-08-06
DAH-2544 - Add the day-zero Ansible host setup for CVM v2

fortunelucky777:
## Describe your changes

Takes a fresh Ubuntu 25.10 / 26.04 Intel TDX host from bare OS to CVM-ready with one
command, replacing `neurons/executor/dstacktee/docs/host-setup.md` §1–§4.

```bash
cd neurons/executor/ansible
sudo ./bootstrap.sh
```

Built from the ralplan-approved plan (`.omc/plans/dah2544-ansible-day-zero.md`, r3).

### What lands

| | |
|---|---|
| `neurons/executor/ansible/` | 11 roles, `bootstrap.sh`, a JSON schema, and a test harness |
| `.github/workflows/test.yml` | two merge-gate jobs (no `paths:` edit needed — `neurons/executor/**` already matches) |
| `.github/workflows/qemu-2604-compile.yml` | `workflow_dispatch` de-risk job, deliberately **not** a merge gate |
| `docs/host-setup.md` | §1–§4 replaced by the one command; §5, §6 and Troubleshooting untouched |
| `.dockerignore` | gains `ansible/` |

`cvmd` and `catalog` ship as variable interfaces only, so day-zero does not wait on
DAH-2575 / DAH-2576. Nothing under `dstacktee/app/**` is touched, and `lium-cvm.sh`
and `dstack.py` are unchanged (Non-Goals).

### Evidence

| Box | Status | Where it was proved |
|---|---|---|
| 1 · lint + syntax + CI job | ✅ | `ansible-lint` passes at `production`, above the required `moderate` |
| 2 · bootstrap installs ansible-core; preflight names each knob | ✅ | Reproduced in a real `ubuntu:26.04` container |
| 3 · GRUB: drop-in, split vars, drift | ✅ | 4 fixtures + 6 edge cases + a 5-case parser unit: render → union → write → regenerate → assert → drift |
| 4 · reboot/resume exactly once | ⚠️ logic only | 9 machine-checked branches. **The real cycle needs the VM runbook — not yet run.** |
| 5 · idempotence | ⚠️ kernel role only | Pass 2 leaves all 4 managed files byte-identical. Whole-playbook idempotence needs the VM. |
| 6 · destructive guard | ✅ | 17 checks, all six plan cases plus the ladder order |
| 7 · verify JSON + schema | ✅ | 34 ids agree across schema enum, README headings and the live report |
| 8 · docs retired | ✅ | 18 checks |
| — · first and second converge both complete | ✅ | 2 new plays in `test_repo_app_pristine.yml` |

The container job runs `docker run ubuntu:26.04` against the committed branch and
asserts positively: `ansible-playbook` absent before and present after, both state
directories created by the run, `--print-expected-preflight-rc` creating nothing, the
exit code matching the constant the script exports, all five BIOS strings present
including the *device-absent* branch, and `os.supported` **not** firing — which is what
makes the assertions above it meaningful.

### Defects found and fixed before this was opened

An adversarial review ran 14 mutation experiments against the tree and confirmed the
tests are real: declaring `destructive` on the role entry instead of the play failed
`test_maintenance_profile.yml` by name, three mutations of the guard ladder were caught,
and mutating the drop-in template to assign `GRUB_CMDLINE_LINUX` failed all four GRUB
fixtures.

Three defects were only reachable as root on a supported OS, so the container job is the
only thing that could have found them: `$SUDO --preserve-env=X cmd` made the *flag* the
command when `SUDO` is empty; `--tree` is adhoc-only and `ansible-playbook` rejects it,
so the plan's invocation would have failed on every run; and `set +e` does not disable an
`ERR` trap, which discarded the playbook's real exit code.

Two more mattered on real hosts:

- **The guard returned `CLEAN` on a host holding a renter's encrypted disk** whenever the
  checkout was not at `/opt/lium-io`. `find -maxdepth 8` covers exactly that one path
  shape; `/home/<user>/lium-io` is depth 9 — and `/home` is in the search list precisely
  to catch a home-directory checkout. We have a host on record with its checkout on
  `/data0`. The budget is now 12, and the guard locates the checkout from its own path.
- **The `vfio-pci.ids` superset rule refused to converge on healthy hosts.**
  `present_pci_ids` is built only from the ids we manage, so any other `vfio-pci` entry
  read as "dropped" on every run: a host also passing an SR-IOV NIC or NVMe, any Blackwell
  host (§6.2 withdrew those ids), and anyone who typed the ids in upper case — the kernel
  parses them with `%x`, so `10DE:2335` works. Ignoring those ids fixes the refusal and
  opens something worse: `vfio-pci.ids=` is a module parameter whose **last occurrence
  wins** and our drop-in lands last, so the unmanaged device would be silently unbound.
  The role now compares only within the managed id universe, lower-cased, and **carries
  unmanaged ids forward** into what it writes.

Also fixed: both `!`-inverted deny rules were structurally inert (the comment stripper
prefixed each line with `path:lineno:`, so `^` bound to the file path) — those guard the
defect that has already shipped once; the pre-reboot `grub.cfg` assertion failed open on a
full `/boot`; the fixture tests were reading the **real host's** `/etc/default/grub`; and
the resume unit could hang a host at boot forever (a bare `sync` in `ExecStartPre` of a
`Type=oneshot` unit, which systemd gives no start timeout).

### The day-zero path did not work, and a four-angle review found why

A cleanup review across reuse, simplification, efficiency and altitude found
**six** further defects. Three broke the primary path outright. Each was
reproduced before the fix and re-broken after it, so each has a test that fails
without the change.

1. **The `vfio-pci.ids` parser silently lost ids.** `local IFS=,` has *function*
   scope, so it was still set on the next trip round the origin loop, and
   `resolve_key` splits its line with an unquoted `for tok in $line`. Every source
   after the first one carrying ids resolved to nothing: ids in
   `GRUB_CMDLINE_LINUX_DEFAULT` were dropped from the carry-forward and unbound on
   the next boot, and an id only in `/proc/cmdline` vanished from the union the
   narrowing refusal is built on — so that safety rule was partly inert too.
   Fixing it exposed a second, semantic bug: the carry-forward unioned both GRUB
   variables, but `10_linux` emits `${GRUB_CMDLINE_LINUX}
   ${GRUB_CMDLINE_LINUX_DEFAULT}` and last occurrence wins, so a host with ids in
   both boots with `_DEFAULT` alone. Carrying the union would re-bind a device the
   provider had stopped passing through — and binding the host's own NVMe to
   `vfio-pci` is not a smaller mistake than unbinding a NIC.
2. **The QEMU source build always failed.** `build-qemu.sh` printed its completion
   marker to stdout and `systemd-run` adds no redirection, so `wait_for` on
   `qemu-build.log` timed out after the full 3600 s *after* a build that
   succeeded. It is the default path. The `workflow_dispatch` compile job stayed
   green because it supplied the missing `tee` itself — the one job meant to
   de-risk this never exercised it.
3. **`verify` reported a correctly converged host as broken, two ways.** `lium` is
   a `set_fact` taken once in play 1, so ~34 of 36 checks described the host
   *before* any role ran. And `verify` rebuilt the expected token set from
   `present_pci_ids` alone while `kernel` writes present + carried-forward, so the
   carry-forward fix turned into two permanent `blocker` checks on exactly the
   hosts it was meant to unblock. The suite could not see it: the only test that
   reaches `verify` runs `--skip-tags destructive`, which skips `kernel`.
4. **The first converge never downloaded the OS image** — gated on
   `lium.repo_present`, a snapshot taken before the clone in the same role.
5. **The second converge aborted in the `repo` role.** `sgx` repoints `pccs_url`
   in a *tracked* file, and the git module fails on any tracked modification when
   `force` is no.
6. **A crashed analyser read as a pass.** The fallback substituted three empty
   lists, rendering `kernel.grub_persisted` as OK and resetting the reboot budget,
   which disabled loop protection.

The same review drove a cleanup pass: seven facts nothing consumed, a QEMU
fingerprint task duplicated across both install arms, six repeated `when:` clauses
collapsed to one `block`, `min_ansible_version` corrected from `2.16` to the
`2.18` the tree requires, and eight plays that no longer run a full `setup` to
supply two values.

### Four findings left for you rather than fixed blind

Each needs a decision or hardware this branch cannot reach:

- **The seam gate freezes `lium_hda_search_roots`.** A provider whose checkout is
  on a data volume must add `/data0`, and doing so aborts the run with *"set to a
  non-production value"*. We have such a host on record — and *not* adding it
  means a `CLEAN` verdict on a host holding a renter's encrypted disk. The fix is
  a superset gate for that one entry (`/data0` may be added, `/opt` may never be
  removed), which is a different rule from every other seam.
- **`lium_pccs_url` defaults to `localhost:8081`** and is written on the keyless
  path too, where nothing listens — leaving a provider worse off than if the role
  had never run.
- **`sgx-dcap-pccs` is installed unconditionally** but its `dpkg --configure`
  repair is gated on the API key, so the degraded path — the one every new
  provider takes — ends with a half-configured package that errors on every later
  apt transaction.
- **The key-provider health check reads cumulative `docker logs`** on a
  `restart: always` container, so a `PRODUCTION mode` line from a previous
  incarnation satisfies the wait while the enclave crash-loops now.

### Deviations from the plan — please review these

Nine in total. The three worth your attention:

- **D-1 — `host-setup.md` §3b does not exist on `main`.** The plan says to preserve it as
  an appendix; it is an uncommitted local edit on another branch. **If §3b lands on `main`
  before this merges, the appendix and its grep must be added.**
- **D-4 — the GRUB drop-in carries the full token set, not just the missing ones.** The
  plan's literal wording is not idempotent: the drop-in is inside the configuration being
  analysed, so run 2 finds nothing missing, empties the file, and run 3 finds them missing
  again.
- **D-7 — `forbidden-patterns.txt` carries a per-pattern scope**, because the plan itself
  scopes `|| true` to CI while scoping the GPU reset flags tree-wide.

### Decisions the plan asks for and this PR does not settle

- **§6.9 — Intel API key distribution.** Provider-supplied, Lium-distributed, or is the
  remote-PCCS degraded path the intended provider experience? The degraded path ships and
  is loud about itself.
- **§6.10 — docs retirement.** Clean retirement now (what this does), or a labelled
  "superseded" appendix for one release cycle?
- **§6.2 — Blackwell.** B200/B300 support is **not** claimed anywhere, because no
  field-verified PCI ids exist. Restoring the claim needs ids.

### Not run here

- **The hardware-gated TDX acceptance run.** Blocks task close, not this merge.
- **`tests/vm/RUNBOOK.md` on multipass** — the real reboot/resume cycle and whole-playbook
  idempotence. Boxes 4 and 5 are honest about this.
- **The QEMU-on-26.04 compile job.** `workflow_dispatch`, 10–40 minutes. Worth running
  before merge — it de-risks the largest unproven assumption, that the pinned 9.2.1 tree
  still builds on 26.04.

[DAH-2544 — CVM v2 — day-zero Ansible host setup (FR-A1–A3)](https://app.notion.com/p/CVM-v2-day-zero-Ansible-host-setup-FR-A1-A3-3b1b8bfdbde98195b140f768113796e2)

- [x] I wrote tests.


fortunelucky777:
## Hardware acceptance run — au11, 2026-08-07

Ran the tree end to end on a real Intel TDX host: `chutes-miner-tee-au11`, Ubuntu 25.10 /
kernel 6.17.0-23, 2011 GB RAM, one 28 TB `/`, 8 CC-capable GPUs (`10de:2335`, which
nvtrust reports as `H100-PCIE`) + 4 NVSwitch, converted from a Chutes SN64 miner.
Checkout at `$HOME/lium-io`, so `--repo-path` and `--no-clone` throughout.

Six defects had to be fixed to get a clean run. Pushed as `cb5f5087..90347447`.

**Reading the runs separately, because they prove different things — the reboot and the
green converge are not the same run:**

| Run | What it was | Outcome |
|---|---|---|
| A | first converge | **reboot task failed** — `ansible.builtin.reboot` refuses a local connection (defect 1). No reboot happened |
| B | converge after fixing the reboot | teardown gate passed, timer fired, **host rebooted**, resume unit started itself and stopped in the `gpu` role (defect 4) |
| C | converge after fixing the `gpu` ordering | `28 OK · 2 WARN · 0 MUST_FIX · 0 UNKNOWN · 4 SKIPPED`, **exit 0** — no reboot needed, the tokens were already live |
| D | `--tags sgx` after fixing the QPL config | `0 MUST_FIX`; `/etc/sgx_default_qcnl.conf` now parses |
| E | `--tags verify`, CVM running | `27 OK · 4 WARN · 0 MUST_FIX · 0 UNKNOWN · 4 SKIPPED` |

⚠️ **Run C reported `0 MUST_FIX` on a host that could not boot a CVM.** That green result
*is* the evidence for defect 6: `verify` was grepping for a key rather than parsing the
file, and the QPL config it was reporting on had already been corrupted by the converge
itself. A CVM launched at that point rebooted at ~160 s. That is why runs D and E exist.

After D, the CVM was rebuilt from the preserved config and booted:

| | |
|---|---|
| reboot (run B) | host went down 25 s into the watch, back at 317 s — ~5 min, 2 TB TDX memory training |
| CVM rebuild | `app-compose.json` byte-identical → `app_id 8ea0f897…` unchanged, whitelist still matches |
| CVM boot | executor answering on `:32000` at t+464 s (404 on every path is the liveness signal) |

All five kernel tokens are live after the reboot: `kvm_intel.tdx=on intel_iommu=on
iommu=pt vfio-pci.ids=10de:2335,10de:22a3 vfio_iommu_type1.dma_entry_limit=16777216` —
with `=on` landing after the provider's `kvm_intel.tdx=1`, so last-occurrence-wins puts
ours in force.

### Box 4 — reboot/resume

The real cycle ran. The reboot fired, the host came back on the new command line, and the
one-shot unit started on its own and disarmed itself (`systemctl is-enabled` →
`disabled`). It fired **once**; `resume.state` is left behind, which is the documented
behaviour when the resumed converge exits non-zero, and `ExecStopPost` disabled the unit
anyway — so the "cannot fire twice" property held on the failure path, which is the path
that matters. `verify` then reset `converge_reboots` to 0 on the clean converge.

⚠️ The **happy** path — reboot → resume → converge completes unattended — was *not*
observed end to end. The resume hit defect 4 below and stopped in the `gpu` role; by the
time that was fixed the host no longer needed a reboot, so the later converges never
re-armed the unit. Box 4 should stay ⚠️ until one run goes reboot → resume → green
without a human in between.

### Box 5 — idempotence

⚠️ Still not proven, and this run could not prove it. Two back-to-back **full** converges
need a `CLEAN` host, and once the CVM is running the guard reports `LIVE` and
`bootstrap.sh` correctly routes to the maintenance profile, which withholds
kernel/gpu/qemu. Unchanged from the PR body.

---

### Defects found — none of these were reachable without real hardware

**1 · `ansible.builtin.reboot` refuses a local connection, so the reboot path had never
worked.**

```
Running ansible.builtin.reboot with local connection would reboot the control node.
```

`inventory/localhost.yml` sets `ansible_connection: local` by design — a provider runs
this on the host it is provisioning — so the task failed on 100% of runs that reached it,
and the day-zero converge could never apply a kernel command line unattended. The
module's model does not fit: it reboots and then reconnects to carry on, and here the
control node *is* the host. It went unnoticed because every reboot on a real host so far
had been issued by hand.

Replaced with a detached `systemd-run --on-active` timer (`lium_reboot_cmd`), then the
play **holds** until the shutdown kills it. The hold is load-bearing: `meta: end_play`
ends only the current play, so Ansible would walk into the `gpu` play — flipping CC mode
across eight GPUs and rebinding them — with seconds to live. Everything past that point
is redone by the resume unit anyway. Surviving the hold means the reboot never fired, and
that now fails loudly. `--on-active` rather than a bare `reboot` so the task returns and
the log flushes instead of the shutdown SIGTERMing Ansible mid-task.

**2 · The reboot could still be issued into an in-flight TDX teardown.**

This is the ar6 incident, and it was reachable from this tree: the guard above the reboot
is waivable with `-e lium_force_reboot=true`. A soft reboot during a teardown asks the
kernel to first finish the very teardown it would then be blocked behind, then shut down
the same GPUs that teardown is still unmapping. On a production host that left the old
kernel alive on the network for 14+ minutes — ping answering, port 22 completing the
handshake and never sending a banner — and took a datacentre power-cycle.

`roles/kernel/tasks/wait_for_teardown.yml` now runs immediately before the reboot,
reading `/proc` fresh rather than the play-1 guard snapshot. A **running** qemu-system is
refused at once (it never drains on its own, so polling would burn the whole budget to
reach the same answer); a **zombie** one is waited for, because a teardown does finish.

It sits *below* `lium_force_reboot` and has **no override of its own**, deliberately:
that variable authorises destroying a drained encrypted disk, which is a trade an
operator can rationally make, and wedging the host is not — it costs an outage and buys
nothing. The only knob is `lium_reboot_drain_timeout_sec`. When waiting genuinely is not
an option the answer is an out-of-band power-cycle, which Ansible cannot and should not
perform.

Reproduced live on au11 while stopping the existing CVM: the zombie ground for ~4.5 min
with cpu ticks climbing steadily at roughly one to two cores of page reclaim, while
`MemAvailable` sat **dead flat** —
which is exactly what makes a healthy teardown look hung. `lium-guard.sh --qemu-procs`
(new) reports `pid state threads cpu_ticks` and skips the disk sweep so the gate can poll
it. The guard also no longer tells the operator that *"a host reboot is the only thing
that frees it"* — the sentence the ar6 reboot followed.

**3 · `bootstrap.sh` rejected the `-e` its own messages tell operators to run.**

Two remediation messages say `sudo ./bootstrap.sh -e lium_reset_reboot_budget=true`, and
the entry point answered `Unknown flag: -e`, a usage dump, and exit 64 — at the moment
the operator is already stuck. `-e/--extra-vars` is now parsed and appended **last**, so
an explicit `-e` beats the values bootstrap derives from its own flags. Secrets stay out
of argv (that is what `--vars-file` is for) and the usage text says so.

`tests/test_bootstrap_args.sh` derives its cases instead of listing them: it greps the
tree for the invocations the tree itself recommends and asserts each is accepted, so a
message added later is covered without anyone remembering.

**4 · The `gpu` role gated on CC-mode readability *before* installing the tool that reads
it.**

```
REFUSING: the confidential-compute mode of 8 GPU(s) could not be read
  0000:19:00.0: nvtrust tool /opt/nvtrust/.../nvidia_gpu_tools.py is not available
Install or repair nvtrust first, then re-run.
```

…while the very next task was the one that installs nvtrust. Re-running changed nothing.
That aborts on every host without a pre-existing `/opt/nvtrust` — every fresh host, the
only kind this playbook exists for.

Moving the clone above the gate is not sufficient on its own: `lium.gpu_cc_mode` is
collected in play 1, before the clone exists, so the bundle still says "not available"
and the gate would keep judging a snapshot the role had just invalidated. The facts are
re-read between the two, conditional on CC mode having been unreadable at entry.

Incidental finding: `nvidia_gpu_tools.py --query-cc-mode` **works on vfio-bound devices**
(it forces D3→D0 and restores power control), so nvtrust does not need the nvidia driver
loaded. CC mode was already `on` on this host.

**5 · The converge corrupted Intel's QPL config into invalid JSON, so every CVM it
produced rebooted at ~160 s.**

The worst one. Two `lineinfile` tasks wrote `/etc/sgx_default_qcnl.conf`. Intel ships that
file in **leading-comma** style — first key bare, every later key prefixed with a comma:

```
{
  "pccs_url": "https://localhost:8081/sgx/certification/v4/"
  ,"use_secure_cert": true
}
```

- `regexp: '^\s*"use_secure_cert"\s*:'` never matched `,"use_secure_cert"`. lineinfile
  **appends** when its regexp matches nothing, so the line landed after the closing brace
  and the file stopped being JSON.
- `"pccs_url"` *did* match and was rewritten with a **trailing** comma, which followed by
  the next key's leading comma produced a double comma.

Either one alone is fatal, and nothing on the host says so. The failure surfaces three
layers away:

```
dstack-prepare.sh: Failed to get sealing key: error sending request for url
  (http://10.0.2.2:39053/api/GetSealingKey?json)
gramine_sealing_key_provider::server] Connection error: DCAP error
[  166.598935] reboot: Restarting system
qemu-system-x86_64: cpus are not resettable, terminating
```

The QPL cannot parse its config → quote generation returns a DCAP error → the sealing-key
provider refuses the guest → the guest cannot unseal its data disk → `panic=1` reboots it
→ a TDX guest cannot be reset, so QEMU exits. Meanwhile the converge reported success.

The file is now **written whole from a template**, which removes the class rather than the
instance: no regexp to fail to match and no house style to guess. The write carries
`validate:` with `json.load`, so an unparseable result is refused instead of installed —
that caught two defects in the template itself while writing it (a JSON file cannot carry
an `ansible_managed` banner, and Ansible's `trim_blocks` makes an inline Jinja comment
join the next key onto the previous line). The pristine copy is kept once at
`.pre-lium-cvm` with `force: false`, so a second converge cannot overwrite the original
with the managed one.

The key provider's own copy keeps `lineinfile`, because it is a tracked file the `repo`
role restores with `git checkout --` and it is plain strict JSON today — so both regexps
match. "Today" is the whole risk, so a `json.load` check now runs after those edits.

**6 · `verify` reported `0 MUST_FIX` on a host that could not boot a CVM.**

`sgx.qpl_secure_cert_false` **greps** for `"use_secure_cert"\s*:\s*(true|false)`, and the
corrupting append left exactly that string findable — after the closing brace. The check
read `false`, said OK, and the QPL could not read a word of the file.

A value check and a well-formedness check are different questions, and the first cannot
answer the second however it is written. `sgx.qpl_config_parses` is therefore a separate
check on a separate fact. Whole-line `//` comments are stripped before parsing, because
Intel's shipped file uses them and their parser accepts them — a pristine platform file is
*correct* and must not be reported broken. Only whole-line comments: guessing where a
`//` sits inside a string is how a stripper corrupts what it is checking.

`lium_system_qcnl_path` moved to `group_vars`; it was a role default of
`sgx_key_provider`, so the verify template referencing it failed with
`'lium_system_qcnl_path' is undefined` — caught only by the one fixture test that renders
a whole report.

---

### One more, about the tests themselves

`test_verify_schema.sh` compared the schema enum, the README headings and the golden
example — **three descriptions of the code, and none of them the code.** A check added to
`checks.json.j2` and nowhere else left all three agreeing with each other while the role
emitted an id its own schema rejected. That is precisely what adding check 6 above did,
and the test passed. It now reads `checks.json.j2` as a fourth source, with a floor on the
parsed id count so a regex that stops matching cannot make the comparison vacuous.

`tests/test_task_ordering.sh` is new and pins both ordering invariants above (nvtrust
before the CC gate; teardown gate → schedule → hold → fail). Both defects were pure
ordering bugs — every task present, every task individually correct, only the sequence
wrong. No syntax error, no lint warning, no failing test; only a converge on real
hardware. Its `line_of()` runs `grep` as an `if` condition rather than in a pipeline,
because the obvious form makes a missing marker kill the script at the assignment under
`set -e` and report exit 1 with **no output at all** — which is the silent-assertion
failure mode this tree keeps removing, and it happened while writing this file.

Every new assertion was negative-tested by breaking the thing it guards and confirming it
fails by name.

### Suggested evidence-table updates

| Box | Was | Now |
|---|---|---|
| 4 · reboot/resume exactly once | ⚠️ logic only | ⚠️ **real reboot + resume observed on au11**; unit fired once and disarmed itself. Unattended happy path still unproven |
| 5 · idempotence | ⚠️ kernel role only | unchanged — needs a CLEAN host for two full converges |
| 7 · verify JSON + schema | 34 ids / 3 sources | 35 ids / **4 sources**, now including `checks.json.j2` |
| — · hardware acceptance | — | ✅ au11: converge exit 0, `0 MUST_FIX`, CVM rebuilt and booted |

### Known, not fixed

- `bootstrap.sh` prints the **previous** run's verify report when a play fails before
  `verify`. After the reboot it showed the pre-reboot report's "5 tokens missing" as
  though current, while `/proc/cmdline` already had them. Check `generated_at` and
  `host.boot_id` before trusting a report. Left alone rather than widening this PR.
- The 4 warnings on the final run are all expected: two exist *because* a CVM is running
  (`guard.host_state: LIVE`, `repo.ref_behind`), `net.acl_probe_required` is not
  answerable from the host, and `kernel.grub_conflicting_tokens` is the Chutes
  `99-tdx.cfg` `kvm_intel.tdx=1` sitting alongside our `=on` — ours lands last and wins.
  Its remediation points at `/etc/default/grub`, where the stale copy is not; minor
  wording issue, not touched here.
- Worth knowing for any cloud-image host: `50-cloudimg-settings.cfg` **assigns**
  `GRUB_CMDLINE_LINUX_DEFAULT`, so edits to `/etc/default/grub` are dead. This tree
  already handles it — `99-lium-cvm.cfg` appends and sorts after `50-` — but
  `update-grub` having run is **not** evidence the tokens took. au11's `grub.cfg` was
  regenerated 3 s after the file was edited and still contained none of them. Read the
  generated `linux` line.

### Caveat

One host, one architecture (H100-PCIE + NVSwitch), one OS (25.10). Nothing here exercises
26.04 on metal, Blackwell, or a host with a genuinely fresh `/opt/lium-io` clone.

fortunelucky777:
## Rebased onto `feature/2462-cvm-v2`

This PR now targets **`feature/2462-cvm-v2`**, the integration branch for the CVM
upgrade v2 epic (DAH-2462), instead of `main`. The branch was cut from `main` at
`30628c4d` and carries no commits of its own, so the umbrella is byte-identical to
`main` at that point.

The 37 commits were rebased from their old base (`b38ea5e0`, 92 commits behind) onto
that branch. `90347447` → `f749f206`.

| | |
|---|---|
| rebase | 37/37 replayed, **zero conflicts** |
| diff | **12,568 additions across 141 files** — unchanged from before the rebase |
| content | `git diff 90347447 f749f206 -- neurons/executor/ansible .github` is **empty** |
| pre-rebase head | `90347447` (rollback point, since the rebase rewrote all 37 SHAs) |

Conflict-free because `main`'s 92 new commits and this branch have **zero file-level
overlap** — `main` touched `neurons/executor/{src,tests}`, the Dockerfile and the
compose files, none of which this PR goes near.

### ⚠️ Checks on this PR are stale — read this before trusting them

`.github/workflows/test.yml` fires on `pull_request: branches: [main, dev]`, and
`branches:` filters the **target** branch. With the base now `feature/2462-cvm-v2`,
**this PR no longer receives CI**, and no workflow change was made to restore it.

The check run currently attached is an artifact of ordering, not evidence of coverage:
the force-push fired `synchronize` at `23:10:31Z` while the base was still `main`, and
the retarget landed about thirty seconds later. So a full suite queued under the old
base and will report against a PR that gets nothing from here on. **Any future push to
this branch will show no checks at all.** Do not read a green tick here as "CI is
running on the umbrella".

This is deliberate, and the gates are deferred rather than lost: when
`feature/2462-cvm-v2` opens its own PR into `main`, that PR's base **is** `main`, so
Ansible (lint/fixtures + supported-OS), Executor Tests and Validators Tests all run
across the whole epic before anything reaches production.

In the meantime the **local suite is the gate**. Re-run in full on the rebased tree,
against current `main`:

- `ansible-lint` — passes at `production` (required: `moderate`)
- `yamllint`, `shellcheck` — clean
- 10 shell suites: forbidden-strings, seam-gate, guard (28 checks), resume-guard,
  verify-schema (35 ids / 4 sources), docs-pointer, cmdline-resolution,
  facts-determinism, bootstrap-args, task-ordering
- 8 fixture playbooks incl. `test_reboot_teardown_gate` and `test_maintenance_profile`
- golden verify report validates against the schema

Sibling CVM PRs #1151 and #1134 were deliberately left on `main` and are unaffected.

PR #1204, opened 2026-08-07
Fix A10 GPU model normalization in validator

pixel29913:
## Summary

- Normalize the raw `NVIDIA A10` NVML alias to the canonical `NVIDIA A10 Tensor Core GPU` model for validator policy and scoring.
- Preserve the raw GPU name while the native capability challenge runs because its challenge key is derived from the executor's raw machine information.
- Canonicalize GPU names only when publishing the completed machine spec to the backend, allowing the machine catalog to populate `machine_name` and `gpu_count` correctly.
- Keep the existing RTX 3090 scrape coverage and add A10 regression coverage for both validation and result publication.

## Root cause

The executor reports `NVIDIA A10`, while the validator and backend machine catalog use `NVIDIA A10 Tensor Core GPU`. Normalizing the stored scrape spec fixed the validator allowlist but changed the native capability challenge input and caused GPU verification to fail. Keeping the raw spec fixed that challenge, but publishing the raw name left the backend unable to match its machine catalog, producing `machine_name: null` and a blank GPU title in the webapp.

This change keeps the raw name only for the challenge and publishes the canonical name after validation completes.

## Impact

A10 executors validate successfully, retain native capability verification, and appear in the API and webapp with the correct GPU model and count.

## Validation

```text
101 passed, 69 warnings
```

Covered suites:

- machine-spec scrape and GPU alias normalization
- GPU model and VRAM prechecks
- native capability and matrix validation
- result publication canonicalization

Staging verification completed with A10 score `1.0`, three successful provider executors, and API output `machine_name: "NVIDIA A10 Tensor Core GPU"`.

### DIFF
--- a/neurons/validators/src/services/task/checks/machine_spec_scrape.py
+++ b/neurons/validators/src/services/task/checks/machine_spec_scrape.py
@@ -8,6 +8,7 @@ from ..messages import MachineSpecMessages as Msg, render_message
 from ..pipeline import CheckResult, Context
 from ..runner import SSHCommandRunner
 from services.file_encrypt_service import ORIGINAL_KEYS
+from services.gpu_spec_table import normalize_gpu_model
 from .network_ema import compute_ema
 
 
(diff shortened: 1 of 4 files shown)


fortunelucky777 (approved):
LGTM. Verified the two-sided split holds: `capability.py`/`verifyx.py` still pass raw `ctx.state.specs` into the challenge builders, while policy/scoring and the published spec use the canonical name — and the backend catalog (`MACHINE_PRICES`, `REQUIRED_DEPOSIT_AMOUNT`, `GPU_ARCHITECTURES`) already carries `NVIDIA A10 Tensor Core GPU`.

Non-blocking:
- `_normalize_gpu_details` and `_canonicalize_published_gpu_names` are the same transform written twice, and only the latter guards `isinstance(detail, dict)` — worth one shared helper next to `normalize_gpu_model`.
- `state.specs["gpu"]["details"]` (raw) and `state.gpu_details` (canonical) now diverge, so the dead `if not gpu_details: gpu_details = specs.get(...)` fallbacks in `gpu_model_valid.py` and `collateral.py` would read raw names if they ever fired — safe to delete.
- The change canonicalizes every `NORMALIZATION_MAP` alias, not just A10 (T4, V100, `Tesla H100 80GB HBM3` also start passing `GpuModelValidCheck`). Backend has all of them, but worth stating in the summary.

PR #1205, opened 2026-08-08
DAH-2575 - Add the cvmd service skeleton, signed two-scope API and its Ansible role

fortunelucky777:
## Describe your changes

DAH-2575 in full: **the cvmd daemon and the Ansible role that deploys it.**

`cvmd` is the host-side control daemon for CVM v2. This PR is the skeleton — transport, authentication, state machine, packaging, and provisioning. The lifecycle handlers land in DAH-2576 and currently answer `501`.

**Base is `feature/2462-cvm-v2`, not `main`.** It sits on top of #1203, which merged into the umbrella.

### The daemon

- **Signed-request auth.** The signed blob is a domain-separated, length-prefixed encoding of method, request target, body, timestamp and nonce, so no two distinct requests share a preimage. Signing the method is what stops a captured `POST /v1/cvm` replaying as `DELETE`.
- **Replay protection** — a persisted nonce window plus a startup floor. The floor is captured once per process and never advances during a lifetime; a per-request floor would reject the second of any two concurrent requests and every client retry. The nonce is fsynced *before* handler dispatch, so a daemon that dies mid-request still blocks the replay.
- **Two scopes** (`validation`, `renter`) across five endpoint classes, with the full 8-cell matrix asserted in `tests/test_scope.py`.
- **Single-parse bodies.** The middleware parses once and passes the object forward via `request.state`. Two parsers over the same bytes need to disagree only once — duplicate keys, unicode escapes, number coercion — for the scope check and the handler to act on different values.
- **Durable state** — tmp write + fsync + `os.replace` + directory fsync, reloaded on start.
- **Packaging** — reproducible tarball (verified: two builds of one commit are byte-identical), hash-pinned `install.sh`, systemd unit.

### The Ansible role

Replaces #1203's stub, which asserted its variables then failed "not implemented". The stub's defining property is kept — **loud in both directions**. Unset, the role says cvmd was not requested and ends. Set, it either produces a daemon answering `/health` or fails the run.

⚠️ **Interface change.** `lium_cvmd_authorized_client_keys` → **`lium_cvmd_authorized_clients`**, now `[{hotkey, scope}]` rather than a flat list whose stub fixture held an SSH public key (`["ssh-ed25519 AAAA fixture"]`). cvmd authenticates against a bittensor hotkey, so an ss58 address paired with a scope is the only thing it can act on. Nothing consumed the old name — setting it always failed the run — so the name and the shape both moved.

The authorized-clients file is validated by **cvmd's own loader**: `template: validate:` runs `load_authorized_clients()` against the candidate file and installs it only on exit 0. That is the same function that refuses at startup, so it does the real base58 checksum, the scope enum and the duplicate check — a file that would leave cvmd unstartable never replaces the working one. Both files are rendered whole from a template, never line-edited.

`cvmd.stub_status` in the verify report becomes **`cvmd.service_state`**, reporting the real unit state and the installed package checksum.

### The bittensor version split

cvmd pins `bittensor==11.0.2` rather than the executor's `9.0.0`, because it runs on the CVM host's own Python (3.13+), not inside the executor image. `tests/fixtures/golden_vector.json` is a blob signed under **9.10.1 — what the validator actually runs** — and verified under 11.0.2 by the suite. That vector is what makes the split safe, so it is committed and never regenerated. Two v11 breaks this codes against: `bittensor.Keypair` is gone (use `bittensor.sp_core.Keypair`), and `verify()` is bytes-only.

## Verification

**Hardware acceptance run on a TDX host** — `au11`, Ubuntu 25.10, kernel 6.17.0-23, `virt/tdx: module initialized`, TDX enabled in BIOS. Full results in the acceptance comment below.

| | |
|---|---|
| cvmd unit suite | **127 passed** |
| Ansible: lint (`production` profile), syntax, yamllint, shellcheck | clean |
| Ansible: 10 shell suites + 9 fixture playbooks | all pass |
| Live daemon over HTTPS, real signed requests | **26/26** |
| Idempotence on hardware | unchanged package → `changed=0`, no restart |
| Wrong checksum on hardware | refused, nothing installed |

**Two defects were found by the hardware run and fixed here** — neither is reachable by the unit suite, because both are about a daemon that is already running:

1. `pip install` resolves by **version, not contents**, so a rebuilt wheel keeping the same version string was "already satisfied" and silently not installed. The venv kept the previous build while the install stamp updated and the health check passed. Fixed with `--force-reinstall` (`f02ca900`).
2. `install.sh` replaces the wheel inside a venv systemd is already running from, so without a restart the host serves the **previous build** while every layer above reports a successful upgrade. The install step now notifies the restart handler (`5b732453`).

**One behaviour was found, documented and pinned rather than changed** (`f22095a1`): the skew window is two-sided, so an accepted future-dated request becomes the persisted floor, and a restart then refuses every caller until wall-clock passes it — at most `skew_seconds`, self-healing. Clamping the floor to `now` republishes that request and defeats the floor's purpose as the backstop for a lost nonce set; narrowing the window to one side rejects any client with a fast clock. Both are protocol changes affecting the DAH-2576/2580 clients, so this wants a reviewer's call, not a drive-by fix.

### ⚠️ This PR runs zero CI

`test.yml` is `on: pull_request: branches: [main, dev]`, and the filter matches the **target** branch. A PR into the umbrella runs no checks — silently, while GitHub still reports `MERGEABLE`. The gate is deferred to the eventual `feature/2462-cvm-v2` → `main` PR. Please do not read an empty check list here as coverage; the evidence is the acceptance comment.

[DAH-2575 — CVM v2 — cvmd service skeleton + signed two-scope API](https://app.notion.com/p/CVM-v2-cvmd-service-skeleton-signed-two-scope-API-3b2b8bfdbde98141b2e0e59f6362a127)

- [x] I wrote tests.


fortunelucky777:
## Hardware acceptance run — TDX host

Run against real Intel TDX hardware, not a simulator. This is the evidence for the PR, because a child PR of the umbrella runs no CI.

```
host      au11 (chutes-miner-tee-au11)
os        Ubuntu 25.10, kernel 6.17.0-23-generic, Python 3.13.7
tdx       virt/tdx: BIOS enabled: private KeyID range [32, 64)
          virt/tdx: module initialized
          /proc/cpuinfo: tdx_host_platform
cmdline   kvm_intel.tdx=on intel_iommu=on iommu=pt
```

The host has a **live production CVM** (1500 GB, 96 vCPU, 12 GPUs passed through, dstack-nvidia-0.5.11). Nothing here went near it. cvmd is host-side userspace bound to `127.0.0.1:8443`; its lifecycle handlers answer `501`, so no CVM was created, no `.env` was read, and no port in the guest's forwarded 32000–32025 range was touched. The role runs under `--tags cvmd`, which selects only the fact collector and the cvmd play — no `kernel`, `gpu` or `qemu`, so no GRUB edit, no GPU rebind and no reboot.

### 1 · The role refuses what it cannot honour

| Case | Result |
|---|---|
| sha256 mismatch (`0000…`) | `get_url` refused, playbook exit **2**, nothing installed |
| Unset `lium_cvmd_package_url` | one line, role ends, exit 0 |

```
The checksum for …/tmp8ttkgfbq did not match 0000000000000000000000000000000000000000000000000000000000000000;
it was b23a06ae5d20ac865c2c5b3d7cf5dc930c4c570edcd4330b47d375d7b5d7757e.
```

### 2 · Install, and idempotence in both directions

```
first install     ok=34  changed=10   →  "cvmd 0.1.0 is serving on 127.0.0.1:8443 with 2 authorised client(s)."
re-run, unchanged ok=25  changed=0    →  "cvmd b23a06ae5d20 is already installed — skipping the download and the venv rebuild."
                                          daemon NOT restarted: Sat 2026-08-08 18:32:12 UTC (unchanged)
changed package   ok=34  changed=7    →  RUNNING HANDLER [cvmd : Restart cvmd]
                                          code installed: 18:32:08   daemon started: 18:32:12
```

The last line is the one that matters: the daemon starts **after** the code it runs was written. Getting there took two fixes — see §5.

### 3 · Live daemon, real signed requests over HTTPS — 26/26

Signed with `bittensor.sp_core.Keypair` from the installed venv, so the signer is the same library as the verifier.

```
PASS  health is unauthenticated and answers                            200
PASS  unsigned /v1/state is refused                                    401
PASS  signed /v1/state by the validation key                           200
PASS  signed /v1/state by the renter key (reads open to either)        200
PASS  an unauthorised hotkey is refused                                401
PASS  a tampered body is refused                                       401
PASS  a POST signature does not replay as DELETE                       401
PASS  a stale timestamp is refused                                     401
PASS  a future timestamp beyond skew is refused                        401
PASS  a short nonce is refused                                         401
PASS  a non-hex signature is refused                                   401
PASS  the first use of a nonce is accepted                             200
PASS  replaying the same nonce is refused                              401
PASS  two requests sharing a timestamp are both accepted          (200, 200)
PASS  out-of-order timestamps from one key are both accepted      (200, 200)
PASS  validation key POST kind=validation -> reaches the handler        501
PASS  renter key     POST kind=validation -> scope violation            403
PASS  renter key     POST kind=renter     -> reaches the handler        501
PASS  validation key POST kind=renter     -> scope violation            403
PASS  renter key     DELETE               -> reaches the handler        501
PASS  validation key DELETE               -> scope violation            403
PASS  an unknown kind is rejected as unprocessable                      422
PASS  a body over the cap is refused before any signature work          413
PASS  the state document survives a restart                            200
PASS  a fresh request after restart is accepted                        200
PASS  a timestamp at or below the startup floor is refused             401
```

The state document read after the restart still carries `entered_at: 2026-08-08T18:22:22.499404+00:00` — the original first-boot timestamp, unchanged across every restart and reinstall in this run.

### 4 · The verify report, generated on the host

```json
{
  "id": "cvmd.service_state",
  "title": "cvmd control daemon",
  "status": "OK",
  "observed": "active, package 2ffcdbb3cf33",
  "expected": "active whenever lium_cvmd_package_url is set",
  "evidence_source": "systemctl is-active cvmd.service",
  "severity": "blocker"
}
```

`2ffcdbb3cf33` is the deployed tarball's checksum, so the report names the build that is running. The host-generated report validates against `schemas/verify-report.schema.json` — the id rename is correct end to end, not only in the golden fixture.

### 5 · Two defects the unit suite cannot reach

Both are about a daemon that is **already running**, which no in-process test has.

**a. `pip install` resolves by version, not contents.** A rebuilt wheel keeping `0.1.0` was "already satisfied" and silently not installed. Caught by checking the installed file rather than trusting the run:

```
=== does the RUNNING code have the new docstring? ===
0                                            ← not installed
installed: 2026-08-08 18:22:18               ← mtime from the PREVIOUS build
```

The tarball checksum differed, Ansible reported `changed`, the install stamp updated, and `/health` passed — while the venv kept the old build. Fixed with `--force-reinstall` (`--no-deps`, so the hash-pinned dependency install is not redone).

**b. The daemon was never restarted after an upgrade.** `install.sh` replaces the wheel inside a venv systemd is already running from. The install step now notifies the restart handler.

Either defect alone means a host that reports a successful upgrade and serves the previous build indefinitely.

### 6 · One behaviour documented and pinned, not changed

A future-dated request **inside** the 60 s skew window is valid, becomes the persisted high-water mark, and is read as the startup floor on the next boot — so a restart refuses every caller until wall-clock passes it. Isolated deliberately:

```
CASE A — only present-time requests before the restart
PASS  a fresh request immediately after restart is accepted            200
PASS  the state document is byte-identical across the restart

CASE B — one request dated 20s ahead before the restart
PASS  a request 20s in the future is accepted (inside the 60s skew)    200
      persisted floor is now +20.0s relative to wall-clock time
PASS  immediately after restart a present-time request is REFUSED      401
PASS  once wall-clock passes it, service resumes with no intervention  200
```

Bounded by `skew_seconds` and self-healing. Not fixed here on purpose: clamping the floor to `now` republishes the accepted request, leaving only the nonce set to refuse it — and the floor exists to be the backstop for a lost or corrupt nonce set. Narrowing the window to one side rejects any client whose clock runs fast. Both are protocol changes that reach the DAH-2576 and DAH-2580 clients, so it wants a decision rather than a drive-by fix. Pinned by `test_an_in_skew_future_timestamp_holds_the_floor_ahead_of_wall_clock` and documented in `auth/replay.py`.

### 7 · Local gate

| Check | Result |
|---|---|
| `pdm run pytest tests/` (cvmd) | **127 passed** |
| `ruff check` + `ruff format --check` | clean, 24 files |
| `ansible-lint` | 0 failures — `production` profile |
| `ansible-playbook --syntax-check site.yml` | pass |
| `yamllint` | clean |
| `shellcheck` — 20 files incl. `cvmd/packaging/*.sh` | clean |
| 10 shell suites | all pass |
| 9 fixture playbooks incl. new `tests/test_cvmd.yml` | all pass |
| `check-jsonschema` on the golden report | pass |
| Tarball reproducibility — two builds of one commit | byte-identical |

### Cleanup

`~/dah2575-acceptance/` on the host holds the tree, tarball and scripts; cvmd is installed at `/opt/cvmd` with state in `/var/lib/cvmd`, bound to loopback only. Say the word and I will remove all of it, or leave it in place for DAH-2576 to build on.

PR #1206, opened 2026-08-08
DAH-2576 - Launch and stop the validation CVM from cvmd, with dstack.py as a library

fortunelucky777:
## Describe your changes

DAH-2576: **cvmd launches and stops the validation CVM**, calling
`neurons/executor/dstacktee/scripts/dstack.py` as a library.

**Base is `feature/2462-cvm-v2`, not `main`.** It sits on top of #1205.

### The launch

```
POST /v1/cvm
{"kind": "validation", "qemu": "9.2.1",
 "os_image_hash": "<64 hex>", "compose_hash": "<64 hex>"}
```

The body names **which software stack to run** and nothing else. Sizing — vCPUs, memory, disk, GPUs, forwarded ports — is provider configuration read from `/etc/cvmd/config.toml`. That was your call on #1205, and it is why `CreateCvmRequest` carries no sizing fields.

The three hashes are a pinned triple, resolved against the local catalog; a triple the catalog does not carry is refused naming the component that was not approved. The request returns when the CVM is up, with the launch report — instance id, forwarded ports, SSH host-key fingerprint — while `/v1/state` moves `RECONCILING → LAUNCHING → VALIDATION_RUNNING`.

`POST {kind: renter}` still answers 501 (DAH-2580 defines that body). `DELETE` is implemented and stays platform-scoped: it carries no `kind`, so a validator holding that right could destroy a **renter's** CVM.

### No behavior fork, and it is checked rather than asserted

cvmd imports the real `dstack.py` and calls the same functions its CLI calls. `tests/test_dstack_library.py` prepares one CVM through cvmd's plan and through `dstack.py new` with the flags `lium-cvm.sh` builds, then compares **every file**. Dropping a single flag from `plan.py` fails it.

Calling the same code makes identical measurements likely. The gate makes them checked: between `setup_instance` writing the VM directory and QEMU starting, cvmd measures what it just built — `sha256(shared/app-compose.json)`, the image's `digest.txt`, `get_qemu_version_string()` — and refuses to launch anything that would not measure as the requested triple. The directory is removed and the node is left exactly where it was. Configuration says what was *intended*; only the files say what was *built*.

### One CVM per node

Checked by reading `/proc` for a running TDX guest, not cvmd's own records — during the rollout the CVM already on a host is most likely one `lium-cvm.sh` started, and cvmd's records cannot see those. Every forwarded port must also be bindable before QEMU starts, which is how a second CVM is stopped from answering on a live one's address.

## Verification

**Hardware acceptance on `au11`** — real Intel TDX, QEMU 9.2.1, `dstack-nvidia-0.5.11`. Full results in the comment below.

| | |
|---|---|
| Measured-byte parity on the host (app-compose.json, 95-line QEMU cmdline, `vm_config`) | **identical** |
| Live launch → teardown → relaunch over signed HTTPS | **37/37** |
| CVM survives a `systemctl restart cvmd` | **11/11** |
| cvmd unit suite | **223 passed** |
| Ansible: lint (`production`), syntax, yamllint, 10 shell suites, 10 fixture playbooks | all pass |

**Three defects were found by the hardware run and fixed here** (`49f4d694`). Each is about a process cvmd started and no longer controls, so none is reachable by an in-process test:

1. **Teardown reported a failure that had not happened, 260s late.** The supervisor was cvmd's direct child and cvmd never calls `wait()`, so a stopped supervisor sat as a zombie — and a zombie is still a process-group member, so `killpg(pgid, 0)` kept succeeding. `os.kill(pid, 0)` has the same blind spot, and that is what `shutdown_instance` waits on. The supervisor now double-forks. Teardown went **260s → 5s** and reports the graceful path.
2. **A cvmd restart killed the node's CVM.** Detaching into its own session is not enough: systemd kills by *cgroup*, and nothing a process does moves it out of its unit's. The journal read `Killing process 218807 (qemu-system-x86) with signal SIGKILL` and the node came back FAILED holding a CVM directory and no CVM. Fixed with `KillMode=process`.
3. **The QEMU command line never reached `console.log`** while the CVM ran — Python block-buffers stdout to a file, so the one line saying what was launched sat in the child's buffer for the VM's whole life. The child now runs with `-u`.

The graceful-poweroff window is now a setting rather than a constant, because a 1.13 TB guest was measured taking 43 minutes to return its memory and a fixed budget would SIGKILL a legitimate slow exit. Per-hardware-class values are DAH-2577's declared open item.

### Re-accepted after review (`a38e11af`)

A careful re-read of the branch found **five more defects**, all on paths the first acceptance did not stage — a lost record, a half-written file, a missing tool, a mismatched port. They are fixed in `a38e11af`, along with eight simplifications (`_exclusively` for the routes' shared lock, `_describe`, `HEX64` moved to `catalog.py`, dead `runtime_pid` / `RUNTIME_FILE` / `Measurements.as_dict` removed).

Because the earlier suite passes either way, each fix was **staged on the hardware** and the old failure asserted against:

| Staged condition | Before | Now |
|---|---|---|
| a stray VM directory with a guest still running under it | the disk was deleted from a live QEMU and success reported | **409**, guest and disk untouched |
| `runtime.json` half-written by a crashed host | dstack raised past the signal ladder — the guest was never signalled at all | falls back to the ladder, torn down in **4s** |
| `ssh-keyscan` absent while `cvm_ssh_guest_port` is set | polled for a fingerprint that could never arrive, then failed the node after the full 600s timeout | **201 in 5s**, null fingerprint, note says why |
| `cvm_ssh_guest_port` that no `cvm_ports` entry forwards | discovered only by timing out | **503 in 0.0s**, naming the setting |
| a malformed `cvm_ports` mapping | valid TOML, so the daemon started looking configured and refused every launch | fatal at startup; the role's `validate:` hook now runs `load_config` |

Re-run on `au11` against the exact bytes of `a38e11af`:

| | |
|---|---|
| Measured-byte parity (app-compose.json, 95-line QEMU cmdline, `vm_config`) | **identical** |
| `e2e_launch.py` | **37/37** |
| `e2e_restart.py` | **11/11** |
| the five fixes, staged on hardware | **39/39** |
| the installed config loader vs 4 malformed mappings + 1 valid | **5/5** |
| converges with a bad SSH port / a bad port range | **both refuse before install**, `changed=0` |
| clean converge afterwards | idempotent, `changed=0` |

The measured triple came back **byte-identical to the pre-review run** — same qemu `9.2.1`, same `os_image_hash`, same `compose_hash` `90d63529…`. That is the evidence the simplifications did not change what cvmd builds.

One caveat, declared here so it is not mistaken for something this PR introduced. The readiness fallback — a plain TCP connect, used when no fingerprint can be obtained — is weak. Measured on one launch: the forwarded port accepts at **1s**, the guest's SSH host key answers at **43s**, because QEMU binds the `hostfwd` listener before the guest boots. So on that path a 201 is returned while the CVM is still booting. It is **not a regression** — that configuration previously failed outright after 600s — but "port accepts connections" is close to vacuous, and the `note` field is the only thing distinguishing it. Worth its own ticket; a caller should not read a 201 with a null fingerprint as proof the guest is up.

### Left for the tasks that own them

- **The catalog is a local file** rendered by the Ansible role. DAH-2578 replaces the loader with the backend's signed-manifest client; only `load_catalog` changes.
- **Teardown confirms the process group is gone, and no more.** The four-condition predicate — VFIO descriptors closed, guest RAM returned, forwarded ports bindable — is DAH-2577, and so is the timeout budget per hardware class.
- **Quote comparison.** The task asks for MRTD/RTMR0-3 equal to a `lium-cvm.sh` launch. Rather than sample two quotes, the acceptance run proves the three artifacts those measurements are computed from are byte-identical — a proof by construction. Reproducing a full quote pair needs a validator in the loop and belongs with DAH-2581.

### ⚠️ This PR runs zero CI

`test.yml` is `on: pull_request: branches: [main, dev]`, and the filter matches the **target** branch, so a PR into the umbrella runs no checks — silently, while GitHub still reports `MERGEABLE`. The gate is deferred to the eventual `feature/2462-cvm-v2` → `main` PR. Please do not read an empty check list here as coverage; the evidence is the acceptance comment.

[DAH-2576 — CVM v2 — cvmd validation-CVM launch path (dstack.py as a library)](https://app.notion.com/p/CVM-v2-cvmd-validation-CVM-launch-path-dstack-py-as-a-library-3b2b8bfdbde981089505c34e85fa0115)

- [x] I wrote tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -242,6 +242,7 @@ jobs:
           ansible-playbook -i inventory/localhost.yml tests/test_sgx_registration.yml
           ansible-playbook -i inventory/localhost.yml tests/test_stubs.yml
           ansible-playbook -i inventory/localhost.yml tests/test_cvmd.yml
+          ansible-playbook -i inventory/localhost.yml tests/test_cvmd_launch.yml
           ansible-playbook -i inventory/localhost.yml tests/test_repo_app_pristine.yml
           ansible-playbook -i inventory/localhost.yml tests/test_reboot_teardown_gate.yml
 
--- a/neurons/executor/ansible/group_vars/all/main.yml
+++ b/neurons/executor/ansible/group_vars/all/main.yml
@@ -306,6 +306,13 @@ lium_cvmd_key_provider_port: 3443
 # the VM.
 lium_cvmd_launch_timeout_seconds: 900
 
+# How long a graceful poweroff is given before cvmd signals the process group.
+# A large-memory TDX guest returns its RAM to the host as it exits, and that is
+# slow — 43 minutes was measured for a 1.13 TB guest on this fleet. Too low and
+# a legitimate slow exit is cut short by a SIGKILL. Per-hardware-class budgets
+# are DAH-2577's open item.
+lium_cvmd_teardown_timeout_seconds: 600
+
 # --- CVM sizing: PROVIDER configuration, never an API field ------------------
 #
 # A cvmd request names which software stack to run; the host decides how big it
--- a/neurons/executor/ansible/group_vars/all/main.yml
+++ b/neurons/executor/ansible/group_vars/all/main.yml
@@ -253,6 +253,97 @@ lium_cvmd_skew_seconds: 60
 # `kind` plus three hashes — hundreds of bytes.
 lium_cvmd_max_body_bytes: 65536
 
+# --- the launch path (DAH-2576) ---------------------------------------------
+#
+# All of the below is optional. cvmd starts and serves /health and /v1/state
+# without it; a launch request on a host that lacks a setting is refused naming
+# that setting. That split is deliberate — a mistyped size should not take the
+# node's state reporting down with it.
+
+# Where dstack.py and host_api.py live on this host. cvmd imports dstack.py as a
+# library rather than shelling out, so this is a directory, not an executable.
+# Typically <repo>/neurons/executor/dstacktee/scripts.
+lium_cvmd_dstack_scripts_dir: ""
+
+# Where cvmd keeps the VM directories it creates. Deliberately NOT dstacktee's
+# own run/vms: a legacy CVM launched by lium-cvm.sh must stay untouched by cvmd
+# during the rollout, and separate directories are what guarantee that.
+lium_cvmd_run_dir: /var/lib/cvmd/vms
+
+# The approved artifact catalog: which software stacks this host may launch.
+# Each entry pins a triple (QEMU build + OS image hash + compose hash) and the
+# local paths that produce it. A request naming a triple that is not here is
+# refused, which is what lets a validator detect a host serving a stack other
+# than the one it expects to attest.
+#
+#   lium_cvmd_catalog:
+#     - id: validation-v3
+#       kind: validation
+#       qemu: "10.1.0"
+#       os_image_hash: a6eafc5f007f642d8ea90c7fa8881f1e6715720ccb531941a28218f4f26d7b02
+#       compose_hash: ab4d14336f0762c0d8ec7631a69148246661de84ceead7a215f8a33b74fd43e6
+#       os_image_path: /opt/lium-io/neurons/executor/dstacktee/run/images/dstack-nvidia-0.5.11
+#       compose_path: /etc/cvmd/composes/validation-v3.yml
+#       init_script: /opt/lium-io/neurons/executor/dstacktee/app/init_script.sh
+#       pre_launch_script: /opt/lium-io/neurons/executor/dstacktee/app/pre_launch_script.sh
+#
+# Hashes must be 64 lowercase hex digits — no `sha256:` prefix, no uppercase.
+# cvmd compares them against values it computes itself, so any other spelling
+# would never match and the launch would fail as "not approved", which sends an
+# operator looking at the wrong thing.
+#
+# This is the DAH-2576 stub. DAH-2578 replaces it with the backend's signed
+# manifest, at which point this list stops being the source of truth.
+lium_cvmd_catalog: []
+
+# dstack's key-provider port, which the guest reaches for its sealing key. The
+# same default lium-cvm.sh relies on.
+lium_cvmd_key_provider_port: 3443
+
+# How long a launch waits for the guest to answer before giving up and failing
+# the node. Generous because a TDX guest with a large memory backing boots
+# slowly; the CVM is left running on timeout, so this bounds the request, not
+# the VM.
+lium_cvmd_launch_timeout_seconds: 900
+
+# --- CVM sizing: PROVIDER configuration, never an API field ------------------
+#
+# A cvmd request names which software stack to run; the host decides how big it
+# is. Settled while reviewing DAH-2575. Everything here is the CVM's shape, so
+# these are the CVM_* values from dstacktee/.env, moved to where the daemon can
+# read them.
+lium_cvm_vcpus: 0
+lium_cvm_memory: ""
+lium_cvm_disk: ""
+
+# PCI slots to pass through, e.g. ["19:00.0", "3b:00.0"], or the single entry
+# ["all"]. An empty list is a legal CVM with no passthrough — so it does not
+# block a launch the way a missing size does.
+lium_cvm_gpus: []
+
+# Forwarded ports as "protocol[:address]:host_port:guest_port". Every one must
+# be bindable before QEMU starts, so a mapping that collides with a CVM already
+# on this host refuses the launch instead of producing two guests fighting over
+# an address.
+lium_cvm_ports: []
+
+# Passed to dstack as --env-file. It lands in the guest's .user-config, which is
+# OUTSIDE app-compose.json — so its contents do not change the compose hash, and
+# two hosts with different env still measure identically.
+lium_cvm_env_file: ""
+
+# The guest-side port carrying SSH. When set, cvmd reads the guest's host key
+# through the matching forward and reports its fingerprint; that read is also
+# what proves the CVM is up, since a TCP accept only proves QEMU is listening.
+# Unset => readiness falls back to a TCP accept and the fingerprint is null.
+lium_cvm_ssh_guest_port: 0
+
+# Both left at what lium-cvm.sh leaves them at: it passes neither flag, so
+# argparse defaults them to false. They change the QEMU command line and
+# therefore the measurements — do not enable one on a single host.
+lium_cvm_pin_numa: false
+lium_cvm_hugepages: false
+
 # catalog variable interface. Unset => the stub says what it is waiting on.
 # Set => the stub asserts its companion then fails "not implemented".
 lium_catalog_manifest_url: ""
--- a/neurons/executor/ansible/roles/cvmd/README.md
+++ b/neurons/executor/ansible/roles/cvmd/README.md
@@ -36,6 +36,7 @@ request.
 | `lium_cvmd_catalog` | The approved artifacts — see below. |
 | `lium_cvmd_key_provider_port` | dstack's key-provider port. `3443`. |
 | `lium_cvmd_launch_timeout_seconds` | How long a launch waits for the guest before failing the node. |
+| `lium_cvmd_teardown_timeout_seconds` | The graceful-poweroff window before cvmd signals the process group. |
 | `lium_cvm_vcpus`, `lium_cvm_memory`, `lium_cvm_disk` | CVM sizing. |
 | `lium_cvm_gpus` | PCI slots to pass through, `["all"]`, or `[]` for none. |
 | `lium_cvm_ports` | Forwarded ports, `protocol[:address]:host:guest`. |
--- a/neurons/executor/ansible/roles/cvmd/README.md
+++ b/neurons/executor/ansible/roles/cvmd/README.md
@@ -41,7 +41,7 @@ request.
 | `lium_cvm_gpus` | PCI slots to pass through, `["all"]`, or `[]` for none. |
 | `lium_cvm_ports` | Forwarded ports, `protocol[:address]:host:guest`. |
 | `lium_cvm_env_file` | Passed as `--env-file`. Lands outside `app-compose.json`, so it does not change the compose hash. |
-| `lium_cvm_ssh_guest_port` | The guest-side SSH port. Set ⇒ cvmd reports the host-key fingerprint and uses reading it as proof the CVM is up. |
+| `lium_cvm_ssh_guest_port` | The guest-side SSH port. Set ⇒ cvmd reports the host-key fingerprint and uses reading it as proof the CVM is up. Must be the guest side (the last field) of one `lium_cvm_ports` entry — the role refuses a value nothing forwards, since there would be nothing to read it through. |
 | `lium_cvm_pin_numa`, `lium_cvm_hugepages` | Both change the QEMU command line and therefore the measurements. `lium-cvm.sh` passes neither. |
 
 **Sizing is provider configuration, never an API field.** A cvmd request names
--- a/neurons/executor/ansible/roles/cvmd/README.md
+++ b/neurons/executor/ansible/roles/cvmd/README.md
@@ -21,6 +21,67 @@ Produce the tarball and its checksum with
 [`cvmd/packaging/build.sh`](../../cvmd/packaging/build.sh), which prints exactly
 the sha256 this role expects.
 
+### The launch path (DAH-2576)
+
+Optional as a whole — cvmd serves `/health` and `/v1/state` without any of it and
+refuses a launch naming the setting it lacks. `lium_cvmd_dstack_scripts_dir` is
+the switch: set it and the role requires the rest, because a host with HALF a
+launch configuration looks configured and refuses every launch at the first
+request.
+
+| Variable | Meaning |
+|---|---|
+| `lium_cvmd_dstack_scripts_dir` | The dstacktee `scripts` directory. cvmd **imports** `dstack.py` from it as a library, so a path to the repo root will not do. |
+| `lium_cvmd_run_dir` | Where cvmd keeps the VM directories it creates. Deliberately not dstacktee's own `run/vms`. |
+| `lium_cvmd_catalog` | The approved artifacts — see below. |
+| `lium_cvmd_key_provider_port` | dstack's key-provider port. `3443`. |
+| `lium_cvmd_launch_timeout_seconds` | How long a launch waits for the guest before failing the node. |
+| `lium_cvm_vcpus`, `lium_cvm_memory`, `lium_cvm_disk` | CVM sizing. |
+| `lium_cvm_gpus` | PCI slots to pass through, `["all"]`, or `[]` for none. |
+| `lium_cvm_ports` | Forwarded ports, `protocol[:address]:host:guest`. |
+| `lium_cvm_env_file` | Passed as `--env-file`. Lands outside `app-compose.json`, so it does not change the compose hash. |
+| `lium_cvm_ssh_guest_port` | The guest-side SSH port. Set ⇒ cvmd reports the host-key fingerprint and uses reading it as proof the CVM is up. |
+| `lium_cvm_pin_numa`, `lium_cvm_hugepages` | Both change the QEMU command line and therefore the measurements. `lium-cvm.sh` passes neither. |
+
+**Sizing is provider configuration, never an API field.** A cvmd request names
+which software stack to run; the host decides how big it is. So there is nothing
+a caller could send to fill a gap here — a host missing a size refuses every
+launch, which is why the role refuses first.
+
+### The catalog pins a triple
+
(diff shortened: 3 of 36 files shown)


fortunelucky777:
## Hardware acceptance run — TDX host

Real Intel TDX, not a simulator. This is the evidence for the PR, because a child PR of the umbrella runs no CI.

```
host      au11 (chutes-miner-tee-au11), Ubuntu 25.10, kernel 6.17.0-23, Python 3.13
qemu      /opt/qemu-dstack/bin/qemu-system-x86_64  9.2.1  (the dstack fork)
image     dstack-nvidia-0.5.11
          digest a6eafc5f007f642d8ea90c7fa8881f1e6715720ccb531941a28218f4f26d7b02
          — the hash already in the validator's TDX_WHITELIST
```

**Isolation.** cvmd's run directory is `/var/lib/cvmd/vms`, never dstacktee's `run/vms`, so this node's production CVM directory was not read, written or removed — `my-executor/hda.img` still carries its pre-run mtime. Forwarded ports were `127.0.0.1:12200` and `127.0.0.1:18001`; nothing in the 32000–32025 range this node's production guest forwards was touched. The test CVM ran a compose with an sshd and nothing else — the executor compose would have registered this node with the subnet.

The production CVM was already stopped when this run began, by `lium-cvm.sh stop my-executor` at 18:23:53, from an interactive shell. Not by this work — see the note at the end.

---

### 1 · Zero measured bytes, proven by construction

The task asks for MRTD and RTMR0-3 equal to a `lium-cvm.sh` launch of the same triple. Rather than sample two quotes, this prepares the same CVM both ways — through cvmd's library call and through `dstack.py new` with the flags `lium-cvm.sh` builds — and compares the three artifacts those measurements are computed **from**. Same sizing as production, GPUs included, nothing launched.

```
=== prepared artifacts ===
  identical   shared/app-compose.json
  identical   shared/.user-config
  identical   vm-manifest.json (minus created_at_ms, id)

=== compose hash (sha256 of app-compose.json) ===
  f590a4f35763386d4cf0e75ad6fa03c00cf84e5a8977c676aa9521566b27ac5c   via-cli
  f590a4f35763386d4cf0e75ad6fa03c00cf84e5a8977c676aa9521566b27ac5c   via-cvmd

=== QEMU command line ===
  identical   95 lines, after normalising the four per-run values

=== vm_config (what the verifier recomputes measurements from) ===
  vm_config IDENTICAL
  {"cpu_count": 16, "hotplug_off": true, "hugepages": false, "image": "dstack-nvidia-0.5.11",
   "memory_size": 68719476736, "num_gpus": 2, "num_nvswitches": 0,
   "os_image_hash": "a6eafc5f…", "ovmf_variant": "pre202505", "pci_hole64_size": 17592186044416,
   "pic": false, "qemu_single_pass_add_pages": false, "qemu_version": "9.2.1", "spec_version": 1}
```

The four normalised values are the VM directory name (it *is* the instance id, and two cvmd launches differ there too), `guest-cid` (randomised per run by dstack.py, a vsock address absent from `vm_config`), the host-API ephemeral port (printed, not a QEMU flag), and `created_at_ms`. Everything else matched byte for byte, including `-machine …confidential-guest-support=tdx`, `-object tdx-guest`, both `vfio-pci` devices, `-append`, `-kernel`, `-initrd` and `-bios`.

### 2 · Launch, teardown, relaunch — 37/37

Live over HTTPS with real signed requests, signed by the installed venv's bittensor so the signer is the verifier's library.

```
--- refusals, before anything is launched ---
PASS  the node starts idle                                          RECONCILING
PASS  no CVM is reported
PASS  a triple the catalog does not carry is refused                422
PASS  an uppercase hash is refused at the edge                      422
PASS  a create with no triple is refused                            422
PASS  the renter key may not create a validation CVM                403
PASS  nothing was launched by any refusal

--- launching a validation CVM ---
PASS  the validator key launches the CVM                            201 in 48s
PASS  the report carries an instance id                472801a2-1964-43a0-b393-43594f3b4a67
PASS  the report carries the forwarded ports                        12200->2200, 18001->8001
PASS  the report carries the guest's SSH host-key fingerprint
      SHA256:npLIxOfv7uM0BBArHVI/FKScEjxyN0ekXcj71AftLrU
PASS  the report carries the measured triple
PASS  the state reached VALIDATION_RUNNING
PASS  /v1/state agrees
PASS  /v1/state describes the running CVM                           pid 218702
PASS  exactly one TDX guest is running on the host

--- one CVM per node ---
PASS  a second launch is rejected                                   409
PASS  the node state is unchanged by the rejection                  VALIDATION_RUNNING
PASS  still exactly one TDX guest

--- the supervisor ---
PASS  the supervisor leads its own session and process group        sid=pgid=218702
PASS  it is not a child of cvmd, so it can never be left a zombie   ppid=1
PASS  QEMU is in the supervisor's process group
PASS  console.log carries the QEMU command line, while running      967 lines
PASS  console.log carries the guest's boot console

--- teardown ---
PASS  the validator key may not tear a CVM down                     403
PASS  the platform key tears the CVM down          200 in 5s  guest powered off on request
PASS  the guest powered off on request, without a signal
PASS  teardown did not burn the signal ladder                       5s
PASS  no zombie was left behind
PASS  no TDX guest remains
PASS  the node is idle again                                        RECONCILING
PASS  the VM directory was removed
PASS  tearing down an idle node is not an error

--- relaunch on the same node ---
PASS  the node relaunches without hand-cleanup                      201 in 51s
PASS  the relaunch measures identically
PASS  it is a new instance, not the old one
```

The refusal for an unapproved triple names the component, not just "not found":

```
compose_hash='eeee…' is not approved for kind='validation'; the catalog allows: 90d63529…
```

### 3 · The CVM outlives cvmd — 11/11

```
PASS  one TDX guest before the restart                              ['224392']
PASS  the guest is the same process after the restart   ['224392'] -> ['224392']
PASS  cvmd came back under it as VALIDATION_RUNNING
PASS  it adopted the same instance rather than relaunching
PASS  the supervisor is still recognised as ours                    pid 224376
PASS  the launch report survives the restart unchanged
PASS  no error is recorded
PASS  the adopted CVM tears down normally           200  guest powered off on request
PASS  no TDX guest remains
PASS  the node is idle
```

### 4 · Three defects the unit suite cannot reach

Each is about a process cvmd started and no longer controls.

**a. Teardown reported a failure that had not happened, 260s late.** The supervisor was cvmd's direct child, and cvmd never calls `wait()`, so a stopped supervisor sat as a zombie:

```
$ ps -eo pid,ppid,stat,comm | awk '$3 ~ /^Z/'
 212756  212142 Zs   python3
```

A zombie is still a process-group member, so `killpg(pgid, 0)` kept succeeding. Every teardown ran the whole signal ladder and ended with

```
{'torn_down': True, 'detail': 'process group 212756 is still present after SIGKILL'}   260s
```

about a guest that had powered off gracefully minutes earlier. The teardown was only *reported* as successful because `is_supervisor()` reads `/proc/<pid>/cmdline`, which is empty for a zombie — luck, not design. `os.kill(pid, 0)` — the obvious implementation, and the one `shutdown_instance` itself waits on — would have reported it failed. The supervisor now double-forks so init reaps it: **260s → 5s**, `guest powered off on request`.

**b. A cvmd restart killed the node's CVM.** Detaching into its own session is not enough — systemd kills by cgroup, and nothing a process does moves it out of its unit's:

```
Aug 08 21:25:23 cvmd.service: Killing process 218807 (qemu-system-x86) with signal SIGKILL
Aug 08 21:25:27 cvmd[218960]: ERROR reconciliation failed: the supervisor for CVM aae51533…
                              (pid 218785) is gone but its directory /var/lib/cvmd/vms/aae51533… remains
```

The node came back FAILED holding a CVM directory and no CVM. Reconciliation was right; the unit was wrong. `KillMode=process` fixes it, and §3 above is the same test passing afterwards. Worth stating plainly: without this, every cvmd upgrade would have taken the node's CVM down with it.

**c. The QEMU command line never reached `console.log`.** `grep -c tdx-guest console.log` returned **0** on a 785-line log of a running guest. Python block-buffers stdout to a file, so the line `run_instance` prints sat in the child's buffer for the VM's whole life while QEMU wrote past it through the same descriptor. `-u` on the child fixes it; the log now carries the command line first and 967 lines of boot console after it.

All three are pinned by `tests/test_supervisor.py`, which drives the real `spawn()` and the real `_detach()`. Reverting the double-fork fails 7 of its tests; removing `-u` fails 1; removing `KillMode=process` fails 1.

### 5 · Local gate

| Check | Result |
|---|---|
| `pytest tests/` (cvmd) | **208 passed** |
| `ruff check` + `ruff format --check` | clean, 42 files |
| `ansible-lint` | 0 failures — `production` profile |
| `ansible-playbook --syntax-check site.yml`, `yamllint` | pass |
| 10 shell suites | all pass |
| 10 fixture playbooks incl. new `tests/test_cvmd_launch.yml` | all pass |
| `check-jsonschema` on the golden report | pass |
| Tarball reproducibility | byte-identical across two builds |

Mutation-checked rather than assumed: disabling the measurement gate fails 3 tests, disabling the port check fails 4, and dropping either `--local-key-provider` or `--pre-launch-script` from `plan.py` fails 4 of the parity tests.

### Cleanup, and one thing to know about the host

The test CVM is torn down, `/var/lib/cvmd/vms` is empty, no TDX guest is running, and 1980 GB of the host's 2011 GB is free. `~/dah2576-acceptance/` holds the tree, tarball and scripts; cvmd remains installed at `/opt/cvmd`, bound to loopback, with a catalog pointing at the test fixtures. Say the word and I will remove it, or leave it as the base for DAH-2577.

**This node's production CVM is down.** It was stopped at `Aug 08 18:23:53` by `lium-cvm.sh stop my-executor`, run from an interactive shell in `~/lium-io/neurons/executor/dstacktee` — before this work started and not by it. Its directory is intact and its `runtime.json` still names pid 56920, long dead, because `lium-cvm.sh stop` gave up waiting and left the file behind. That is the DAH-2577 acceptance bug in the wild, and it is why nothing here trusts a recorded pid.

fortunelucky777:
## Re-acceptance on `au11` — 2026-08-09, against `a38e11af`

Reviewer note first: **this PR shows no checks, and that is the base, not a broken pipeline.** `test.yml` is `on: pull_request: branches: [main, dev]` and filters on the *target* branch, so a PR into `feature/2462-cvm-v2` runs nothing while GitHub still reports `MERGEABLE`. The local suite is the gate until the umbrella → `main` PR. It was run in full and is green; the evidence is below.

`a38e11af` is a third commit added after a line-by-line re-read of the branch. It closes five defects and applies eight simplifications. The earlier acceptance passes either way — none of the five is on a path it stages — so each condition was staged deliberately on the hardware and the *old* failure asserted against.

The tarball installed for this run was built from the exact bytes of `a38e11af` with a clean working tree.

### The daemon under test is the new build

Worth stating explicitly, because a cvmd upgrade can report success at every layer while the host keeps serving the previous build (`pip` resolves by version, not contents; replacing a wheel does not restart systemd — both measured on this host during #1205).

```
code mtime   2026-08-09 16:36:12 +0000
daemon start 2026-08-09 16:36:17 UTC          <- later than the code, as it must be

manager._assert_ssh_port_is_forwarded : 2     sshkey.keyscan_available  : 1
manager._describe                     : 1     routes._exclusively       : 1
catalog.HEX64                         : 2     supervisor.runtime_pid    : 0  (removed)
config ports.parse_all                : 1     measure.as_dict           : 0  (removed)
```

### Parity, unchanged

```
identical   shared/app-compose.json
identical   shared/.user-config
identical   vm-manifest.json (minus created_at_ms, id)
identical   95 lines of QEMU command line, after normalising the four per-run values

compose hash  f590a4f35763386d4cf0e75ad6fa03c00cf84e5a8977c676aa9521566b27ac5c   (both sides)
```

And the triple this host is approved to produce came back **byte-identical to the pre-review run** — qemu `9.2.1`, `os_image_hash a6eafc5f…`, `compose_hash 90d63529…`. That is what says the simplifications did not change what cvmd builds.

### The existing suites

`e2e_launch.py` **37/37**, `e2e_restart.py` **11/11**, both exit 0.

```
PASS  the node relaunches without hand-cleanup  201 in 48s
PASS  the relaunch measures identically
PASS  it is a new instance, not the old one

PASS  the guest is the same process after the restart  ['261446'] -> ['261446']
PASS  cvmd came back under it as VALIDATION_RUNNING
PASS  it adopted the same instance rather than relaunching
```

### The five fixes, staged — 39/39

**1. A CVM whose record cvmd has lost.** `destroy()` used to delete a stray VM directory without asking whether a guest was running under it. "No record" usually means the record naming its pid went missing — so it removed the disk from a live QEMU and reported a teardown that never happened. Staged by taking `instance.json` away while cvmd was stopped:

```
PASS  cvmd came back with no record of it and said so  FAILED
PASS  teardown REFUSES while a guest is still running  409 … holds a CVM cvmd has no record
      of and a confidential guest is still running (pid 262027 (qemu-system-x86_64))
PASS  the guest is still alive — nothing was pulled out from under it  ['262027'] -> ['262027']
PASS  its disk was left in place
PASS  the orphan stops on one signal to its process group
PASS  with the guest gone, teardown removes the directory  200
```

**2. A `runtime.json` a crashed host left half-written.** `shutdown_instance` reads that file and indexes `cid` outside its own `try`, and `run_instance` writes it with a plain `open()`. The exception propagated past cvmd's signal ladder — so the guest was never signalled at all, the one outcome teardown exists to prevent.

```
PASS  teardown still succeeds  200 in 4s  process group stopped by SIGTERM
PASS  cvmd logged the fallback rather than propagating
PASS  no TDX guest remains
```

**3. A host with no `ssh-keyscan`.** `read_host_key` returns `None` when the tool is absent, which is indistinguishable from "the guest is not up yet". The loop polled for a fingerprint that could never arrive and failed a booted CVM after the full 600s timeout. The lookup is now made once, before the loop.

```
PASS  the launch still succeeds  201 in 5s
PASS  and returns promptly, not after the 600s launch timeout  5s
PASS  the report is honest about having no fingerprint  None
PASS  and says why readiness was weaker  port accepts connections; ssh-keyscan is not installed
```

**4. A `cvm_ssh_guest_port` nothing forwards.** Readiness and the report's fingerprint are both read through that forward; with no match there is nothing to probe.

```
PASS  cvmd starts on the config — this is not a startup refusal
PASS  the launch is refused  503 in 0s  this host's cvm_ssh_guest_port is 2222, which none of
      its cvm_ports forwards (tcp:127.0.0.1:12200:2200, tcp:127.0.0.1:18001:8001)
PASS  it is refused immediately, not after a timeout  0.0s
PASS  nothing was launched
```

**5. A malformed `cvm_ports` mapping.** `LaunchConfig` promises malformed values are fatal at startup, and every numeric setting honoured that — but `cvm_ports` was only coerced to strings, never parsed. So `tcp:0.0.0.0:70000:22` was valid TOML, the daemon started looking configured, and the first launch got a 503. The role had felt the gap and approximated it with `[0-9]{1,5}`, which accepts both `0` and `70000` — drifted from the Python on day one.

Run against the **installed** loader, which is what the role's `validate:` hook now executes:

```
tcp:12200                     REFUSED: `cvm_ports`: 'tcp:12200' is not 'protocol[:address]:from:to'
tcp:0.0.0.0:0:22              REFUSED: `cvm_ports`: … has a port outside 1-65535
tcp:0.0.0.0:70000:22          REFUSED: `cvm_ports`: … has a port outside 1-65535
sftp:0.0.0.0:2200:2200:2200   REFUSED: `cvm_ports`: … is not 'protocol[:address]:from:to'
a valid pair                  loaded
```

And the role now refuses both cases before it downloads anything:

```
bad SSH port    fatal at "The SSH guest port is one of the forwarded ports"   changed=0
bad port range  fatal at "Each forwarded port is … with ports in 1-65535"     changed=0
clean converge afterwards                                                     changed=0  (idempotent)
```

Not staged here: the `_settle()` illegal-edge fix. Reaching `destroy()`'s stray branch on a real host always goes through startup reconciliation first, which lands the node in `FAILED` — and `FAILED → RECONCILING` is a legal edge, so the old code survived it. The unreachable-on-hardware cases are covered by `TestSettling`.

### One caveat, declared

The readiness fallback — a plain TCP connect, used when no fingerprint can be obtained — is weaker than it reads. Measured on a single launch:

```
POST /v1/cvm returned 201 after            59 s
the forwarded port first accepted at        1 s
the guest's SSH host key first answered at 43 s
```

QEMU binds the `hostfwd` listener before the guest boots, so "port accepts connections" is satisfied ~42s early and cvmd reports the CVM running while it is still booting. **Not a regression** — that configuration previously failed outright after 600s, so the fallback is strictly better — but it is close to vacuous, and the `note` field is the only thing distinguishing it. It deserves its own ticket. A caller should not read a 201 with a null `ssh_host_key_fingerprint` as proof the guest is up.

### Host left clean

cvmd active with `KillMode=process`, node `RECONCILING`, no VM directories, no TDX guest, no recorded error. `dstacktee/run/vms/my-executor` was never read or written, and nothing in the run went near ports 32000–32025 or the production `.env`.

PR #1207, opened 2026-08-09
DAH-2577 - Verify a teardown released the node, and give the switch its own state

fortunelucky777:
## Describe your changes

DAH-2577: **a teardown reports complete only when the node's hardware is actually free**, and
the crossing between modes gets its own state.

**Base is `feature/2462-cvm-v2`, not `main`.** It sits on top of #1206.

A stopped CVM is not a free node. `lium-cvm.sh stop` reports success when its wait loop gives
up and QEMU is still holding the guest's memory afterwards — that is the acceptance bar the
task names. DAH-2576 could only confirm the process group was empty, which is necessary and
nowhere near sufficient, because the expensive part of a TDX teardown happens *after* the
process exits.

### The four conditions (FR-C6)

`DELETE /v1/cvm` now returns `200` only when all four hold **in one evaluation** — a condition
that held a minute ago is not evidence now. They are read off the host, never off cvmd's own
records:

| | |
|---|---|
| `process_reaped` | the supervisor's process group has no members left, **zombies included**, and no TDX guest runs anywhere on this host |
| `vfio_released` | nothing holds a descriptor under `/dev/vfio`. A device stays bound to `vfio-pci` across a teardown by design, so the driver says nothing about availability; an open group descriptor says exactly that |
| `memory_returned` | the host has the guest's RAM back, and its hugepages are back in the pool |
| `ports_free` | every port this CVM forwarded can be bound again — the same check the launch path makes, so "torn down" and "launchable" cannot disagree |

`DELETE /v1/cvm` **is** the FRD's `/teardown` endpoint. The API shape was settled in
DAH-2575/2576 and adding a second route that does the same thing would be worse than reusing
the one that already exists.

### The switch window (FR-C7, FR-I3)

`SWITCHING` is the state between the process going and the hardware coming back. `TEARDOWN` is
cvmd still stopping the CVM; both are the window FR-C7 calls "switching" — neither rentable nor
reachable, and not offline either. **`SWITCHING` has no edge to `LAUNCHING`**, which is how
FR-C6 is enforced structurally rather than by remembering to check it.

The first moment each condition holds is recorded, so the report says *which* part was slow
rather than only how long the whole crossing took — that per-condition breakdown is what FR-I3
needs to tell a switching node from an offline one. `/v1/state` reports the last switch under
`last_switch`, present whether or not a CVM is.

```
DELETE /v1/cvm  ->  200
{"torn_down": true, "instance_id": "...", "detail": "guest powered off on request",
 "switch": {"outcome": "released", "complete": true, "duration_seconds": 0.2,
            "conditions": {"process_reaped": {"satisfied_after_seconds": 0.2, "detail": "..."},
                           ...}}}
```

A teardown that stops the CVM but leaves the node holding hardware answers **504** naming the
conditions that did not hold, fails the node, and **keeps the VM directory** — which is what
`_assert_node_is_free` reads, so the next launch is refused rather than merely discouraged.

The switch record is written **before the first signal**, carrying the memory baseline. A
repeat teardown continues the crossing it finds unfinished rather than opening a new one: the
baseline is only meaningful taken while the guest was still running, and the window FR-I3
budgets for began at the platform's first call, not at whichever retry worked.

### Two budgets, because they measure two different things

`teardown_timeout_seconds` bounds how long the **guest** is given to power itself off.
`teardown_verify_timeout_seconds` (new, default 1800) bounds how long the **host** then takes to
return the hardware. They are not proportional — a guest that exits in seconds can leave its
memory draining for tens of minutes.

`teardown_memory_tolerance` (new, default 0.9) is a tolerance on a kernel estimate, not a
licence to launch into memory that has not come back. Above 1.0 no teardown could ever complete,
so the loader and the Ansible role both refuse it.

## Verification

**Hardware acceptance on `au11`** — real Intel TDX, QEMU 9.2.1, `dstack-nvidia-0.5.11`. Full
results in the comment below.

| | |
|---|---|
| Measured-byte parity with `dstack.py new` (app-compose.json, 95-line QEMU cmdline, `vm_config`) | **identical** |
| `e2e_launch.py` (DAH-2576's suite, unchanged) | **37/37** |
| `e2e_restart.py` (DAH-2576's suite, unchanged) | **11/11** |
| DAH-2577 fault-injection matrix, five blocks | **57/57** |
| cvmd unit suite | **275 passed** |
| Ansible: lint (`production`), syntax, yamllint, shellcheck, 10 shell suites, 11 fixture playbooks | all pass |

The triple this host produces came back **byte-identical to the DAH-2576 run** — qemu `9.2.1`,
`os_image_hash a6eafc5f…`, `compose_hash 90d63529…`. That is the evidence this change did not
alter what cvmd builds.

### The faults, staged on the hardware

| Staged condition | Before | Now |
|---|---|---|
| a **zombie** in the recorded supervisor's process group | `is_supervisor()` reads a zombie as gone — it has no `cmdline` — so teardown answered `200 {"torn_down": true}` | **504** `process group N still has pid N (python, zombie)`, disk kept, node FAILED |
| a `/dev/vfio` descriptor held by another process | nothing looked, so the GPUs could be handed to the next CVM while still held | **504** naming the pid and `/dev/vfio/101`; a launch into that node is refused **409** |
| a forwarded port taken after the CVM stopped | nothing looked | **504** naming the port |
| a **SIGSTOPped** supervisor | — | the ladder gets past it and teardown completes in 21s, and nothing reported the node free while the process was still there |

The zombie case is the headline, and it is the DAH-2576 bug reproduced deliberately: a zombie is
still a process-group member, so `killpg(pgid, 0)` keeps succeeding, while `is_supervisor()`
reads it as gone. The old teardown reported success on exactly that state.

### Measured switch windows

The per-hardware-class budget the task asks for, recorded in `roles/cvmd/README.md`:

| Host | Guest | Guest powered off | Hardware back afterwards | `DELETE` end to end |
|---|---|---|---|---|
| au11 | 8 vCPU, 16 GiB, no GPU passthrough | ~5 s | **0.2 s** | **5–6 s** |
| the same fleet under `lium-cvm.sh` | 1.13 TB | — | ~43 min | — |

## Declared, not hidden

- **One unexplained outlier.** A single au11 teardown held on `ports_free` for **161 s** while
  the other three conditions settled in 0.2 s. It did not recur across three later runs of the
  same shape, and a probe across a teardown found the forwarded ports bindable 2 s after the
  guest stopped — with *and* without `SO_REUSEADDR`, so it was not `TIME_WAIT`. I have not
  explained it, so it is recorded rather than explained away. `verify_released` now logs its
  outstanding conditions every 30 s, which is exactly what was missing to diagnose it at the
  time (`bedf14a6`).
- **The SIGSTOP fault cannot end in a timeout** on a host where cvmd is root and the
  supervisor's parent is init: SIGKILL always reaches a stopped process, and init reaps it at
  once. What that block proves is the other half — nothing reported the node free while the
  process was there. The zombie block is the one that reaches FAILED.
- **The zombie and port blocks stage a supervisor record rather than a real guest.** Not a
  shortcut: a zombie inside a real supervisor's group cannot outlive the SIGKILL that ends the
  group, because the kernel reparents it to init, which reaps it. The only way to hold that
  condition open is the shape the original bug had — a group leader whose parent is alive and
  outside the group.
- **`ports_free` uses the launch path's own check.** Deliberate: if the release check were more
  lenient than the launch check, a teardown could report the node free while the next launch was
  refused for the same ports.
- **The 1.13 TB figure is from DAH-2544's observation under `lium-cvm.sh`**, not from a run of
  this code. au11 has no TB-class guest configured, and sizing is provider configuration.
- **The cost of looking.** Two of the conditions scan `/proc`. Measured on au11 — 1601 processes
  holding 962 open descriptors — one full evaluation takes **60 ms**, so a 2 s poll is ~3% of one
  core, and only while a teardown is in flight. Measured rather than assumed, because a scan of
  every descriptor of every process is the kind of thing that is cheap until it is not.

[DAH-2577 — CVM v2 — verified teardown + SWITCHING state (FR-C6/C7)](https://app.notion.com/p/CVM-v2-verified-teardown-SWITCHING-state-FR-C6-C7-3b2b8bfdbde98194bbc4ef03bc663b78)

## ⚠️ This PR runs zero CI

`test.yml` is `on: pull_request: branches: [main, dev]`, and the filter matches the **target**
branch, so a PR into the umbrella runs no checks — silently, while GitHub still reports
`MERGEABLE`. The gate is deferred to the eventual `feature/2462-cvm-v2` → `main` PR. Please do
not read an empty check list here as coverage; the evidence is the acceptance comment.

- [x] I wrote tests.

### DIFF
--- a/neurons/executor/ansible/group_vars/all/main.yml
+++ b/neurons/executor/ansible/group_vars/all/main.yml
@@ -309,9 +309,25 @@ lium_cvmd_launch_timeout_seconds: 900
 # How long a graceful poweroff is given before cvmd signals the process group.
 # A large-memory TDX guest returns its RAM to the host as it exits, and that is
 # slow — 43 minutes was measured for a 1.13 TB guest on this fleet. Too low and
-# a legitimate slow exit is cut short by a SIGKILL. Per-hardware-class budgets
-# are DAH-2577's open item.
+# a legitimate slow exit is cut short by a SIGKILL.
 lium_cvmd_teardown_timeout_seconds: 600
+
+# How long the node's hardware then gets to come back: QEMU reaped with no
+# zombie, every /dev/vfio descriptor closed, the guest's memory returned and the
+# forwarded ports bindable. A teardown that runs out of this budget fails the
+# node naming the condition that did not hold, rather than reporting a success
+# the next launch would discover was false.
+#
+# A separate budget from the one above because it measures a different thing —
+# that one bounds how long the GUEST is given to power itself off, this one
+# bounds how long the HOST takes to get its hardware back afterwards. The two
+# are not proportional. See roles/cvmd/README.md for the measured values.
+lium_cvmd_teardown_verify_timeout_seconds: 1800
+
+# How much of the guest's configured memory must be back before the node counts
+# as free. Below 1.0 because MemAvailable is the kernel's own estimate and reads
+# a little short even on a host holding nothing.
+lium_cvmd_teardown_memory_tolerance: 0.9
 
 # --- CVM sizing: PROVIDER configuration, never an API field ------------------
 #
--- a/neurons/executor/ansible/roles/cvmd/README.md
+++ b/neurons/executor/ansible/roles/cvmd/README.md
@@ -37,6 +37,8 @@ request.
 | `lium_cvmd_key_provider_port` | dstack's key-provider port. `3443`. |
 | `lium_cvmd_launch_timeout_seconds` | How long a launch waits for the guest before failing the node. |
 | `lium_cvmd_teardown_timeout_seconds` | The graceful-poweroff window before cvmd signals the process group. |
+| `lium_cvmd_teardown_verify_timeout_seconds` | How long the node's hardware then gets to come back — see below. |
+| `lium_cvmd_teardown_memory_tolerance` | How much of the guest's configured memory must be back before the node counts as free. |
 | `lium_cvm_vcpus`, `lium_cvm_memory`, `lium_cvm_disk` | CVM sizing. |
 | `lium_cvm_gpus` | PCI slots to pass through, `["all"]`, or `[]` for none. |
 | `lium_cvm_ports` | Forwarded ports, `protocol[:address]:host:guest`. |
@@ -48,6 +50,34 @@ request.
 which software stack to run; the host decides how big it is. So there is nothing
 a caller could send to fill a gap here — a host missing a size refuses every
 launch, which is why the role refuses first.
+
+### The two teardown budgets
+
+They measure different things, which is why there are two.
+
+`lium_cvmd_teardown_timeout_seconds` bounds how long the **guest** is given to
+power itself off before cvmd signals its process group.
+`lium_cvmd_teardown_verify_timeout_seconds` bounds how long the **host** then
+takes to get its hardware back: QEMU reaped with no zombie left in the group,
+every `/dev/vfio` descriptor closed, the guest's memory returned, and the
+forwarded ports bindable again. A teardown reports success only when all four
+hold together; running out of the second budget fails the node naming the
+condition that did not.
+
+The two are not proportional. A guest that exits in seconds can leave its memory
+draining for tens of minutes, because that work happens in the kernel after the
+process is gone — which is exactly why confirming the process group is empty was
+never enough.
+
+Measured switch windows, for setting the budget per hardware class:
+
+| Host | Guest | Process reaped | Node fully released |
+|---|---|---|---|
+| au11 (Intel TDX, dstack-nvidia-0.5.11) | 16 GiB, no GPU passthrough | _measured during acceptance_ | _measured during acceptance_ |
+| — | 1.13 TB | — | ~43 min (observed under `lium-cvm.sh`, DAH-2544) |
+
+The default of 1800 s is sized for the ordinary case with room, not for the
+largest guests on the fleet. A host running 1 TB-class CVMs needs its own value.
 
 ### The catalog pins a triple
 
--- a/neurons/executor/ansible/roles/cvmd/README.md
+++ b/neurons/executor/ansible/roles/cvmd/README.md
@@ -69,15 +69,29 @@ draining for tens of minutes, because that work happens in the kernel after the
 process is gone — which is exactly why confirming the process group is empty was
 never enough.
 
-Measured switch windows, for setting the budget per hardware class:
+Measured switch windows, for setting the budget per hardware class. The two
+columns are what the two budgets bound, so read them against the settings above.
 
-| Host | Guest | Process reaped | Node fully released |
-|---|---|---|---|
-| au11 (Intel TDX, dstack-nvidia-0.5.11) | 16 GiB, no GPU passthrough | _measured during acceptance_ | _measured during acceptance_ |
-| — | 1.13 TB | — | ~43 min (observed under `lium-cvm.sh`, DAH-2544) |
+| Host | Guest | Guest powered off | Hardware back afterwards | `DELETE` end to end |
+|---|---|---|---|---|
+| au11 — Intel TDX, QEMU 9.2.1, `dstack-nvidia-0.5.11` | 8 vCPU, 16 GiB, no GPU passthrough | ~5 s | **0.2 s** | **5–6 s** |
+| the same fleet, under `lium-cvm.sh` | 1.13 TB | — | ~43 min | — |
 
-The default of 1800 s is sized for the ordinary case with room, not for the
-largest guests on the fleet. A host running 1 TB-class CVMs needs its own value.
+On a small guest the hardware is back before the first evaluation finishes: the
+process, the VFIO groups, the memory and the ports were all released 0.2 s after
+the stop, across three runs. The 43-minute figure is the other end of the range
+and the reason the default is 1800 s rather than a minute — it is sized for the
+ordinary case with room, not for the largest guests on the fleet. **A host
+running TB-class CVMs needs its own value**, and the measurement to set it from
+is the `memory_returned` timing in its own `last_switch` report.
+
+One outlier is worth knowing about: a single au11 teardown held on `ports_free`
+for 161 s while the other three conditions settled in 0.2 s. It did not recur in
+three later runs of the same shape, and a probe of the forwarded ports across a
+teardown showed them bindable 2 s after the guest stopped — with and without
+`SO_REUSEADDR`, so it was not `TIME_WAIT`. It is recorded rather than explained.
+`verify_released` now logs what it is still waiting for every 30 s, which is what
+was missing to diagnose it at the time.
 
 ### The catalog pins a triple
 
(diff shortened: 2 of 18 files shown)


fortunelucky777:
## Hardware acceptance on `au11` — 2026-08-09, against `bedf14a6`

Reviewer note first: **this PR shows no checks, and that is the base, not a broken pipeline.**
`test.yml` is `on: pull_request: branches: [main, dev]` and filters on the *target* branch, so a
PR into `feature/2462-cvm-v2` runs nothing while GitHub still reports `MERGEABLE`. The local
suite plus this run is the gate until the umbrella → `main` PR.

Real Intel TDX, QEMU 9.2.1, `dstack-nvidia-0.5.11`, 8 vCPU / 16 GiB guest, no GPU passthrough,
ports 12200/18001 on loopback. `dstacktee/run/vms/my-executor` was never read or written, and
nothing went near 32000–32025 or the production `.env`.

### The daemon under test is the new build

Worth stating, because a cvmd upgrade can report success at every layer while the host keeps
serving the previous build — `pip` resolves by version, not contents, and replacing a wheel does
not restart systemd. Both measured on this host during #1205.

```
code mtime   2026-08-09 19:02:07 +0000
daemon start 2026-08-09 19:02:11 UTC        <- later than the code, as it must be

installed cvm/: release.py and switching.py present; PROGRESS_SECONDS present
/etc/cvmd/config.toml: teardown_verify_timeout_seconds = 1800, teardown_memory_tolerance = 0.9
```

### Nothing about what cvmd builds changed

```
identical   shared/app-compose.json
identical   shared/.user-config
identical   vm-manifest.json (minus created_at_ms, id)
identical   95 lines of QEMU command line, after normalising the four per-run values

compose hash  f590a4f35763386d4cf0e75ad6fa03c00cf84e5a8977c676aa9521566b27ac5c   (both sides)
```

And the triple this host is approved to produce came back **byte-identical to the DAH-2576
run** — qemu `9.2.1`, `os_image_hash a6eafc5f…`, `compose_hash 90d63529…`.

`parity.sh` exits 2 on a cosmetic bug of its own: the *display* block after the comparison runs
`grep -E -- '<pattern>' -A1 <file>`, and grep reads `-A1` as a filename. It is a host-side
scratch harness, not a repo file, and every comparison above ran and passed.

### DAH-2576's suites, unchanged: 37/37 and 11/11

Both exit 0 against this build.

### The switch window, measured — 5–6 s end to end

```
DELETE returned after               6.1 s
verification alone                  0.1 s
  process_reaped        0.1 s   no process group members and no TDX guest
  vfio_released         0.1 s   nothing holds a /dev/vfio descriptor
  memory_returned       0.1 s   MemAvailable 1980.8 GiB of the 14.4 GiB this guest needs
  ports_free            0.1 s   all 2 forwarded port(s) can be bound again
MemAvailable 1978.5 -> 1980.8 GiB
```

Consistent across three runs (5.2 s, 6 s, 6.1 s). On a guest this size the hardware is back
before the first evaluation finishes; the ~5 s is the guest powering itself off. That is what
sets the per-hardware-class budget, and it is why the two budgets are separate settings — the
same fleet's 1.13 TB guest took ~43 min to return its memory.

Observed live, in order: `VALIDATION_RUNNING -> TEARDOWN -> SWITCHING -> RECONCILING`, and
`transition_id` advanced by exactly 3.

### The faults, staged — 57/57

**1. A zombie in the recorded supervisor's process group.** The headline, and the DAH-2576 bug
reproduced deliberately: a zombie is still a group member so `killpg(pgid, 0)` keeps succeeding,
while `is_supervisor()` reads it as gone because a zombie has no `cmdline`.

```
PASS  pid 295020 is a zombie  295020 (python) Z 295019 295020 295020 0
PASS  with an empty cmdline, so cvmd reads its supervisor as gone
PASS  cvmd reconciles to FAILED
PASS  teardown REFUSES to report complete  504 in 90s
      process_reaped: process group 295020 still has pid 295020 (python, zombie)
PASS  the disk is kept — this node has not finished giving anything back
PASS  with its parent gone, init reaps the zombie
PASS  and the repeat teardown completes  200
```

**Before this change the same state answered `200 {"torn_down": true}`.**

**2. A `/dev/vfio` descriptor held by a process cvmd did not start.**

```
PASS  a process is holding /dev/vfio/101  pid 294882
PASS  teardown REFUSES to report complete  504 in 95s
PASS  and the node reads as SWITCHING for the whole window
      VALIDATION_RUNNING -> TEARDOWN -> SWITCHING -> FAILED
PASS  vfio_released: pid 294882 (python) holds /dev/vfio/101
PASS  the VM directory is KEPT — the node still owes hardware
PASS  a launch into the un-released node is refused  409
PASS  with the descriptor closed, teardown completes  200
PASS  and continues the SAME switch window rather than restarting the clock
```

**3. A SIGSTOPped supervisor**, detached with init as its parent exactly as the real one is.

```
PASS  its parent really is init, as the real supervisor's is  ppid 1
PASS  pid 301752 is stopped  301752 (python) T 1 301752 301752
PASS  the predicate refuses to call the node free
      process group 301752 still has pid 301752 (python)
PASS  the signal ladder gets past SIGSTOP, so teardown completes  200
PASS  and the process really is gone, not a zombie
PASS  reported complete only after that, 21s in  process group stopped by SIGKILL
```

SIGKILL always reaches a stopped process and init reaps it at once, so this fault cannot end in
a timeout on a host where cvmd is root. What it shows is the other half: nothing reported the
node free while the process was still there.

**4. A forwarded port taken after the CVM stopped.**

```
PASS  teardown REFUSES to report complete  504
      ports_free: these forwarded ports are already in use on this host: tcp:127.0.0.1:12200:2200
PASS  with the port released, teardown completes  200
```

### The progress log, live

Added in `bedf14a6` after the run below, and here it is doing its job on the hardware:

```
19:09:00 INFO cvmd.cvm.release 30s after the stop this node is still held by
              vfio_released: pid 301462 (python) holds /dev/vfio/101
19:10:35 INFO cvmd.cvm.release 31s after the stop this node is still held by
              process_reaped: process group 301603 still has pid 301603 (python, zombie)
```

### One outlier I did not explain

The **first** acceptance run had a teardown hold on `ports_free` for **161 s** while the other
three conditions settled in 0.2 s. I did not want to fix that on a hypothesis, so I measured it:

```
                                          run 1    run 2    run 3    run 4
ports_free satisfied after                161.4s     0.2s     0.1s     0.1s
```

A dedicated probe sampled a real teardown once a second, comparing a bind **without**
`SO_REUSEADDR` (what `cvm/ports.py` does) against one **with** it (what QEMU's own `inet_listen`
does), plus every socket whose *local* address was the forwarded port:

```
   0s  plain=False reuse=False  1xLISTEN
   2s  plain=True  reuse=True   none
```

Both binds succeeded at the same moment, 2 s after the guest stopped, and nothing was in
`TIME_WAIT` on the local address. So my `TIME_WAIT` hypothesis was **wrong**, and the change it
would have justified — loosening the bind probe — would have been a fix for a cause that does
not exist. It is recorded in `roles/cvmd/README.md` as an unexplained observation. The progress
log above is the part I *could* act on: it is what was missing to diagnose it while it was
happening.

### The cost of looking

Two conditions scan `/proc`. On this host — 1601 processes holding 962 open descriptors — one
full evaluation takes **60 ms**, so a 2 s poll is ~3% of one core and only while a teardown is in
flight.

### Host left clean

cvmd active on its own config with the test drop-in removed, node `RECONCILING`, no VM
directories, no TDX guest, no recorded error.

PR #1208, opened 2026-08-09
DAH-2578 - Fetch the approved catalog as a signed manifest

fortunelucky777:
## Describe your changes

DAH-2578, host side: **the catalog a host launches from is now a document the platform
signed**, not a JSON file Ansible wrote whose integrity rested on its filesystem
permissions.

**Base is `feature/2462-cvm-v2`, not `main`.** It sits on top of #1207.

Until now the approved set lived in each host's Ansible variables. That means a revoked
image stays launchable on every host nobody remembered to edit — and nobody can tell from
the platform which hosts are still carrying it.

### The four checks that make a manifest evidence

`cvmd/catalog/` replaces the single loader with four modules split by what each is allowed
to trust: `artifacts.py` (what an approved stack is — pure), `manifest.py` (the signed
document and every check on it), `store.py` (this host's copy), `client.py` (getting the
bytes, trusting nothing about them).

| | |
|---|---|
| **signer** | Verified against the ss58 the **host** was configured with, never the one the document carries. A document checked against its own claimed signer proves only that somebody owns a key. |
| **freshness** | `expires_at` must be in the future, or a revocation is defeated by simply never delivering the next manifest. |
| **serial** | Monotonic. Without it a validly signed *older* manifest replays a revoked artifact back into the catalog. |
| **floors** | Monotonic too. The floor is the backend's version ratchet; accepting a lower one would undo it remotely. |

A manifest is accepted as a whole or not at all. Dropping the entries that look wrong and
keeping the rest would leave a host launching from a document nobody wrote.

### The measured inputs travel as content

The compose and both guest scripts are carried **in** the manifest, not referenced by
path. cvmd writes them under `state_dir` on every launch and puts back anything edited on
the host — so an operator who adds a service to the compose has their edit reverted before
it can be measured, rather than discovered afterwards by the attestation.

The OS image is the exception because it is gigabytes: an entry names a directory under
`catalog_images_dir`, and the existing measurement gate checks that directory's own
`digest.txt` against the pinned hash before QEMU starts.

### Two new routes

`GET /v1/catalog` reports what this host is approved to launch and why, answering **200
even when the catalog is unusable** — a node that will not launch is exactly when someone
needs the reason, and a 503 would only say something is wrong with a daemon that is
plainly answering.

`POST /v1/catalog/refresh` fetches now instead of waiting for the timer. It grants no
authority over *what* the host runs: the bytes go through the same checks either way.

### Ansible

The `catalog` role stops being a stub and stages the seed manifest, validated through
cvmd's own `parse_manifest`. It moves into the `cvmd` play — the seed is cvmd
configuration, and handlers do not cross a play boundary, so a borrowed one would make the
role runnable only in the exact play order `site.yml` happens to use.

`catalog.manifest` in the verify report is a **blocker** once cvmd is requested, read
through cvmd's verifier against cvmd's configured signer. `observed` is `usable`,
`expired`, `none` or `error`, with the serial and expiry — which is what tells an operator
whether a revocation has reached this host yet.

## Verification

| | |
|---|---|
| cvmd unit suite | **327 passed** (275 before) |
| Ansible lint (`production`), syntax, yamllint, shellcheck | all pass |
| Ansible shell suites | **10/10** |
| Ansible fixture playbooks | **10/10** |
| Cross-repo, against the real backend | **13/13** |
| **Hardware acceptance on au11 (real TDX)** | **22/22** — see below |

### Cross-repo, because two implementations agreeing in one repo is not evidence

The real backend catalog service (Datura-ai/lium-io-backend#882, Python 3.11 venv) signing,
and this client (3.13 venv) verifying, over HTTP. Nothing shared but the wire format.

```
1. an empty backend catalog is a valid, usable manifest      2/2
2. approving in the backend reaches cvmd on the next fetch   3/3
3. a NEW image hash appears on the next fetch                1/1   <- the task's E2E case
4. one flipped byte is refused, last good copy kept          2/2   <- the task's tamper case
5. a manifest signed by another key is refused               1/1
6. revoking in the backend removes it from this host         1/1
7. raising the backend floor retires a generation            1/1
8. an expired manifest refuses a launch, and says why        2/2
```

### Hardware acceptance on `au11` — 2026-08-09, real Intel TDX

**22/22.** Real TDX, QEMU 9.2.1, `dstack-nvidia-0.5.11`, 8 vCPU / 16 GiB guest, no GPU
passthrough, ports 12200/18001 on loopback. `dstacktee/run/vms/my-executor` was never read
or written, and nothing went near 32000–32025 or the production `.env`.

The headline is that the manifest's `compose_hash` was **derived by DAH-2579's backend
module and never measured**, and the manifest itself was **signed by DAH-2578's backend
service** — so one launch proves both:

```
pinned compose_hash 2704baf19789533e1b5e60f719112baccd68690b4588a894b0f307d91f1b51b3

2. THE HEADLINE: a real TDX launch from the derived hash
   PASS  the measurement gate accepted the DERIVED hash          HTTP 201 in 48s
   PASS  what was measured IS what was pinned
   PASS  and the OS image measured as pinned
   PASS  the guest booted                                        SHA256:aiAv2LiFE4u4LGk3y8Ujh
   PASS  the node is VALIDATION_RUNNING
   PASS  the compose came from the manifest
   PASS  the compose on disk is the manifest's, byte for byte

3. teardown returns the node                DELETE 200 in 6s, four conditions in 0.2s
4. a tampered manifest is refused, the good one stays            still serial 10
5. a manifest signed by another key is refused
6. an older serial is refused as a rollback                      still serial 10
7. a newly approved image reaches this host                      serial 11, 2 images
8. revocation removes it again                                   serial 12, 1 artifact
9. an expired manifest is refused at INSTALL, not at use
10. an expired CACHED manifest refuses a launch, and says why
```

The daemon under test is the new build — code mtime 21:15:26, `ExecMainStartTimestamp`
21:15:27, later than the code as it must be. A cvmd upgrade can otherwise report success at
every layer while the host keeps serving the previous build.

**One correction to my own test, worth stating.** Block 9 first asserted that an expired
manifest would install (higher serial) and be refused at launch. It is not: `install`
verifies freshness *before* it compares serials, so an expired manifest never reaches the
cache at all. That is the stronger behaviour — the host keeps a catalog it can still
confirm — and the block now asserts it, with block 10 covering the other half through a
cached manifest that has aged out.

Host left clean: cvmd active, node `RECONCILING`, no CVM, no VM directories, no recorded
error, and the manifest URL removed so it does not poll an endpoint that no longer exists.

## Declared, not hidden

- **The first acceptance attempt failed on readiness, not on measurement.** My initial
  manifest carried a compose with only `executor-runner` and no sshd, so the guest booted
  and the launch then timed out waiting for port 12200. The measurement gate had already
  passed. Re-run with the compose DAH-2576/2577 booted on this host.
- **`lium_cvmd_catalog` is gone**, and with it `roles/cvmd/templates/catalog.json.j2`. A
  host that had one keeps working — cvmd ignores unknown config keys — but it will have no
  catalog until `lium_catalog_signer` is set, and it says so.
- **`lium_catalog_manifest_sha256` is no longer required** with the URL. The stub demanded
  it on the grounds that a manifest fetched without an integrity check is not a working
  configuration; the signature is a strictly better version of that check, and a pinned
  checksum goes stale the moment the platform publishes a new serial. It is still accepted
  and still pins the bytes when set. What is required now is the **signer**.
- **`_require_str` stripped whitespace, and a stripped trailing newline is a different
  compose hash.** Caught by a test that compared the materialized compose byte-for-byte.
  It would have surfaced on hardware as "this stack is not approved", with nothing anywhere
  naming whitespace as the cause. `_require_text` exists for exactly the two fields whose
  bytes are the point.
- **The `stubs` tag is now `catalog`** in `site.yml`, `run-vm-tests.sh`, the RUNBOOK and
  `bootstrap.sh`, because the role is no longer a stub.
- **A host that cannot reach the backend eventually stops launching.** That is the intended
  trade, stated plainly: it is what makes a revocation reach a host that has gone quiet.
  The verify report goes `MUST_FIX` with the expiry in `observed` before it happens.

[DAH-2578 — CVM v2 — artifact catalog: backend SSOT + signed host manifest](https://app.notion.com/p/CVM-v2-artifact-catalog-backend-SSOT-signed-host-manifest-3b2b8bfdbde9815cb512e8ea9145d9a5)

## ⚠️ This PR runs zero CI

`test.yml` is `on: pull_request: branches: [main, dev]`, and the filter matches the
**target** branch, so a PR into the umbrella runs no checks — silently, while GitHub still
reports `MERGEABLE`. The gate is deferred to the eventual `feature/2462-cvm-v2` → `main`
PR. Please do not read an empty check list here as coverage.

- [x] I wrote tests.


fortunelucky777:
Superseded by #1212, which carries DAH-2578 and DAH-2579 together in one review.

The commits are the same two, cherry-picked unchanged onto the same base — verified byte-identical: for each of the two original branches, `git diff <branch> <combined>` over that branch's own files is empty, and the combined branch changes exactly 39 + 12 = 51 files.

Closing rather than deleting the branch, so this PR's history stays readable.

PR #1209, opened 2026-08-09
DAH-2579 - Add the renter CVM's attest-agent

fortunelucky777:
## Describe your changes

DAH-2579, host side: **the renter CVM's attest-agent** — the one thing the platform can
talk to inside a rental, and the only thing it should be able to talk to.

**Base is `feature/2462-cvm-v2`, not `main`.** It is independent of #1208 (no shared
files); either can land first.

FR-E6 puts one small Lium agent inside a renter CVM whose only job is answering trust-check
requests, and which gives the platform no access to the customer's workload.

### It ships in the compose, not in the image

That is the point of the task rather than an implementation detail. Baked into the OS
image:

- **updating the agent would mint a new image hash** — a one-line fix becomes a fleet-wide
  image stage-and-approve before it reaches a single host;
- **the customer could not see it** — something would be running in their CVM that they
  have to take on trust.

Shipped in the compose *they* submitted, it is a service they can read, next to a
`read_only` root, no docker socket and no volume into their data. The cost is that the
agent is inside the compose measurement, so changing it changes the `compose_hash` — which
is the correct direction: a changed agent *should* be a changed measurement.

### What the 64 bytes are spent on

```
report_data = sha256(agent_tls_public_key ‖ gpu_uuid_digest) ‖ nonce
              |________________ 32 bytes _________________|  |_ 32 _|
                      identity                                freshness
```

**The TLS key, because the quote and the channel have to be the same thing.** The
validator's only way into a renter CVM is this agent's TLS endpoint. A quote that did not
name the key proves some TDX guest exists somewhere — an attacker who can terminate TLS
relays a genuine quote from a CVM it does not own, and the verifier cannot tell. The key is
generated *inside* the guest on first boot, so its private half never leaves encrypted CVM
memory. It is the renter-side counterpart of the validation CVM's `SSH_HOST_KEY:` binding
in `executor/src/services/tdx_service.py`.

**The GPU UUIDs, because of FR-G6.** Those identifiers travel today on a channel no proof
covers, so a node can claim GPUs it does not hold. The digest is sorted, so a reboot that
reorders NVML does not change what the node attests to; a UUID containing the separator is
refused, because otherwise `{"a,b"}` and `{"a", "b"}` digest identically.

### Idempotent, not amnesiac

Asked twice with one nonce the agent returns the identical answer and takes no second
quote; it never produces two different quotes for one challenge. Refusing a repeat would
break a verifier's retry after a network timeout and would add nothing — **rejecting a
*reused* nonce is the verifier's half**, because only the party that issued one knows
whether it has seen it. `TestTheVerifiersHalf` pins that half next to the agent whose
answers it judges. The ledger is bounded: it is keyed on caller-supplied input.

### Also here: the golden loop's measuring half

`tests/fixtures/measure_compose_corpus.py` runs the **real**
`DStackManager.setup_instance` over the corpus that lium-io-backend#883 derives, and
records what dstack actually wrote. That is what confirms the backend's byte-exact
`app-compose.json` reproduction against dstack rather than against itself — a second
implementation here would agree with the backend's for exactly as long as both were wrong
in the same way.

**All 12 cases agree.** The golden file is committed in the backend repo, next to the test
that reads it: one copy, one owner.

## Verification

| | |
|---|---|
| attest-agent suite | **33 passed** |
| ruff (repo-pinned 0.5.1), lint and format | clean |
| golden corpus measured by real `dstack.py` | **12/12 agree** |

The TDX quote and the NVIDIA evidence are the seam — there is no TDX and no CC-capable GPU
in a test process. Everything around them is real, including the EC key generated by the
real code. What the suite checks is the part hardware cannot: that the bytes the agent
*says* it bound are the bytes it asked the hardware to sign, and that it never signs two
different ones for one challenge.

## Declared, not hidden

- **The agent itself has NOT run inside a real CVM.** The quote path, the NVML read and the
  NVIDIA evidence collection have not executed against real hardware — only their degraded
  paths have, which is the opposite of what matters. That needs the image built and pullable
  from inside a guest, and it belongs with DAH-2580's renter launch. **Outstanding.**
- **The derivation half DID run on hardware.** `measure_compose_corpus.py` was run on au11
  against that host's own `dstack.py`: **12/12 agree with the backend's derivation, and
  byte-identical to the dev-box golden file.** Separately, a real TDX CVM was launched on
  au11 from a manifest whose `compose_hash` this task's backend module derived and never
  measured — cvmd's measurement gate accepted it (HTTP 201 in 48 s) and the measured hash
  equalled the derived one. That is the task's core claim, confirmed on real hardware.
- **The image is not built or published.** The Dockerfile is here; no registry push, and no
  digest to pin in a catalog entry yet. That pinning is what makes the agent part of an
  approved measurement, and it belongs with the rollout.
- **The `nv_attestation_sdk` call is written from the executor's existing
  `gpu_attestation_service.py`**, not from a run. If that SDK's shape has drifted, this
  finds out in a CVM, not here — the dependency is not installable outside one.
- **Nothing calls this yet.** The agent is injected by lium-io-backend#883's compose code,
  and the launch that would run it is DAH-2580's. This PR is the agent and the measuring
  half of the golden loop.
- **The verifier's half is a test, not production code.** The validator that will issue and
  reject nonces is DAH-2580's consumer; what is here is the property, pinned where the
  agent's behaviour can break it.

[DAH-2579 — CVM v2 — renter attest-agent + compose-hash reproduction](https://app.notion.com/p/CVM-v2-renter-attest-agent-compose-hash-reproduction-3b2b8bfdbde981b2b2fbe2d300291632)

## ⚠️ This PR runs zero CI

`test.yml` is `on: pull_request: branches: [main, dev]`, and the filter matches the
**target** branch, so a PR into the umbrella runs no checks — silently, while GitHub still
reports `MERGEABLE`. The gate is deferred to the eventual `feature/2462-cvm-v2` → `main`
PR.

- [x] I wrote tests.


fortunelucky777:
Superseded by #1212, which carries DAH-2578 and DAH-2579 together in one review.

The commits are the same two, cherry-picked unchanged onto the same base — verified byte-identical: for each of the two original branches, `git diff <branch> <combined>` over that branch's own files is empty, and the combined branch changes exactly 39 + 12 = 51 files.

Closing rather than deleting the branch, so this PR's history stays readable.

PR #1211, opened 2026-08-10
DAH-2620: WireGuard cluster overlay for multi-node InfiniBand (validator + template)

taiberium:
Notion: https://app.notion.com/p/Multi-node-IB-clusters-atomic-group-rental-WireGuard-overlay-template-3b8b8bfdbde98175b792ca05b63c78b1

## What this is

Multi-node IB clusters, part 1 of the feature: the transport that lets a rented group act as one
job. The IB data path already works after DAH-2571 (376 Gb/s on one queue pair from a plain
container), but NCCL never reaches it — its bootstrap advertises the container's private
`172.17.0.2`, the peer dials something unreachable, and the IB transport never starts. This gives
every node of a group a routable address on a private WireGuard overlay, for the low-bandwidth
socket bootstrap only; the tensors still travel over InfiniBand.

## What is in this PR (validator + template)

- **`services/cluster_fabric.py`** — the full-mesh WireGuard generator. Given the N nodes of a group
  it mints a keypair per node and renders each node's `wg-quick` config (all-to-all peers, one
  overlay address each on `10.42.0.0/24`). Pure and tested; this is what the group-rental caller
  invokes. Plus `cluster_env_and_ports`, the tiny injector the validator uses.
- **`docker_service`** — when a pod carries `cluster_membership` it gets the config injected as
  `LIUM_WIREGUARD_CONF_B64` and the WireGuard UDP port (51820) published. An ordinary rental has no
  `cluster_membership` and is byte-for-bit unchanged.
- **`payload_models`** — `ClusterMembership{node_index, wireguard_conf}` on the create request. The
  config is minted once for the whole group by the caller (so every node's keys agree) and only
  injected here — the validator generates nothing itself.
- **`cluster-template/`** — the customer-facing image. Its entrypoint raises `wg0` from the injected
  config and exports `NCCL_SOCKET_IFNAME=wg0` (+ the 7.3x socket-thread vars). The renter runs
  `torchrun --nnodes=N` and the IB transport engages on its own. No `LIUM_WIREGUARD_CONF_B64` → the
  entrypoint no-ops and the pod is a normal node.

UDP publishing needed no new plumbing: `PortBinding.protocol` already threads through the SDK; it
was simply never set. WireGuard's handshake is UDP, so without this the overlay never forms.

## Not in this PR — the other half, in the backend

**Atomic group rental** (lium-io-backend): a customer requesting N nodes on one IB fabric as a
single all-or-nothing order, and at reservation time calling `build_cluster_wireguard_configs` with
the group's nodes and sending each node its `wireguard_conf` in the payload. That is where the
"these N nodes are one group" fact is born; without it this PR has nothing to inject. Scoped in the
ticket, not started.

## Tests

`tests/test_cluster_fabric.py`, 9 cases — full mesh, unique sequential overlay addresses, peer
endpoint = host public IP on the WireGuard port, the advertised public key derives from the private
key the node holds, keys unique per node, a valid `wg-quick` render, single-node and duplicate-index
rejection, and the env/UDP injection. Full validator suite: 1503 passed (the 3
`test_sshd_bootstrap_script.py` failures are pre-existing, confirmed on clean main; the golden
wire-format snapshot was regenerated for the new optional field).

## Not verified end to end

A real two-node NCCL job over `wg0` on the 400G H200 trio — needs the group-rental half plus two of
the trio free at once. That is the acceptance test, called out in the ticket.


neurons/validators/src/services/cluster_fabric.py
```diff
@@ -0,0 +1,41 @@
...
+# host's public address on this same port. UDP is not a choice: WireGuard has no TCP mode.
+WIREGUARD_LISTEN_PORT = 51820
+
+
+@dataclass(frozen=True, slots=True)
+class ClusterPodNetworking:
+    """What a cluster node needs on top of an ordinary pod."""
+
+    environment: dict[str, str]
+    published_udp_ports: tuple[int, ...]
```
arhangel66, line 30: The PR body says this module exposes `build_cluster_wireguard_configs` and `cluster_env_and_ports` for the group-rental caller, but the file only has `cluster_pod_networking`, which takes an already-rendered config. The module docstring is the accurate one — the backend mints the mesh. Could you fix the body so part 2 does not go looking for a generator that lives nowhere, and say explicitly which repo owns it? Non-blocking, the code here is self-consistent.

arhangel66 (approved):
Verified the validator side: with `cluster_membership=None` the run spec is unchanged, and `NET_ADMIN` the entrypoint relies on is already granted to every rental container, so the overlay path is additive. One question about the PR body inline.

PR #1214, opened 2026-08-10
DAH-2629 + DAH-2630 + DAH-2632 + DAH-2633 - Launch validation CVMs from the validator, account for the switch window, publish the attest-agent by digest, and drive it all with one E2E command

fortunelucky777:
## Describe your changes

Four tasks of the CVM v2 epic, stacked on #1213 (DAH-2580–2582) because they are its missing callers and accounting:

**DAH-2629 — the validator launches validation CVMs via cvmd.** New `services/cvmd_client.py` signs this validator's own cvmd calls (`validation` scope; the blob formula mirrors the backend's signer and is pinned by test to the same bytes). `services/cvm_whitelist.py` now retains the signed manifest's full stack entries so the launch triple comes from the same document the measurement check reads (`newest_validation_entry()`). New `services/cvm_lifecycle.py` keeps a Redis registry of cvmd hosts (fed by every attested cycle and by the relay), and launches the validation CVM on an idle, empty host — pinned triple, verified launch report, per-host cooldown, fail-closed without a catalog. Two triggers: the per-cycle sweep, and immediately after a successful renter-teardown relay (`compute_client._schedule_validation_cvm_return`). A launch failure is a scoring signal (the node stays unverifiable), not an operator page.

**DAH-2630 — the switch window is availability-accounted (FR-I3).** `miner_service._record_and_grace_cvm_hosts` runs after real results, exactly like the manual-rental forced pass: a registered host that is mid-switch (`SWITCHING`/`TEARDOWN` with a readable `started_at`) and inside its per-hardware-class budget gets a synthesized grace JobResult (`spec=None`, so the backend's executor row is untouched); beyond budget is a flagged event (`CVM_SWITCH_BUDGET_EXCEEDED`), never a longer window. Budgets: `CVM_SWITCH_BUDGET_SECONDS_DEFAULT` + per-GPU-model JSON overrides.

**DAH-2632 — the attest-agent image is published by CI, pinned by digest.** `.github/workflows/attest-agent-publish.yml` builds `neurons/executor/attest-agent` and pushes `ghcr.io/datura-ai/lium-attest-agent`; the job summary prints the exact `CVM_ATTEST_AGENT_IMAGE=...@sha256:...` line to deploy. `PUBLISHING.md` documents why a new digest is a catalog rollout, never a swap.

**DAH-2633 — the local full-flow E2E harness.** `tools/cvm-local-e2e/run.py`: one command that drives renter order → measured launch → scope proofs → verified teardown (switch window observed) → validator-signed validation-CVM relaunch → restored host, against a remote cvmd over an SSH tunnel, using the real modules from both repos. Hard guards: aborts on a busy host, aborts if the remote port range could allocate 32000, restores the host at the end.

Flags (all default-off, DAH-2581 posture): `ENABLE_CVM_LIFECYCLE`, `ENABLE_SWITCHING_GRACE`; plus `CVMD_PORT`, `CVMD_URL_OVERRIDE` (harness only), `CVM_LAUNCH_COOLDOWN_SECONDS`, `CVMD_HOST_REGISTRY_TTL_SECONDS`.

**Verified on hardware (au11, 2026-08-10):** the DAH-2633 harness against au11 (2026-08-10): **24/24 checks passed** — renter launch measured the derived compose hash (golden loop), validator key 403 on renter create AND delete, verified teardown with the switch window observed as RENTER_RUNNING → TEARDOWN → SWITCHING in 12.3 s (inside the 300 s budget, started_at readable), validator-signed validation-CVM relaunch matching the pinned catalog triple, and the host restored to idle with zero leftover disks or staged composes. No forwarded port was 32000 at any point.

Test suites: validators 1602 passed, cvmd 357 passed. No cvmd code changes — the daemon already implements everything these clients call.

Known boundary, tracked in the epic: a virgin cvmd host enters the lifecycle registry with its first attested cycle; the very first CVM on a brand-new host still comes from day-zero provisioning (DAH-2634 owns the staging bring-up).

[DAH-2629](https://www.notion.so/DAH-2629) · [DAH-2630](https://www.notion.so/DAH-2630) · [DAH-2632](https://www.notion.so/DAH-2632) · [DAH-2633](https://www.notion.so/DAH-2633)

Backend counterpart (DAH-2628 + DAH-2631): https://github.com/Datura-ai/lium-io-backend/pull/893. The two merge in either order — the wire contract (`cvm_expectations`, CVM response message types) is additive on both sides.

- [x] Full validators + cvmd suites green locally (this PR targets a feature branch, so CI does not run — the umbrella → main PR is the CI gate)
- [x] Hardware acceptance on au11 via the DAH-2633 harness
- [x] Every new check is flag-gated off by default

### DIFF
--- a/.github/workflows/attest-agent-publish.yml
+++ b/.github/workflows/attest-agent-publish.yml
@@ -1,0 +1,83 @@
+# DAH-2632 — build and publish the renter CVM's attest-agent image, pinned by digest.
+#
+# The digest this job prints is THE deployment artifact: CVM_ATTEST_AGENT_IMAGE must name
+# the image by digest, never by tag, because the injected compose is measured — a mutable
+# tag would mean the approved bytes stay approved while the image behind them changes.
+#
+# Re-publishing is a CATALOG event, not a swap: a new digest changes every derived renter
+# compose hash, so it rolls out through new catalog rows (with grace for in-flight
+# rentals) and the old digest stays valid until its rows are deprecated. See
+# neurons/executor/attest-agent/PUBLISHING.md.
+
+name: "Publish: attest-agent image"
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - "attest-agent-v*"
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: datura-ai/lium-attest-agent
+
+concurrency:
+  group: attest-agent-publish
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to Container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push
+        id: build
+        uses: docker/build-push-action@v5
+        with:
+          context: ./neurons/executor/attest-agent
+          push: true
+          # Provenance attestations wrap the image in an OCI index whose digest differs
+          # from the plain manifest's; disabled so the digest below is the one a docker
+          # pull resolves.
+          provenance: false
+          tags: |
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ github.ref_name }}
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:sha-${{ github.sha }}
+          cache-from: type=gha,scope=attest-agent
+          cache-to: type=gha,mode=max,scope=attest-agent
+
+      - name: Summary — the digest to pin
+        env:
+          DIGEST: ${{ steps.build.outputs.digest }}
+          IMAGE: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          COMMIT: ${{ github.sha }}
+        run: |
+          echo "### attest-agent published" >> $GITHUB_STEP_SUMMARY
+          echo "" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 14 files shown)


fortunelucky777:
## Live full-stack E2E verified — and a switch-back bug it caught

Ran the whole flow through a **local backend + validator + connector + miner**, with au11's real cvmd reached over an SSH tunnel (`CVMD_URL_OVERRIDE`). au11 restored to idle afterwards (RECONCILING, serial 10, original `authorized_clients`; no forwarded port ever 32000).

**Validation flow (plumbing, live):** local validator → miner → the au11 CVM's executor over SSH → machine-spec scrape ran → a **real TDX quote** was collected → verifier call → job result → published to Redis `MACHINE_SPEC_CHANNEL` → connector → backend. Every hop works against the real node. (The final attestation *verdict* depends on the local dstack-verifier's TCB collateral and the guest's intermittent quote generation — an environment limit, not this code; the verifier parses and cryptographically processes the real quote once its version matches the guest's dstack 0.5.11.)

**Rental + switch-back (this PR's code, live, twice):** order → **switch teardown** of the validation CVM → `CvmTornDown` → renter CVM launched and measured (`compose_hash` = the backend's derived value, a compose no catalog holds) → verified → pod RUNNING → **SSH into the renter CVM** → delete → teardown → **the validation CVM relaunched automatically** (`CVM_VALIDATION_LAUNCH_STARTED → VERIFIED`), `TEARDOWN(renter) → LAUNCHING(validation) → VALIDATION_RUNNING`. Scope separation re-proved: validator key `DELETE` → 403, platform key → 200.

### Fix pushed (`65f817f3`)

The first live delete surfaced a real bug: the immediate switch-back runs in the **connector** process (`compute_client`), which never runs an attestation cycle — so its `whitelist_source.current()` was perpetually `None` and every relaunch fail-closed with `CVM_VALIDATION_LAUNCH_FAILED — no signed catalog entry`. This would have failed in production too, not just locally. `ensure_validation_cvm` now `refresh()`es the catalog before reading it (fail-closed and is-due-gated, safe on both the sweep and switch-back paths). Regression test added.

Also `262fa05c`: `CvmTeardownRequest.intent` — a `"switch"` teardown (the backend clearing a node for a renter launch it is about to send) must **not** trigger the switch-back, or the relaunch would race the renter order.

Suite: **1605 passed** (adds the switch-back-refresh regression).

PR #1216, opened 2026-08-11
DAH-2501: vast_api — Vast host management API for the executor box

arhangel66:
- `vast_api` module mounted **inside the executor app** (`attach_vast_api` in `src/executor.py`): routes ride port 8001 behind `MinerMiddleware`. **Key-split architecture** (`plan-key-split`): the Vast account key never touches the box — the box receives only a per-setup, backend-minted `machine_key` inside the signed `POST /vast/setup` body. Market operations (list/unlist/price/self-test) were removed from the node; they belong to the backend (DAH-2653). Box-side `VastClient` = `identify()` only; the `vastai` CLI is out of the executor image.
- Surface: `GET /healthz` (open) · `POST /vast/setup` (idempotent ladder: nested-dockerd cage → kaalia → GPU ×2-around-daemon-reload → register → report; async run model with per-stage journal and crash-safe resume) · `GET /vast/runs*`, `GET /vast/status` (signed `x-signature`/`x-timestamp` GETs bound to the concrete path — they expose kaalia journal and machine id) · `DELETE /vast` (local teardown; purge also tears down the nested data-root). Safety rails as 409 codes: `rental_running` (a Vast contract is a `C.<id>` container in the nested dockerd, counted in ANY state — a client-stopped instance still holds the contract), `run_in_progress`, `port_range_overlap` (vs `RENTING_PORT_RANGE`), `state_mount_missing`.
- Auth: dedicated `VAST_ADMIN_HOTKEY` trusted by `MinerMiddleware` for `/vast/*` paths only; `/vast*` requests get a 180 s timeout, the rest keep 30 s. Miner install instruction is byte-identical.
- Hardened after a triple review pass (bugs / silent failures / test coverage): fail-closed exec into a stopped `vast-uns`, `restart_policy=unless-stopped`, no-fallback network resolve, rc checks on `daemon-reload` and `vastai` restart, `lsattr`-verified `daemon.json` immutability, `losetup` rc check, status degradation slots.
- Proven live on a staging exam (2026-08-12, hyperstack A6000 → Vast machine 147260, deleted after): signed setup with a Mac-minted key end-to-end, one `machine_key` serves both the kaalia installer and identify, Lium rent smoke green on the same box while enrolled. Exam-driven fixes: publish range 40000-40300 reserved from the kernel ephemeral span (`ip_local_reserved_ports`), a never-started (`created`) `vast-uns` treated as absent by the rental rail, `report` stage verifies the daemon reports by itself (manual `send_mach_info.py` is permanently nonce-blocked server-side).
- 72 vast_api unit tests (docker/vast mocked), full executor suite green (188).


arhangel66:
CodeQL finding addressed: raw exception text no longer flows into HTTP responses (/vast/status sections and run-doc error envelopes). Own error types (ApiRefusal/ApiFailure) keep their curated detail; anything else is logged server-side and exposed as the class name only.

arhangel66:
Three-agent review pass (correctness/security, silent-failure hunt, test-coverage) applied in 00f66aa8 — 17 fixes, suite 146→159:

**Security:** empty `VAST_API_TOKEN` no longer opens every route (min_length=16 at startup + empty-bearer reject); the account key left `vastai` argv (world-readable via /proc under `--pid host`) — the CLI now reads `~/.vast_api_key` (0600); machine_key passes to the kaalia installer as an env var instead of shell interpolation; `run_ladder` crash text goes through `safe_error` like everything else.

**Correctness/rails:** a stale persisted machine id can no longer fall through to guessing another machine on a shared account; setup and DELETE rental rails refuse (409) when the rental state can't be verified (Vast outage ≠ no rental); mutating sync ops refuse while a setup/self-test run is active; purge deletes the Vast record before destroying local identity; a typo'd `machine_id` fails before identify mints an orphan record; `chattr +i` on the nested daemon.json is now verified; blkid rc is discriminated so a broken blkid can't green-light `mkfs.xfs -f`; ghost-GPU `[Unknown Error]` output refuses loudly instead of a bare 500.

**Tests:** the previously-dead rails are pinned — `listing_not_confirmed` polling, DELETE×rental×force matrix, purge semantics, resolve precedence, ApiFailure→500 envelope, state-mount rail, fresh-box setup, empty token, run-in-progress guard.

neurons/executor/src/middlewares/miner.py
```diff
@@ -14,6 +14,24 @@
...
+    if path.startswith("/vast"):
+        return VAST_REQUEST_TIMEOUT_SECONDS
+    return AUTHENTICATED_REQUEST_TIMEOUT_SECONDS
+
+
+def trusted_hotkeys(path: str) -> list[str]:
+    # the Vast admin key opens /vast/* only — never the rest of the executor API
+    hotkeys = [settings.MINER_HOTKEY_SS58_ADDRESS, settings.DEFAULT_MINER_HOTKEY]
+    if path.startswith("/vast"):
+        hotkeys.append(settings.VAST_ADMIN_HOTKEY)
```
surcyf123, line 33: This key now opens mutating /vast routes, but the middleware only checks `signature` over the free-form `data_to_sign`; path, method, time and the rest of the body are unbound. One observed backend request can be replayed to any executor with a replaced `machine_key`, `events_url` or `purge: true`. For `/vast*` require `data_to_sign` to be canonical JSON of path, method, timestamp and a sha256 of the other body fields, and verify all four, like `verify_signed_get` does for the GETs.

neurons/executor/src/vast_api/host_ops.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def write_state_file(self, name: str, content: str) -> None:
+        # atomic write (tmp + mv) under STATE_DIR_HOST, surviving executor-container
+        # restarts; content travels base64-encoded so it never meets the shell
+        path = f"{self.settings.STATE_DIR_HOST}/{name}"
+        encoded = base64.b64encode(content.encode()).decode()
+        self.run([
+            "sh", "-c",
+            f"mkdir -p {self.settings.STATE_DIR_HOST} && "
+            f"echo '{encoded}' | base64 -d > {path}.tmp && mv {path}.tmp {path}",
```
surcyf123, line 54: The base64 of `events_token` lands in the `sh -c` argv, world-readable via /proc on the host under `--pid host` (same class as the vastai key fix in 00f66aa8). Pass the content on stdin instead: add `input=` to `HostOps.run` and use `cat > "$0.tmp" && mv "$0.tmp" "$0"` with the path as `$0`.

neurons/executor/src/vast_api/schemas.py
```diff
@@ -0,0 +1,75 @@
...
+class SetupRequest(MinerAuthPayload):
+    # machine_key is backend-minted per call and rotates on every mint
+    # (plan-key-split); the account key never reaches this box
+    machine_key: str
+    machine_id: int | None = None
+    force: bool = False
+    # contract-events delivery config (plan-contract-events): persisted under
+    # STATE_DIR_HOST for the poller; re-setup overwrites it (idempotent)
+    events_token: str | None = None
+    events_url: str | None = None
```
surcyf123, line 52: Unbounded and any scheme accepted; the poller will POST the bearer token here every 30 s. Suggest `HttpUrl` with an explicit https check, and `max_length` on `machine_key` and `executor_id`.

neurons/executor/src/core/config.py
```diff
@@ -25,6 +25,12 @@ class Settings(BaseSettings):
         env="DEFAULT_MINER_HOTKEY",
         default="5D4jX4TqUkZwNwKAjjYrbk2FHFNN2U1TgFF6ZMuNPnjnKJVU"
     )
+    # trusted ONLY for /vast/* routes: the backend signs Vast enrollment calls
+    # with this key (plan-key-split); the account key never touches the box
+    VAST_ADMIN_HOTKEY: str = Field(
```
surcyf123, line 30: With a hardcoded default the /vast surface (nsenter, mkfs, losetup, sysctl) is live on every executor after upgrade. Consider `VAST_API_ENABLED: bool = False` and only call `attach_vast_api` when set.

neurons/executor/src/vast_api/README.md
```diff
@@ -0,0 +1,36 @@
+# vast_api — Vast.ai side manager for a Lium executor box
+
+Local Vast operations for a machine that already runs the Lium executor: builds and runs
+the `vast-uns` nested-dockerd container, enrolls kaalia, registers the machine
+(setup ladder G0→G2, the proven SUCCESS-PATH recipe, idempotent), reports local status
+and tears the install down. Mounted into the executor app (`attach_vast_api` in
+`wiring.py`), so the routes ride the executor's own port and `MinerMiddleware` auth:
+every non-GET body carries a `MinerAuthPayload` signature; GET routes are open.
```
surcyf123, line 8: "GET routes are open" is no longer true: status and runs verify `x-signature`/`x-timestamp`; only `/healthz` is open. The `wiring.py` docstring has the current wording.

surcyf123 (changes requested):
Not approving yet: this needs a rebase on main first (CONFLICTING since the 26 Aug head), and there is one auth gap to close before the /vast routes ship to every executor. Read the whole diff plus the middleware at head.

1. **[high]** `neurons/executor/src/middlewares/miner.py:33`: `MinerMiddleware` verifies `payload.signature` over `payload.data_to_sign` only. `data_to_sign` is a free string the client picks, and nothing binds it to the path, the method, a timestamp, or the rest of the body (`machine_key`, `events_url`, `events_token`, `purge`, `force`). So one signature the VAST_ADMIN key ever produced (a miner sees the backend's setup request on their own box, port 8001 is plain HTTP) can be replayed against any other executor with a different body: enroll that box into the replayer's own Vast account with their `machine_key`, point `events_url` and the bearer token at their server, or `DELETE /vast {"purge": true, "force": true}` to mkfs the data root and drop the machine identity. The body says the setup body is signed; only `data_to_sign` is. Fix: for `/vast*` paths require `data_to_sign` to be canonical JSON of `{"path", "method", "timestamp", "body_sha256"}` and check path, method, freshness (reuse `SIGNATURE_MAX_AGE_SECONDS`) and the hash of the remaining fields before trusting the request, the same way `verify_signed_get` already binds the GET routes. The existing miner routes share the weakness, but this PR is what hands a fleet-wide key destructive host commands, so it should not inherit it.

2. **[medium]** `neurons/executor/src/vast_api/host_ops.py:54`: `write_state_file` puts the base64 of the content into the `sh -c` argv. For `events_config.json` that content is `events_token`. Under `--pid host` the argv is world-readable through `/proc/<pid>/cmdline` on the host while the write runs, the same class you fixed for the vastai account key in 00f66aa8 (base64 hides nothing). Fix: pass the content on stdin. Add `input: str | None = None` to `HostOps.run` (forwarded to `subprocess.run(..., input=input)`) and run:
```python
self.run(
    ["sh", "-c", 'mkdir -p "$(dirname "$0")" && cat > "$0.tmp" && mv "$0.tmp" "$0"', path],
    input=content,
)
```

3. **[low]** `neurons/executor/src/vast_api/schemas.py:52`: `events_url` and `machine_key` are unbounded strings and `events_url` accepts any scheme. The poller then POSTs the bearer token to that URL every 30 s for as long as the box is enrolled. Fix: `machine_key: str = Field(max_length=256)`, `events_url: HttpUrl | None` plus an explicit `https` scheme check, `executor_id: str | None = Field(max_length=64)`.

4. **[low]** `neurons/executor/src/core/config.py:30`: `VAST_ADMIN_HOTKEY` ships with a hardcoded default, so the whole `/vast/*` surface (nsenter into the host, `mkfs.xfs -f`, `losetup`, sysctl writes, a 400 GB image on /ephemeral) is live on every executor after the next upgrade, and the body confirms the miner install text does not change. Fix: gate `attach_vast_api` behind `VAST_API_ENABLED: bool = False` (or treat an empty hotkey as disabled) so the routes exist only where a miner opted in.

5. **[low]** `neurons/executor/src/vast_api/README.md:8`: says "GET routes are open". The router signs `/vast/status` and `/vast/runs*` with `x-signature`/`x-timestamp`; only `/healthz` is open. Fix: reuse the sentence from the `wiring.py` docstring.

6. **[low]** `neurons/executor/src/vast_api/docker_ops.py:25`: `status()` and `delete()` run sync exec calls bounded only by the 3600 s client timeout, while the middleware gives up at 180 s. The threadpool worker keeps blocking after the 180 s response, so a wedged nested dockerd plus a status poller can eat the executor's shared anyio pool (default 40) and slow the rent path. Fix: give `exec_run` calls used by status/delete their own short-lived client (timeout 60 s) or run them with `timeout` inside the container (`["timeout", "60", ...]`).

7. **[info]** `neurons/executor/src/vast_api/router.py:64`: on the two CodeQL alerts. Alert 24 (stack trace exposure) last fired on commit 75da22f at line 52; CodeQL at head 28f1dea passes and has no instance there, so it does not hold at head. `ApiFailure` details still carry `nsenter` command lines and `stderr[:500]` into the signed status and run responses, which is acceptable for an admin route as long as those routes stay signed. Alert 25 (clear text logging at :48) was dismissed as a false positive on 12 Aug: that line logs the public hotkey plus the traceback, not a secret. Dropping `exc_info=True` there would keep it from coming back after the rebase.

Checked and fine: `machine_key` travels only as an exec env var and is kept out of run docs (`test_run_doc_args_never_contain_machine_key`); the events spool persists last-seen before delivery and retries with the same `event_id`; the rental rail counts `C.*` containers in any state.

PR #1219, opened 2026-08-12
DAH-2622: floor a live miner's weight at one u16 unit funded from burn

arhangel66:
## Problem

A miner whose machine is online and passes validation can still compute to an incentive of exactly `0` — for
example an idle machine whose GPU model is not in the unrented program. The SDK's `process_weights_for_netuid`
drops every uid whose weight is exactly zero before our own `_convert_weights_with_positive_floor` ever sees
the vector, so that hotkey is simply absent from what we publish. The chain then deregisters the UID with the
lowest emission (`get_neuron_to_prune`: `(emission, registration block, uid)` ascending), which means the
longer a miner has honestly held a UID, the sooner it is pruned once it drops to zero.

Measured on 2026-08-12 (block 8825772): 54 miner hotkeys have active executors, 48 earn, **6 sit at exactly
zero** — queue positions 49/125/160/189/191/193 among 193 zero-emission non-immune UIDs.

## Change

After quantisation, every hotkey that had at least one live executor in the validator's last completed cycle
and is missing from the u16 vector gets **one u16 unit**, taken from the largest burn entry:

- `_select_floor_candidates` / `_apply_eligibility_floor` / `_largest_burn_index` in `subtensor_client.py` —
  pure functions that build new lists, they never mutate their arguments;
- `Validator.active_hotkeys` — the hotkeys with a successful `JobResult` last cycle, overwritten every cycle,
  never persisted, passed into `set_weights`;
- an unconditional `[set_weights] eligibility floor applied` log line carrying `floor_hotkeys` and
  `floor_count`, so every published vector says who got the minimum from us.

The transfer is integer and post-quantisation on purpose: `_convert_weights_with_positive_floor` upscales by
`round(w / max_w * 65535)`, so deducting from burn *before* quantisation would move the anchor and silently
rewrite every other miner's u16 value. Here the mass is conserved exactly — earning miners' entries are
untouched integers.

Nothing upstream moves: `miner_scores`, `cycle_scores` and the `NORMALIZED_SCORE_CHANNEL` publish keep showing
the honest zero. Only the vector that goes on chain and its `send_weights_to_lium` mirror carry the floor.

Cost: one floored miner receives 1/65535 of the vector ≈ $0.041/h (measured from four UIDs currently sitting at
u16 = 1), so today's six candidates cost ≈ $178/month — drawn from burn, which currently takes ~72 % of the
miner pool.

## Testing

- `pdm run pytest tests/test_incentive_flow.py tests/test_rental_price_incentive_flow.py tests/prod_snapshot/`
  → 97 passed, 4 skipped (87 before this change: the 10 new tests, nothing else moved).
- Full validator suite: 1529 passed; the 3 failures in `tests/test_sshd_bootstrap_script.py` reproduce on
  `origin/main` and are unrelated.
- The floor was temporarily neutered to confirm the new tests really catch its absence:
  `test_set_weights_floors_active_hotkey_with_zero_score` and
  `test_set_weights_floor_does_not_dilute_earning_miners` fail without it.
- No new `JobResult` field, so the prod-snapshot baselines are untouched.

Notion: DAH-2622.

### DIFF
--- a/neurons/validators/src/core/validator.py
+++ b/neurons/validators/src/core/validator.py
@@ -48,6 +48,9 @@ class Validator:
         self.default_extra = {}
 
         self.miner_scores = {}
+        # Hotkeys with at least one live executor in the last completed cycle. Rebuilt
+        # every cycle, never persisted — set_weights floors them to u16=1.
+        self.active_hotkeys: set[str] = set()
         # Number of fully-completed sync cycles since this process started.
         # Used to skip the first post-restart set_weights when the in-memory
         # accumulator may have been re-built from a partial cycle.
@@ -185,7 +188,10 @@ class Validator:
                                 )
                             )
                         else:
-                            await self.subtensor_client.set_weights(miner_scores=self.miner_scores)
+                            await self.subtensor_client.set_weights(
+                                miner_scores=self.miner_scores,
+                                active_hotkeys=self.active_hotkeys,
+                            )
                         self.miner_scores = {}
             except Exception as e:
                 logger.error(
@@ -406,6 +412,14 @@ class Validator:
                             ),
                         ),
                     )
+
+                    # DAH-2622: a miner whose machine passed validation must keep its UID even
+                    # when it earned nothing this cycle. Overwrite, never accumulate.
+                    self.active_hotkeys = {
+                        miner_hotkey
+                        for miner_hotkey, results in all_job_results.items()
+                        if any(result.is_successful for result in results)
+                    }
 
                     incentive = IncentiveFactory.create(
                         config=self.incentive,
--- a/neurons/validators/src/clients/subtensor_client.py
+++ b/neurons/validators/src/clients/subtensor_client.py
@@ -83,6 +83,61 @@ def _convert_weights_with_positive_floor(
     return out_uids, out_vals, floored
 
 
+def _select_floor_candidates(
+    uint_uids,
+    miners,
+    active_hotkeys: set[str],
+    burn_uids: set[int],
+) -> list[tuple[int, str]]:
+    """Pick the hotkeys that had a live executor last cycle but are missing from the
+    u16 vector. Burn uids are never candidates — they fund the floor, they don't take it.
+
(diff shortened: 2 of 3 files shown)


taiberium (changes requested):
Requested changes.

neurons/validators/src/core/validator.py
```diff
@@ -407,6 +413,14 @@ async def sync(self):
                         ),
                     )
 
+                    # DAH-2622: a miner whose machine passed validation must keep its UID even
+                    # when it earned nothing this cycle. Overwrite, never accumulate.
+                    self.active_hotkeys = {
+                        miner_hotkey
+                        for miner_hotkey, results in all_job_results.items()
+                        if any(result.is_successful for result in results)
```
arhangel66, line 421: You are right that the hard price cap zeroes only `actual_score` and leaves `job_score` at 1.0 — it is the one check that touches a single field. But `score > 0 or job_score > 0` is the established "node is alive" predicate, not something this PR invents: it also gates the unrented pool in `incentive/rental_price.py:718`, and compute-app keys `executor.active` off the same pair (`worker.py:287`, where `synthetic_job_score` is this very `JobResult.job_score`). So an over-cap idle node whose GPU model is in the unrented programme keeps earning from that pool, stays in the weight vector, and never reaches the floor at all — `_select_floor_candidates` only picks uids that are *not* in the vector.

The suggested `result.score > 0` would take the floor away from rented machines: `_format_return` (`score_calculator.py:110`) forces `job_score = 1.0` for a rented executor while `actual_score` can be 0 from collateral, an old contract version or a failed CVM attestation. Those are machines with a paying customer on them, which is exactly what this PR exists to keep registered.

The tail you are pointing at is real but narrow — over-cap price *and* a GPU model outside the unrented programme — and it is a policy question rather than a bug: today a price penalty costs the miner money, not its registration. It also cannot be separated here, because `JobResult` carries no zeroing reason code, only the `score_warning` string. Leaving the condition as is; happy to open a follow-up if you think the price-penalised case should lose the floor.

taiberium (approved):

PR #1220, opened 2026-08-12
DAH-2462 - Land the CVM provisioning v2 epic (subnet side)

fortunelucky777:
## Describe your changes

Integration PR for the CVM provisioning v2 epic. `feature/2462-cvm-v2` was cut from `main` at `30628c4d` and collected seven child PRs; this lands them together.

**What ships**

| Child | What it adds |
|---|---|
| #1203 | DAH-2544 — day-zero Ansible host setup (TDX/SGX, kernel cmdline, GRUB, fail-closed guard ladders) |
| #1205 | DAH-2575 — `cvmd` service skeleton, signed two-scope API, and its Ansible role |
| #1206 | DAH-2576 — launch and stop the validation CVM from `cvmd`, with `dstack.py` usable as a library |
| #1207 | DAH-2577 — verified teardown, and `SWITCHING` as its own state |
| #1212 | DAH-2578 + DAH-2579 — fetch the approved catalog as a signed manifest; add the renter attest-agent |
| #1213 | DAH-2580 + DAH-2581 + DAH-2582 — launch a renter CVM from the customer's compose, verify it against the signed catalog, refuse two nodes that are one machine |
| #1214 | DAH-2629 + DAH-2630 + DAH-2632 + DAH-2633 — validator-driven validation-CVM lifecycle, switch-window availability accounting, attest-agent published by digest, one-command E2E |

250 files, +32363 / −134, against the merge base with `main`. By area: `neurons/executor/cvmd` (56 files), `neurons/executor/ansible` (39, plus roles and tests), `neurons/validators/src` (12), `neurons/executor/attest-agent` (5), `tools/cvm-local-e2e`.

Non-goals held throughout: nothing written under `neurons/executor/dstacktee/app/**`, and no behaviour change to `lium-cvm.sh` or `dstack.py`'s CLI entry point.

## Testing

Measured on the pre-merge stack tips, whose trees are byte-identical to these umbrella heads:

- Full local suites green in both repos — validators 1605 tests, backend 2289.
- Full-stack live E2E run twice through the deployed `POST /rent-cvm`, against real Intel TDX hardware (host au11).
- Verified teardown and CVM switch measured on au11; switch window 5–6 s.
- attest-agent executed inside a real TDX CVM: 15/15.
- The live run caught a real switch-back bug that would otherwise have reached production: the switch-back runs in the connector, which never refreshes the catalog, so `current()` stayed `None` forever. Fixed in `65f817f3`.

## Known gaps, stated rather than hidden

- **No child of this epic has ever run CI.** `.github/workflows/test.yml` fires on `pull_request: branches: [main, dev]`, so every umbrella-based PR reported no checks by design; the gates were deferred to this PR. **This is the first CI run for the whole epic** — expect it to surface things the local suites did not.
- **GPU-host acceptance has not run** (DAH-2637). au11 has no GPUs, so the GPU half of the flow is unexercised on hardware.
- **The attest-agent image is pinned to a personal test namespace.** Publishing it to a real namespace is the remaining half of DAH-2632.
- Staging bring-up (DAH-2634) and rollout telemetry for the DAH-2583 phase gates (DAH-2636) are separate tasks and are not in this PR.

## Merge coordination

Paired with `lium-io-backend` PR for the same umbrella. The backend side carries an Alembic two-head fork that must be fixed before either merges — see that PR's Risks section.

Opened as a draft on purpose: review is wanted now, merge is gated on the first CI run and on the backend fix landing with it.

[DAH-2462 — CVM upgrade: research & planning](https://app.notion.com/p/CVM-upgrade-research-planning-3a3b8bfdbde98193b9bffb871440a5bc)

Landing task: [DAH-2635 — CVM v2: land the umbrellas — coordinated merge, first CI run, Alembic fix](https://app.notion.com/p/CVM-v2-land-the-umbrellas-coordinated-merge-first-CI-run-Alembic-fix-3b8b8bfdbde9812085efe46608db95f6)

- [x] I wrote tests.


neurons/executor/attest-agent/src/attest_agent/identity.py
```diff
@@ -0,0 +1,100 @@
...
+            )
+    joined = _UUID_SEPARATOR.join(sorted(uuids))
+    return hashlib.sha256(joined.encode()).digest()
+
+
+def identity_digest(tls_public_key: bytes, uuids: list[str]) -> bytes:
+    """The 32-byte identity half: this agent's TLS key bound to this CVM's GPUs."""
+    if not tls_public_key:
+        raise IdentityError("the agent has no TLS public key, so it can prove nothing")
+    return hashlib.sha256(tls_public_key + gpu_uuid_digest(uuids)).digest()
```
kvinwang: Could we add an explicit, versioned domain-separation prefix here (and define the public-key encoding), for example `SHA256("LIUM_RENTER_ATTEST_TLS_V1\0" || tls_spki_der || gpu_uuid_digest)`? The current construction has no cryptographic context distinguishing this TLS-channel binding from other present or future uses of the same key material. A versioned purpose prefix would also make future recipe migrations unambiguous.

neurons/executor/attest-agent/src/attest_agent/identity.py
```diff
@@ -0,0 +1,100 @@
...
+    GPU binding of the same shape.
+    """
+    for uuid in uuids:
+        if not uuid or _UUID_SEPARATOR in uuid:
+            raise IdentityError(
+                f"{uuid!r} is not a usable GPU identifier: it is empty or contains "
+                f"{_UUID_SEPARATOR!r}, which would make two different GPU sets digest alike"
+            )
+    joined = _UUID_SEPARATOR.join(sorted(uuids))
+    return hashlib.sha256(joined.encode()).digest()
```
kvinwang, line 97: NVML UUIDs are guest-observed identifiers, not identities authenticated by the NVIDIA attestation evidence. Binding their digest into the TDX quote proves what this measured guest observed, but it does not make those UUIDs hardware-attested or establish that they correspond to the GPUs accepted by NRAS. For platform decisions about GPU existence, authenticity, counting, and cross-executor uniqueness, could we use the verified per-GPU `ueid` claims (and reject missing or duplicate UEIDs) rather than treating this NVML-derived set as trusted identity? If this digest is retained as observational metadata, the comments should avoid describing it as a hardware-attested GPU set.

neurons/executor/attest-agent/src/attest_agent/identity.py
fortunelucky777: Done — thank you, this was right on both counts.

The recipe is now

```
sha256("LIUM_RENTER_ATTEST_TLS_V1\0" ‖ tls_spki_der ‖ gpu_uuid_digest(uuids))
```

with three things pinned deliberately:

- **The tag is versioned and inside the hashed bytes**, so a future recipe is a new tag a verifier accepts or refuses by name, never the same 32 bytes reinterpreted.
- **The NUL terminates it**, for exactly the reason you imply: without it a longer future tag beginning with this one could be made to collide by shifting bytes across the tag/key boundary.
- **The public-key encoding is named in the contract**: SubjectPublicKeyInfo DER, taken from the certificate, which is what a TLS client sees on the wire. It was already what the agent hashed, but only by accident of implementation — now it is stated in `identity.py`, in `renter_identity_digest` on the validator, and in the README.

Agent and validator hold the constant separately (same reasoning as `GPU_UUID_SEPARATOR`: the agent is a package that does not ship with the validator), and each side has a test asserting the exact bytes, plus one asserting that the *untagged* digest is refused.

No transition window: nothing in production produces the old construction yet, so this ships as one lockstep change rather than an accepted-recipes list.

Commit: a87e78b5 (on `feat/2675-rented-cvm-quote-verify`, #1225 — the top of the stack that merges into this branch).

neurons/executor/attest-agent/src/attest_agent/identity.py
fortunelucky777: Agreed, and taken the first branch of your either/or — the ueids now decide, rather than the comments being softened around an NVML-derived set.

**The claim is corrected everywhere it was overstated.** `identity.py`, the agent README, `gpu_uuid_digest`, `_bound_gpu_digest` and `_expected_identity_digests` now say what the binding actually buys: the digest travels inside the hardware signature, so the host cannot rewrite the guest's NVML reading in flight or lift it onto another quote. That is an integrity property, not authentication. (One of those comments also described a `scraped_uuids` cross-check that has never existed in the code — deleted.)

**The rented path decides on `ueid` now.** `verify_rented_cvm_quote` submits the agent's GPU evidence to NRAS and takes the GPU set from the verified per-GPU claims:

- sold-vs-held count is compared against the ueids, not the UUID list;
- a per-GPU token with **no** ueid fails, and two tokens sharing **one** ueid fails — silently shrinking or deduplicating would quietly narrow every decision taken on the set (this lands in `_check_gpu_claims`, so the validation path gets it too);
- if the guest reports more devices than NVIDIA attests, that mismatch is itself a rejection;
- evidence NRAS examined and refused → REJECTED; NRAS unreachable → UNAVAILABLE (our condition, not the node's); no evidence supplied → observe-only until `ENABLE_GPU_ATTESTATION_ENFORCEMENT`, the same flag and posture as `_verify_gpu`.

The UUID list keeps only the job it can do: contradicting the order.

**Two things this exposed.** The agent could not produce usable evidence at all — its image shipped without the NVIDIA stack, and `collect_gpu_evidence` returned a bare evidence list, which NRAS will not judge. It now installs `nv-ppcie-verifier==1.5.0` (the executor image's package and version) and returns the submittable `{nonce, evidence_list, arch}` payload through the same collection flow. And while testing that image: the `verifier` package opens a log file in the working directory *as it is imported*, so under the agent's read-only rootfs the import raised `PermissionError` from any non-writable CWD. The container now starts in its state volume, and a load failure degrades to "no evidence" instead of 500-ing the route.

**What is not proven yet:** the NRAS leg of the rented path has unit coverage only — a GPU TDX host is DAH-2637's dependency and I do not have one. Until then, and while enforcement is off, a rented CVM whose agent returns no evidence still verifies on the quote with `GPU set unattested (...)` stated in the verdict reason rather than passed over in silence.

Cross-executor uniqueness for rented CVMs is still not wired — `AttestedIdentityRegistry` only sees the validation path today. That is a real gap, and I would rather track it than half-wire it here.

Commit: a87e78b5 (on `feat/2675-rented-cvm-quote-verify`, #1225).

surcyf123:
Triage: the CVM v2 epic (DAH-2462, +36k lines, draft, conflicting) has had no push for 20 days and its author is no longer active on this repo. @arhangel66: is your CVM work (DAH-2780/2859/2860) built from this branch, or should the epic be closed and the ticket re-planned? Also decides lium-io-deployment#33 (its staging flags). Not closed — an epic is the owner's call.

surcyf123:
Value audit (SO §69, Rustam asked 15 Sep 11:30Z to re-check every open PR): Waris's CVM v2 epic (cvmd + ansible) conflicts since Aug with nobody to carry 264 files; prod runs the dstacktee path; close the integration PR and let the team re-cut any child it wants

Numbers: size +36553/-193 files=264; value author offboarded 8 Sep; risk prod-config.

PR #1224, opened 2026-08-13
DAH-2671: shadow spec-truth checks — kernel GPU verdict, CPU truth, rental CPU limit, all-cards matmul

arhangel66:
Closes the class where an executor advertises 8 GPUs / 176–240 CPU cores while really having one GPU and 20–44 cores and still passes rental verification. Every verification surface reads the same tampered NVML on the judged host; the fix is built on **a lie is exposed by what you demand, not by what you enumerate** — verification asks for access, rental asks for resources.

**All four checks ship shadow-first.** Each pair mirrors `FILLER_LIVENESS_*`: `*_CHECK_ENABLED` defaults `True` (compute + emit, but `passed=True`, no score change), `*_ENFORCEMENT_ENABLED` defaults `False`. Nothing reduces an honest miner's score until enforcement is flipped from a week of fleet data — an operator decision at a halt, not a follow-up commit.

### What changed
- **1 · surface-kernel-verdict** (`nvidia_devices.py`) — the validator already computes a truth verdict from the kernel's own `/proc/driver/nvidia/gpus/` list (a source the NVML shim does not author) and throws it away (243×/7d in prod). Emit it as a stable, greppable structured event, and treat a kernel-vs-nvidia-smi-XML disagreement as the finding: under enforcement refuse the spoofable XML overwrite so the kernel truth stands (fail closed); the unreadable-`/proc` honest case is recorded distinctly.
- **2a · rental-cpu-truth** (new non-fatal `CpuTruthCheck`) — compares the advertised `CPU(s)` count against the kernel-present population (`/sys/devices/system/cpu/present`, like-for-like with `lscpu`), `/proc/cpuinfo` and docker `NCPU`, over SSH. Emits the RAW counts so the enforcement predicate is chosen from data. Any inability to measure passes with a distinct `*_unknown` reason.
- **2b · rental-cpu-limit** (`rental_verification.py`, `backend_client.py`) — send the advertised core count in the rental-verification health check so the backend creates the probe with `--cpus=<advertised>`. The host's **own Docker daemon** rejects a count above its physical cores (`CPU_QUOTA_EXCEEDS_HOST`) — a lie the lscpu wrapper cannot cover, because dockerd reads the kernel directly. Shadow observes the verdict; enforcement zeroes the score. Skipped on TDX hosts (a real rent skips `--cpus` there too). **Requires the backend half first:** Datura-ai/lium-io-backend#918.
- **3 · full-matmul-all-cards** (`matrix_validation_service.py`) — runs the GPU work-proof on **every** claimed card concurrently, each pinned `CUDA_VISIBLE_DEVICES=<index>`, each with its own challenge, sized from the registry SKU table (`gpu_spec_table`) instead of host self-report, timed in aggregate by the validator. A count lie fails device selection or serialises past a wide wall-clock. Only reaches idle executors (customer rentals halt at `TenantEnforcement`; filler-only nodes skip in `CapabilityCheck`). No `.so` rebuild — the change is validator-side command construction only. Legacy single-card / unknown-model hosts skip the probe.

### Companion PR
Backend half of 2b (declare + apply `cpu_count`, classify the daemon's CPU-400): Datura-ai/lium-io-backend#918 — **merge that first**, then this.

### Out of scope (own tickets)
The per-card cryptographic seal in `libdmcompverify.so` (staging proved `CUDA_VISIBLE_DEVICES` breaks per-card decrypt, so item 3 is a count-check, not per-card crypto), NCCL/NVLink chassis-binding, and collateral/clawback.

### Testing
`pdm run pytest tests/` → 1543 passed, 28 new (`test_cpu_truth_check.py`, `test_nvidia_devices_kernel_verdict.py`, `test_matmul_allcards.py`, +5 in `test_rental_verification_check.py` for 2b). The 3 `test_sshd_bootstrap_script.py` failures are pre-existing (fail identically on `main`). Items 1/2a live on staging; item 3 ran live on a real 2×A6000 (passes honest, no false positive). 2b: honest-box no-false-positive + the daemon→classify→shadow chain to be exercised on staging.

### DIFF
--- a/neurons/validators/src/services/matrix_validation_service.py
+++ b/neurons/validators/src/services/matrix_validation_service.py
@@ -30,6 +30,11 @@ MATRIX_VERIFY_TIMEOUT_SECONDS = 120
 # each keep the MATRIX_VERIFY_TIMEOUT_SECONDS cap, so this only fails clear serialisation.
 MATMUL_ALLCARDS_WALL_CLOCK_SECONDS = 90
 
+# Cards probed at once. Each card takes an SSH channel on one connection and MAX_GPU_COUNT is 14,
+# while OpenSSH defaults to MaxSessions 10 — 8 keeps the concurrency signal (a count lie still
+# serialises) and stays clear of that ceiling on the widest honest hosts.
+MATMUL_ALLCARDS_MAX_CONCURRENT_CARDS = 8
+
 
 class DMCompVerifyWrapper:
     def __init__(self, lib_name: str):
@@ -295,12 +300,20 @@ class ValidationService:
             params.cipher_text = self._generate_card_cipher(machine_info, params)
             challenges.append(params)
 
+        # One SSH channel per card on a single connection, so the fan-out has to stay under the
+        # executor's sshd MaxSessions (OpenSSH defaults to 10). A 12-14 card host would otherwise
+        # push the tail channels into the ssh_error branch and read as a failed work-proof.
+        gate = asyncio.Semaphore(MATMUL_ALLCARDS_MAX_CONCURRENT_CARDS)
+
+        async def run_gated(index: int, params: VerifierParams) -> dict:
+            async with gate:
+                return await self._run_one_card(
+                    ssh_client, executor_info, script_path, index, params
+                )
+
         start = time.perf_counter()
         per_card = await asyncio.gather(
-            *(
-                self._run_one_card(ssh_client, executor_info, script_path, index, params)
-                for index, params in enumerate(challenges)
-            )
+            *(run_gated(index, params) for index, params in enumerate(challenges))
         )
         elapsed = time.perf_counter() - start
 
--- a/neurons/validators/src/services/matrix_validation_service.py
+++ b/neurons/validators/src/services/matrix_validation_service.py
@@ -298,6 +298,17 @@ class ValidationService:
             params.generate()
             params.dim_k = int(self.get_max_matrix_dimensions(sized_vram_mb, params.dim_n))
             params.cipher_text = self._generate_card_cipher(machine_info, params)
+            # A native error yields an empty cipher, which would go out as a bare `--cipher_text `
+            # and read as a failed work-proof on every card. Our own fault, not the host's: skip
+            # the probe and let the caller's single-card path decide, as the legacy path does.
+            if not params.cipher_text:
+                logger.warning(
+                    _m(
+                        "All-cards matmul skipped: cipher text generation failed (native error)",
+                        extra=get_extra_info(default_extra),
+                    )
+                )
+                return None
             challenges.append(params)
 
         # One SSH channel per card on a single connection, so the fan-out has to stay under the
--- a/neurons/validators/src/services/matrix_validation_service.py
+++ b/neurons/validators/src/services/matrix_validation_service.py
@@ -386,10 +386,15 @@ class ValidationService:
             return {"card_index": index, "ok": False, "reason": f"exit_status={exit_status}"}
         if not stdout.strip():
             return {"card_index": index, "ok": False, "reason": "empty_output"}
-        # Record whether the executor's decrypt recovered THIS card's challenge uuid. Emitted for
-        # observability, NOT yet gated on (verdict stays exit-status + non-empty): it lets the shadow
-        # data show per-card decrypt efficacy — chiefly that machine_info reconstructs correctly under
-        # CUDA_VISIBLE_DEVICES on a real multi-GPU host. Gate on it before enforcement (DAH-2671).
+        # Observability only — do NOT gate the verdict on this. Confirmed on a staging 2xA6000
+        # (DAH-2671): under CUDA_VISIBLE_DEVICES the .so's getGPUInfo() reconstructs machine_info from
+        # the masked (single) card, so its AES key diverges from the validator's full-N machine_info
+        # key and decrypt recovers no uuid — uuid_match is expected False on EVERY honest multi-GPU
+        # host (the same box's no-CVD legacy CapabilityCheck verified fine). Item 3 therefore proves
+        # card COUNT via device-selection failure (an index past the real cards -> CUDA_ERROR_NO_DEVICE
+        # -> non-zero exit -> ok False) plus the aggregate wall-clock, NOT per-card crypto; the per-card
+        # cryptographic seal that survives CVD is the named follow-on ticket. The field is kept so the
+        # shadow data records this, and so a future seal-aware .so would flip it True.
         returned_uuid = _extract_result_uuid(stdout)
         return {
             "card_index": index,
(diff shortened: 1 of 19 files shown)


arhangel66:
Review pass. The intent, the shadow-first rollout and the test coverage are solid — requesting changes on four points, the first of which is a real bypass.

**1. `_probe_cpu_limit` (rental_verification.py) — remove the cap entirely; it defeats both CPU checks at once.**

The cap logic is readable by miners (this repo is public) and hands them a precise recipe:
- bind-mount a fake `/sys/devices/system/cpu/present` (same effort class as the lscpu wrapper we already observed in prod), keep `/proc/cpuinfo` real;
- item 2a passes (advertised == present);
- the cap condition (`advertised - present <= margin`) triggers, so the validator sends `min(advertised, online) = online` — the daemon accepts it and `CPU_QUOTA_EXCEEDS_HOST` can never fire.

One next-generation spoof silences 2a and 2b together. Fix: delete `_probe_cpu_limit` and the `_count_cpu_list` / `CPU_TRUTH_MARGIN_CORES` imports; send the advertised count as-is (under the flag, non-TDX unchanged). The honest SMT-off host this cap was protecting moves to the backend classifier instead: it compares advertised against the daemon's own N ("only N CPUs available") and only classifies at >2xN + margin — see the companion request on Datura-ai/lium-io-backend#918. Please add a regression test named for the bypass: fake present (176) + real cpuinfo (44) → 176 goes out uncapped. This also removes 2 SSH round-trips per verification.

**2. `_ssh_peer` (nvidia_devices.py) — test-only code in production.**

The `inspect.iscoroutine(peer)` branch exists only because an AsyncMock hands back a coroutine in tests. Remove it (and the `inspect` import); in the affected tests set `get_extra_info` to a plain function, as test_nvidia_devices_kernel_verdict.py already does.

**3. `CpuTruthCheck` — drop `_read_online` and `_read_ncpu`.**

Only `present` participates in the verdict; online and docker NCPU are recorded but never gated on (and both are host-authored, as the docstring itself notes). That is 2 extra SSH round-trips per executor per cycle for log fields nobody reads. A single advertised-vs-present comparison halves the check.

**4. Item 3 — drop `uuid_match` / `returned_uuid` / `_extract_result_uuid` until the seal-aware .so exists.**

A field documented as "expected False on EVERY honest multi-GPU host" is dead weight in the shadow data; reintroduce it in the follow-on seal ticket where it becomes meaningful. Also worth one line in the PR description: the real protection in item 3 is device-ordinal failure plus OOM from concurrent near-full-VRAM allocations — the 90 s wall-clock alone does not catch 8 fast cards serialised (8 x 4 s = 32 s < 90 s).

Everything else (item 1 kernel verdict, the 8-flag shadow-first pattern, the backend retry loop) is good to keep as is.

arhangel66:
Pushed the simplifications from the review comment above: c0033657 — cap removed (advertised goes out as-is; the ratio decision moved to the backend classifier, see #918), CpuTruthCheck reduced to advertised-vs-present, uuid_match dropped until the seal-aware .so, _ssh_peer test-hack removed. Bypass regression test added.

neurons/validators/src/services/nvidia_devices.py
```diff
@@ -147,25 +269,55 @@ async def _query_gpu_nodes_for_uuids(
...
+                    gpu_uuids,
+                    uuid_to_minor,
+                    xml_uuid_to_minor,
+                    proc_unreadable=proc_unreadable,
+                    enforce=enforce,
+                    executor_id=executor_id,
+                    default_extra=default_extra,
+                )
+                if enforce:
+                    overwrite_with_xml = False
```
taiberium, line 308: NIT: when procfs is unreadable the map stays empty, so line 311 raises a plain `RuntimeError` and the caller falls back to legacy `--gpus`. Fail-closed does not hold.

neurons/validators/src/services/task/checks/cpu_truth.py
```diff
@@ -0,0 +1,137 @@
...
+        cpu = (ctx.state.specs or {}).get("cpu") or {}
+        try:
+            count = int(cpu.get("count"))
+        except (TypeError, ValueError):
+            return None
+        return count if count > 0 else None
+
+    async def _read_present(self, ssh) -> int | None:
+        # kernel-present CPU population — the authoritative like-with-like comparison
+        res = await ssh.run("cat /sys/devices/system/cpu/present")
```
taiberium: NIT: this `ssh.run` has no timeout, so a stalled host burns the executor's whole 13-minute budget. Add `timeout=` like the matmul service does.

neurons/validators/src/services/task/checks/rental_verification.py
```diff
@@ -271,6 +286,55 @@ async def run(self, ctx: Context) -> CheckResult:
                 ctx.ssh, ctx.executor.uuid
             )
 
+    @staticmethod
+    def _advertised_cpu_count(ctx: Context) -> int | None:
```
taiberium: NIT: this duplicates `CpuTruthCheck._advertised_count`, so the same specs mapping exists twice and can drift. Share one helper.

neurons/validators/src/services/matrix_validation_service.py
```diff
@@ -221,13 +252,181 @@ async def _kill_remote_matmul(self, ssh_client, script_path: str, log_extra: dic
                 )
             )
 
+    async def _probe_all_claimed_cards(
```
taiberium, line 255: NIT: 96 lines and two phases here, building the challenges and then running and judging them. Extract the challenge building.

neurons/validators/src/services/matrix_validation_service.py
```diff
@@ -221,13 +252,181 @@ async def _kill_remote_matmul(self, ssh_client, script_path: str, log_extra: dic
...
+            "enforced": settings.MATMUL_ALLCARDS_ENFORCEMENT_ENABLED,
+        }
+        log = logger.info if passed else logger.warning
+        log(_m("All-cards GPU work-proof", extra=get_extra_info(log_extra)))
+
+        if any(outcome.get("reason") == "timeout" for outcome in per_card):
+            # asyncssh's timeout only abandons the local wait; a timed-out remote matmul keeps
+            # running and holding VRAM. The fatal single-card capability matmul runs next in this
+            # same call, so sweep the orphans first or it OOMs on device 0 — mirrors the legacy path.
+            await self._kill_remote_matmul(ssh_client, script_path, log_extra)
```
taiberium, line 350: NIT: an `ssh_error` card can leave the same matmul holding VRAM, and the fatal single-card run follows on device 0. Sweep on that reason too.

neurons/validators/src/services/matrix_validation_service.py
```diff
@@ -221,13 +252,181 @@ async def _kill_remote_matmul(self, ssh_client, script_path: str, log_extra: dic
...
+        finally:
+            if gen_ptr is not None:
+                try:
+                    self.wrapper.free(gen_ptr)
+                except Exception:
+                    pass
+
+    async def _run_one_card(
+        self, ssh_client, executor_info, script_path: str, index: int, params: "VerifierParams"
+    ) -> dict:
```
taiberium, line 388: NIT: the per-card outcome has fixed named fields, so make it a small dataclass and `asdict` it at the log boundary.

taiberium (approved):

PR #1228, opened 2026-08-17
DAH-2697: preserve provider routing during portal outages

pixel29913:
## Summary

Keep the last successful provider-portal response available across validator restarts so a portal outage does not remove central-routed providers from the existing serving-miner set.

## Change

- Validate portal records as `OptedInMiner`; hotkey, central IP, and central port are required, and unusable endpoints make the response unavailable.
- Preserve the distinction between portal failure (`None`) and a successful empty response (`[]`).
- Cache every valid portal response in Redis without a TTL. On portal failure, use the Redis snapshot; if Redis is unavailable, retain the in-memory miner list.
- Apply the live or cached portal response only as IP/port routing overrides. Miner eligibility remains unchanged: serving axons plus configured burner UIDs.
- If no live, Redis, or in-memory snapshot exists, skip the validator iteration explicitly while keeping accumulated scores persisted.
- Report scored registered hotkeys missing from the final miner set as diagnostics only. Weight submission continues.
- Log cache source, provider count, and age; emit a critical alert once the cached response is one hour old, and warn when a previously non-empty provider response becomes empty.

## Scope

A successful `200 []` remains authoritative and is cached. Detecting a portal response that is syntactically successful but semantically incomplete requires a stronger portal API contract and is outside this validator fix.

This change does not redesign scoring or miner eligibility. It only prevents a failed portal request from being interpreted as an empty routing response.

## Validation

- `tests/test_validator_portal_fail_safe.py`: 19 passed.
- The regression coverage verifies that cached portal routing survives a failed refresh, serving non-opted-in miners remain eligible, and non-serving miners remain excluded.
- Full validator suite on commit `6b84d253`: 1599 passed, 9 skipped, with the same 3 unrelated SSH bootstrap failures in the untouched test file.
- `git diff --check` passed.

Corrected staging smoke is pending deployment of commit `6b84d253`.

### DIFF
--- a/neurons/validators/src/clients/subtensor_client.py
+++ b/neurons/validators/src/clients/subtensor_client.py
@@ -160,7 +160,7 @@ def _apply_eligibility_floor(
 
 def _classify_missing_scored_hotkeys(
     miner_scores: dict[str, float],
-    snapshot_miner_hotkeys: set[str],
+    selected_miner_hotkeys: set[str],
     registered_hotkeys: set[str],
     active_hotkeys: set[str],
 ) -> MissingScoredHotkeys:
@@ -169,7 +169,7 @@ def _classify_missing_scored_hotkeys(
         for hotkey, score in miner_scores.items()
         if score > 0
         and hotkey in registered_hotkeys
-        and hotkey not in snapshot_miner_hotkeys
+        and hotkey not in selected_miner_hotkeys
     )
     active_missing_hotkeys = [
         hotkey for hotkey in missing_hotkeys if hotkey in active_hotkeys
@@ -497,7 +497,7 @@ class SubtensorClient:
         self._report_cached_opted_in_miners_usage(cached_snapshot)
         return cached_snapshot.miners
 
-    def _build_miners_from_opted_in_snapshot(
+    def _build_serving_miners_with_opted_in_routing(
         self,
         opted_in_miners: Sequence[OptedInMiner],
     ) -> list[bittensor.NeuronInfo]:
@@ -539,7 +539,7 @@ class SubtensorClient:
                     )
                 )
                 return
-            miners = self._build_miners_from_opted_in_snapshot(opted_in_miners)
+            miners = self._build_serving_miners_with_opted_in_routing(opted_in_miners)
 
         logger.info(
             _m(
@@ -583,16 +583,16 @@ class SubtensorClient:
         except Exception as e:
             logger.error(_m("[send_weights_to_lium] Failed to post latest-set-weights", extra=get_extra_info({"error": str(e)})))
 
-    def _log_scored_hotkeys_missing_from_snapshot(
+    def _log_scored_hotkeys_missing_from_selected_miners(
         self,
         miner_scores: dict[str, float],
-        snapshot_miners: Sequence[bittensor.NeuronInfo],
+        selected_miners: Sequence[bittensor.NeuronInfo],
         registered_miners: Sequence[bittensor.NeuronInfo],
         active_hotkeys: set[str],
     ) -> None:
         missing_hotkeys = _classify_missing_scored_hotkeys(
             miner_scores=miner_scores,
-            snapshot_miner_hotkeys={miner.hotkey for miner in snapshot_miners},
+            selected_miner_hotkeys={miner.hotkey for miner in selected_miners},
             registered_hotkeys={miner.hotkey for miner in registered_miners},
             active_hotkeys=active_hotkeys,
         )
@@ -609,7 +609,7 @@ class SubtensorClient:
         )
         log_diagnostic(
             _m(
-                "[set_weights] scored miners missing from provider snapshot",
+                "[set_weights] scored miners missing from selected miner set",
                 extra=get_extra_info(
                     {
                         **self.default_extra,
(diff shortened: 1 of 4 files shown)


pixel29913:
## Staging smoke status

The earlier smoke results were for superseded commit `59144c8` and exercised its fail-closed behavior. They do not validate the corrected Redis fallback in `da200ebb`.

A corrected staging smoke is pending deployment of the latest commit.

neurons/validators/src/clients/subtensor_client.py
```diff
@@ -436,6 +449,32 @@ async def set_weights(self, miner_scores: dict[str, float], active_hotkeys: set[
             )
             return
 
+        metagraph = self.get_metagraph()
```
arhangel66, line 658: this is **dangerous**. If miner with score in last cycle with get is_service=False it will be excluded from the list of miners and you will get endless critical error. 
Please don't do it. you can add it as dry_run with notificaiton only, but not raise

neurons/validators/src/clients/validator_portal_api.py
```diff
@@ -68,16 +69,24 @@ async def get_opted_in_miners() -> list[dict]:
...
+                                    "Invalid opted-in miners response from portal",
+                                    extra=get_extra_info({"url": url}),
+                                )
+                            )
+                            return None
+                        return data
                 except asyncio.TimeoutError:
                     logger.error(_m("Timeout fetching opted-in miners from portal", extra=get_extra_info({"url": url})))
-                    return []
+                    return None
```
arhangel66: usually return None is bad, if you want to pass unexpected value - pass the raise exception with catch it on upper stage

arhangel66 (changes requested):
Instead of break the validation cyce in case of miner portal not working - use the latest miner portal api response.

like
get_opted_in_miners
try:
    miners = get_miners()
    if not ok
          raise
    save_to_redis()
except:
    miners = get_from_redis()
return miners

----
And be carefull with saving scores logic. Ideally - don't break it.

neurons/validators/src/clients/subtensor_client.py
pixel29913: Fixed in `da200ebb`. The submission-blocking raise is removed. Missing registered positive-score hotkeys are now classified as active or inactive and logged at critical or warning level, but Redis publication, backend mirroring, and the chain submission continue. The regression tests use a metagraph with populated neurons and assert both the classification and the completed submission.

neurons/validators/src/clients/validator_portal_api.py
pixel29913: I kept the `None` sentinel intentionally because this call has two valid states that must not collapse: `None` means the portal response is unavailable or malformed, while `[]` means the portal successfully returned no opted-in providers. There is already the same precedent in `validator.py` for `get_all_rented_executors()`. The return is now typed as `list[OptedInMiner] | None`; the caller falls back to Redis and then memory, and raises `ProviderPortalDataUnavailable` only when all three sources are absent.

pixel29913:
Staging smoke test passed on `6b84d25`.

### Test procedure

1. Confirmed the validator had fetched a healthy portal response and persisted one provider routing record in validator Redis with no TTL.
2. Scaled `deployment/miner-portal-backend` in `staging-miner-portal-backend` from 1 replica to 0, leaving the service with no endpoints.
3. Restarted `deployment/validator` in `staging-subnet` to exercise the cold-process/restart path while the portal was unavailable.
4. Observed the expected portal connection failure and Redis fallback.
5. Restored the portal deployment to its original 1 replica and waited for the validator to return to the live portal source.

### Verified behavior

- The replacement validator loaded the Redis snapshot with `provider_count=1` and `snapshot_source=redis_cache`.
- It rebuilt the same 36-miner set; it did not enter the no-snapshot path or reduce the set to zero.
- The cached provider `5DfbmGgkgYoYKJk7rA9UFgeJFacfkBMhaeFZgJk7c4mCv7DT` was dispatched through its cached central route `3.214.83.201:8000`.
- The provider returned two executors, completed their tasks, and produced a `Job_Result` with two results.
- The validator successfully submitted weights at `20:17:24 UTC` while the portal was unavailable.
- No `No reliable provider snapshot`, cache read/write/validation, missing-scored-miner, or weight-submission errors were emitted.
- After restoration, the validator logged `snapshot_source=portal`, retained 36 miners, and refreshed the persistent Redis snapshot.

### Final state

- Staging validator: 1/1 ready, image `6b84d25`.
- Staging miner portal backend: restored to its original 1/1 ready state and original image.
- Redis snapshot remains present with TTL `-1`.
- No smoke-test resources were left behind.

There were unrelated connection failures from some ordinary staging miners during the cycle; the cached portal-provider path itself completed successfully.

neurons/validators/src/clients/subtensor_client.py
```diff
@@ -323,45 +369,177 @@ def get_my_uid(self):
...
+                        "[fetch_miners] using in-memory provider snapshot",
+                        extra=get_extra_info(
+                            {
+                                **self.default_extra,
+                                "provider_count": len(self.miners),
+                            }
+                        ),
+                    )
+                )
+                return
```
arhangel66, line 541: Portal plus Redis down means the metagraph is never refreshed here, so new registrations stop landing. Could we cache the last opted-in list and rebuild from a fresh metagraph?

arhangel66 (approved):

neurons/validators/src/clients/subtensor_client.py
pixel29913: This is a very rare case and i think its fine to defer this.

PR #1230, opened 2026-08-18
DAH-2703: withhold unrented incentive when a host kills the filler at create time

taiberium:
## Why

ISSUE-050 / DAH-2439 punishes a host that kills a **RUNNING** filler. It does nothing when the container is removed seconds after `docker run`: the run never reaches RUNNING, the backend reports no filler container for the executor, `RentalVerificationCheck` takes the normal path, and the node collects unrented incentive while never running the default job it was assigned.

Prod, 2026-08-16..18: one miner, 6 executors on two hosts, 37 runs killed this way, zero fillers ever running. Timeline for one of them — `23:10:12.932 Created Docker Container` → `23:10:14.315 SDK op failed` → `23:10:15.969 inspect: no such object`. Nothing on our side removed it.

## What this PR does

**Detect** — the create flow now distinguishes "the container we created disappeared from the host" from any other create failure:
- `ContainerDeathDiagnostics.container_missing`, set when docker answers *no such object / no such container*;
- `cleanup_failed_container_creation()` returns that verdict;
- the failure is reported to the backend as `FailedContainerErrorCodes.ContainerVanished`, but **only** if `docker run` had already succeeded — before that there is no container to remove, and a failed pull must never read as a kill. Unproven cases (SSH dead, diagnostics failed) stay `UnknownError`: never accuse a host on missing evidence.

**Punish** — `RentalVerificationCheck` fails the fatal check for an unrented executor the backend lists in `filler_create_kill_executor_ids` (3+ create kills in 24 h), emitting a customer-facing `FILLER_KILLED_AT_CREATE` with remediation. A customer rental on the node is never punished for its default-job history.

Gated by the existing DAH-2439 pair: `FILLER_LIVENESS_CHECK_ENABLED` (master switch) and `FILLER_LIVENESS_ENFORCEMENT_ENABLED` (withholds). One switch now covers both offences — killed at start, killed while running.

Two things the review changed: the streak is judged **before** the liveness probe (a GPU-split node running one surviving bundle would otherwise return a clean verdict for a host that destroys every other bundle it is given), and **shadow mode only logs and falls through** to the normal verification — returning a pass there would have skipped the backend health check the node otherwise has to survive, making shadow an upgrade rather than an observation.

## Deploy order

Backend (Datura-ai/lium-io-backend#931) first — it must accept the new error code and publish the streak. Ships in shadow; enforcement is flipped in the deploy repo afterwards.

## Tests

`test_rental_verification_check.py`: streak → shadow logs / enforcement fails / check-disabled ignores / customer rental exempt. `test_failed_container_diagnostics.py`: vanished vs died-on-its-own, cleanup reports the verdict and still runs, unproven → False. `test_docker_service.py`: create failure maps to the right error code, and a `docker run` failure is never reported as a kill. Full validator suite: 1591 passed (3 pre-existing sshd-bootstrap failures, unrelated).

Notion: DAH-2703

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -2043,9 +2043,8 @@ class DockerService:
     ) -> bool:
         """Remove the failed container's artifacts; report whether it was already gone.
 
-        DAH-2703: "already gone" is the host-reaper signature — the container existed (we created
-        it) and something outside the platform removed it before this cleanup ran. Unproven cases
-        (SSH dead, diagnostics failed) report False: never accuse a host on missing evidence.
+        DAH-2703: unproven cases (SSH dead, diagnostics failed) report False — never accuse a host
+        on missing evidence.
         """
         container_missing = False
         try:
@@ -3873,9 +3872,8 @@ class DockerService:
         log_tag = "container_creation"
         current_step = "start"
         # DAH-2703: the container reached the host and then disappeared from it — the host-reaper
-        # signature, and the error code the failure is reported with. `container_created` is what
-        # keeps it honest: before `docker run` succeeds there is nothing to remove, so a missing
-        # container there is an ordinary create failure, not a kill.
+        # signature. `container_created` keeps it honest: before `docker run` succeeds there is
+        # nothing to remove, so a missing container there is an ordinary create failure.
         container_created = False
         container_vanished = False
         volume_encryption_status = VolumeEncryptionStatus.DISABLED
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -2040,16 +2040,24 @@ class DockerService:
         container_name: str,
         volume_name: str | None = None,
         remove_volume: bool = False,
-    ) -> None:
+    ) -> bool:
+        """Remove the failed container's artifacts; report whether it was already gone.
+
+        DAH-2703: "already gone" is the host-reaper signature — the container existed (we created
+        it) and something outside the platform removed it before this cleanup ran. Unproven cases
+        (SSH dead, diagnostics failed) report False: never accuse a host on missing evidence.
+        """
+        container_missing = False
         try:
             # DAH-2395: read the death evidence before `rm -fv` destroys it —
             # without this line, "why did the container die" (OOM vs entrypoint
             # crash) is unanswerable: the host is the miner's, not ours.
-            await self.capture_failed_container_diagnostics(
+            diagnostics = await self.capture_failed_container_diagnostics(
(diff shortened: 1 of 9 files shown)


neurons/validators/src/services/docker_service.py
```diff
@@ -4782,13 +4797,14 @@ async def create_container(
...
                 except Exception:
-                    await self.cleanup_failed_container_creation(
+                    container_missing = await self.cleanup_failed_container_creation(
                         ssh_client=ssh_client,
                         default_extra=default_extra,
                         container_name=container_name,
                         volume_name=local_volume,
                         remove_volume=created_local_volume,
                     )
+                    container_vanished = container_created and container_missing
```
arhangel66, line 4807: `container_vanished` only means the container is gone, but `clean_existing_containers` force-removes a sibling `filler_*` during a concurrent create on the same executor. Does the backend's create-kill counter exclude removals we did ourselves?

neurons/validators/src/services/task/messages.py
```diff
@@ -1052,6 +1052,22 @@ class RentalVerificationMessages:
             "Contact Lium support if this container was not removed by you."
         ),
     )
+    FILLER_KILLED_AT_CREATE = MessageTemplate(
+        event="Lium default job container is removed on the host right after it is created",
+        reason="FILLER_KILLED_AT_CREATE",
+        severity="error",
+        category="policy",
+        impact=(
+            "Unrented incentive withheld: every Lium default job assigned to this machine is "
```
arhangel66: `impact` says every default job on this machine is destroyed, but the rule fires on a backend streak threshold. Can we word it as repeated create-time kills in the last 24h?

arhangel66 (approved):

neurons/validators/src/services/docker_service.py
taiberium: Checked. Since DAH-2465 the backend lists every ACTIVE filler run in `active_container_names` (pod.py, `_active_filler_container_names`), so a concurrent create on the same executor no longer wipes a sibling. That one was fixed after it happened live on staging.

One path still drops them on purpose: a rent on a profiling-tainted host (executor.py:653) force-removes every filler before the customer's run. If that races an in-flight create we would count it as a kill. It takes three such races in 24h on one node to withhold anything, and the shadow week will show whether it ever happens.

neurons/validators/src/services/task/messages.py
taiberium: Fair, reworded: "Lium default jobs assigned to this machine were destroyed seconds after they started, repeatedly over the last 24 hours".

PR #1232, opened 2026-08-18
DAH-2705: scrape whether the executor container can apply a GPU power cap

taiberium:
On some hosts `nvidia-smi -pl` exits 4 (no permission), so the fail-closed cap path refuses the PEARL filler every cycle and the node just idles — 223 refusals across 22 executors since 2026-07-24.

This collects the two raw readings the backend needs: the executor container's `CapEff` (`/proc/self/status`) and the owner uid of `/dev/nvidiactl`. Both are needed — a sysbox host keeps every capability but maps the device to `nobody` (uid 65534), so neither reading alone separates a healthy host from a blocked one. No verdict on the machine side; unreadable values stay empty/None and the backend fails open on them.

Verified on staging 149.36.1.151: the executor container reads `000001ffffffffff` / uid 0, while a plain unprivileged container on the same host reads `00000000a80425fb` / uid 0 and `nvidia-smi -pl` there fails with `Insufficient Permissions`, exit=4.

Backend side: Datura-ai/lium-io-backend#933 (merge this one first — the gate there is inert until specs carry the probe).

Tests: `tests/test_scrape_gpu_power_cap_probe.py`.

### DIFF
--- a/neurons/validators/tests/helpers.py
+++ b/neurons/validators/tests/helpers.py
@@ -38,6 +38,18 @@ def build_scrape_namespace(
     return namespace
 
 
+def dict_literal_keys(module: ast.Module, dict_name: str) -> list[str]:
+    """Keys of the single dict literal assigned to `dict_name`, in source order."""
+    for node in ast.walk(module):
+        if (
+            isinstance(node, ast.Assign)
+            and getattr(node.targets[0], "id", "") == dict_name
+            and isinstance(node.value, ast.Dict)
+        ):
+            return [key.value for key in node.value.keys]
+    raise AssertionError(f"{dict_name} dict literal not found")
+
+
 class DummyScoreCalc:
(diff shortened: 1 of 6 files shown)


neurons/validators/tests/helpers.py
```diff
@@ -38,6 +38,18 @@ def build_scrape_namespace(
     return namespace
 
 
+def dict_literal_keys(module: ast.Module, dict_name: str) -> list[str]:
```
arhangel66, line 41: nit: `_dict_literal_keys` still lives in `test_scrape_ncu_profiling.py:95` and `test_scrape_infiniband.py:259`, so this makes three copies. Point both at this helper instead? non-blocking, fine as a follow-up.

arhangel66 (approved):

PR #1233, opened 2026-08-18
DAH-2640: Retry temporary backup repository access denial

pixel29913:
## Describe your changes

Retry `AccessDenied` while the executor probes or initializes a newly created backup repository.

- Use a shared, bounded readiness deadline of 180 seconds with exponential backoff.
- Continue reporting heartbeats and honor cancellation while waiting.
- Re-probe after an ambiguous initialization result so an already-created repository is never initialized again.
- Preserve the complete redacted Restic/S3 error if readiness retries are exhausted.

### Root cause

A production user's first backup was dispatched immediately after their S3 bucket, IAM user, and policy were created. The repository probe ran before AWS had propagated the new IAM permissions and failed with `Access Denied`. A later backup using the same pod, repository, and credentials completed successfully.

Restic retries ordinary transient backend failures internally, but deliberately treats `AccessDenied` as permanent. During initial platform-managed IAM setup, that normally permanent error can be temporary while the new policy propagates. This change retries only that known exception during backup repository setup.

### Impact

Normal backups and restores are unchanged. Other S3, credential, repository, and endpoint failures do not receive an additional executor-level retry. This requires publishing the updated executor image; there are no validator protocol, database, IAM, or backend changes.

### Validation

- `ruff check src/storage/restic.py tests/test_storage_runner.py`
- `pytest tests/test_storage_runner.py -q` — 34 passed
- Covered probe recovery, initialization recovery, ambiguous initialization confirmation, heartbeats, cancellation, bounded exhaustion with the redacted cause preserved, and immediate return for non-`AccessDenied` failures including `InvalidAccessKeyId` and HTTP 503.

[DAH-2640: Improve backup and restore reliability to 90% first-attempt success](https://app.notion.com/p/Epic-improve-backup-and-restore-reliability-to-90-first-attempt-success-3b8b8bfdbde981259ae5fa1155239ae1?pvs=23)

- [x] I wrote tests.

### DIFF
--- a/neurons/executor/src/storage/restic.py
+++ b/neurons/executor/src/storage/restic.py
@@ -31,6 +31,28 @@ TAR_RECORD_SIZE_BYTES = 20 * 512
 TAR_CHECKPOINT_RECORDS = 1024
 MEBIBYTE = 1024 * 1024
 RESTIC_TMPFS_BYTES = 512 * MEBIBYTE
+REPOSITORY_RETRY_DELAYS_SECONDS: tuple[float, ...] = (
+    2.0,
+    4.0,
+    8.0,
+    16.0,
+    30.0,
+    30.0,
+    30.0,
+    30.0,
+    30.0,
+)
+REPOSITORY_RETRY_POLL_SECONDS: float = 1.0
+RETRYABLE_REPOSITORY_ERROR_MARKERS: tuple[str, ...] = (
+    "access denied",
+    "request timeout",
+    "slow down",
+    "status code: 429",
+    "status code: 500",
+    "status code: 502",
+    "status code: 503",
+    "status code: 504",
+)
 ENCRYPTED_BACKUP_SCRIPT = 'cd "$1"; shift; exec "$@"'
 ENCRYPTED_BOOTSTRAP_SCRIPT = r"""
 set -eu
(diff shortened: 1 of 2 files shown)


pixel29913:
## Staging smoke-test report

Tested the PR executor image on staging against both a natural first-use race and a controlled temporary-access failure.

### Natural first-user reproduction

- Created a new isolated staging account with no pre-existing backup resources.
- Created its first backup configuration and triggered the first backup immediately.
- The backup IAM identity and the backup operation were created/started within the same second, reproducing the production timing window without manually modifying IAM.
- The executor received genuine temporary S3 access failures during the repository probe:

  ```text
  Transient repository probe failure; retrying in 2s: ... Access Denied.
  Transient repository probe failure; retrying in 4s: ... Access Denied.
  ```

- The same operation recovered automatically once the new credentials/policy became usable. It did not require a restart or a second backup request.
- Result: backup completed at 100% in 18 seconds for one 8 MiB file, with no final error.
- Restored that snapshot through the normal product flow. Restore completed at 100% in 11 seconds, and the restored file SHA-256 matched the source exactly.

This directly exercises the user-reported failure class: the first repository access happens before newly issued AWS permissions have fully propagated.

### Controlled temporary-access test

- Captured the test identity's original IAM state, then temporarily removed only its test bucket policy attachment.
- Started a backup while access was unavailable.
- The same operation retried after 2s, 4s, 8s, 16s, then at the capped 30s interval while continuing to report heartbeats.
- Restored the unchanged policy attachment.
- The in-flight operation recovered and completed without being restarted.
- A normal follow-up backup also completed, and a restore from the recovered snapshot matched the source checksum.
- Verified the IAM state after rollback matched the captured pre-test state.

### Cleanup

All disposable test resources were cleaned up after verification: snapshots, backup configuration, pod, backup volume, generated S3 bucket, and generated IAM identity/policy. The isolated account was then disabled. No credentials or reusable identifiers are included in this report.

### Result

**Pass.** The retry handles both the natural first-user IAM propagation race and a longer temporary AccessDenied period, while preserving operation liveness and backup/restore integrity.

arhangel66 (approved):

PR #1234, opened 2026-08-18
DAH-2828, feat(validator): quote path for customer pods on CVM nodes (dstack socket behind a quote-only broker)

fortunelucky777:
## Why

A renter running a TEE workload on a Confidential Computing node expects the dstack guest-agent socket at `/var/run/dstack.sock` **inside the pod** (dstack SDK default: `DstackClient().get_quote(report_data)` / `.info()`). On a CVM node the pod is a container on the guest's docker, created by the validator; the executor container already has that socket, the rental container never did.

The raw socket cannot be mounted as-is (review of 2026-08-19): besides `GetQuote`/`Info` the guest agent serves `GetKey`/`GetTlsKey`/`Sign` (derive the guest's app keys, shared by the executor and every pod on the node) and `EmitEvent` (extends RTMR3).

## What

Validator-only change — no executor release, no compose_hash / RTMR3 change, no `TDX_WHITELIST` entry, no CVM redeploy on provider nodes.

- `services/cvm_quote_broker.py` (new): one nginx container per CVM guest, `lium-dstack-quote-broker`, image pinned by digest, config passed through an env var and written by a `sh` entrypoint. It listens on `/var/run/lium-dstack/dstack.sock` and proxies only `GetQuote`, `Info`, `Version` to `/var/run/dstack.sock`; every other path gets HTTP 403. `restart: unless-stopped`, no GPU, no caps.
- `ensure_quote_broker()`: `docker inspect` fast path (running at this release's revision → nothing to do), else `rm -f` + pull-if-missing + run, then wait until the socket exists (a pod bind issued earlier would make dockerd create a directory at the path). Revision = sha256(image + entrypoint + nginx.conf), so a broker from an older validator release is replaced, not trusted.
- `create_container`: for `CUSTOMER_RENTAL` on an executor with `tdx_quote` (the existing CVM marker) the validator ensures the broker, then mounts the broker socket **read-only** at `/var/run/dstack.sock` in the pod. Best-effort: broker failure → pod still created, warning streamed to the pod log (`TDX quote socket unavailable in this pod`), `logger.error` on the validator. Fillers and bare metal are untouched.
- Kill switch `ENABLE_CVM_POD_QUOTE_SOCKET` (default `True`).

## Tests

- `tests/test_cvm_quote_broker.py` (new): allow-list regex vs. every SDK method, nginx conf wiring, run spec (digest pin, both binds, entrypoint), revision follows the conf, `ensure_quote_broker` (fast path, cold start + pull + socket wait, stale revision replaced, parallel-create race, start failure, socket never appears), pod mount, run-spec builder, `_wants_quote_socket` matrix.
- `tests/test_docker_service.py`: `create_container` gates — CVM customer rental brokers the socket (raw socket never mounted, `--cpus` quirk preserved), broker failure still rents with a warning, bare metal, filler on CVM, kill switch.
- Verified on local docker: nginx creates the listen socket 0666; a non-root client over a read-only bind connects; `/GetQuote`, `/Info`, `/prpc/DstackGuest.Info` proxy, `/GetKey`, `/EmitEvent`, `GET /` → 403.

## Rollout

See the Notion task (DAH-2828, "Rollout"): merge → `prod_deployment` → canary rental on a free CVM node with `dstack-sdk` from inside a sysbox pod → reply to the renter → merge lium-docs `DAH-2828-cvm-pod-quote-socket-docs`.

## Known limitations

- File bind: if the broker container restarts, pods started before keep the old socket inode until they restart. A guest reboot re-registers the executor and ends the rental anyway.
- Pods on a split CVM node share one broker (stateless).
- The validator-side `report_data` binding (any quote whose report_data starts with `sha256("SSH_HOST_KEY:"+host_key)` is accepted) is a separate task; enforcement flags are off in prod.
- A renter probing `/dev/tdx_guest` instead of the dstack socket gets nothing here; device passthrough would be a small follow-up in the same run spec if a customer needs it.


ivangors:
when can we merge the pr?

fortunelucky777:
Screenshot from 2026-08-19 18-22-17

Haven't  tested with real CC node but it's working fine on the node in staging.

neurons/validators/src/services/docker_service.py
```diff
@@ -444,6 +444,25 @@ def _build_cache_volume_mounts(payload: ContainerCreateRequest, occupied_targets
...
+
+def _build_dstack_socket_mounts(payload: ContainerCreateRequest, *, in_cvm: bool) -> list[VolumeMount]:
+    # Only a customer rental on a dstack CVM guest gets the socket. `in_cvm` is the validator's
+    # existing CVM marker (the executor answered the SSH-key upload with a TDX quote it could only
+    # have obtained through this very socket) — outside a CVM the source path does not exist and
+    # dockerd would silently create an empty DIRECTORY there. A FILLER (DPHN/PEARL) has no
+    # attestation use and the socket is one shared per guest, so it never receives it.
+    if not in_cvm or payload.workload_kind != WorkloadKind.CUSTOMER_RENTAL:
+        return []
+    return [VolumeMount(source=DSTACK_GUEST_SOCKET_PATH, target=DSTACK_GUEST_SOCKET_PATH)]
```
arhangel66: Only report_data binds a quote to a machine (attestation_service.py:565), so a pod that sets it freely can mint quotes for any host key. Broker in front that rewrites report_data?

arhangel66 (changes requested):
See the inline comments.

neurons/validators/src/services/docker_service.py
fortunelucky777: Ty for your comment. And, it totally makes sense.

arhangel66:
Taking this over under DAH-2828 (Notion: Rental pods on CVM nodes cannot produce a TDX quote). Plan: keep the validator-side mount and the tdx_quote / CUSTOMER_RENTAL gate as they are, and put a quote-only broker in front of the socket (GetQuote + Info only; GetKey / GetTlsKey / EmitEvent never reach the pod). The report_data binding raised in review is tracked as a separate task before attestation enforcement is enabled — rewriting report_data in the broker would break the very verifiers this path exists for.

arhangel66 (approved):

arhangel66:
**Verification status before merge**

Done:
- Unit tests: `tests/test_cvm_quote_broker.py` (allow-list, run spec, revision, `ensure_quote_broker` paths) and the `create_container` gates in `tests/test_docker_service.py`. CI green.
- Local docker: nginx creates the listen socket, a non-root client connects over a read-only bind, `GetQuote`/`Info`/`Version` proxy, `GetKey`/`EmitEvent`/`GET /` return 403.
- Independent review pass; findings fixed (agent socket mounted read-only, socket wait on both paths, bounded wait).

Not done yet:
- No real CVM node in staging, so the end-to-end path (validator -> broker on the guest -> pod -> dstack-sdk -> verifier) has not been exercised. This is the canary right after deploy: rent a free CVM node, `get_quote` from inside the pod, verify the quote against `dstack-verifier` (same service the validator uses), confirm `get_key` is 403.

Rollback: `ENABLE_CVM_POD_QUOTE_SOCKET=false` + redeploy; a broker failure never blocks pod creation.

neurons/validators/src/services/docker_service.py
```diff
@@ -4673,7 +4716,18 @@ async def create_container(
...
                 # --cpus flag isn't working inside cvm. skip to use it when tdx_quote is present
                 # TODO: remove this when cvm is fixed
-                cpu_count = None if executor_info.tdx_quote else payload.cpu_count
+                in_cvm = bool(executor_info.tdx_quote)
+                cpu_count = None if in_cvm else payload.cpu_count
+
+                quote_socket = False
+                if _wants_quote_socket(payload, in_cvm=in_cvm):
+                    current_step = "cvm_quote_broker"
+                    quote_socket = await self._ensure_pod_quote_socket(
```
taiberium, line 4725: NIT: the broker cold start (image pull plus the socket wait) lands inside the `PORT_CHECK_WAIT` profiler step. Reset `prev_timestamp` after this block.

neurons/validators/src/services/docker_service.py
```diff
@@ -1047,6 +1063,33 @@ def _build_rental_container_run_spec(
             entrypoint=custom_options.entrypoint,
         )
 
+    async def _ensure_pod_quote_socket(
+        self,
+        *,
+        docker_client,
```
taiberium: NIT: `docker_client` has no type. `RentalDockerSdkClient` is already imported here.

```suggestion
        docker_client: RentalDockerSdkClient,
```

taiberium (approved):

neurons/validators/src/services/docker_service.py
arhangel66: Fixed in a953a844e5d0ab85303c8810d23337cd15260b38.

neurons/validators/src/services/docker_service.py
arhangel66: Fixed in a953a844e5d0ab85303c8810d23337cd15260b38.

arhangel66:
**Prod canary after deploy**

Validator `096534c3a33a9576ea9bb57b4e0d7971c0470063` (run [33626044125](https://github.com/Datura-ai/lium-io-deployment/actions/runs/33626044125)), pod `validator-647796ddf5-g2m2m` ready, 0 restarts.

Rented a real CVM node (executor `295f6cbd-82e1-401f-9390-f38b33ed0e89`, 8×H200, pod `46ed440f-43c3-4ff9-83a3-02c1bd04d855`, default PyTorch template):

- `/var/run/dstack.sock` present in the pod; `/proc/cpuinfo` model `06/cf` (TDX guest).
- `dstack-sdk` `info()` → app_id `8224d58801af6333561f116e2d566b179b399f1d`, compose_hash `8224d58801af6333561f116e2d566b179b399f1d1d700f0e8a5ab9326ae901d9`.
- `get_quote(b"lium-canary-dah2828")` → 17582-byte response; verified with the prod `dstack-verifier` (the same service the validator uses): `is_valid=true`, `quote_verified=true`, `event_log_verified=true`, `os_image_hash_verified=true`, `tcb_status=UpToDate`, `report_data` = the nonce, os_image_hash `a6eafc5f007f642d8ea90c7fa8881f1e6715720ccb531941a28218f4f26d7b02`.
- `Version` → 200 `{"version":"0.5.11"}` through `nginx/1.27.5`; `GetKey` and `GET /` → 403 `{"error":"lium quote broker: only GetQuote, Info, Version reach the dstack guest agent from a pod"}`.
- Validator log: `Starting the CVM quote broker` with `broker_revision=ee963af46c77fcf6`; no `quote broker unavailable` lines since the deploy. Two other customer rentals on CVM nodes (11:48 and 11:52 UTC) got the broker the same way.

Pod deleted after the check. Docs: Datura-ai/lium-docs#151 merged, v1.0.80.

PR #1235, opened 2026-08-19
DAH-2706: fix jlab token leak

jam6099:
## Describe your changes
- Stop leaking the renter Jupyter token in logs and argv

[DAH-2706](https://app.notion.com/p/Renter-Jupyter-token-is-readable-by-the-host-provider-container-logs-docker-inspect-ps-3c0b8bfdbde98180834ac51d95ee66fd?v=34ab8bfdbde980eb8c88000c8ab9f75f&source=copy_link)

- [ ] I wrote tests.

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -1336,6 +1336,7 @@ class DockerService:
         log_extra: dict = {},
         timeout: int = 0,
         raise_exception: bool = True,
+        stdin_data: str | None = None,
     ) -> tuple[bool, str]:
         logger.info(
             _m(
@@ -1354,6 +1355,9 @@ class DockerService:
         error = ''
         try:
             async with ssh_client.create_process(command) as process:
+                if stdin_data is not None:
+                    process.stdin.write(stdin_data)
+                    process.stdin.write_eof()
                 if timeout != 0:
                     status, error = await asyncio.wait_for(self._stream_process_output(process, log_tag), timeout=timeout)
                 else:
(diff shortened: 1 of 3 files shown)


neurons/validators/src/services/docker_service.py
```diff
@@ -2959,8 +2963,8 @@ async def run_jupyter(
             )
 
             command = (
-                f"/usr/bin/docker exec -u 0 {container_q} sh -c "
-                f"{shlex.quote(f'{script_q} --password={jupyter_token} --port={jupyter_port}')}"
+                f"/usr/bin/docker exec -i -u 0 {container_q} sh -c "
+                f"{shlex.quote(f'read JUPYTER_PASSWORD; export JUPYTER_PASSWORD; {script_q} --password=$JUPYTER_PASSWORD --port={jupyter_port}')}"
```
arhangel66: `export JUPYTER_PASSWORD` is a no-op: run_jupyter.sh:14 resets it to `""`, so the token still reaches the script only via `--password=` in argv. Drop the export, or make the script honour the env var instead.

```suggestion
                f"{shlex.quote(f'read JUPYTER_PASSWORD; {script_q} --password=$JUPYTER_PASSWORD --port={jupyter_port}')}"
```

neurons/validators/tests/test_jupyter_command_quoting.py
```diff
@@ -55,8 +55,6 @@ async def test_renter_volume_path_stays_one_token_at_both_layers(encrypted: bool
...
     for command in exec_commands:
         inner_script: str = _inner_script_of(command)
-        # punctuation_chars=True makes shlex surface ';' as its own token, the way
-        # the container's sh would — without it an injected path looks like one word
-        inner_tokens: list[str] = list(shlex.shlex(inner_script, punctuation_chars=True))
-        assert ";" not in inner_tokens, f"path broke out into commands: {inner_script!r}"
-        assert HOSTILE_PATH in inner_script, inner_script
+        # the renter path must stay inside single quotes so its ';' never splits commands
+        assert f"'{HOSTILE_PATH}" in inner_script, f"path not quoted: {inner_script!r}"
+        assert "tok" not in command, "jupyter token leaked into the exec command"
```
arhangel66, line 62: Nothing asserts the new stdin path here, the recording double swallows `stdin_data` in `**_`. Can you record it and check both, since losing `-i` fails silently (the script errors on stdout only, so `status` stays `True`)?

```python
assert " -i " in command, "docker exec needs -i or the container's `read` gets EOF"
assert stdin_data == "tok\n"
```

arhangel66 (approved):

PR #1237, opened 2026-08-19
DAH-2715: withhold the unrented incentive from executors that cannot apply a GPU power cap

taiberium:
[DAH-2715](https://app.notion.com/p/Penalize-executors-that-cannot-apply-a-GPU-power-cap-no-CAP_SYS_ADMIN-device-not-root-owned-3c1b8bfdbde981e0bd08f763ca1f0304)

Since DAH-2705 every scrape carries whether the executor container can apply a GPU power cap. A node that cannot is silently skipped for capped fillers (PEARL) but keeps its unrented incentive, so the provider has no reason to fix the host.

This adds a fourth unrented gate next to the price / disk / flagship ones: an idle executor whose container provably cannot cap forfeits the unrented incentive and stays active. The miner is told why and how to fix it (`privileged: true` / `CAP_SYS_ADMIN`, and `/dev/nvidiactl` owned by root — the sysbox case needs the executor out of the user namespace).

**Enforced on merge.** `ENABLE_UNRENTED_POWER_CAP_LIMIT` defaults to `True` — the provider fix is one compose flag, so there is no shadow window. Set the env to `false` to fall back to logging the breach without withholding anything.

**Blast radius (prod, 2026-08-20):** 379 active executors probed — 371 capable, 3 incapable, 5 with no probe (never penalized, the gate fails open on every unknown). Of the 3, one is rented (untouched) and two earn idle incentive today: `178.105.229.97` (4x RTX PRO 6000, 0.0617 α / 24 h) and `88.22.127.152` (RTX 5090, sysbox uid 65534, 0.0057 α / 24 h).

**Staging-verified end to end** on a rented A6000: capable → routed to the unrented pool, gate silent; the same node recreated without `privileged` (CapEff `00000000a80425fb`) → excluded from the pool, cycle stored with `incentive=0`, `reason=cannot_apply_gpu_power_cap` and the full fix text, which is the row the miner portal renders. Shadow mode (flag off) was verified too: breach logged, incentive kept.

Tests: `pytest tests/test_unrented_power_cap_capability.py` (29) + full validator suite green.

### DIFF
--- a/neurons/validators/src/incentive/rental_price.py
+++ b/neurons/validators/src/incentive/rental_price.py
@@ -413,18 +413,38 @@ class RentalPriceIncentive(DefaultIncentive):
         # reading, not a breach. bool is excluded explicitly - it passes isinstance(int)
         # and would otherwise read as uid 1, i.e. "not root", i.e. a penalty.
         if not isinstance(cap_eff, str) or not isinstance(owner_uid, int) or isinstance(owner_uid, bool):
+            self._log_power_cap_unmeasured(result, "cap_eff or nvidiactl owner missing from the scrape")
             return None
         try:
             capability_mask: int = int(cap_eff, 16)
         except ValueError:
+            self._log_power_cap_unmeasured(result, f"unreadable capability mask: {cap_eff!r}")
             return None
         has_sys_admin: bool = bool(capability_mask >> CAP_SYS_ADMIN_BIT & 1)
         if has_sys_admin and owner_uid == NVIDIACTL_ROOT_UID:
             return None
         return PowerCapIncapable(container_cap_eff=cap_eff, nvidiactl_owner_uid=owner_uid)
 
+    def _log_power_cap_unmeasured(self, result: JobResult, cause: str) -> None:
+        # gate could not run: a shadow report must not read this executor as "can be capped".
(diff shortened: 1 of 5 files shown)


neurons/validators/src/incentive/rental_price.py
```diff
@@ -762,6 +843,23 @@ async def calculate_executor_score(
                 )
                 job_result.record_incentive_log(reason)
 
+        # DAH-2715 power cap gate: an idle machine whose container cannot apply a GPU power
+        # cap is not fully usable for Lium's own jobs, so it forfeits the unrented incentive
+        # (node stays active). While the flag is off we only log the would-be exclusion.
+        # Last in the chain, so a node already excluded by an ENFORCED gate above is not
+        # measured here - read the shadow numbers against the flags that were on that cycle.
+        power_cap_incapable: PowerCapIncapable | None = (
```
arhangel66, line 851: PEARL is the only capped filler and its `unsupported_gpu_names` skips RTX PRO 6000 and B300, but this gate still withholds their unrented incentive. Scope it by model or reframe the rule before the flag flips? Non-blocking for the shadow merge.

arhangel66 (approved):

PR #1239, opened 2026-08-20
DAH-2728: cancel an in-flight create when its delete arrives

taiberium:
## Problem

A delete that lands while the create for the same pod is still in flight finds no container yet, so `_is_missing_docker_container_error` reads "No such container" as "already gone" and the pod is reported deleted. Nothing stops the create: it finishes minutes later, binds the host ports, and the orphan `filler_*` container blocks the next paying rental with `port is already allocated`.

Reference incident (2026-08-19, executor `d4093d0c`): filler create starts 18:35:03, delete at 18:35:09 skips the teardown, the create completes 18:37:36 — the customer pod burns 13 × `PORT_ALREADY_ALLOCATED_RETRY`, dies at 18:39:09 and only gets the H100 x8 back at 19:18:40.

## Fix

Only the create can undo itself, so the delete no longer guesses:

- `inflight_creates` — a small in-process registry of creates running right now. `MinerService.handle_container` registers the create before the miner handshake; `delete_container` flags the matching create instead of inferring absence.
- The create checks the flag at three checkpoints: before the image pull/build (nothing on the host yet, and this is where a cancelled create would otherwise spend its minutes), before the ports get bound, and after the container is up — the last one just before the pod is cached as rented. Each aborts through the cleanup path that is already there (`cleanup_failed_container_creation`) and fails the create with `failure_step=cancelled_by_delete`.
- A cancelled FILLER also restores its GPU power limits — the DAH-2356 cap is applied before `docker run`, and the delete that cancelled it had no record to restore.

A delete with no create in flight changes nothing, so a later create for the same pod id (edit-pod) is untouched. The registry is deliberately in-memory, not Redis: a create and its delete run in the same validator event loop, and a Redis outage must not break the cancellation.

The backend already expects this message — `DefaultJobLauncher.handle_failed` names "create failure of a preempted run" as expected fallout and ignores it for a terminal run, so the normal ordering (delete callback first) costs the provider nothing. If the create-failed callback wins the race the run is still STOPPING and becomes STOP_FAILED: a 15-min soft retry delay, not a launch strike.

## Not in this PR

- `wait_for_port_check_containers` still ignores `filler_*`. Adding the prefix would let any customer create force-remove a live filler on the box, skipping the power-cap restore and the wedged-GPU sweep — worth a separate decision.
- `renting_in_progress` still filters by pod id, not by executor.
- `redis_service.add_pending_pod` is a second, overlapping "a create is in flight" fact. It is registered too late (inside `create_container`, after the handshake) and its presence is load-bearing for the `RentingInProgress` decline gate, so it cannot carry the cancel bit — the two are deliberately separate.
- Residual race: a create that passes the last checkpoint in the same instant the delete probes for the container. Milliseconds now, minutes before.

## Tests

`tests/test_delete_cancels_inflight_create.py` — 5 new tests. Full validator suite: 1629 passed, 9 skipped (`test_sshd_bootstrap_script.py` fails the same way on a clean tree here — local env).

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -10,7 +10,7 @@ import secrets
 import shlex
 import time
 from collections.abc import Iterator
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Annotated, Any
 from uuid import UUID, uuid4
@@ -387,6 +387,7 @@ class _InflightCreate:
     # entry outlives any single one of them and only the last to leave clears it.
     running: int = 0
     cancelled_by_delete: bool = False
+    done: asyncio.Event = field(default_factory=asyncio.Event)
 
 
 class _InflightCreateRegistry:
@@ -412,6 +413,7 @@ class _InflightCreateRegistry:
             create.running -= 1
             if create.running <= 0:
                 self._creates_by_pod_id.pop(pod_id, None)
+                create.done.set()
 
     def cancel(self, pod_id: str) -> bool:
         """Flag the in-flight create for this pod. False when no create is running."""
@@ -428,10 +430,26 @@ class _InflightCreateRegistry:
     def is_running(self, pod_id: str) -> bool:
         return pod_id in self._creates_by_pod_id
 
+    async def wait_until_done(self, pod_id: str, timeout: float) -> bool:
+        """Wait for the create(s) on this pod to leave. False when the wait ran out."""
+        create = self._creates_by_pod_id.get(pod_id)
+        if create is None:
+            return True
+        try:
+            await asyncio.wait_for(create.done.wait(), timeout)
+        except asyncio.TimeoutError:
+            return False
+        return True
+
 
 # In-process: a pod's create and delete are driven by the same validator event loop. Move it to
 # Redis if the two ever land in different processes.
 inflight_creates = _InflightCreateRegistry()
+
+# How long a delete waits for the create it just cancelled. The create reads the flag at its next
+# checkpoint, and the only checkpoint gap that can orphan a container is the short one before
+# `docker run` — a pull-length wait would hold the customer's delete for nothing.
+CANCELLED_CREATE_ABORT_TIMEOUT_SECONDS = 30.0
 
 
 def _is_missing_docker_container_error(exc: Exception) -> bool:
@@ -5803,6 +5821,14 @@ class DockerService:
         # refuses cannot kill a healthy create.
         if inflight_creates.cancel(payload.pod_id):
             log.info("Cancelled the in-flight create for this pod")
+            # Let the create finish tearing itself down before this teardown runs, so the last
+            # actor on the host is never a create that ran after we answered "deleted". On timeout
+            # we tear the container down ourselves, which is what this delete did before DAH-2728.
+            create_aborted = await inflight_creates.wait_until_done(
+                payload.pod_id, CANCELLED_CREATE_ABORT_TIMEOUT_SECONDS
+            )
+            if not create_aborted:
+                log.warning("The cancelled create is still running; deleting without it")
 
         private_key = self.ssh_service.decrypt_payload(keypair.ss58_address, private_key)
         pkey = asyncssh.import_private_key(private_key)
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -381,6 +381,14 @@ class _CreateCancelledByDelete(Exception):
     """Raised at a create checkpoint when this pod's delete has already reported it gone."""
 
 
+@dataclass
+class _InflightCreate:
+    # A retried rent can put a second create on the same pod while the first still runs, so the
+    # entry outlives any single one of them and only the last to leave clears it.
+    running: int = 0
+    cancelled_by_delete: bool = False
+
+
 class _InflightCreateRegistry:
     """Creates running right now, and whether a delete has already cancelled them.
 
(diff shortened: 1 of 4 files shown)


taiberium:
Tested live on staging — validator running this branch (deployed sha `8984c28`), board held 06:52–07:33 UTC.

Reproduced the race on the A4000 node: freed it, then rented it ~3s after the filler launcher's 5-minute boundary, so the preemption delete landed while the filler create was still running.

```
07:15:00,623  Create Docker Container                       (filler 71a68d7c create starts)
07:15:04,182  Cancelled the in-flight create for this pod   (delete side)
07:15:05,779  Create cancelled by an in-flight delete       (checkpoint before docker run)
07:15:07,351  Failed create_container  step=cancelled_by_delete
07:15:08,470  Deleted Docker Container
```

What this proves: the cancel crosses from the delete's task to the create's task end to end, the create aborts itself at the pre-run checkpoint, its cleanup removes the volume (the delete's own `remove_volume_local` then 404s, non-fatal), no `filler_*` container is left on the host, and the customer pod comes up 12s later with zero `PORT_ALREADY_ALLOCATED_RETRY`. The filler run ends STOP_FAILED — a 15-minute soft retry, not a 12h launch strike.

What it does not prove: that this particular run would have orphaned on master. The staging filler image is cached, so the create window is ~1.6s (measured on the untouched 07:10 launch) and the delete's force-remove landed 0.94s after the checkpoint — master's `docker rm -f` would most likely have caught the container. The prod incident needed a 2.5-minute window from an uncached image pull. The mechanism is verified here; the long-window case stays covered by the unit tests.

Separate finding, not fixed here: on the customer path the backend finalizes a pod locally and never sends a delete to the subnet when `pod.container_name` is empty (`apps/server/src/services/pod.py:341`, "there is nothing for the validator to delete") — which is untrue while a create is in flight. Watched it live: pod deleted at 07:05:17, the validator knew nothing and created the container at 07:08:16. Nothing reaches the validator there, so this PR cannot cancel it — worth its own ticket.

taiberium:
Backend half is up: [lium-io-backend#946](https://github.com/Datura-ai/lium-io-backend/pull/946). It derives `pod_{pod_id}` itself, so the two PRs no longer have a release order — the derivation here stays as the boundary guard for a delete that arrives with no name at all.

taiberium:
Second staging run, this time on the A6000 (`ea343cba`, PEARL filler) with the current head deployed (sha `48f9d52b`). Rented the node 5s after the filler run row appeared, so the delete landed after the container was already up.

```
08:25:02,573  gpu power limit cap ok            (filler caps the A6000 to 210W)
08:25:03,468  Created Docker Container
08:25:06,349  Cancelled the in-flight create for this pod
08:25:08,521  Failed create_container  step=set_environment
08:25:08,596  gpu power limit restore ok
08:25:09,274  Deleted Docker Container
```

Outcome: no filler container left, the card back at 300W (its default), the customer pod up, zero `PORT_ALREADY_ALLOCATED_RETRY`, filler run STOP_FAILED — a 15-minute soft retry, not a launch strike.

Worth being precise about what this proves. The cancel reached a create that was already past `docker run`, and the teardown was clean. It did **not** exercise the third checkpoint: the delete's own `docker rm` reached the container first, so the create died at its next SSH step (`step=set_environment`) instead. That is the normal shape once a container exists — the third checkpoint only matters when the delete's teardown ran before the container appeared, which cannot be forced from outside the process. Same for the first checkpoint: it runs ~0.9s into the create, and a delete cannot travel backend→validator that fast. Both stay covered by unit tests.

One consequence for the post-deploy check: in this interleaving the Loki signal is `Cancelled the in-flight create for this pod`, not `failure_step=cancelled_by_delete`. Grep the first one.

neurons/validators/src/services/docker_service.py
```diff
@@ -5686,6 +5798,12 @@ async def delete_container(
             log.error("Invalid Docker volume name", error=str(exc))
             return self._failed_delete(payload, msg="Invalid Docker volume name")
 
+        # DAH-2728: a create still in flight would otherwise finish behind our back and leave an
+        # orphan container holding the ports. Past the validation above, so a delete this validator
+        # refuses cannot kill a healthy create.
+        if inflight_creates.cancel(payload.pod_id):
+            log.info("Cancelled the in-flight create for this pod")
```
arhangel66, line 5823: The problem is that the delete function currently calls "cancel" on the create function and hopes it will stop creating and clean up afterward, without checking the result. The response from the "delete" function may indicate "deleted," but this is misleading - it actually just means "I requested to stop creating."

The proposed solution is for the delete function to request the cancellation of the creation process, wait for the result with a timeout, and then react based on the outcome.

```
async def delete():
    if inflight_creates.is_running():
        cancel_result = await cancel_creation(timeout=1800)
        if cancel_result.success:
            return success

    delete_created_container
```

arhangel66 (changes requested):
look to the comment

neurons/validators/src/services/docker_service.py
taiberium: Fair point, fixed in 9d7303da. You are right that the ack was optimistic: the delete flipped the flag and then answered "deleted" while the create was still alive, so the backend could hand the node to the next renter with the old create still running.

The delete now waits for the create to actually leave before it tears anything down. I kept the wait at 30s instead of 1800s. The create reads the flag only at a checkpoint, and the long gap between checkpoints is the image pull, but a create inside a pull has no container yet, so nothing can be orphaned there. The only gap that can orphan is the short one right before `docker run`. A 30-minute wait would hold a customer's delete through a 34 GB pull and buy nothing. If the wait runs out we log a warning and tear the container down ourselves, which is exactly what the delete did before this PR.

Three tests cover it: no create running, the create leaves in time, the create never stops.

arhangel66 (approved):

PR #1240, opened 2026-08-20
DAH-2730: a filler that exited on its own is not a host kill

taiberium:
## Why

The ISSUE-050 liveness probe decides "the host killed our filler" from `docker ps -q -f name=<container>` alone (`core/docker_utils.py:92`), and that lists **running** containers only. A container still sitting on the host that exited by itself is indistinguishable from one that was removed — so flipping `FILLER_LIVENESS_ENFORCEMENT_ENABLED` today would withhold unrented incentive from nodes that removed nothing.

Prod shadow data, 48 h to 2026-08-20 — 7 flagged runs on 7 executors:

- **2 real removals**: diagnostics `inspect: error: no such object`, both on `178.105.229.97` (`5EjovEqPR…`), later closed by the reconciler as "container is gone".
- **5 false positives**: `container_status=exited, exit_code=255, OOMKilled=false` — the object is still there. One carries docker's own `error while mounting volume … VolumeDriver.Mount`, and its log tail shows the miner working normally right up to the exit. All 5 were then closed by **our own** PEARL power-cap guard (278 such runs on 36 executors in 7 days).

Docker is unambiguous here: `docker stop` exits 143, `docker kill` 137, a removed container has no state at all. Exit 255 with a recorded `State.Error` is the container's own failure.

## What this PR does

`_filler_killed_result` now judges the diagnostics it already collects instead of ignoring them:

Every `State.Status` docker can report is mapped explicitly — a state falling through to a default is exactly how this check mislabeled deaths:

- **absent from the host, or `removing`** → `FILLER_CONTAINER_MISSING`, unchanged (fails the fatal check under enforcement). `removing` is the offence caught a moment early; our own ISSUE-050 data has 5 create-time kills in that state.
- **`exited`** → new `FILLER_CONTAINER_EXITED`, passes, carrying status / exit code / `State.Error` so the population stays observable.
- **`running`, `created`, `restarting`, `paused`, `dead`, or unreadable** → `FILLER_STATE_UNKNOWN`, no penalty. Fillers carry `restart: unless-stopped`, so docker can bring one back between the ps probe and the inspect; `dead` means docker failed to remove it, which does not prove the host succeeded. Unproven is never an offence.

No new evidence is gathered: `container_missing` has been on `ContainerDeathDiagnostics` since DAH-2703; this only stops throwing it away.

## Tests

`test_rental_verification_check.py`: exited container passes under enforcement with the exit code on the event; a removed container still fails; unreachable diagnostics never penalize. The filler SSH mock now answers `docker inspect` for both shapes. Full validator suite: 1665 passed (3 pre-existing sshd-bootstrap failures, unrelated).

## Rollout

This is the gate for the enforcement flip. After it ships, re-run the 48 h shadow count: `FILLER_CONTAINER_MISSING` should keep only the real removals, and the power-cap-stopped nodes should show up as `FILLER_CONTAINER_EXITED`.

Notion: DAH-2730

### DIFF
--- a/neurons/validators/src/services/task/checks/rental_verification.py
+++ b/neurons/validators/src/services/task/checks/rental_verification.py
@@ -48,10 +48,15 @@ _GPU_RUNTIME_QUARANTINE: dict[str, GpuRuntimeQuarantine] = {
 # bundle but the pipeline emits a single event, so the node must be judged by its worst container:
 # a confirmed kill outranks an indeterminate probe, which outranks a healthy one. Anything
 # unlisted sorts as healthy.
-# States `docker inspect --format '{{json .State}}'` reports. "removing" is a removal already in
-# flight — the offence itself, caught a moment early (seen in prod on the ISSUE-050 class).
-_DOCKER_STATUS_RUNNING = "running"
+# Every state `docker inspect --format '{{json .State}}'` can report, mapped to what it proves about
+# the filler. Listed exhaustively on purpose: a state that fell through to a default is how this
+# check mislabeled deaths before (DAH-2730).
+# "removing" is the offence itself caught a moment early — a removal already in flight (observed in
+# prod on the ISSUE-050 class). "dead" means docker failed to remove the container, so it is not
+# proof the host succeeded either — it stays unproven rather than punished.
 _DOCKER_STATUS_REMOVING = "removing"
+_DOCKER_STATUS_EXITED = "exited"
+_DOCKER_STATES_PROVING_NO_DEATH = frozenset({"running", "created", "restarting", "paused", "dead"})
 
 _FILLER_VERDICT_SEVERITY: dict[str, int] = {
     Msg.FILLER_KILLED.reason: 3,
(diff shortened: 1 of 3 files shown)


neurons/validators/src/services/task/checks/rental_verification.py
```diff
@@ -498,28 +512,59 @@ async def _filler_killed_result(
...
+                    ctx,
+                    filler_container,
+                    reason=(
+                        f"container state {diagnostics.status} is not a death"
+                        if diagnostics.status in _DOCKER_STATES_PROVING_NO_DEATH
+                        else "container state could not be read"
+                    ),
+                    details=diagnostics.to_log_fields(),
+                )
+            event = render_message(
```
arhangel66, line 587: Every `exited` passes regardless of exit code, so `docker stop` (143) and `docker kill` (137) become free ways to drop the filler. Split them out before the enforcement flip? Non-blocking.

arhangel66 (approved):

neurons/validators/src/services/task/checks/rental_verification.py
taiberium: Good catch. Checked prod for the last 7 days: 40 flagged runs, 15 with the container removed, 16 exit 255, 5 exit 128, 4 exit 143. All four 143s come from the one miner running a container auto-killer (5HKyXqfZ), none from our own stops. Ours never reach this branch anyway, because the run leaves RUNNING before the probe gets there.

So 137 and 143 now count as a kill, except when OOMKilled is set. Fillers run with `restart: unless-stopped`, so a container still sitting in `exited` after a signal was stopped on purpose. dc760bf4

PR #1242, opened 2026-08-20
DAH-2735: fail idle executors holding foreign GPU workloads

taiberium:
Nodexo (SN106) is a competing GPU marketplace. It resells the GPUs of nodes that are listed on Lium, and we continue to pay idle incentive for those nodes. Its workloads stay below our percentage gates. One rental container held 22.4 GB of VRAM at approximately 0% reported load. Other Nodexo workers run as bare host processes under PM2, in a plain user session.

Ticket: DAH-2735 (Notion).

## What changed

1. **Ownership gate** (`gpu_usage.py`). On an idle executor, a GPU process fails validation if the platform does not expect its holder. The reason code is `GPU_FOREIGN_PROCESS`. Utilization does not matter. The expected holders are the fillers and pods that the backend reports for this node (`rented_data`), and the executor's own cgroup namespace. A container-name prefix is not used, so `docker rename nodexo-rental filler_x` gives no advantage.
2. **Absolute VRAM floor** (`GPU_VRAM_HELD`). The node fails when no process is visible, the node runs none of our GPU containers, and `memory_used_mb` is more than 2048 MB. NVML `memory_utilization` is bus load, and it reads approximately 0 on a loaded but idle card. `memory_used_mb` was already collected, but no gate used it. The validator samples the card again before it gives this verdict. The scrape reads NVML memory before it walks `/proc`, so a process that exits between the two steps can leave its memory with no owner.
3. **Shadow first.** Both verdicts are behind two flags: `FOREIGN_GPU_WORKLOAD_CHECK_ENABLED` observes and logs, and `FOREIGN_GPU_WORKLOAD_ENFORCEMENT_ENABLED` withholds the score. Enforcement is off by default. Every other money-withholding gate in `core/config.py` uses this pattern, for example `CPU_TRUTH_*` and `FILLER_LIVENESS_*`.
4. **No verdict when we cannot measure.** Three conditions give no verdict: no backend data this cycle, a failed `nvidia-smi` query, and an unreadable card.

The gate runs on idle nodes only. For rented nodes the pipeline stops at `TenantEnforcementCheck`.

## Verified on the production fleet

We replayed the rule over the specs of every active production executor after each change:

- Four nodes were marked as foreign, and all four are Nodexo. Their cgroups are `pm2-lichsl.service` and `user.slice/session-*.scope`, and they run a `nodexo-monitor-*` container. They use the two known IP addresses.
- No node was marked on held VRAM.
- No false positives occurred among the approximately 690 GPU processes of customer pods and fillers.
- Every real GPU process on the fleet has a cgroup that leaves the namespace of the scrape (`0::/../…`). None has the clean `0::/…` shape that the executor exemption uses, so that exemption costs no detection.

We found three false positives this way, and we corrected them before merge:

- The GPU work of the validator itself.
- A node whose filler worked at 34% while NVML reported no processes.
- A stale VRAM sample.

## Verified on staging, in both flag states

We deployed the branch to the staging validator. Then we reproduced the Nodexo condition on an idle A6000 node that runs one of our PEARL fillers. A foreign container held approximately 3 GB of VRAM, and the card read 0% utilization.

- **Shadow, the default.** The validator logged `GPU_FOREIGN_PROCESS` at severity `warning`, with `process_count: 2`. It named only the container that the backend does not know. Our filler held VRAM at the same moment, and the validator did not name it. The same cycle logged `SCORE_COMPUTED actual_score=1.0`, so nothing was withheld.
- **Enforcement on.** We set the flag on a temporary deployment branch, and we did not change `main`. The validator gave the same verdict at severity `error`, with the impact `Validation skipped; score set to 0`. No `SCORE_COMPUTED` and no `VALIDATION_COMPLETED` followed for that node, because the fatal check stopped the pipeline. This is the score-0 path.

We then restored staging: we removed the test container, we resumed the filler, and we deployed the validator again from `main`. The flag is gone.

## Not in this PR

These items belong to the ticket:

- An active availability probe. The validator must try to allocate all free VRAM on a node that is listed as free. This gap is the most important one. One flagged node now runs a `nodexo-rental-*` container. That container holds no VRAM and shows no process, so no check of processes or memory can see it.
- A sweep of the listening ports inside the `port_range` of the machine.
- A container that is renamed to the exact filler or pod name that the backend holds for this node still passes. To close this gap we must make sure that the image digest of the container is correct.
- Two conditions that exist today and that this PR does not change. First, a GPU process at 5% utilization or more takes the older `GPU_USAGE_HIGH` path before the gate reads the owner. Second, the rented path (`TenantEnforcementCheck`) still judges the same evidence by utilization only.
- A node with a score of 0 stays listed, and our fillers continue to run on it. We stop the payment to the provider, and we continue to sell and use the node. This is a product decision, not a code correction.

## Tests

`tests/test_gpu_usage_check.py` has 18 new cases. They cover these conditions:

- Rename evasion and bare host processes.
- The executor cgroup, and held VRAM with the live confirmation.
- Stale samples and failed queries.
- The three flag states, missing backend data, and payload serialization. The full validator suite passes 1679 tests. Three tests in `test_sshd_bootstrap_script.py` fail, and they fail on `main` without this branch.


neurons/validators/src/services/task/checks/gpu_usage.py
```diff
@@ -46,15 +65,112 @@ async def run(self, ctx: Context) -> CheckResult:
...
+        enforce: bool = settings.FOREIGN_GPU_WORKLOAD_ENFORCEMENT_ENABLED
+        workload_containers: set[str] = _lium_workload_containers(ctx)
+        foreign_processes: list[ForeignGpuProcess] = [
+            ForeignGpuProcess(
+                pid=process.get("pid"),
+                container_name=process.get("container_name"),
+                cgroup=process.get("info"),
+            )
+            for process in gpu_processes
+            if process.get("container_name") not in workload_containers
```
arhangel66, line 101: A filler the backend starts after the per-cycle `rented_data` snapshot is absent from this allowlist and reads as foreign at 0% util. Could the shadow run count that case (our prefix, not in the snapshot) before enforcement goes on?

arhangel66 (approved):

neurons/validators/src/services/task/checks/gpu_usage.py
taiberium: Done in 571a70e6. The verdict does not change, but the payload now has a separate `lium_named_outside_snapshot` list. It holds the flagged processes whose container carries our `pod_`/`filler_` prefix and is absent from this cycle's snapshot. The shadow week can count that race before we turn enforcement on.

PR #1244, opened 2026-08-21
DAH-2742: stop transient checks from clearing verified job info

arhangel66:
A check that returns `updates={"clear_verified_job_info": True}` makes the backend deactivate the
executor immediately and fire `EXECUTOR_INACTIVE_MID_RENTAL` when a rental is open. Without that
flag a failed cycle is self-healing: score 0 means no spec upsert, `updated_at` freezes, and the
hourly sweep only deactivates after a full hour — four consecutive failures. A signal that flaps
once never gets there.

Four branches were taking the instant door on a signal that can be wrong while the host is fine, or
that the ordinary path already handles. They no longer set the flag. Everything else keeps it.

Executor `9cb90025-0314-445f-8fb2-e63355a5f958` logged 93 `SYSBOX_REQUIRED_OK` against 3
`SYSBOX_REQUIRED_MISSING` in the 24h to 2026-08-21T02:30Z, with an OK cycle 14 minutes after the
miss at 01:42:52Z. Each miss deactivated a healthy machine. `sysbox_runtime` is forced to `False`
by `ConnectivityOrchestrator.verify` on *any* DinD probe failure — an SSH timeout included — not
only on a genuinely missing runtime.

### What changed

| branch | instant reset | why |
|---|---|---|
| `SysboxRequiredCheck` | **dropped** | the DinD probe reports "no sysbox" for any probe failure |
| `NvmlDigestCheck`, driver absent from the digest map | **dropped** | our map lagging behind an NVIDIA release is not evidence of tampering |
| `NvmlDigestCheck`, driver on `nvml_invalid_drivers` | kept | the backend already confirmed that one as a spoof |
| `NvmlDigestCheck`, `DIGEST_MISMATCH` | kept | driver known, digest differs — that is tampering |
| `CpuTruthCheck` under enforcement | **dropped** | still zeroes the score through `cpu_truth_passed`; a host that really understates its cores fails every cycle |
| `RentalVerificationCheck`, CPU quota under enforcement | **dropped** | same reasoning; still zeroes the score |
| `RentalVerificationCheck`, GPU-runtime quarantine | kept | see below |
| everything else | kept | ban lists, duplicates, spec and fingerprint pinning, `POD_NOT_RUNNING` |

The two CPU branches ship disabled (`CPU_TRUTH_ENFORCEMENT_ENABLED`,
`RENTAL_CPU_LIMIT_ENFORCEMENT_ENABLED` are both `False`), so those rows are inert in prod today.

Two `impact` strings in `messages.py` promised the miner that verification was cleared, which the
two dropped branches no longer do. Corrected. `result_handler.py`, `redis_service.py`,
`pipeline.py` and compute-app are untouched.

### The GPU-runtime quarantine kept its instant reset

The first pass dropped it and that was wrong:

- The GPU-runtime codes come from specific NVIDIA-hook signatures — `driver/library version
  mismatch`, `requires reset`, `unknown device` — states that persist until a reboot or a host
  reset. The classifier's own comment records hosts failing this way *for weeks across images*, and
  it deliberately excludes renter-controlled output. A generic probe failure never reaches this
  branch; it lands in the `FAILED` branch, which never set the flag. The message also promises "new
  rentals are disabled", and `active=false` is the only thing that actually disables them — score 0
  does not delist.
### The DRIVER_UNKNOWN split

The `DRIVER_UNKNOWN` split came out of the same review: dropping the flag from that whole branch
also softened drivers the backend had already confirmed as spoofs. Since the staleness sweep needs
an unbroken hour, a miner alternating between a spoofed driver and a clean one would refresh
`updated_at` every other cycle and stay listed indefinitely. Now only the never-seen-before case
falls back.

### Why it is safe

`clear_verified_job_info` is read in exactly one place, `result_handler.py:274`. Dropping it routes
these failures into `set_verified_job_info(success=False)`, which preserves the pinned `spec` and
`uuids` exactly as the clear did and only bumps a `failed` counter that nothing reads. The score
path never consults the flag. Both branches keep halting the pipeline and zeroing the score; only
the timing of the deactivation changes.

### What it costs

A genuinely sysbox-less host stays rentable for up to an hour instead of 15 minutes. DAH-2313 was
written as "never appear on the network"; this deliberately relaxes it to "not after four
consecutive failures". Machines pushed onto the hourly sweep meet that sweep's own false positive,
which is DAH-2725 — this PR does not fix that one.

### Verification

`pdm run pytest tests/ -q --tb=short --strict-markers` from `neurons/validators`: 1662 passed, 9
skipped. The 3 failures in `tests/test_sshd_bootstrap_script.py` are identical before and after this
change — BSD `sed` on macOS does not accept the `-i -E` / `I` combination used in
`sshd_bootstrap.sh:215`, so they fail on a clean `main` on any mac. Nothing new broke.

The nvml split is pinned by the existing parametrized table, which carries `DIGEST_MISMATCH` and
`DRIVER_UNKNOWN` rows side by side with opposite expectations, plus an assertion in
`test_known_spoof_driver_is_not_reported`. Every change was checked failing first, then passing.

### DIFF
--- a/neurons/validators/src/services/task/messages.py
+++ b/neurons/validators/src/services/task/messages.py
@@ -312,7 +312,7 @@ class NvmlDigestMessages:
         reason="NVML_DRIVER_UNKNOWN",
         severity="error",
         category="env",
-        impact="Score set to 0; previous verification cleared",
+        impact="Score set to 0 for this cycle; the previous verification is kept",
         remediation="Update to a supported NVIDIA driver version. Your current driver version is not recognized.",
     )
     DIGEST_OK = MessageTemplate(
@@ -410,7 +410,7 @@ class SysboxRequiredMessages:
         reason="SYSBOX_REQUIRED_MISSING",
         severity="warning",
         category="policy",
-        impact="Score set to 0; verification cleared; executor kept off the network",
+        impact="Score set to 0 for this cycle; the previous verification is kept; repeated failures take the executor off the network",
         remediation="Install the sysbox runtime; unrented machines without sysbox are not allowed on the network.",
     )
     SYSBOX_OK = MessageTemplate(
--- a/neurons/validators/src/services/task/messages.py
+++ b/neurons/validators/src/services/task/messages.py
@@ -312,7 +312,7 @@ class NvmlDigestMessages:
         reason="NVML_DRIVER_UNKNOWN",
         severity="error",
         category="env",
-        impact="Score set to 0 for this cycle; the previous verification is kept",
+        impact="Score set to 0 for this cycle",
         remediation="Update to a supported NVIDIA driver version. Your current driver version is not recognized.",
     )
     DIGEST_OK = MessageTemplate(
@@ -410,7 +410,7 @@ class SysboxRequiredMessages:
         reason="SYSBOX_REQUIRED_MISSING",
         severity="warning",
         category="policy",
-        impact="Score set to 0 for this cycle; the previous verification is kept; repeated failures take the executor off the network",
+        impact="Score set to 0 for this cycle; verification is kept until repeated failures deactivate the executor",
         remediation="Install the sysbox runtime; unrented machines without sysbox are not allowed on the network.",
     )
     SYSBOX_OK = MessageTemplate(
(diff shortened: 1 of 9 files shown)


neurons/validators/src/services/task/messages.py
```diff
@@ -312,7 +312,7 @@ class NvmlDigestMessages:
         reason="NVML_DRIVER_UNKNOWN",
         severity="error",
         category="env",
-        impact="Score set to 0; previous verification cleared",
+        impact="Score set to 0 for this cycle; the previous verification is kept",
```
taiberium: **NIT:** the confirmed-spoof branch renders this same template but still clears the verification. Give that branch its own `impact`, or remove the clause.

taiberium (approved):

taiberium:
**NIT:** two docstrings still describe the old behavior. `sysbox_required.py` says the verification is cleared, and `_cpu_quota_verdict` calls this the same quarantine as the GPU-runtime fault, which kept the flag.

PR #1245, opened 2026-08-21
DAH-2734, provider-side CPU/disk on rented executors as triage signals

taiberium:
### Task Link

[DAH-2734](https://app.notion.com/p/3c2b8bfdbde981d9ab00c822c93c7ee8)

---

## Problem

A provider kept his own SN13 (Data Universe) miner on a machine listed on Lium. It burned ~9 of 32 CPU threads and grew to 1.5 TB until the disk hit 100% while the machine was rented the whole time. Nothing in validation looks at provider-side CPU or disk, so the damage was invisible until the renter complained.

## Solution

Judge a machine by what its OWNER takes from it — the CPU/disk twin of the foreign-GPU gate in #1242. A provider who mines another subnet on a listed machine (SN13/Data Universe: 9 of 32 threads, 1.5 TB, GPU untouched) sells capacity he already uses himself, and the GPU gate cannot see him because he never touches the card.

The scrape collects per-container CPU (`docker stats`) plus a host CPU sample over the same window. `ProviderSideLoadCheck` computes host-minus-containers CPU cores and used-minus-docker-breakdown disk, stores the result in specs as `provider_side_load`, and judges the node above 2 cores / 100 GB outside docker — rented or idle alike, because the provider cheats the renter in the first case and the marketplace in the second.

Shadow-first like every other money-withholding gate: `PROVIDER_SIDE_LOAD_CHECK_ENABLED` (default on) observes and logs; only `PROVIDER_SIDE_LOAD_ENFORCEMENT_ENABLED` (default off) zeroes the score. An inability to measure never withholds money.

Follows #1242 (DAH-2735) — the VRAM/GPU side of the same hole, already merged. This PR now targets `main` and carries only the DAH-2734 commit.

## Changes

- `machine_scrape.py`: collect per-container CPU core-percent, attach as `each_cpu_percent` (fail-open)
- `file_encrypt_service.py`: obfuscation entries for the new key
- `checks/provider_side_load.py`: `ProviderSideLoad` + `compute_provider_side_load()` + `ProviderSideLoadCheck` (shadow-first gate)
- `core/config.py`: `PROVIDER_SIDE_LOAD_CHECK_ENABLED` (on) and `PROVIDER_SIDE_LOAD_ENFORCEMENT_ENABLED` (off), mirroring the DAH-2735 pair
- `pipeline.py` + `score_calculator.py`: `provider_side_load_passed` context flag and the score gate, same mechanics as the CPU-truth gate
- `messages.py` / `pipeline_factory.py` / `checks/__init__.py`: event templates + registration in both pipelines
- tests: `test_provider_side_load.py` (attribution math, malformed/negative/non-finite inputs, enforcement, shadow, idle, below-limit, disabled)

## Deployment Steps

Use GitHub Action.

## Review Request

- [x] Just code review

## Other PRs

- #1242 (DAH-2735) — merged; this work continues on top of it

## Risks

- One extra `docker stats --no-stream` call per scrape (~2 s); wrapped in try/except, a failure only drops the CPU part of the signal
- `provider_side_load.disk_kb` includes the OS baseline (it is "outside docker", not "provably the provider's"), hence the 100 GB floor
- The check is non-fatal and exception-guarded: a malformed specs value can never fail validation or withhold money
- Enforcement is OFF by default. The floors (2 cores / 100 GB) need a shadow week over the live fleet: a busy provider host with heavy housekeeping can land above them, and the CPU floor is absolute, not proportional to core count
- Backend `MachineSpecs` strips undeclared keys on validation, so `provider_side_load` needs a mirror field in lium-io-backend to become queryable there (same as DAH-2514 did for the disk breakdown) — follow-up PR; until then the signal lives in the validation events

### Test Case 1

**Actions**

Run `pytest tests/ --ignore=tests/integration --deselect tests/test_sshd_bootstrap_script.py` in `neurons/validators` (that file's 3 failures are pre-existing and fail on clean main too).

**Expected Output**

1696 passed, 5 skipped.

## Checklist

- [x] Risks are documented
- [x] Tests are described


taiberium:
Backend companion: https://github.com/Datura-ai/lium-io-backend/pull/961 — it declares the new specs fields, otherwise `MachineSpecs` strips them and the readings never reach the DB.

neurons/validators/src/services/task/checks/provider_side_load.py
```diff
@@ -0,0 +1,409 @@
...
+            )
+    except (KeyError, TypeError, ValueError, AttributeError, OverflowError):
+        pass
+
+    try:
+        hard_disk = specs.get("hard_disk") or {}
+        used_kb = int(hard_disk.get("used"))
+        docker_parts_kb = [int(hard_disk.get(key)) for key in ("images", "containers", "volumes")]
+        if used_kb >= 0 and all(part_kb >= 0 for part_kb in docker_parts_kb):
+            disk_kb = max(0, used_kb - sum(docker_parts_kb))
```
arhangel66, line 315: Container json logs are not in images/containers/volumes (`SizeRw` is the writable layer only), and we set no log rotation on executors, so a chatty renter pod is booked to the provider. Could the disk half stay observational until that is subtracted? Non-blocking.

arhangel66 (approved):

neurons/validators/src/services/task/checks/provider_side_load.py
taiberium: Good catch, fixed in e2708731. The scrape now walks /var/lib/docker/containers/*/*.log and adds it to hard_disk_containers, so the logs land on docker's side of the split instead of the provider's. A host on the journald driver writes no json log at all, so an empty walk reports 0 rather than failing the breakdown. Two tests in test_scrape_disk_breakdown.py.

taiberium:
@arhangel66 could you take another look? The rule itself moved since your approval, so the old one no longer covers what is here.

What changed:
- Only the CPU half withholds money now. The disk figure is reported and never scored, for the reason you raised: it is read once, has no second look, and every docker category nobody enumerated lands in it. Container json logs and BuildCache are subtracted now, but the next missing category would zero an honest node.
- dockerd's own CPU is subtracted. Pulling a Lium image belongs to no container row and a big pull outlives both samples.
- The allowlist grew: a pod the backend started after the snapshot is confirmed with it (the DAH-2757 helper, now shared), the whole executor stack, the s3fs backup helper and the RoCE probe. Everything matched by name is capped together at the floor, so wearing our prefix buys a forger nothing.
- 1.96 cores used to round up to the 2.0 limit and zero a node below it. It floors now.
- The check runs after StaleContainerCleanupCheck: a pod we just ended is our leftover, not the provider's load.

Proven on staging: a foreign 3-core container is caught, the same load renamed executor-sn13 is still caught, honest nodes read 0.0-0.1 cores. Suite 1793.

neurons/validators/src/services/task/checks/provider_side_load.py
```diff
@@ -0,0 +1,523 @@
...
+                    container_id=short_id, name=name, cpu_percent=float(raw_percent.rstrip("%"))
+                )
+            )
+    except (TypeError, ValueError, OverflowError):
+        return None
+    window_seconds = total_delta / (_JIFFIES_PER_SECOND * core_count)
+    dockerd_cores = (last_dockerd - first_dockerd) / _JIFFIES_PER_SECOND / window_seconds
+    host_cores = (total_delta - (last_idle - first_idle)) / total_delta * core_count
+    return provider_cores_outside_lium(
+        host_cores - max(0.0, dockerd_cores), stats_rows, lium_container_keys
```
arhangel66: The dockerd subtraction is uncapped and keyed on process comm, so `cp sn13-miner dockerd && ./dockerd` hides nine cores on the reading that decides money. Cap it like the name tier before enforcement is armed?

```suggestion
        host_cores - min(max(0.0, dockerd_cores), CLAIMED_EXCUSE_CORES),
        stats_rows,
        lium_container_keys,
```

neurons/validators/src/services/task/checks/provider_side_load.py
```diff
@@ -0,0 +1,523 @@
...
+            "lium_named_outside_snapshot_cores": lium_named_cores_outside_snapshot(
+                ctx.state.specs or {}, lium_container_keys
+            ),
+        }
+
+    async def measure_provider_side_load(
+        self, ctx: Context, lium_container_keys: set[str] | None
+    ) -> ProviderSideLoad:
+        """The scrape's reading, confirmed over ssh whenever it is high enough to cost money."""
+        core_count: int | None = advertised_cpu_count(ctx)
```
arhangel66, line 467: `core_count` is the lscpu value CpuTruthCheck distrusts, and it only catches over-reporting: a host claiming 8 of its 32 threads reports a quarter of its load. Could the resample count /proc/stat's `cpuN` lines instead?

arhangel66 (approved):

neurons/validators/src/services/task/checks/provider_side_load.py
taiberium: Capped at the same 2 cores as the name tier, commit 481b0d46. A real pull can exceed that, so the node reads high while it runs; better that than a copied binary hiding nine cores on the reading that decides money. Test: test_a_miner_renamed_dockerd_is_capped_like_every_other_forged_name.

neurons/validators/src/services/task/checks/provider_side_load.py
taiberium: Done, same commit. The sample now counts /proc/stat's cpuN lines and the resample uses that, so the verdict no longer depends on the advertised count at all. Verified on the staging box: it reports 4, which is what that host has. The scrape's first reading still uses the advertised value, but it only decides whether to look again.

PR #1246, opened 2026-08-24
DAH-2757: confirm a Lium-named container with the backend before calling it foreign

taiberium:
Follow-up to #1242. The first prod shadow day showed the race arhangel66 asked about in that review, so enforcement stays off until this lands.

**What happens now.** `rented_data` is read once, at cycle start. A filler or pod the backend starts later in the same cycle carries our own name but is absent from the allowlist, so it reads as foreign at 0% load. 7 events on 5 executors of 5 different miners in 24 h. One example: executor `a09d2771` (rented) flagged `filler_6c6be0dd-b698-4fdd-9dde-d0e7ec055519`, its own live filler.

**What this does.** A container named `filler_<uuid>` or `pod_<uuid>` that the snapshot misses is now asked about directly, with the same endpoints the filler-liveness check already uses. The name still proves nothing on its own — the prefix only decides whom to ask, the backend gives the answer:

- `filler_<uuid>` in STARTING, RUNNING or STOPPING is ours. Terminal states are not: a container that outlives its run is an orphan.
- `pod_<uuid>` is ours while the backend reports the rental active.
- An id the backend never issued comes back `active=false` and stays foreign. A suffix that is not a uuid is never asked about, because a 422 would read as "backend unreachable".
- An unreachable backend never withholds money.

**Verification.** Full validator suite: 1700 passed, 9 skipped. The 3 `test_sshd_bootstrap_script.py` failures also fail on `main`. 7 new cases cover the late filler, a filler mid-transition, a late pod, a stopped filler, a forged uuid, a non-uuid name, and an unreachable backend.

The gate is still shadow-only in prod (`FOREIGN_GPU_WORKLOAD_ENFORCEMENT_ENABLED=false`). The plan is another shadow week after this merges, then the flip.

### DIFF
--- a/neurons/validators/src/services/task/checks/gpu_usage.py
+++ b/neurons/validators/src/services/task/checks/gpu_usage.py
@@ -397,17 +397,21 @@ async def _the_backend_still_owns(ctx: Context, container_name: str) -> bool:
         filler_run: FillerRunActiveResponse | None = await ctx.services.backend.get_filler_run_active(
             backend_issued_id
         )
+        if filler_run is None:
+            return True
+        if not _the_answer_is_about_this_node(ctx, filler_run.executor_id):
+            return False
         # `active` alone means RUNNING. The status carries the two transitional states, which is
         # where a filler the snapshot missed usually sits — but a backend that sends no status
         # must not turn a run it calls active into a foreign workload.
-        if filler_run is None or filler_run.active:
-            return True
-        return filler_run.status in LIVE_FILLER_RUN_STATUSES
+        return filler_run.active or filler_run.status in LIVE_FILLER_RUN_STATUSES
 
     pod_rental: PodRentalActiveResponse | None = await ctx.services.backend.get_pod_rental_active(
         backend_issued_id
     )
-    return pod_rental is None or pod_rental.active
+    if pod_rental is None:
+        return True
+    return _the_answer_is_about_this_node(ctx, pod_rental.executor_id) and pod_rental.active
 
 
 def _uuid_after_prefix(container_name: str, prefix: str) -> str | None:
@@ -423,6 +427,18 @@ def _uuid_after_prefix(container_name: str, prefix: str) -> str | None:
     except ValueError:
         return None
     return canonical_uuid if canonical_uuid == suffix else None
+
+
+def _the_answer_is_about_this_node(ctx: Context, owning_executor_id: str | None) -> bool:
+    """Whether the run the backend described lives on the node the check is looking at.
+
+    The endpoints answer about a run id, which is public on its own host, so a provider with two
+    nodes could name a squatter container after a live run of their honest one. A backend that
+    predates the field sends nothing, and then the fleet snapshot is the only owner test we have.
+    """
+    if owning_executor_id is None:
+        return True
+    return str(owning_executor_id) == str(ctx.executor.uuid)
 
 
 def _containers_claimed_by_another_executor(ctx: Context) -> set[str]:
--- a/neurons/validators/src/services/task/checks/gpu_usage.py
+++ b/neurons/validators/src/services/task/checks/gpu_usage.py
@@ -5,6 +5,7 @@ from uuid import UUID
 
 from core.config import settings
 from protocol.vc_protocol.compute_requests import (
+    LIVE_FILLER_RUN_STATUSES,
     FillerRunActiveResponse,
     PodRentalActiveResponse,
 )
@@ -14,7 +15,6 @@ from services.const import (
     GPU_MEMORY_UTILIZATION_LIMIT,
     GPU_UTILIZATION_LIMIT,
     GPU_WEDGE_MEMORY_MAX,
-    LIVE_FILLER_RUN_STATUSES,
     POD_CONTAINER_PREFIX,
 )
 from services.gpu_wedge import (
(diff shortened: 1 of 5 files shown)


taiberium:
Replayed the shadow events against the prod DB, so the change is measured against what actually happened, not only against tests.

Every Lium-named container in those events was alive at the moment the validator called it foreign:

| Container | Event | Live from | Live to |
|---|---|---|---|
| `filler_6c6be0dd` | 08-23 00:33:44 | 00:30:31 | 01:10:26 |
| `filler_ceb7d20d` | 08-23 19:21:52 | 19:20:23 | 20:03:03 |
| `pod_d8f198ac` | 08-21 22:30:20 | 08-21 19:45:57 | 22:31:03 |
| `pod_3512598b` | 08-23 19:04:54 | 16:55:10 | 19:34:08 |
| `pod_9b22db3c` | 08-23 20:07:20 | 20:03:00 | 22:23:50 |

All five are cleared by the backend re-check. The sixth name in the window is `filler_tensorcash_fallback`, which is not ours — `tensorcash` appears nowhere in lium-io or in the backend. It keeps our prefix with no uuid behind it, and it stays foreign.

That last one also found a hole worth naming. The endpoints answer like this on a live backend:

```
GET /internal/filler-runs/<unknown-uuid>/active   -> 200 {"active":false,"status":null}
GET /internal/pods/<unknown-uuid>/rental-active   -> 200 {"active":false}
GET /internal/filler-runs/tensorcash_fallback/active -> 422
```

The client reports any non-200 as "backend unreachable", and the gate reads that as a pass. So a name with no uuid behind it would have earned a free pass from the very re-check meant to catch it. That is why the uuid is tested locally before anything is sent.

taiberium:
A second review pass found three ways the re-check could be turned into a free pass. All three are fixed in `c237113e`.

**A live run of another node.** The backend answers about the run, not about where it runs. A provider with two nodes can read a live filler id off the honest one (`docker ps` shows it) and name a foreign container after it on the second node. The re-check would have cleared it. Now a name that this cycle's snapshot already gives to a different executor is foreign without asking — the snapshot carries the whole fleet, so it is the thing that separates two nodes.

**A non-canonical spelling of the same id.** `UUID()` also parses `filler_9622D623-...` and `filler_9622d623dcb3...`, and both are legal docker names, so both can sit beside the genuine container. Only the canonical spelling is accepted now.

**A filler the backend calls active with no status.** The schema permits it. The status set exists to catch STARTING and STOPPING, which `active` does not cover, but a backend that sends no status must not have its own live filler turned into a foreign workload.

**Left open, needs your call.** One narrow case survives: an id born after the snapshot and copied to a second node inside the same cycle. Closing it needs `executor_id` on `FillerRunActiveResponse` and `PodRentalActiveResponse` so the validator can compare it with the node in front of it — a backend change, so I did not make it here. The gate is shadow-only, so nothing is at stake until the flip.

Tests: 1707 passed, 9 skipped; the 3 `test_sshd_bootstrap_script.py` failures also fail on `main`.

neurons/validators/src/services/task/checks/gpu_usage.py
```diff
@@ -338,6 +346,111 @@ def _carries_a_lium_prefix(process: ForeignGpuProcess) -> bool:
...
+    if is_filler:
+        filler_run: FillerRunActiveResponse | None = await ctx.services.backend.get_filler_run_active(
+            backend_issued_id
+        )
+        # `active` alone means RUNNING. The status carries the two transitional states, which is
+        # where a filler the snapshot missed usually sits — but a backend that sends no status
+        # must not turn a run it calls active into a foreign workload.
+        if filler_run is None or filler_run.active:
+            return True
+        return filler_run.status in LIVE_FILLER_RUN_STATUSES
```
arhangel66: A run wedged in STARTING drops out of the snapshot after 15 min (`RENTED_EMISSION_TRANSITIONAL_MAX_AGE`) and nothing closes it, so its id is in no executor's entry and the cross-executor guard misses it, letting a provider name a squatter container on their second node after it. Could the backend answer with the run's executor_id so we can compare it to `ctx.executor.uuid` before enforcement goes on? Non-blocking, fine as a follow-up.

arhangel66 (approved):

neurons/validators/src/services/task/checks/gpu_usage.py
taiberium: Good catch — the wedged STARTING run makes it permanent, not a one-cycle window, so I did it now rather than after the flip.

Backend: https://github.com/Datura-ai/lium-io-backend/pull/967 adds `executor_id` to both answers, resolved from the row's foreign key to the uuid the subnet knows. Validator side compares it with `ctx.executor.uuid` in `40ed4925`.

Additive in either deploy order: the field defaults to `None`, and a validator that gets no owner falls back to the fleet-snapshot test.

Tests: backend 2568 passed, validator 1710 passed.

taiberium:
Tested on staging, both halves deployed together.

**The attack, blocked on live hardware.** On staging executor `66e9c110` (149.36.1.151) I put a container named `pod_0de97386-8879-47c8-bede-ac4db3cf0ca5` on the card, holding 440 MiB. That pod rental is ACTIVE, but it belongs to executor `a702c21e` — exactly the case you described. The verdict:

```
GPU_FOREIGN_PROCESS | severity=warning | Shadow observation only: score was NOT changed
foreign_processes=[{pid 2911956, container_name pod_0de97386-8879-47c8-bede-ac4db3cf0ca5}]
```

Before the change the re-check would have cleared it as ours.

**The identifier, confirmed against prod.** The value the backend returns is `Executor.executor_id`. For the node at 151.185.43.26 that is `d51b8008-7338-4c49-a5ae-d37e876ff79f`, and the prod validator logs the same string as `executor_uuid`. This was the one assumption that unit tests could not settle: had the two identifiers differed, enforcement would have failed every honest container instead of none.

**The contract, from the deployed staging backend.** A live filler and a live pod both carry the owner, and an id the backend never issued carries `null` in its place — so a name a provider invents still finds no owner to hide behind.

**One thing that is not ours.** `/internal/pods/{id}/rental-active` returns 500 for one DELETING pod on staging. The traceback stops one line before the new code, at `find_latest_for_pod`: `LookupError: 'VAST' is not among the defined enum values. Enum name: rentalsource`. The staging database carries a `rental_source` column that neither master nor prod has, so an unmerged branch left it there. Worth someone's attention, but not this PR's.

Staging is back on main and master, the test container is gone, the filler is untouched, and the board is released.

arhangel66 (approved):

PR #1247, opened 2026-08-25
DAH-2667: measure a RoCE fabric before the backend sells it as a cluster

taiberium:
A RoCE fabric has no subnet manager, so the fleet can only infer one: same miner, same IPv4 segment,
one address per host. That is evidence, not proof. A switch without lossless queueing (PFC/ECN)
carries no RDMA, NCCL falls back to TCP over the cluster overlay, and the renter pays cluster price
for a job nothing in the logs explains.

This measures the wire instead. At the end of a miner's cycle — while its key is still installed —
the validator pairs the miner's FREE hosts by RoCE segment and runs `ib_write_bw` between each pair
from `daturaai/lium-rdma-probe`, over the same rail and GID `lium-fabric-env` hands NCCL inside a
cluster pod. The result is written into the specs of BOTH hosts, each naming the other.

Two facts set when this runs. The measurement matters only for a pair that can be sold, and a pair
is sellable only while both hosts are free — which is exactly when the probe can take the cards
without touching a renter's job. A rented host is never probed.

Behind `ROCE_LINK_PROBE_ENABLED`, off by default. Best effort throughout: a probe that fails leaves
the specs without a measurement, the backend then does not sell that pair, and the cycle carries on.

The backend half (lium-io-backend, DAH-2667-roce-cluster) requires a fresh measurement naming
another member of the same wire before it groups a RoCE pair, behind its own flag, and shows the
measured speed to the buyer.

Tests: `neurons/validators/tests/test_roce_link_probe.py` — 13, covering which hosts are measurable
(dual-fabric and two-segment hosts are not), how free hosts pair, the `ib_write_bw` parse, and
where the result lands.

### DIFF
--- a/neurons/validators/src/services/miner_service.py
+++ b/neurons/validators/src/services/miner_service.py
@@ -249,6 +249,9 @@ class MinerService:
             )
         
         loop = asyncio.get_event_loop()
+        # DAH-2667: what the RoCE probe subtracts from the cycle's own wait, so a job that already
+        # spent most of it takes a shorter sweep instead of overrunning and scoring the miner zero.
+        job_started_at: float = time.monotonic()
         my_key: bittensor.Keypair = settings.get_bittensor_wallet().get_hotkey()
         default_extra = {
             "job_batch_id": payload.job_batch_id,
@@ -383,6 +386,7 @@ class MinerService:
                             my_key.ss58_address, private_key.decode("utf-8")
                         ),
                         default_extra,
+                        job_started_at,
                     )
 
                     logger.info(
@@ -1853,6 +1857,9 @@ class MinerService:
         default_docker_image_digests: dict[str, str],
     ):
         """REST API version of request_job_to_miner."""
+        # DAH-2667: see the WebSocket path — the RoCE probe measures the cycle's remaining time
+        # from here.
+        job_started_at: float = time.monotonic()
         my_key: bittensor.Keypair = settings.get_bittensor_wallet().get_hotkey()
         default_extra = {
             "job_batch_id": payload.job_batch_id,
@@ -1961,6 +1968,7 @@ class MinerService:
                         my_key.ss58_address, private_key.decode("utf-8")
                     ),
                     default_extra,
+                    job_started_at,
                 )
 
                 logger.info(
--- a/neurons/validators/src/services/miner_service.py
+++ b/neurons/validators/src/services/miner_service.py
@@ -64,6 +64,7 @@ from protocol.vc_protocol.compute_requests import RentedExecutorsResponse
 from services.attestation_service import AttestationService
 from services.docker_service import DockerService, inflight_creates
 from services.redis_service import MACHINE_SPEC_CHANNEL, RedisService
+from services.roce_link_probe import measure_and_attach
 from services.ssh_service import SSHService
 from incentive.config import BASE_GPU_MAP
 from services.task_service import TaskService, JobResult
@@ -371,6 +372,17 @@ class MinerService:
                     results = self._filter_task_results(msg.executors, raw_results, default_extra)
                     results.extend(
                         self._build_manual_rental_results(payload, rented_data, existing=results)
+                    )
+
+                    # DAH-2667: while the miner's key is still installed and its idle hosts are
+                    # free, measure every RoCE pair. Best effort — an unmeasured pair is simply not
+                    # sold as a cluster.
+                    await measure_and_attach(
+                        results,
+                        self.ssh_service.decrypt_payload(
+                            my_key.ss58_address, private_key.decode("utf-8")
+                        ),
+                        default_extra,
                     )
 
                     logger.info(
@@ -1938,6 +1950,17 @@ class MinerService:
                 results = self._filter_task_results(msg.executors, raw_results, default_extra)
                 results.extend(
                     self._build_manual_rental_results(payload, rented_data, existing=results)
+                )
+
+                # DAH-2667: while the miner's key is still installed and its idle hosts are free,
+                # measure every RoCE pair. Best effort — an unmeasured pair is simply not sold as a
+                # cluster.
+                await measure_and_attach(
+                    results,
+                    self.ssh_service.decrypt_payload(
+                        my_key.ss58_address, private_key.decode("utf-8")
+                    ),
+                    default_extra,
                 )
 
                 logger.info(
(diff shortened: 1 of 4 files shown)


neurons/validators/src/services/miner_service.py
```diff
@@ -373,6 +374,17 @@ async def request_job_to_miner(
                         self._build_manual_rental_results(payload, rented_data, existing=results)
                     )
 
+                    # DAH-2667: while the miner's key is still installed and its idle hosts are
+                    # free, measure every RoCE pair. Best effort — an unmeasured pair is simply not
+                    # sold as a cluster.
+                    await measure_and_attach(
```
arhangel66, line 383: No time budget here: an overrun gets the whole job cancelled at `core/validator.py:315` and the miner scores zero for the cycle. Can we wrap this in `asyncio.wait_for` and swallow the timeout locally, so it never reaches the outer handlers that return `_build_failed_job_result`?

neurons/validators/src/services/roce_link_probe.py
```diff
@@ -0,0 +1,341 @@
...
+    if match is None:
+        return None
+    return float(match.group(1))
+
+
+def attach_measurement(result: JobResult, measurement: RoceLinkMeasurement) -> None:
+    """Carry the measurement to the backend in this host's specs, where the grouping reads it."""
+    if result.spec is None:
+        result.spec = {}
+    result.spec[SPEC_KEY] = measurement.model_dump(mode="json")
```
arhangel66: Each pair overwrites both hosts' single spec key, so only the last write per host survives the mesh and a pair that failed cannot be recorded at all. Store a peer list, or walk a chain of N-1 runs instead?

arhangel66 (changes requested):
See the inline comments.

neurons/validators/src/services/miner_service.py
taiberium: Fixed. The sweep sits in one asyncio.wait_for with a 180 second budget and swallows the timeout, so it can never reach the outer handler that scores the miner zero. There is a test with a hanging sweep.

neurons/validators/src/services/roce_link_probe.py
taiberium: Right. Each host now keeps one entry per peer, failures included, so the backend can ask whether the exact two nodes a renter picked were proven against each other. The probe also measures every pair of the segment now, not disjoint pairs.

neurons/validators/src/services/roce_link_probe.py
```diff
@@ -0,0 +1,387 @@
...
+# A write count, not a duration: the probe proves the wire carries RDMA at all, and a validator
+# cycle has seconds for a miner's whole segment, not minutes. 1000 writes of 64 KiB finish in well
+# under a second on any wire worth selling; the reported rate is a side effect kept for the logs.
+PROBE_ITERATIONS = 1000
+# Covers the one slow step — the first pull of the 38 MB image on a host that has never seen it.
+PROBE_TIMEOUT_SECONDS = 120
+# The whole sweep's budget. The caller's timeout cancels the miner's entire job and scores it
+# zero, so the probe stops itself well before that: a cycle is 15 minutes and every other check
+# has run by now.
+PROBE_BUDGET_SECONDS = 180
```
arhangel66, line 59: Executor tasks get `JOB_TIME_OUT - 120` and the outer wait is `JOB_TIME_OUT - 50`, so 180s here can overrun that 70s slack and zero a slow miner's cycle. Can the budget come from the time actually left?

neurons/validators/src/services/roce_link_probe.py
```diff
@@ -0,0 +1,387 @@
...
+
+def measured_gigabits_per_second(ib_write_bw_output: str) -> float | None:
+    """The average bandwidth of a finished run, or None when the run printed no result row."""
+    match = _BANDWIDTH_ROW.search(ib_write_bw_output)
+    if match is None:
+        return None
+    return float(match.group(1))
+
+
+def attach_measurement(result: JobResult, measurement: RoceLinkMeasurement) -> None:
```
arhangel66, line 206: The backend replaces specs wholesale each cycle, so nothing accumulates and every pair of a segment has to fit one 180s sweep. Is that sized against the biggest real segment?

neurons/validators/src/services/roce_link_probe.py
```diff
@@ -0,0 +1,387 @@
...
+        if entry.get("peer_address") != measurement.peer_address
+    ]
+    result.spec[SPEC_KEY] = kept + [measurement.model_dump(mode="json")]
+
+
+def _server_command() -> str:
+    return (
+        f"docker rm -f {PROBE_CONTAINER_NAME} >/dev/null 2>&1; "
+        f"docker run --rm -d --name {PROBE_CONTAINER_NAME} --network host --device /dev/infiniband "
+        f"{RDMA_DOCKER_FLAGS} {PROBE_IMAGE} server {PROBE_HANDSHAKE_PORT} {PROBE_ITERATIONS}"
```
arhangel66: When the budget expires mid-pair the `finally` cleanup is cancelled too, and the server keeps waiting for a client forever. Make it self-limiting:

```suggestion
        f"{RDMA_DOCKER_FLAGS} --entrypoint timeout {PROBE_IMAGE} {PROBE_TIMEOUT_SECONDS} /usr/local/bin/lium-roce-probe server {PROBE_HANDSHAKE_PORT} {PROBE_ITERATIONS}"
```

arhangel66 (approved):

neurons/validators/src/services/roce_link_probe.py
taiberium: Fixed. The budget is now the smaller of 180s and what this job has left: JOB_TIME_OUT, minus the 50s the cycle keeps for its own wait, minus the time the job already spent, minus a 30s tail for removing the key. Under 20 seconds left it skips the sweep instead of starting one it cannot finish. Both ends have a test.

neurons/validators/src/services/roce_link_probe.py
taiberium: Applied your suggestion. The server runs under timeout now, so it ends itself even when the sweep is cancelled before the cleanup runs. A test pins the flag.

neurons/validators/src/services/roce_link_probe.py
taiberium: No, it is not sized against the biggest segment, and I am leaving it for now. A pair costs a few seconds, so 180s covers roughly six or seven hosts; a bigger segment measures what fits and the rest wait for the next cycle. The order is fixed, so the tail pairs of a very large segment would keep waiting. If prod grows one, the fix is running pairs concurrently, not a bigger budget. I will open a follow-up rather than widen this PR.

PR #1248, opened 2026-08-27
DAH-2757: a BROKEN pod's container is ours, not a foreign GPU workload

taiberium:
Validator half of the last DAH-2757 piece. Needs lium-io-backend#975.

**The problem.** `_force_close_broken` marks a pod BROKEN and closes its rental, but the container stays on the host until the validator's own stale-container reaper collects it (~15 min). During that window the ownership re-check asked the backend, got `active=false`, and the gate called our own leftover a foreign GPU workload. With enforcement on, the provider is scored 0 for a container our platform left behind — on a rental we already penalised. Seen once in two days of prod shadow: pod `c1bb0859-…` on executor `7cdbaaa3`, rental closed 17:39:28, foreign verdict 17:55:42.

**The change.** The backend now sends the pod's `status` alongside `active`, and a pod reported BROKEN counts as ours — the same shape the filler path already uses (`filler_run.active or filler_run.status in LIVE_FILLER_RUN_STATUSES`). The cross-executor guard still applies first, so a BROKEN pod of a different node stays foreign.

**Why not widen `active` on the backend instead.** That was the first attempt, and it breaks a second consumer. `rented_machine.py:180` reads the same field as its stand-down signal: `active=false` is what lets a rented-but-missing pod pass with `STALE_POD_NOT_RUNNING`. A pod that goes BROKEN after the cycle's snapshot is still in that snapshot, so widening `active` would have sent it to `_recover_downed_pod`, failed the recovery, and returned `POD_NOT_RUNNING` with `passed=False` — a real score-0 on the exact event the fix exists to protect. Keeping `active` narrow and letting each caller widen it leaves both consumers honest.

**Tests.** `pytest` in neurons/validators — 1768 passed, 15 skipped; the 3 `test_sshd_bootstrap_script.py` failures also fail on `main`. Three new cases: a BROKEN pod of this node is ours, a BROKEN pod of another node is still foreign, and a pod the backend never issued is still foreign.

Deploy order is free: the field defaults to `None`, and a validator that gets no status behaves exactly as it does today.

### DIFF
--- a/neurons/validators/src/protocol/vc_protocol/compute_requests.py
+++ b/neurons/validators/src/protocol/vc_protocol/compute_requests.py
@@ -191,6 +191,9 @@ class RentedExecutorsResponse(BaseModel):
 
 class PodRentalActiveResponse(BaseModel):
     active: bool
+    # DAH-2757: the pod's own state. `active` answers "is the rental live", which is a different
+    # question from "is this container ours" — see BROKEN_POD_STATUS below.
+    status: str | None = None
     # DAH-2757: the executor the pod belongs to. The answer is about the POD, not about where it
     # runs, so without this a provider can name a squatter container after a live pod of their
     # second node. Absent means a backend that predates the field — then the fleet snapshot is the
@@ -221,6 +224,11 @@ class FillerRunActiveResponse(BaseModel):
 # container that outlives its run is an orphan, not a workload we started. Mirrors FillerRunStatus
 # in the backend (models/filler_run.py) — a new transitional state there belongs here too.
 LIVE_FILLER_RUN_STATUSES = frozenset({"STARTING", "RUNNING", "STOPPING"})
+
+# DAH-2757: a pod the backend marked BROKEN has a closed rental, so it reports active=False, but its
+# container stays on the host until the stale-container reaper collects it. That leftover is ours to
+# tolerate, not a foreign workload to score against the provider.
+BROKEN_POD_STATUS = "BROKEN"
 
 
 class ExecutorUptimeResponse(BaseModel):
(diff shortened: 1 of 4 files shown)


taiberium:
Review pass (Claude + Codex, 3 rounds, clean on the third) found two holes in the first shape of this fix. Both are closed in `3b5fa8a6`.

**The exemption had no end.** A BROKEN pod's ROW lives 24 hours, but the container it left behind is gone within the stale-container grace. Keying the exemption on the row handed a provider a whole day in which the name of their own broken pod launders a foreign workload: break a pod, wait for the reaper to take the real container, then create your own under the same name. It is now time-boxed to `BROKEN_POD_CONTAINER_GRACE_MINUTES` (twice the reaper's grace) measured from `rental_closed_at`, which cuts that window from 24 hours to 30 minutes and still covers the real case — the prod event was 16 minutes after the close.

**A timezone would have crashed the check.** `rental_closed_at` is a datetime parsed off the wire. The backend writes it naive today, but an offset in the JSON makes pydantic hand back an aware datetime, and subtracting a naive `utcnow()` from it raises — inside a check that is fatal, so the executor scores 0 on a crash. The value is now read as the UTC it is and both sides are compared in one frame.

**Left alone as pre-existing:** if the pod ROW is deleted while its container still runs (a customer deleting a BROKEN pod is DB-only, `services/pod.py:227`), the container reads as foreign again. That is true without this change too — there the whole window reads as foreign — so it is a follow-up, not part of this PR.

Tests: 1771 passed, 15 skipped; the 3 `test_sshd_bootstrap_script.py` failures also fail on `main`. Three more cases: a stale BROKEN pod no longer shields its name, a BROKEN pod with no close time does not either, and a close time that arrives with a timezone is still read.

arhangel66 (approved):

PR #1249, opened 2026-08-27
DAH-2786: withhold the unrented incentive from a host that reverts our GPU power cap

taiberium:
Case 3 of the three power-cap problems: the cap applies, the validator reads it back, and then the host raises the limit again. The backend grace check kills the PEARL filler, the node runs no Lium job — and it still collects the full unrented incentive. Measured on 2026-08-18: 113 of 536 cap/restore pairs, 103 of them landing exactly on the miner's own pre-cap value.

## Signal
`services/gpu_power_limit.py`. The restore path already reads the live limit before it writes the miner's value back.

- `GpuPowerRestoreRecord` gains `capped_to_watts` — the clamped watts we really applied. Old records carry `None` and never accuse.
- `_reverted_cap_target()` flags a revert only when the cap really lowered the limit AND the live limit is back at or above the miner's own value. The first bound covers a miner who already ran below our target; the second stops a frozen record (older cap target after a failed restore) from inventing a breach.
- Each observation goes to Redis `gpu_power_cap_revert:<executor_id>`, 24 h window, bounded history. Best-effort — teardown never fails over it.

## Gate
`incentive/rental_price.py`, fifth unrented gate, same pattern as DAH-2520 / 2546 / 2715.

- Flags at 3 reverts in 24 h. One revert is not enough: a driver reload can undo a cap on its own.
- `ENABLE_UNRENTED_POWER_CAP_REVERT_LIMIT`, **default False** — shadow only. This gate reads provider behaviour, not a capability, so the prod numbers get read first.
- Miner-facing reason `reverts_gpu_power_cap` with the count, the window and both watt readings. No backend change: the portal humanizes the code generically.
- Fail-open on every unknown — Redis down, corrupt record, missing cap target, missing live reading.

## Tests
`tests/test_gpu_power_cap_revert.py`, 24 cases: the pure discriminator (both prod shapes plus the three false-positive traps), recording at restore time, the reader, and the gate. Full validator suite 1790 passed; the 3 `test_sshd_bootstrap_script.py` failures also fail on clean `main`.

Docs: Datura-ai/lium-docs#132.

## Before the flag flip
1. `default_job/strategies.py` has no incentive input, so a flagged node still gets PEARL fillers it will revert. Same shape as the DAH-2715 finding — grade the sanction first.
2. Shadow mode tells the provider nothing, so the docs and the announcement must land before enforcement.
3. Not verified on staging yet.

### DIFF
--- a/neurons/validators/src/core/config.py
+++ b/neurons/validators/src/core/config.py
@@ -301,6 +301,15 @@ class Settings(BaseSettings):
     # to logging the breach without withholding anything.
     ENABLE_UNRENTED_POWER_CAP_LIMIT: bool = Field(env="ENABLE_UNRENTED_POWER_CAP_LIMIT", default=True)
 
+    # DAH-2786 — GPU power cap revert gate. When True, an unrented executor whose host put the
+    # GPU power limit back while a Lium idle job was running, often enough inside the window,
+    # loses the unrented rental incentive while staying active. Ships in shadow mode: unlike
+    # DAH-2715 this gate reads provider behaviour rather than a capability, so the prod numbers
+    # are read first. Set to True to enforce.
+    ENABLE_UNRENTED_POWER_CAP_REVERT_LIMIT: bool = Field(
+        env="ENABLE_UNRENTED_POWER_CAP_REVERT_LIMIT", default=False
+    )
+
     # DAH-2467 — mixed scoring for a partially rented GPU-split node: the rented GPUs earn in
     # the mining pool and the free GPUs in the unrented pool. Set to False to fall back to
     # scoring the whole box as rented.
--- a/neurons/validators/src/incentive/miner_incentive_log.py
+++ b/neurons/validators/src/incentive/miner_incentive_log.py
@@ -54,7 +54,12 @@ from pydantic import BaseModel, Field
 from core.utils import _m, _StructuredMessage, get_extra_info
 
 if TYPE_CHECKING:
-    from incentive.rental_price import InsufficientDisk, MissingFlagshipCapability, PowerCapIncapable
+    from incentive.rental_price import (
+        InsufficientDisk,
+        MissingFlagshipCapability,
+        PowerCapIncapable,
+        PowerCapReverted,
+    )
     from services.task_service import JobResult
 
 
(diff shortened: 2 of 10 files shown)


taiberium:
Updated after the host-side investigation of the two provider guards. Two things in the first commit were wrong.

**The discriminator missed the main shape.** Guard script A restores the card default every 30 s, so our cap dies during the filler's cold start and the pre-cap reading we recorded IS the default. It then pins `0.91 x default`, which sits *below* that reading — so the old `watts_before >= pre_cap` test never fired on those hosts. The line is now `0.9 x default`: the same one the backend guard kills the run at, and the one both operators deliberately sit a point above (4090 409/450, 5090 518/575).

**Counting was per GPU.** One guard tick resets every GPU on the node, so an 8x host wrote 8 breaches for one event. Now one revert per restore batch, grouped by executor — a whole-host restore sweeps every executor on the box, and 109.174.15.2 alone carries four.

Threshold moved 3 → 2 in 24 h. The backend arms a 12 h backoff after each killed run, so an affected node only gets ~2 PEARL runs a day; 3 was unreachable. Recovery stays 24 h after the last one.

Not covered here, worth its own ticket: script B `docker exec`s into our filler and `kill -STOP`s the miner. This gate only catches those hosts because they also raise the limit — a freeze on its own leaves `filler_run` reading RUNNING while the card idles.

taiberium:
Detection moved out of the restore path and into the run.

Restore-time detection could not tell "the host raised our cap while the job ran" from "the host reset the limit after the job ended". Only the first kills the job; the second harms nothing. That is constructible from the guard script we pulled off an affected host: its miner branch clamps the target to the saved value, so it can leave our cap alone for the whole job, while its no-miner branch restores the default within 30 s of the container going away. That provider would have lost the idle payout for two jobs it never touched.

The revert is now observed in `GpuPowerLimitCheck`, in the branch that already skips the power-limit penalty for a live Lium filler. It runs every validation cycle while our job is on the node and already holds Redis, the executor id and the live `power_limit` / `power_default_limit`. It reads, it does not write to the host, so the filler keeps running either way; the dry-run pipeline writes nothing.

Four more bugs fixed on the way here, all in the same direction — an innocent provider losing the payout:

- a cap applied without persistence mode was stamped as attributable, but the driver unloads on an idle GPU and restores the stock limit by itself (DAH-2702);
- a record frozen by an earlier failed restore kept the first capping executor and pod, so the breach was charged to the wrong provider or deduped away;
- a transient Redis read failure erased the stored history instead of leaving it alone;
- after our own restore raised the limit, a record surviving a failed delete let the next pass read our raise as the host's.

`_restore_records` no longer judges anything, so the whole teardown-window class is gone by construction. Per-pod dedup now means what it should: one job counts once, however many cycles observe it.

Tests: 33 in the revert suite, full validator suite 1799 passed. One existing assertion in `test_gpu_power_limit_check.py` said "never touch records under a live filler" — that is the behaviour this reverses on purpose, so it was inverted with a comment.

PR #1250, opened 2026-08-27
DAH-2787, withhold the unrented incentive when somebody enters the filler container

taiberium:
### Task Link

[DAH-2787](https://www.notion.so/3c9b8bfdbde981e5aa52ee8e5e54da94)

---

## Problem

An idle node earns the unrented incentive while it runs Lium's own job in a `filler_*`
container. Some providers work inside that container: a systemd timer on hosts
`5.130.180.236` / `109.174.15.2` runs `docker exec <filler> pgrep -x peakminer` every 300 s and
then `docker exec <filler> kill -STOP <pid>`, which freezes our PEARL miner while `filler_run`
still reads RUNNING. The node keeps the incentive, the GPU is free for the provider's own work,
and nothing in our telemetry records the visit.

## Solution

The scrape reports every visit inside a filler container, and the unrented-incentive gate
withholds the incentive for an executor that had one. Two detection shapes, because one alone
misses the real attacker:

- `docker_exec` — the docker daemon's own event log (`GET /events?since=&until=`). It keeps
  finished visits with the command line, so a visit that lasts milliseconds is still a fact one
  cycle later.
- `open_session` — process parentage: a process in the filler's PID namespace whose parent is
  outside it. This is the only way to see an `nsenter`, which never reaches the daemon.

Both report the seconds between the container's start and the visit, so one rule judges both.
The validator execs into a filler only while creating it (public keys, sshd bootstrap) and never
enters one afterwards, so a visit is judged only after `FILLER_ENTRY_GRACE_SECONDS` (900 s).

## Changes

- `machine_scrape.py`: `probe_filler_container_entries()` and its helpers; new spec keys
  `filler_entries` / `filler_entry_scrape_error`, registered in both obfuscation tables.
- `rental_price.py`: `FillerContainerEntry` + `_filler_container_entry()` gate, next to the
  DAH-2715 power-cap gate; new `ZeroIncentiveReason.FILLER_CONTAINER_ENTERED`.
- `miner_incentive_log.py`: the provider-facing reason line.
- `config.py`: `ENABLE_UNRENTED_FILLER_ENTRY_LIMIT` (default **False**, shadow mode) and
  `FILLER_ENTRY_GRACE_SECONDS` (default 900).
- One reader for the docker socket (`docker_api_read`), used by both the JSON and the
  event-log call.

## Deployment Steps

Use GitHub Action. The gate ships **off**: the first deploy only measures. Read the shadow
numbers in Loki (`reason=filler_container_entered`, `enforced=false`) before flipping
`ENABLE_UNRENTED_FILLER_ENTRY_LIMIT`.

## Review Request

- [x] Just code review

## Other PRs

None.

## Risks

- The scrape runs on every executor each cycle and `get_machine_specs` has no guard around it,
  so a probe that raised would cost a miner every executor's score. The probe therefore never
  raises: it returns an empty result plus a scrape error, and a test pins that.
- `/proc` is walked once per node and only when the host actually runs a filler container.
- Every unknown reading fails open: no spec, no key, not a list, unknown kind or wrong types
  leave the incentive untouched.
- The daemon keeps only its last events in memory, so on a host with chatty healthchecks an old
  visit can roll out of the window before the scrape reads it. A timer that keeps entering the
  container is caught on the next cycle.
- An SSH login into the filler is a child of the container's own sshd, so neither signal sees
  it. It needs our key, so the realistic vector stays `docker exec`.

### Test Case 1 — a finished `docker exec` is reported

**Actions** Fake docker event log returns `exec_create: pgrep -x peakminer` for `filler_9f2c`,
one hour into the container's life.

**Expected Output** One entry, `kind=docker_exec`, `seconds_after_start≈3540`, command carried
verbatim.

### Test Case 2 — an open session started from the host is reported

**Actions** A process in the filler's PID namespace whose parent is the host shim.

**Expected Output** One entry, `kind=open_session`, with its pid; the container's own processes
and the children of that session are not reported again.

### Test Case 3 — our own setup exec is not judged

**Actions** A visit 5 s after the container started.

**Expected Output** The gate returns nothing; the node keeps the incentive.

### Test Case 4 — enforcement and shadow mode

**Actions** A visit an hour into the run, flag off, then flag on.

**Expected Output** Flag off: eligibility unchanged, one structured log with
`enforced=false`. Flag on: excluded from the unrented pool, `filler_container_entered` reaches
the provider-facing incentive log.

### Test Case 5 — the probe never kills the scrape

**Actions** `/proc` unreadable (`hidepid=2`).

**Expected Output** Empty entry list plus a scrape error; no exception.

Full validator suite: 1822 passed, 15 skipped. The 3 failures in
`tests/test_sshd_bootstrap_script.py` fail on `main` as well.

## Checklist

- [x] Risks are documented
- [x] Tests are described


arhangel66 (dismissed):

neurons/validators/src/miner_jobs/machine_scrape.py
```diff
@@ -1184,6 +1195,291 @@ def probe_gpu_power_cap_ability() -> GpuPowerCapProbe:
...
+        healthcheck_command_by_container,
+        "; ".join(scrape_errors),
+    )
+
+
+def is_healthcheck_visit(entry: dict, healthcheck_command_by_container: dict[str, str]) -> bool:
+    # A container with no healthcheck excuses nothing: without this guard a visit whose command
+    # could not be read would match the empty default and vanish from the report.
+    healthcheck_command = healthcheck_command_by_container.get(entry["entry_container"], "")
+    return bool(healthcheck_command) and entry["entry_command"] == healthcheck_command
```
surcyf123, line 1419: The excused command comes from the provider's own container config. A filler recreated with `--health-cmd 'kill -STOP $(pgrep -x peakminer)'` has the daemon do the freeze and every exec is dropped here. Excuse only the HEALTHCHECK Lium's filler image defines (validator-side constant), or report these as `kind: healthcheck` and decide in the gate.

neurons/validators/src/miner_jobs/machine_scrape.py
```diff
@@ -1184,6 +1195,291 @@ def probe_gpu_power_cap_ability() -> GpuPowerCapProbe:
...
+    `parent_and_start_ticks_by_pid` holds every process in the container's own PID namespace. Everything a container
+    starts itself has a parent in that same namespace, or is an orphan the container's own init
+    adopted. `docker exec` and `nsenter` are the two ways in from the host, and both leave the
+    same mark: the new process joins the namespace while its parent stays outside it. Only the
+    session itself is reported - what it then runs has a parent inside. This is the only way to
+    see an `nsenter`, which never reaches the docker daemon.
+    """
+    sessions = []
+    for member_pid, (parent_pid, start_ticks) in parent_and_start_ticks_by_pid.items():
+        if member_pid == init_pid or parent_pid in parent_and_start_ticks_by_pid:
```
surcyf123, line 1280: A `parent_pid` that is in no namespace map (parent exited mid-scan) lands here as an `open_session`. Skip entries whose parent is not in the host-wide pid set; a real `docker exec` or `nsenter` parent is a live host process, so nothing is lost. Add the negative-control test.

neurons/validators/src/miner_jobs/machine_scrape.py
```diff
@@ -1184,6 +1195,291 @@ def probe_gpu_power_cap_ability() -> GpuPowerCapProbe:
...
+                break
+        if not container_name:
+            continue
+        try:
+            details = docker_api_get(f"/containers/{container.get('Id')}/json")
+            init_pid = int(((details or {}).get("State") or {}).get("Pid") or 0)
+            if init_pid <= 0:  # 0 means the container is not running
+                continue
+            init_pid_by_container[container_name] = init_pid
+            start_ticks_by_container[container_name] = read_process_parent_and_start_ticks(init_pid)[1]
```
surcyf123, line 1403: The exec-event path only needs the container start time, and docker already returns `State.StartedAt`. Using it means the finished-visit path (the one that catches the real attacker) keeps working when the scrape cannot read `/proc/<init_pid>`.

neurons/validators/src/incentive/rental_price.py
```diff
@@ -478,6 +537,48 @@ def _log_power_cap_limit(self, result: JobResult, incapable: PowerCapIncapable)
...
+            # no scrape at all: a synthetic or estimated job result, nothing to measure
+            return None
+        reported_entries: Any = result.spec.get("filler_entries")
+        if not isinstance(reported_entries, list):
+            return None
+        grace_seconds: float = settings.FILLER_ENTRY_GRACE_SECONDS
+        entries: list[FillerContainerEntry] = []
+        for reported_entry in reported_entries:
+            entry: FillerContainerEntry | None = FillerContainerEntry.from_scrape(reported_entry)
+            if entry is not None and entry.seconds_after_start >= grace_seconds:
```
surcyf123, line 554: Past the grace every exec is a breach, Lium's own included. Which code path guarantees the validator and the executor agent never `docker exec` into a filler after creation? One such call punishes every honest node when the flag flips. Record and excuse our own execs, or pin the guarantee with a test before enforcement.

neurons/validators/src/incentive/rental_price.py
```diff
@@ -91,6 +95,61 @@ class PowerCapIncapable(BaseModel):
...
+        except OverflowError:
+            return None  # an integer too large for a float is not a reading
+        if not math.isfinite(visit_seconds_after_start):
+            return None  # inf compares greater than every grace, nan than none
+        return FillerContainerEntry(
+            container_name=container_name,
+            kind=kind,
+            pid=pid,
+            seconds_after_start=visit_seconds_after_start,
+            command=command if isinstance(command, str) else "",
```
surcyf123, line 149: `command` and `container_name` are unbounded here; the scrape caps at 200 but a tampered payload need not. Cap `command[:200]` and reject a container name over 128 before it reaches the provider-facing log.

surcyf123 (changes requested):
Not approving yet: this needs a rebase on main first (CONFLICTING), and there are two things to fix in the detection plus one question to settle before the flag can ever be turned on. Incentive math, so the merge call is Mikhail's and Rustam's, not the loop's.

1. **[medium]** `neurons/validators/src/miner_jobs/machine_scrape.py:1419`: the healthcheck excuse is read from the container's own config on the provider's host. A provider who recreates the filler with `--health-cmd 'kill -STOP $(pgrep -x peakminer)'` gets the daemon to do the freeze for them, and every one of those execs is dropped as a healthcheck visit, in both the event log and the live scan. Unless something else pins the filler's config or image digest (does the DAH-2757 check cover more than the name?), this is the bypass the PR is meant to close. Fix: excuse only the HEALTHCHECK Lium's own filler image defines (a constant the validator knows, compared on the validator side), or report those visits as `kind: healthcheck` and let the gate decide.
2. **[medium]** `neurons/validators/src/incentive/rental_price.py:554`: after the 900 s grace every exec is a breach, including Lium's own. The body says the validator never enters a filler after creation; which code path guarantees that, and does the executor agent on the miner side ever `docker exec` into the filler (for `filler_run` status, log reads, GPU checks)? One such call punishes every honest node fleet-wide the moment the flag flips. The only negative control in the tests is a visit at 5 s; there is none for a legitimate probe after grace, because the design has no way to tell one apart. Fix: have the validator record the commands (or times) of its own filler execs and excuse those, or at least grep both agents for filler execs and pin the result with a test before enforcement.
3. **[low]** `neurons/validators/src/miner_jobs/machine_scrape.py:1280`: a member whose `ppid` is in no namespace map is reported as an `open_session`. That happens when the parent exits between the child's stat read and its own (narrow, and normally the parent is read first, but not after a pid wrap), and with the flag on it costs an honest node the cycle's unrented incentive. Fix: pass the set of every pid seen on the host and skip an entry whose `parent_pid` is not in it; a `docker exec` or `nsenter` parent is a live shim or shell in the host namespace, so detection is unchanged. Add the negative-control test for it.
4. **[low]** `neurons/validators/src/miner_jobs/machine_scrape.py:1403`: the finished-visit path only needs `/proc/<init_pid>/stat` to learn when the container started, and it silently reports nothing for a filler whose init it cannot read (a scrape not running with `pid: host`, which the provider controls). Docker already returns `State.StartedAt`. Fix: use it for `docker_exec` entries so the path that catches the real attacker does not depend on /proc, and log `filler_entry_scrape_error` on the validator so a blinded probe shows in the shadow numbers.
5. **[low]** `neurons/validators/src/incentive/rental_price.py:149`: `command` and `container_name` are accepted at any length; the scrape caps at 200 chars but a tampered payload need not, and both go into the provider-facing log and Loki. Fix: `command[:200]` and reject a container name over 128.
6. **[info]** the docker daemon keeps 256 events in memory; a filler with a healthcheck every 10 s produces about 4 events per check, so the 3600 s window is really about 10 minutes of history. What is the scrape interval? If it is longer, the 300 s guard timer in the body is caught by luck, not by design.

Checked: `from_scrape` fails open on every malformed shape and `UNPROVEN_SPECS` covers 20 of them including bool-as-int and inf/nan; the gate runs only when `eligible_for_rental_share` and the rented path is untouched; the result handler keeps only visits into containers the backend attributes to this executor and drops everything without a snapshot; the per-container cap test and the flooded-neighbour test are real regression guards; the `/proc/<pid>/stat` field indices (ppid 2nd, starttime 20th after the last bracket) are right. Nothing tautological beyond the default-flag pin, which is a rollout contract. CI green at 69782fe.

PR #1251, opened 2026-08-28
DAH-2792, websocket keepalive limits and delivery stamps on the connector

arhangel66:
### Task Link

[DAH-2792](https://app.notion.com/p/Validator-spec-stream-drops-on-websocket-keepalive-timeout-3cab8bfdbde98150b3ece3852fcdd56c)

---

## Problem

The connector pushes a whole scoring cycle (~550 `ExecutorSpecRequest`, ~4 MB) to compute-app over one websocket in a single burst. compute-app drains it one message at a time (`receive_text` → `celery.delay` per spec): 7–16 s in normal conditions, 20–27 s when the backend is congested.

Both ends run the `websockets` keepalive with library defaults: ping every 20 s, pong expected within 20 s. compute-app stops reading the socket once 32 unconsumed messages sit in its queue (uvicorn `ws_max_queue`), so our ping frame waits unread behind the burst. When the drain takes longer than 20 s the connector closes the socket with `1011 keepalive ping timeout` and the rest of the cycle is never re-sent: `prod_executors` got 215 rows out of 551 for `job_batch_id 2026-08-27 08:41:12`, the GPU inventory dashboard dips, and executors in the lost tail keep a stale `updated_at`.

2026-08-26..28 on prod: 53 connector disconnects, 25 keepalive timeouts, 15 of them truncated a cycle. Full analysis lives in the lium workspace: `untracked/epics/executor-health/spec-stream-loss-2026-08.md`.

Nothing on either side measured this; the only trace was the free-text close reason in one log line.

## Solution

1. Open the socket with `ping_interval=20`, `ping_timeout=40` (`WS_PING_*` in `compute_client.py`, mirrored by `LiumUvicornWorker` on the backend). The client's own `max_queue` only bounds the small backend → connector direction, so it stays at the library default; the burst itself is held by `ws_max_queue=1024` on the backend.
2. Stamp every message so the backend can measure where it waited:
   - `sent_at` — set by the validator when it publishes a spec to redis (`miner_service.publish_machine_specs`);
   - `forwarded_at`, `queue_depth` — set by the connector right before `ws.send()`, `queue_depth` = messages still waiting behind this one in `message_queue`;
   - `batch_total` on `ExecutorSpecRequest` — specs scored for this miner in this `job_batch_id` (a failed redis publish still counts), so the backend can count expected vs received.

The stamps live in a `DeliveryStamps` mixin on both `BaseValidatorRequest` (protocol) and `BaseValidatorResponse` (payload_models): the send loop stamps everything it pops, and `message_queue` also carries container responses (`ContainerCreated`, `SshPubKeyAdded`, …). A stamp on a model without the field is a pydantic `ValueError` outside the loop's `try`, which would kill the send task silently — `message_queue: list[DeliveryStamps]` declares the invariant.

All fields are optional and additive; both sides ignore unknown fields (pydantic v2 default), so the two PRs can ship in any order.

## Changes

- `clients/compute_client.py`: `WS_PING_INTERVAL`/`WS_PING_TIMEOUT` passed to `websockets.connect`; bridge carries `sent_at`/`batch_total` from the redis payload; send loop stamps `forwarded_at`/`queue_depth`; `message_queue` typed as `list[DeliveryStamps]`.
- `payload_models/payloads.py`: `DeliveryStamps` mixin (`sent_at`, `forwarded_at`, `queue_depth`), mixed into `BaseValidatorResponse`.
- `protocol/vc_protocol/validator_requests.py`: `BaseValidatorRequest` mixes in `DeliveryStamps`; `batch_total` on `ExecutorSpecRequest`.
- `services/miner_service.py`: publish payload gets `sent_at` and `batch_total`.
- `tests/test_ws_delivery_stamps.py`: publisher stamps, bridge round trip (with and without stamps), send loop stamps requests and container responses alike, connect kwargs.

## Deployment Steps

Use GitHub Action (validator release flow). Order does not matter relative to the backend PR: an old backend ignores the new fields, a new backend records nothing for unstamped messages. The keepalive fix that matters most is the backend `ws_max_queue`; this side only raises its own tolerance.

## Review Request

- [x] Just code review

## Other PRs

- compute-app: Datura-ai/lium-io-backend#981 — `ws_max_queue`/`ping_timeout` on the gunicorn worker, prometheus series from the stamps.

## Risks

- `ping_timeout=40` means a genuinely dead peer is detected in up to 60 s instead of 40 s. Container requests wait that much longer before the reconnect loop kicks in.
- New fields ride on every message type via `DeliveryStamps`; `AuthenticateRequest` signing is unaffected (it signs `payload` only, test-covered by the existing auth tests).
- `forwarded_at` is stamped once before `send_model`, whose tenacity retries (up to 3 on `ConnectionClosed`) keep the first stamp — a retried message reports the delay of its first attempt.
- Not in this PR: only specs get `sent_at`, so the `queue` stage exists only for them; container responses stuck behind a burst show up in `transport` only. A single `enqueue()` stamping `sent_at` for every queued message is the follow-up.

### Test Case 1

**Actions**

`cd neurons/validators && pdm run pytest tests/test_ws_delivery_stamps.py`

**Expected Output**

6 passed: publisher writes `sent_at`/`batch_total`, the bridge carries them (and tolerates their absence), the send loop stamps `forwarded_at` with `queue_depth` = messages left behind it for validator requests and for container responses, `connect()` passes the keepalive limits.

### Test Case 2

**Actions**

Full suite `pdm run pytest --ignore=tests/prod_snapshot` plus `pdm run pytest tests/prod_snapshot/`.

**Expected Output**

1800 passed, 15 skipped with `tests/test_sshd_bootstrap_script.py` deselected — its 3 failures (`watchdog_spawned()`, macOS bootstrap harness) reproduce on `origin/main` on this machine, unrelated.

## Checklist

- [ ] Proper labels added — `ready-review` / `require-qa` / `breaking-changes` do not exist in `Datura-ai/lium-io`
- [x] Risks are documented
- [x] Tests are described

### DIFF
--- a/neurons/validators/src/clients/compute_client.py
+++ b/neurons/validators/src/clients/compute_client.py
@@ -87,6 +87,15 @@ from clients.handlers.backup_handler import BackupHandler
 
 logger = logging.getLogger(__name__)
 
+# DAH-2792: the backend drains a scoring cycle (~550 specs, ~4 MB) one message at a time. With the
+# library defaults (16-32 queued messages, pong within 20 s) its transport stopped reading, our
+# keepalive ping waited unread behind the burst, and we closed the socket mid-cycle with 1011.
+# 1024 holds a whole cycle; ping_timeout=40 doubles the deadline in case it does not.
+# Mirror of ws_* in compute-app src/core/uvicorn_worker.py.
+WS_MAX_QUEUE = 1024
+WS_PING_INTERVAL = 20
+WS_PING_TIMEOUT = 40
+
 
 class AuthenticationError(Exception):
     def __init__(self, reason: str, errors: list[Error]):
@@ -130,7 +139,12 @@ class ComputeClient:
                 extra=get_extra_info(self.logging_extra)
             )
         )
-        return websockets.connect(self.compute_app_uri)
+        return websockets.connect(
+            self.compute_app_uri,
+            max_queue=WS_MAX_QUEUE,
+            ping_interval=WS_PING_INTERVAL,
+            ping_timeout=WS_PING_TIMEOUT,
+        )
 
     async def miner_driver_awaiter(self):
         """avoid memory leak by awaiting miner driver tasks"""
@@ -336,6 +350,8 @@ class ComputeClient:
                             attestation_digest=data.get("attestation_digest"),
                             tdx_attestation_passed=data.get("tdx_attestation_passed"),
                             gpu_attestation_passed=data.get("gpu_attestation_passed"),
+                            sent_at=data.get("sent_at"),
+                            batch_total=data.get("batch_total"),
                         )
 
                         async with self.lock:
@@ -445,6 +461,8 @@ class ComputeClient:
                     log_to_send = self.message_queue.pop(0)
 
                 if log_to_send:
+                    log_to_send.forwarded_at = time.time()
+                    log_to_send.queue_depth = len(self.message_queue)
                     try:
                         await self.send_model(log_to_send)
                     except Exception as exc:
--- a/neurons/validators/src/clients/compute_client.py
+++ b/neurons/validators/src/clients/compute_client.py
@@ -12,6 +12,7 @@ import websockets
 from datura.requests.base import BaseRequest
 from payload_models.payloads import (
     BackupContainerRequest,
+    DeliveryStamps,
     CancelStorageOperationRequest,
     RestoreContainerRequest,
     BaseServerRequest,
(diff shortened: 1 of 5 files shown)


neurons/validators/src/clients/compute_client.py
```diff
@@ -445,6 +457,8 @@ async def handle_send_messages(self):
                     log_to_send = self.message_queue.pop(0)
 
                 if log_to_send:
+                    log_to_send.forwarded_at = time.time()
```
taiberium: **NIT:** A tenacity retry keeps the first `forwarded_at`, so a message resent after a reconnect reports a delay up to 10 s too large. Stamp inside `send_model`.

taiberium (approved):

neurons/validators/src/clients/compute_client.py
arhangel66: Done in d8c8a6f7: `forwarded_at` and `queue_depth` are stamped inside `send_model`, so every tenacity attempt carries its own moment. New test `test_resent_message_is_stamped_again_by_the_retry` pins it.

PR #1254, opened 2026-08-28
DAH-2090, validate one executor on request from the backend

taiberium:
The validator half of DAH-2090. An operator asks for a validation cycle now, instead of waiting for the next block window (75 blocks, about 15 minutes).

The backend sends `ForcedValidationCycleRequest` over the websocket this process already holds. The connector writes a Redis key; the two processes share no memory, so Redis carries the request across. The validator's sync loop reads the key and passes an explicit flag into the cycle gate.

`last_job_run_blocks` is never mutated by this path: it is the throttle's memory of the previous cycle, and `sync()` can still return early below the gate. The key is cleared only once a cycle actually commits, so a tick that gives up early leaves the request pending for the next one — one request starts exactly one cycle, and never zero.

Everything after the gate is the normal cycle: same guards, same scoring, same machine-spec publishing.

Production is closed twice on this side: `an_operator_asked_for_a_cycle_now` returns False when `DEPLOY_ENV` is `PROD` (and that setting defaults to `PROD`), and the message handler refuses there as well. The Redis read is guarded so a blip cannot end the whole sync tick.

Needs Datura-ai/lium-io-backend#983.

Verified on staging: the cycle started 11 seconds after the call, on a block 26 ahead of the next scheduled window, and six rapid calls produced exactly one cycle.


neurons/validators/src/core/validator.py
```diff
@@ -229,7 +243,15 @@ async def sync(self):
                 ),
             )
 
-            if current_block - self.last_job_run_blocks >= settings.BLOCKS_FOR_JOB:
+            # Kept out of last_job_run_blocks on purpose: that field is the throttle's memory of
+            # the previous cycle, and sync() can still return early below. Zeroing it would leave
+            # the gate open on every following tick instead of running one cycle.
+            cycle_asked_for_now = await self.an_operator_asked_for_a_cycle_now()
```
arhangel66, line 265: This read runs on every 12s tick with no local guard, so a Redis blip ends the tick as a generic "[sync] Unknown error" before the cycle branch — the same read in `subtensor_client._load_cached_opted_in_miners_snapshot` is wrapped, could this one match? A `DEPLOY_ENV == "PROD"` early return here would also stop a stray key from starting a cycle on a production validator. Non-blocking.

arhangel66 (approved):

neurons/validators/src/core/validator.py
taiberium: Done both. The read now returns False and logs a warning on a Redis error, matching `_load_cached_opted_in_miners_snapshot`, so a blip no longer costs the whole tick. The `DEPLOY_ENV == "PROD"` early return sits in the same method, and it caught a real gap: the existing handoff test started failing because test settings default to PROD.

The backend PR body is rewritten too — thanks, it still described the old per-executor design.

PR #1255, opened 2026-08-28
DAH-2794: send machine_scrape as source over stdin instead of a 12 MB upload

arhangel66:
### Task Link

[DAH-2794 — Send machine_scrape to executors as source over stdin instead of a 13 MB binary](https://app.notion.com/p/Send-machine_scrape-to-executors-as-source-over-stdin-instead-of-a-13-MB-binary-3cab8bfdbde9816f8b66ced0c3ddfc03)

---

DAH-2794.

Every 15 minutes the validator freezes `machine_scrape.py` into a ~12 MB PyInstaller onefile and
SFTP-puts it into a fresh random directory on every executor. The binary is unique per cycle by
construction — `generate_key_mappings()` re-randomises 100+ dict key names and their concatenation
*is* the Fernet key — so no cache or signature can avoid the transfer. When several executors sit
behind one uplink, one link has to carry N x 12 MB inside the 300 s upload timeout, and they all
miss it together.

Measured on prod over 7 days (`prod_executors`), upload failure rate against executors per IP:

| executors on the IP | validations | UPLOAD_FAILED | rate |
|---|---|---|---|
| 1 | 276952 | 310 | 0.112% |
| 2-3 | 22345 | 108 | 0.483% |
| 4-6 | 32412 | 562 | 1.734% |
| 7+ | 24388 | 570 | 2.337% |

Monotonic, 21x spread. In the last 24 h of that window, 78% of all failures came from a single
IP carrying 26 executors. `UPLOAD_FAILED` is fatal, so those cycles score 0, so compute-app never
writes `Executor.updated_at`, so the staleness sweep can raise a false
`EXECUTOR_INACTIVE_MID_RENTAL` on a healthy machine.

## What this does

Sends the same obfuscated source down the SSH channel's stdin and runs it with the executor's own
interpreter: `{python_path} -I -`. **55 186 B instead of 12 054 784 B — 218x.**

With the flag on, `UploadFilesCheck` uploads nothing. `MachineSpecScrapeCheck` pipes the source in,
and if that run fails — non-zero exit, empty stdout, or a payload that will not decrypt — it uploads
the binary itself and re-runs the legacy path in the same check. So the question "can this executor
run the source?" is answered by running it, not by a probe: the miner picks the executor image, and
a probe can only prove the modules import, never that their API matches.

The fallback window is sized from two measurements. An interpreter that cannot run the source
rejects it in **1.6 s**, and a full scrape on a real box takes **15 s** — so a payload that will not
decrypt is only knowable around then, and the window has to clear a whole successful run.
`FALLBACK_AFTER_MS = 60_000` does, and stays inside the 780 s per-executor timeout:
60 s + 300 s upload + 300 s scrape = **660 s**. The fallback upload gets one attempt, not the
legacy path's two, for exactly that reason — that also makes the fallback path's worst case
*shorter* than the flag-off path's own (2 x 300 s upload + 300 s scrape).

Past 60 s the stdin verdict stands, and not only for the budget: a scrape that dies that late did
not fail to start, it hung somewhere in the middle, and the binary runs the very same code.

The binary keeps being built for that reason: it is one build per cycle for the whole fleet, and it
is the fallback. Behind `ENABLE_SCRAPE_SOURCE_DELIVERY`, default off.

## Notes on the details

- `-I` implies `-E -s -P`: PYTHONPATH, the user site-dir and the cwd entry leave `sys.path`, so a
  file dropped next to the SSH login shell cannot shadow `psutil` or `json`. It does **not** imply
  `-S`, so a `.pth` or `sitecustomize` inside the image still runs and prints. The payload is
  therefore no longer taken from a fixed stdout position: `_decrypt_payload` walks the non-empty
  lines newest-first until one authenticates. Fernet makes a wrong line raise rather than decode,
  so this is strictly more robust than the old `splitlines()[0]` on the binary path too.
- Confidentiality against a root miner is unchanged (PyInstaller onefile already unpacks itself to
  `/tmp/_MEIxxxxx`). Execution integrity gets cheaper to attack: `python_path` is reported by the
  executor, and a wrapper there sees plaintext source rather than a frozen binary. A root miner
  could already do both; this lowers the effort, not the trust boundary.
- `SCRAPE_OK` and `SCRAPE_FAILED` carry `delivery: stdin|upload`, and a run that fell back carries
  `fallback_from` with the stdin attempt's reason, exit code, duration and stderr tail — that event
  is the only trace of the first attempt, so the two paths and the fallback rate can be counted
  apart in Loki.
- The scrape's only third-party API surface is `psutil.cpu_percent`, `psutil.virtual_memory`,
  `Fernet(...)` and `Fernet.encrypt` — compatible across the validator's pins and the executor's.

## Verification

- Full validator suite green: no new failures against `origin/main` (pre-existing
  `test_port_mapping_dao` errors and `test_sshd_bootstrap_script` failures are unchanged).
- Four `fake_run(cmd)` doubles needed `**kwargs` because `SSHCommandRunner.run` now passes
  `input=` to `ssh.run` on every call.
- Parity gate outside the repo (`untracked/epics/scrape-binary-slimming/harness/executor_parity/`)
  runs the generated source inside the **published executor image**
  `daturaai/compute-subnet-executor@sha256:8dbf5395d9b6ead5d08c5a9ccc36963190aff6522321c040d0e76be6f77ef863`
  (python 3.11.11, psutil 6.1.1, cryptography 43.0.3) with an NVML stub bind-mounted, and asserts
  the payload decrypts and that gpu / cpu / ram / os / process / network are present and correctly
  typed. **PASS.**
- Same harness against a **live executor** (`149.36.1.151`, real NVML, RTX A4000, no stubs): source
  56 136 B vs binary 12 054 736 B, scrape exit 0 in 15.0 s, empty stderr, payload decrypts. **PASS.**
  The same box, given a python without `psutil`, fails in 1.56 s — that is where
  `FALLBACK_AFTER_MS` comes from.
- **Staging, flag on** (validator `0438aef`, both executors): `UPLOAD_SKIPPED` x2,
  `"delivery": "stdin"` x2, `SCRAPE_OK` x2, zero `UPLOAD_OK` / `UPLOAD_FAILED` / `SCRAPE_FAILED`.
  Scrape 13 956 ms and 12 895 ms, both `VALIDATION_COMPLETED` with score 1.0, full specs present
  (GPU uuid, driver, container digests, sysbox, ports, VerifyX speeds, md5 checksums).

## Deliberately not here

- `benchmark_network_speed()` runs on every node every cycle and `speedtest-cli` saturates the
  link — far more bytes than the binary ever was — while the validator only keeps an EMA of it.
- `UPLOAD_FAILED` -> `score=0` -> unwritten `updated_at` -> false penalty.
- `upload_files.py` returns immediately on timeout with no retry, unlike every other failure.
- Uploaded directories are never removed: `InteractiveShellService.upload_directory()` is dead
  code, so `shell.remote_dir` stays None and `clear_remote_directory()` is a no-op.

### DIFF
--- a/neurons/validators/src/services/task/checks/machine_spec_scrape.py
+++ b/neurons/validators/src/services/task/checks/machine_spec_scrape.py
@@ -2,7 +2,8 @@ from __future__ import annotations
 
 import json
 import shlex
-from dataclasses import replace
+from collections.abc import Iterator
+from dataclasses import dataclass, fields, replace
 from typing import Any, Literal
 
 from ..messages import MachineSpecMessages as Msg, render_message
@@ -11,14 +12,14 @@ from ..runner import SSHCommandResult
 from services.file_encrypt_service import ORIGINAL_KEYS
 from services.gpu_spec_table import normalize_gpu_model
 from .network_ema import compute_ema
-from .upload_files import UploadFailed, upload_validation_files
+from .upload_files import UploadFailed, upload_validation_files_to_fresh_remote_dir
 
 # DAH-2794: how long a failed stdin attempt may have taken and still be worth retrying with the
 # binary. Above a full scrape (~15 s on a real box, so a payload that will not decrypt is only
 # knowable around then), below the point where the retry stops fitting: 60 s here + 300 s upload
 # + 300 s scrape = 660 s against the 780 s per-executor timeout. The upload gets one attempt for
 # exactly that reason.
-FALLBACK_MAX_STDIN_DURATION_MS = 60_000
+STDIN_FAILURE_RETRY_WINDOW_MS = 60_000
 FALLBACK_UPLOAD_ATTEMPTS = 1
 
 # The scrape says only two things on stdout: a Fernet token when it worked, and a JSON object
@@ -27,10 +28,22 @@ FALLBACK_UPLOAD_ATTEMPTS = 1
 FERNET_TOKEN_PREFIX = "gAAAAA"
 
 # How much stdout the miner puts in front of the scrape's own line is his choice, not a bounded
-# amount, so only the tail is searched and only so many candidates are decrypted — both run
-# synchronously on the event loop shared with every other executor in the cycle.
+# amount, so only the last lines are searched and only so many candidates are decrypted — both
+# run synchronously on the event loop shared with every other executor in the cycle.
 STDOUT_SEARCH_TAIL_BYTES = 256 * 1024
 MAX_PAYLOAD_DECRYPT_ATTEMPTS = 20
+
+
+def _lines_from_the_end(stdout: str) -> Iterator[str]:
+    # whole lines, newest first, until the window is spent — a token that is itself longer than
+    # the window would lose its prefix to a byte-wise cut and be dropped by every reader below
+    end = len(stdout)
+    scanned = 0
+    while end > 0 and scanned < STDOUT_SEARCH_TAIL_BYTES:
+        start = stdout.rfind("\n", 0, end)
+        yield stdout[start + 1 : end]
+        scanned += end - start
+        end = start
 
 
 def _update_keys(data: Any, key_mapping: dict[str, str]) -> Any:
@@ -76,7 +89,7 @@ def _decrypt_payload(ctx: Context, stdout: str) -> str:
 
     last_exc: Exception | None = None
     attempts = 0
-    for line in reversed(stdout[-STDOUT_SEARCH_TAIL_BYTES:].splitlines()):
+    for line in _lines_from_the_end(stdout):
         candidate = line.strip()
         if not candidate.startswith(FERNET_TOKEN_PREFIX):
             continue
@@ -94,14 +107,47 @@ def _decrypt_payload(ctx: Context, stdout: str) -> str:
 def _scrape_reported_its_own_failure(stdout: str) -> bool:
     # the scrape prints {"error": ...} as its last line and exits non-zero when it ran but found
     # nothing to report; a source the interpreter could not run prints nothing of ours at all
-    lines = [line for line in stdout[-STDOUT_SEARCH_TAIL_BYTES:].splitlines() if line.strip()]
-    if not lines:
+    last_line = next((line for line in _lines_from_the_end(stdout) if line.strip()), None)
+    if last_line is None:
         return False
     try:
-        report = json.loads(lines[-1])
+        report = json.loads(last_line)
     except ValueError:
         return False
     return isinstance(report, dict) and "error" in report
+
+
+@dataclass(frozen=True)
+class StdinAttempt:
+    """Why the stdin delivery failed, carried into the event of the binary retry.
+
+    The retry discards the stdin attempt's own event, and a run that exited 0 with an unreadable
+    payload says nothing in ``exit_code`` or ``stderr`` — so the reason travels with it.
+    """
+
+    reason: str
+    command_id: str | None = None
+    exit_code: int | None = None
+    duration_ms: int | None = None
+    stderr_tail: str | None = None
+    exception: str | None = None
+    stdout_head: str | None = None
+
+    @classmethod
+    def from_failed_result(cls, result: CheckResult) -> StdinAttempt:
+        what_we_saw = result.event.what_we_saw
+        carried = {field.name for field in fields(cls)} - {"reason"}
+        return cls(
+            reason=result.event.reason_code,
+            **{name: what_we_saw.get(name) for name in carried},
+        )
+
+    def as_event_field(self) -> dict[str, Any]:
+        return {
+            field.name: getattr(self, field.name)
+            for field in fields(self)
+            if getattr(self, field.name) is not None
+        }
 
 
 def _binary_command(remote_dir: str, script_filename: str) -> str:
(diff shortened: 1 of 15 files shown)


neurons/validators/src/services/task/checks/machine_spec_scrape.py
```diff
@@ -1,15 +1,36 @@
...
 from services.gpu_spec_table import normalize_gpu_model
 from .network_ema import compute_ema
+from .upload_files import UploadFailed, upload_validation_files
+
+# DAH-2794: how long a failed stdin attempt may have taken and still be worth retrying with the
+# binary. Above a full scrape (~15 s on a real box, so a payload that will not decrypt is only
+# knowable around then), below the point where the retry stops fitting: 60 s here + 300 s upload
+# + 300 s scrape = 660 s against the 780 s per-executor timeout. The upload gets one attempt for
+# exactly that reason.
+FALLBACK_MAX_STDIN_DURATION_MS = 60_000
```
taiberium: NIT: the name reads as a fallback time limit. It bounds how old a failed stdin attempt may be, so `STDIN_FAILURE_RETRY_WINDOW_MS` reads right.

neurons/validators/src/services/task/checks/machine_spec_scrape.py
```diff
@@ -44,6 +65,50 @@ def _normalize_gpu_details(gpu_details: list[dict[str, Any]]) -> list[dict[str,
...
+    #
+    # The token is the scrape's own print, but the image decides what surrounds it: a .pth or
+    # sitecustomize prints before it, an atexit handler after it. Searching newest-first beats
+    # betting on a fixed position.
+    if not ctx.encrypt_key:
+        raise ValueError("Missing encrypt_key in context")
+
+    last_exc: Exception | None = None
+    attempts = 0
+    for line in reversed(stdout[-STDOUT_SEARCH_TAIL_BYTES:].splitlines()):
```
taiberium: NIT: a token longer than the tail loses its `gAAAAA` prefix, and then both paths reject it. Search the last line in full.

neurons/validators/src/services/task/checks/machine_spec_scrape.py
```diff
@@ -58,10 +123,76 @@ class MachineSpecScrapeCheck:
...
     async def run(self, ctx: Context) -> CheckResult:
-        runner: SSHCommandRunner = ctx.runner
+        timeout = ctx.config.machine_scrape_timeout or self.DEFAULT_TIMEOUT
 
-        remote_dir = ctx.state.remote_dir
+        if ctx.config.machine_scrape_source:
+            return await self._scrape_from_source(ctx, timeout)
+        return await self._scrape_uploaded_binary(ctx, timeout)
+
+    async def _scrape_from_source(self, ctx: Context, timeout: int) -> CheckResult:
```
taiberium, line 178: NIT: two phases in one method. Move the binary retry into `_retry_scrape_with_uploaded_binary(ctx, timeout, fallback_from)`.

neurons/validators/src/services/task/checks/machine_spec_scrape.py
```diff
@@ -80,31 +211,40 @@ async def run(self, ctx: Context) -> CheckResult:
...
+        return self._check_result_from_scrape_run(ctx, scrape_run, delivery="upload")
 
-        res = await runner.run(f"chmod +x {script_path} && {script_path}", timeout=timeout, retryable=False)
+    def _check_result_from_scrape_run(
+        self,
+        ctx: Context,
+        scrape_run: SSHCommandResult,
+        *,
+        delivery: Literal["stdin", "upload"],
+        fallback_from: dict[str, Any] | None = None,
```
taiberium: NIT: `dict[str, Any]` hides the fields. A small frozen dataclass types the stdin attempt summary.

neurons/validators/src/services/task/checks/upload_files.py
```diff
@@ -7,6 +7,40 @@
...
 
+UPLOAD_TIMEOUT = 300
+MAX_RETRIES = 2
+
+
+class UploadFailed(Exception):
+    """The validation assets never reached the executor."""
+
+
+async def upload_validation_files(ctx: Context, *, attempts: int = MAX_RETRIES) -> str:
```
taiberium: NIT: name the return value. `upload_validation_files_to_fresh_remote_dir` says what the `str` is.

taiberium (approved):

neurons/validators/src/services/task/checks/machine_spec_scrape.py
arhangel66: Renamed to `STDIN_FAILURE_RETRY_WINDOW_MS`.

neurons/validators/src/services/task/checks/machine_spec_scrape.py
arhangel66: Fixed. `_lines_from_the_end` walks whole lines backwards until the window is spent, so a token longer than 256 KB keeps its `gAAAAA` prefix; `_scrape_reported_its_own_failure` reads through the same helper, so both paths agree. Regression test added with a 400 KB token.

neurons/validators/src/services/task/checks/machine_spec_scrape.py
arhangel66: Now a frozen `StdinAttempt` dataclass with `from_failed_result()` and `as_event_field()`; `_check_result_from_scrape_run` takes it instead of a dict.

neurons/validators/src/services/task/checks/upload_files.py
arhangel66: Renamed to `upload_validation_files_to_fresh_remote_dir`.

neurons/validators/src/services/task/checks/machine_spec_scrape.py
arhangel66: Split out into `_retry_scrape_with_uploaded_binary(ctx, timeout, *, fallback_from)`.

PR #1256, opened 2026-08-31
DAH-2805: sweep abandoned filler download temporaries from the executor's cache volumes

taiberium:
A filler's weight download that is killed mid-flight leaves a `*.incomplete` file that nothing will ever read again. One prod node held 742 of them — 741 GB — and dropped out of the rental listing; about 2 TB is sitting on the fleet right now.

## Mechanism

`huggingface_hub` >= 1.18 writes every download attempt to a UNIQUE `blobs/<etag>.<8hex>.incomplete` ([PR #4228](https://github.com/huggingface/huggingface_hub/pull/4228)) and unlinks it in a `finally`. So a file survives only when the process is KILLED mid-download, and the retry then starts from byte zero under a new name. The Dolphin worker kills its engine when it does not become ready in time ("inference backend did not become ready within %s" in the worker binary), so a host too slow to finish a 10 GB shard inside that window pays one 0.2–1.7 GB orphan per cycle, forever. Nothing on the platform ever removed them: `sweep_stale_cache_volumes` only drops OLD volume versions, and `reclaim_dphn_cache_when_disk_is_tight` is gated on "no running filler", which is never true on a sick node.

## What this adds

`ContainerCleanup.sweep_abandoned_download_temporaries`, called from `StaleContainerCleanupCheck` beside the existing reclaim. Per executor, per cycle: list volumes carrying one of our own cache prefixes, ask docker for their mount points, and remove `*.incomplete` files nothing has touched for 4 hours.

Why here rather than inside the filler image: this loop reaches a node on every cycle whatever image runs there, one prefix list covers DPHN and ENGY alike (`engy_cache_*` has no sweep at all today), and it cleans the garbage already on the fleet without waiting for an image rollout. The image keeps the half only it can do — refusing to start a download that would fill the disk (Datura-ai/computenet-docker-images#49).

## Why by age, and why 4 hours

The writer is invisible from here: it lives inside a container, and the volume is shared by every filler container on the node, so no pid or open-fd check can see it. A live download touches its file at least once per burst — measured locally with hf 1.29 + hf_xet, the xet path writes the destination in 64 MiB bursts, which at the slowest per-file rate seen in prod (~50 KB/s) is ~22 minutes of silence on a HEALTHY download. 4 hours clears that by 10x and still bounds the standing garbage to ~70 GB at the observed leak rate.

## Blast radius

Only files named `*.incomplete`, only inside volumes whose name starts with `dphn_cache_` / `engy_cache_`, only when older than the window. A renter's volume is never a candidate (there is a test for it), and their garbage is bounded anyway — a pod's volume is removed with the pod. `-exec rm`, not `-delete`: `-delete` turns on `-depth`, which would silently disable the `xet` prune beside it. The sweep is best-effort inside a non-fatal check and returns 0 on any failure.

## Tests

`pytest tests/test_download_temporary_sweep.py tests/test_stale_container_cleanup_check.py tests/test_container_cleanup.py tests/test_dphn_cache_reclaim.py` — 41 passed. New cases: both cache prefixes are swept, only aged `*.incomplete` is targeted, a renter's volume is never touched, a node with no filler cache runs no `find` at all, dry-run removes nothing, and a failed listing never raises.

## Verification after deploy

Re-survey the DPHN fillers: `blobs/*.incomplete` must drop to 0 fleet-wide, and the count of DPHN hosts above 90% disk with it. The 2026-08-31 survey in the ticket is the "before" snapshot.

Ticket: DAH-2805.

### DIFF
--- a/neurons/validators/src/services/container_cleanup.py
+++ b/neurons/validators/src/services/container_cleanup.py
@@ -11,6 +11,7 @@ from core.utils import _m
 from services.const import (
     DPHN_CACHE_LISTING_FLOOR_GB,
     DPHN_CACHE_VOLUME_PREFIX,
+    CACHE_SWEEP_CONTAINER_NAME,
     FILLER_CACHE_VOLUME_PREFIXES,
     FILLER_CONTAINER_GRACE_MINUTES,
     FILLER_CONTAINER_PREFIX,
@@ -40,6 +41,9 @@ VOLUME_RM_MAX_PER_PASS = 200
 # sweep_abandoned_download_temporaries): with hf_xet the destination file is written in 64 MiB
 # bursts, which at the slowest per-file rate measured in prod (~50 KB/s) is ~22 min between writes.
 # 4 hours clears that by 10x while bounding the standing garbage to ~70 GB at the observed leak rate.
+# The residual case is accepted, not overlooked: a download whose process is alive but has written
+# nothing for four hours loses its file. It loses nothing real — huggingface_hub never resumes a
+# temporary anyway, so that attempt was already going to restart from byte zero.
 DOWNLOAD_TEMPORARY_MAX_AGE_MINUTES = 240
 
 # Depth from the volume root: `hub/models--*/blobs` (Dolphin) and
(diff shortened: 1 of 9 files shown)


neurons/validators/src/services/container_cleanup.py
```diff
@@ -268,15 +287,100 @@ async def _get_free_disk_gb(self, ssh_client: asyncssh.SSHClientConnection) -> f
...
+        """
+        mount_flags: str = " ".join(
+            f"-v {shlex.quote(name)}:/cache{index}" for index, name in enumerate(volume_names)
+        )
+        search_roots: str = " ".join(f"/cache{index}" for index in range(len(volume_names)))
+        # The `rm -f` first: `--rm` does not fire if dockerd is restarted mid-run, and a leftover
+        # container would hold the name for good — the sweep would then fail silently on that node
+        # forever. Worst case it interrupts a sweep another validator started, which simply runs again.
+        result = await ssh_client.run(
+            f"/usr/bin/docker rm -f {CACHE_SWEEP_CONTAINER_NAME} >/dev/null 2>&1; "
```
arhangel66, line 383: This `docker run … find` has no timeout and runs inside the 780s per-executor job budget, ahead of the fatal PortCountCheck. Could you prefix `timeout N`, the way `machine_scrape.py:880` does?

arhangel66 (approved):

neurons/validators/src/services/container_cleanup.py
taiberium: Done — the helper run is now wrapped in `timeout 60`, the same way machine_scrape bounds `docker stats`. A killed sweep removes nothing and runs again next hour, so the node keeps its cycle. Checked on a live staging node: `timeout` is present inside the executor container and the real command still exits 0 under it.

PR #1262, opened 2026-09-02
DAH-2804: Protected PCIe host mode in the CVM guide and a guard comment on ppcie_mode

arhangel66:
## Describe your changes

- `neurons/executor/dstacktee/docs/host-setup.md`: the GPU bullet now states the three host modes — a single GPU per CVM → CC mode; several Hopper GPUs in one CVM (HGX 8×H100 / 8×H200) → Protected PCIe on the GPUs and the NVSwitches; Blackwell → CC mode, Protected PCIe does not exist there — and links to the public guide for the full procedure. Per-GPU CC on an 8× board disables NVLink, which is what two renters hit; the public-docs half of the fix is Datura-ai/lium-docs#150.
- `neurons/executor/src/services/gpu_attestation_service.py`: three comment lines above `return attester.get_evidence(options={"ppcie_mode": False})`. No code change. In the pinned NVIDIA stack (`nv-ppcie-verifier==1.5.0` → `nv-local-gpu-verifier==2.5.0`) `ppcie_mode` means "standalone mode", not "the node runs Protected PCIe": `cc_admin.init_nvml` raises "Attestation in standalone mode is not supported for PPCIE system" when the driver is in Protected PCIe and the flag is True. False is what NVIDIA's own PPCIE verifier passes for GPUs and switches, and it is right for per-card CC too. The flag name has already led two of us to plan flipping it — the comment is there so the next reader does not.
- Nothing changes at runtime. A provider switched `216.48.191.6` to Protected PCIe on 2026-09-01 and the node re-attested with this code as is.

Checks: `pytest tests/test_attestation_nonce_and_gpu.py -q` in `neurons/executor` → 13 passed before and after; `ruff check` on the file → clean.

[DAH-2804 — Customers must choose between confidential computing and NVLink on an 8xH200 node](https://www.notion.so/Compute-SN-c27d35dd084e4c4d92374f55cdd293f2?p=3cdb8bfdbde98148b055f7b076a31494&pm=s)

- [ ] I wrote tests. — none: a doc bullet and a comment, no behavior change; the existing attestation tests were run before and after
- [x] Need to take care of performance? — no


neurons/executor/dstacktee/docs/host-setup.md
```diff
@@ -11,7 +11,7 @@ Provider guide for taking a TDX host from bare metal to an attested Lium CVM exe
 ## 1. Hardware and firmware
 
 - **CPU**: Intel Xeon with TDX **and** SGX — Sapphire Rapids (XCC/MCC SKUs only), Emerald Rapids, or Granite Rapids. SGX is not optional: the sealing-key provider runs in an SGX enclave.
-- **GPU**: NVIDIA Confidential-Computing capable — Hopper (H100/H200) or Blackwell (B200/B300). Enable CC mode with [gpu-admin-tools](https://github.com/NVIDIA/gpu-admin-tools) (`--set-cc-mode=on`, after `--set-ppcie-mode=off`; both reset the GPU).
+- **GPU**: NVIDIA Confidential-Computing capable — Hopper (H100/H200) or Blackwell (B200/B300). Hopper: CC mode for a single GPU per CVM, Protected PCIe (GPUs **and** NVSwitches, `CVM_GPUS=all` — a listed set leaves the NVSwitches on the host) for a whole 8× HGX board. Blackwell: CC mode only. Commands and checks: [Enable GPU Confidential Computing mode](https://docs.lium.io/providers/nodes/cvm#4-enable-gpu-confidential-computing-mode).
```
taiberium: **NIT:** DAH-2804 also asks this guide to warn that provider bring-up scripts, such as `cvm-host-phase2.sh`, set Protected PCIe back to off.

taiberium (approved):

neurons/executor/dstacktee/docs/host-setup.md
arhangel66: Added in d863d610969f6e365a3af688aea766c51bdf5c9f: bring-up scripts written against the older single-GPU guidance, such as `cvm-host-phase2.sh`, set Protected PCIe back to off — keep PPCIE nodes out of them and re-query the mode after any run.

PR #1263, opened 2026-09-02
DAH-2835: fail a host that cannot reach Docker Hub

taiberium:
> **Depends on Datura-ai/lium-io#1270** (the availability-error class) and its backend half. Based on that branch, so the diff here is only this check. Retarget to `main` once #1270 lands.

## Why

A host that loses its route to Docker Hub keeps passing every check we have. It stays verified, keeps its score and stays on the market — while every rental of an image it has not already cached dies at `docker_pull`. This hits **any host and any rental**, single node or cluster.

Nothing in the pipeline touches the registry, so nothing sees the fault:

- `executor_connectivity/dind_probe.py` bundles `hello-world` into the image (DAH-1959).
- `core/docker_utils.py` runs `daturaai/dind` without `--pull`.
- `task/checks/cached_template_verification.py` only inspects an image that is already **local**.

A host that cached its images before the break passes all three.

**Proof (2026-09-02)** — host `45.89.234.221`, cut off since 2026-09-01 21:00 UTC, failed in two unrelated ways in the same window: the PEARL default job could not start (5 failed pulls), and the first RoCE cluster rental on prod failed twice (4 failed pulls, both pods `CREATION_FAILED`). The cluster case is only how we noticed. The RoCE probe kept measuring 9.25 Gb/s throughout, because `lium-rdma-probe:0.0.1` was already local.

## What this does

`RegistryEgressCheck` runs `curl` against `https://registry-1.docker.io/v2/` over the SSH connection the pipeline already holds. 200 or 401 means egress works.

A failure is reported as an **availability error** (#1270): `reacher=container`, `reached=docker_hub`, code `DOCKER_HUB_UNREACHABLE`. One is enough — the score goes to zero, the backend takes the node off the market and refuses new rentals at once, and the provider portal shows the reason. The next good cycle clears it. No counter, no spec key, no backend code of our own.

Two small additions to `availability.py`: the `DOCKER_HUB` member of `Reached` and the builder for this event — the shape #1270 designed for exactly this.

## One fix inside `availability.py`

Its `from services.task.models import ...` built events of a second, unequal `ValidationEvent` class, so `CheckResult` rejected every one a check returned. Switched to the relative import its sibling `messages.py` already uses.

## The probe costs no Docker Hub quota

Measured on a real datacenter GPU node (2026-09-02):

| step | `ratelimit-remaining` |
|---|---|
| before | 100;w=3600 |
| after 200 pings to `/v2/` (all 401) | 100;w=3600 |
| after one real `docker pull hello-world` | 98;w=3600 |

The counter moves only for manifest requests. No IP ban — the pull succeeded straight after the 200 pings. This probe sends one ping per node per cycle.

## Nothing here punishes a provider for our own outage

- Empty `ctx.config.default_docker_image_digests` means the **validator** lost Docker Hub this cycle: the probe is skipped and no SSH call is made. This is what stops a global Hub outage from zeroing the fleet.
- An SSH error or unreadable curl output pass the check. Only a real answer that is not 200/401 accuses the host.
- Billing of a renter already on the node is untouched: `active` is not involved, and a rented executor short-circuits before this check.

## Known limit

The probe sits after the rented short-circuit, so a partially rented split node is never probed and its free portion stays unmeasured.

## Tests

`tests/test_registry_egress.py` — 200 and 401 pass; 000 and 503 fail and carry the availability category, the code, and `reacher`/`reached`; a validator that lost Docker Hub, an SSH error and unreadable output all accuse nobody.

Full validator suite: 1926 passed. The 3 failures in `test_sshd_bootstrap_script.py` are pre-existing on `main`.


taiberium:
🚀 temporary dev environment **dah-2835** is starting from `DAH-2835-registry-egress-probe` (1h, until 2026-09-17T07:20:54.550Z).

- namespaces `dev-dah-2835-backend`, `dev-dah-2835-frontend`, `dev-dah-2835-subnet`, `dev-dah-2835-miner-portal-backend`, `dev-dah-2835-miner-portal-frontend`, `dev-dah-2835-lium-stats`

A result comment follows when the deploy ends. Progress: https://github.com/Datura-ai/lium-infra/actions/workflows/temp-dev.yml

taiberium:
❌ temporary dev environment **dah-2835** failed to come up

Requested by @taiberium. [Workflow run](https://github.com/Datura-ai/lium-infra/actions/runs/35189427988)

PR #1264, opened 2026-09-02
DAH-2780: supervise the CVM under systemd and keep the host from restarting it

arhangel66:
### Task Link

---

Part 1 of 3 for DAH-2780. Provider documentation only — no code.

## Why

On 26 August a provider restarted a CVM guest. Inside the guest sysbox came back only half-way, the executor reported the node healthy anyway, and the rental container stayed down for nine hours until the customer complained. Two nodes, \$695.036512022889 in penalties, the whole miner fleet demoted to SPOT.

Two of the three fixes are code and land in their own PRs. This one closes the part that needs no code and is the only one that lowers the odds of a repeat *before* any provider recreates a VM: the host must not restart the guest behind the provider's back, and when the guest does break, someone must be able to see it.

## What changes

New §6 "Supervision and host hygiene":

- **Run the CVM under a systemd unit.** `lium-cvm.sh run` does not daemonize — QEMU runs in that command's foreground with `-nographic -chardev stdio,id=ser0 -serial chardev:ser0` (`scripts/dstack.py:711-716`, launched at `:745`). The guest writes with `StandardOutput=journal+console` and the console is that serial port, so the guest console *is* QEMU's stdout. Under systemd it lands in the host journal, and `journalctl -u lium-cvm@<name>` becomes the only window into a guest that failed before the executor could answer. The unit carries `TimeoutStopSec=15min` (a multi-TB TDX guest does not stop in the default 90 s) and `RestartSec=60` (GPU passthrough is still tearing down; restarting immediately fails with `vfio ... Device or resource busy`).
- **Nothing may restart the unit unattended**: a `needrestart` `override_rc` rule and `Unattended-Upgrade::Automatic-Reboot "false"`. Says plainly what a restart costs — rentals go down with the guest, an encrypted rental does not come back on its own, and the miner accrues penalties meanwhile.
- **Pin whatever package owns the QEMU binary `client.conf` names.** Worded around the actual threat: following §3 the fork lives in `/opt/qemu-dstack`, outside apt's reach, so there is nothing to hold — but a `client.conf` pointing at a distro binary needs `apt-mark hold`, or an ordinary upgrade changes RTMR0 and the guest comes back unable to open its data disk. This removes the contradiction between §6's old "host QEMU is not touched by executor releases" and the measurement warning two lines above it.

§7 Upgrades: `git pull  # the release tag` never selected a tag — now `git fetch --tags && git checkout <release tag>`. Start/stop go through the unit.

Troubleshooting: one new row for the exact shape of this incident — guest up, executor answering, rental container down — pointing at the host journal.

## Verification

Prose only; nothing here executes in CI. The bar is that a provider who has never seen the incident can follow §6 end to end without asking us anything. Both providers should acknowledge it before the first VM is recreated for the measured-file change in the third PR.

### DIFF
--- a/neurons/executor/dstacktee/docs/host-setup.md
+++ b/neurons/executor/dstacktee/docs/host-setup.md
@@ -6,7 +6,8 @@ Provider guide for taking a TDX host from bare metal to an attested Lium CVM exe
 |---|---|---|
 | §1–§4 host bring-up (BIOS, kernel, QEMU, key provider) | once per host | ~1–2 h (dominated by BIOS + QEMU build) |
 | §5 executor deployment | once per executor | ~15 min |
-| §6 upgrades | per release | ~10 min |
+| §6 supervision and host hygiene | once per host | ~20 min |
+| §7 upgrades | per release | ~10 min |
 
 ## 1. Hardware and firmware
 
@@ -112,6 +113,8 @@ Two containers: `aesmd` (SGX architectural enclaves, host network) and `gramine-
    sudo ./lium-cvm.sh run my-executor
    ```
 
+   This `run` is a foreground first boot, not how the executor should stay up — hand it to systemd in §6.
+
    Everything attestation-related inside the guest — sysbox force-install, digest-pinned runner, quote generation — is baked into the measured compose; there is nothing to configure in the guest.
 
 3. Verify: the executor API answers on `EXTERNAL_PORT` and SSH banners on `SSH_PORT` within ~3–5 minutes of boot. `./lium-cvm.sh list` shows the VM.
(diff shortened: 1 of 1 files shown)


neurons/executor/dstacktee/docs/host-setup.md
```diff
@@ -120,17 +123,76 @@ Two containers: `aesmd` (SGX architectural enclaves, host network) and `gramine-
...
+After=network-online.target docker.service
+Wants=network-online.target
+
+[Service]
+Type=simple
+WorkingDirectory=/opt/lium-io/neurons/executor/dstacktee
+ExecStart=/opt/lium-io/neurons/executor/dstacktee/lium-cvm.sh run %i
+# --timeout is not optional: `stop` waits 30 s by default, then returns success and leaves the
+# guest running, and systemd SIGTERMs QEMU — a hard power-off of a guest that is still writing.
+ExecStop=/opt/lium-io/neurons/executor/dstacktee/lium-cvm.sh stop %i --timeout 840
```
taiberium, line 143: **Critical:** `ExecStop=… stop %i --timeout 840` cannot prevent the hard power-off it claims to. `shutdown_instance` (`scripts/dstack.py:754`) returns success on every path, including a failed vsock RPC, so ExecStop exits 0 within about 10 s and systemd then SIGTERMs a still-running QEMU — exactly the wedged-guest-agent case this section exists for.

neurons/executor/dstacktee/docs/host-setup.md
```diff
@@ -120,17 +123,76 @@ Two containers: `aesmd` (SGX architectural enclaves, host network) and `gramine-
...
+WantedBy=multi-user.target
+```
+
+Point both paths at the checkout you deploy from — the unit must run the same tree §7 upgrades,
+or `systemctl start` will quietly boot the old one. Stop the foreground `run` from §5 first: a
+second QEMU on the same GPUs dies with `vfio ... Device or resource busy` and the unit then
+retries every 60 s.
+
+```bash
+sudo ./lium-cvm.sh stop my-executor        # the §5 foreground run, if it is still up
```
taiberium, line 164: **Critical:** this block tells a provider to `lium-cvm.sh stop my-executor` before the "drain rentals before any planned restart" warning that follows it, so anyone adopting this on a live host kills the rentals they were told to drain. The command also omits `--timeout`, so on a multi-TB guest it returns success at 30 s and `systemctl enable --now` starts a second QEMU into `vfio … busy`.

neurons/executor/dstacktee/docs/host-setup.md
```diff
@@ -120,17 +123,76 @@ Two containers: `aesmd` (SGX architectural enclaves, host network) and `gramine-
...
+**Pin whatever package owns the QEMU binary `client.conf` names.** RTMR0 only reproduces under the exact QEMU the verifier's oracle was built from (§3). Following §3 you build it into `/opt/qemu-dstack`, outside apt's reach, and there is nothing to hold. If your `client.conf` instead points at a distro binary, `apt-mark hold` that package — an ordinary upgrade would silently change the measurement and the guest would come back unable to open its data disk.
+
+## 7. Upgrades
 
 Per release:
 
 ```bash
-sudo ./lium-cvm.sh stop my-executor
-git pull                                   # the release tag
+sudo systemctl stop lium-cvm@my-executor   # or: sudo ./lium-cvm.sh stop my-executor
```
taiberium, line 189: **NIT:** the fallback `./lium-cvm.sh stop` can return success while the VM still runs, and the next line is `rm -rf run/vms/<name>`.

neurons/executor/dstacktee/docs/host-setup.md
```diff
@@ -120,17 +123,76 @@ Two containers: `aesmd` (SGX architectural enclaves, host network) and `gramine-
...
+ExecStop=/opt/lium-io/neurons/executor/dstacktee/lium-cvm.sh stop %i --timeout 840
+# A multi-TB TDX guest does not finish shutting down in the default 90 s; keep this above the
+# ExecStop timeout so systemd's SIGKILL is never what ends the guest.
+TimeoutStopSec=15min
+Restart=on-failure
+# GPU passthrough is still being torn down when the unit restarts. Restarting immediately
+# fails with "vfio ... Device or resource busy" and burns the restart budget.
+RestartSec=60
+StandardOutput=journal
+StandardError=journal
```
taiberium, line 152: **NIT:** `Restart=on-failure` makes the unit itself an unattended restarter, which the next paragraph tells providers to eliminate. Say why a crash restart is the exception. `After=docker.service` at `:132` also only orders and never pulls docker in — add `Wants=docker.service`.

taiberium (dismissed):
Findings inline.

neurons/executor/dstacktee/docs/host-setup.md
surcyf123, line 143: Rustam's point still holds at this head: `shutdown_instance` (`scripts/dstack.py:754-818`) exits 0 on a failed RPC and on a timed-out wait, so this ExecStop succeeds and systemd SIGTERMs QEMU anyway. Fix in code: exit non-zero when the guest did not stop; until then say here that exit 0 does not prove the guest is down.

neurons/executor/dstacktee/docs/host-setup.md
surcyf123, line 164: Rustam's point still holds: this stops live rentals before the drain warning below, and without `--timeout` it returns success at 30 s so `enable --now` starts a second QEMU into `vfio ... busy`. Move the drain paragraph above this block and use `stop my-executor --timeout 840`.

neurons/executor/dstacktee/docs/host-setup.md
```diff
@@ -120,17 +123,76 @@ Two containers: `aesmd` (SGX architectural enclaves, host network) and `gramine-
...
+Type=simple
+WorkingDirectory=/opt/lium-io/neurons/executor/dstacktee
+ExecStart=/opt/lium-io/neurons/executor/dstacktee/lium-cvm.sh run %i
+# --timeout is not optional: `stop` waits 30 s by default, then returns success and leaves the
+# guest running, and systemd SIGTERMs QEMU — a hard power-off of a guest that is still writing.
+ExecStop=/opt/lium-io/neurons/executor/dstacktee/lium-cvm.sh stop %i --timeout 840
+# A multi-TB TDX guest does not finish shutting down in the default 90 s; keep this above the
+# ExecStop timeout so systemd's SIGKILL is never what ends the guest.
+TimeoutStopSec=15min
+Restart=on-failure
```
surcyf123, line 147: Rustam's point still holds: `Restart=on-failure` is an unattended restarter in a section that forbids them. Say in one sentence why a crash restart is the exception, and add `Wants=docker.service` next to `After=` if docker must be up.

neurons/executor/dstacktee/docs/host-setup.md
surcyf123, line 189: Rustam's point still holds: the fallback `stop` can return success while the VM runs, and `rm -rf run/vms/my-executor` is four lines down. Add `--timeout 840` to the fallback and a `./lium-cvm.sh list` check before the `rm -rf`.

surcyf123 (changes requested):
Not approving yet: this needs a rebase on main first, and the systemd unit it tells providers to install cannot deliver what its own comments promise. Rustam's four inline findings from 7 Sep are all still true at this head (the branch has not moved since 4 Sep).

1. **[high]** `neurons/executor/dstacktee/docs/host-setup.md:143`: Rustam's point at :143 still holds at this head. `shutdown_instance` (`scripts/dstack.py:754-818`) returns None on every path: a failed vsock RPC returns after about 10 s, and a guest still running after `--timeout 840` only logs a warning. Both exit 0, so `ExecStop` succeeds and systemd SIGTERMs QEMU anyway, which is the hard power-off the comment says this line prevents. Fix: make `dstack.py stop` exit non-zero when the guest did not stop (`sys.exit(1)` after the failed RPC and after the timeout warning) in a code PR, and until that lands say here that an `ExecStop` exit 0 does not prove the guest is down.
2. **[medium]** `neurons/executor/dstacktee/docs/host-setup.md:164`: Rustam's point still holds. This block stops the running CVM before the drain warning that follows it, and without `--timeout` the stop returns success after 30 s while a large guest is still shutting down, so `systemctl enable --now` starts a second QEMU into `vfio ... busy`. Fix: move the drain paragraph above this block and use `sudo ./lium-cvm.sh stop my-executor --timeout 840`.
3. **[low]** `neurons/executor/dstacktee/docs/host-setup.md:147`: Rustam's point still holds. `Restart=on-failure` is an unattended restarter inside a section headed "Nothing may restart this unit behind your back". Add one sentence saying why a crash restart is the exception (the guest is already down, so a restart loses nothing more). `After=docker.service` only orders; add `Wants=docker.service` if docker must be up.
4. **[low]** `neurons/executor/dstacktee/docs/host-setup.md:189`: Rustam's point still holds. The fallback `./lium-cvm.sh stop` can return success while the VM runs, and the next line is `rm -rf run/vms/my-executor`. Fix: add `--timeout 840` to the fallback and a `./lium-cvm.sh list` check before the `rm -rf`.
5. **[low]** `neurons/executor/dstacktee/docs/host-setup.md:140`: the body says nothing here has executed. Under systemd QEMU's `-chardev stdio` gets stdin from `/dev/null`. Start this unit once on a real host before providers do and say in the PR what happened; if QEMU needs a terminal, add `StandardInput=tty` or move the console to a `-serial file:` chardev.

PR #1265, opened 2026-09-02
DAH-2780: stop docker from restarting encrypted pods behind the volume mount

arhangel66:
### Task Link

---

Part 2 of 3 for DAH-2780. Depends on nothing; safe to merge on its own.

## Why

When a host or a CVM guest restarts, dockerd restores every container carrying `unless-stopped` before anything else runs. For an encrypted pod that restore is wrong by construction: the gocryptfs mount is made by a `docker exec` *inside* the container and dies with it, and nothing on the docker-side restore path remounts it. The container comes back with the renter's path as an ordinary directory, and everything they write lands unencrypted on the miner's disk.

## What changes

**1. No `restart_policy` for encrypted pods** (`docker_service.py:1036`). The policy was unconditional; it is now `None` when the run spec carries an encrypted local volume. `_restart_policy()` in the SDK already maps `None` to "no policy" (`rental_docker_sdk.py:856-859`), but the dataclass default is `"unless-stopped"` (`:111`), so the call site passes `None` explicitly rather than omitting the argument.

What this costs, stated plainly: `unless-stopped` restarts on *any* exit, not only at daemon boot, so an OOM-killed or crashed encrypted pod now stays exited and reports `POD_NOT_RUNNING` until the pod is recreated. On a crash the restart it loses never worked anyway — it produced the plaintext window above. **After a host reboot it did work, and this PR takes it away:** DAH-2306 recovery is entered only when `container_error` (docker's `State.Error`) carries the stale-vloopback signature, and docker fills that field only when it attempted the restart itself. With no policy the container exits with an empty `State.Error`, `recover_pod_after_stale_vloopback_mount` returns at its first line, and the encrypted pod stays down — and the miner penalised — for the rest of the rental. Closing that needs a second recovery signal that does not come from `State.Error` (container exited before the current `boot_id`); it is not in this PR, so this is a leak-versus-downtime trade to sign off on, not a free win.

Fillers are unaffected in both directions: they never encrypt (`_should_encrypt_local_volume` excludes `WorkloadKind.FILLER`) and they keep their policy.

**2. Refuse a recreate that would mount an encrypted volume as plaintext.** The payload decides encryption from `is_sysbox`, which the backend reads off executor specs — and specs freeze whenever a validation cycle scores zero. A pod created while the machine was known to be sysbox can therefore be recreated with `is_sysbox=False`, and the ciphertext tree gets mounted straight at the renter's path: gocryptfs internals in view, plaintext written next to them.

The wire carries no field for the pod's established encryption state, and the backend column that does (`pod.volume_encryption_status`) is nulled on exactly this path, so the truth has to come from the machine. `_local_volume_encryption_state` already derives it from `docker inspect` mounts and is used by the recovery path; the new `_assert_no_encryption_downgrade` asks the container that is about to be replaced, **before** the cleanup removes it, and refuses on a mismatch. A first create has no such container and no volume to lose — it returns before the inspect, so the deploy path pays nothing on a fresh rental.

The same answer is carried down to the image-label check, which is the second way a recreate ends up plain: an edit that swaps in an image without `lium.volume_encryption.enable=1` used to fall back to a plain mount — of a volume that already holds gocryptfs ciphertext — and by then the container that could have proved it is gone. That fallback now refuses too.

Note the deliberate direction: with `ENABLE_VOLUME_ENCRYPTION` globally off, a recreate of an already-encrypted pod now fails instead of silently exposing it. Failing is the safe side of that trade. The refusal message names every input that decided it — `workload_kind`, `is_sysbox`, `enable_volume_encryption` and the global setting — so an operator can tell frozen specs from the kill switch.

## Tests

`tests/test_docker_service.py`:
- run spec drops the policy for an encrypted volume and keeps `unless-stopped` for a plain one;
- an encrypted old container plus a plain-computing payload raises;
- **the refusal lands before `clean_existing_containers` is awaited** — the old container is the only witness, so a recreate that removed it first would leave the operator with a deleted pod and no evidence;
- an unlabelled image on an encrypted volume fails at `encrypted_volume_image_inspect` instead of falling back to plain;
- a first create (no `local_volume`) never even runs the inspect;
- a recreate whose container is already gone does **not** raise. This one matters: `enable_volume_encryption` defaults to `True` for every rental while `is_sysbox` is `False` on most of the fleet, so a guard written as a naive payload-vs-payload comparison would have refused every ordinary rental.

`pdm run pytest tests/` — 1945 passed, 15 skipped. `tests/test_sshd_bootstrap_script.py` fails 3 tests on this machine both with and without these changes (pre-existing, untouched here).

## Not in this PR

The migration that strips `unless-stopped` from already-running encrypted pods; there is no restart-policy mutation path in the validator today. So this protects new pods only, and an existing encrypted pod keeps both halves of the incident mechanism until its rental churns. The recovery matcher and `RECOVER_ENCRYPTED_VOLUMES` are untouched.

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -1052,8 +1052,10 @@ class DockerService:
             # DAH-2780: an encrypted pod must never be restarted by docker. The gocryptfs mount
             # lives inside the container and dies with it, and nothing remounts it on a docker-side
             # restore -- so the container comes back with the plaintext path as an ordinary
-            # directory and the renter's writes land unencrypted on the miner's disk. Recovery goes
-            # through the validator, which mounts before it starts.
+            # directory and the renter's writes land unencrypted on the miner's disk. The cost is
+            # real: DAH-2306 recovery only fires on the State.Error docker writes after a restart it
+            # attempted itself, so an encrypted pod that exits now stays down until it is recreated
+            # -- a host reboot included.
             restart_policy=None if encrypted_local_volume else "unless-stopped",
             runtime="sysbox-runc" if payload.is_sysbox else None,
             cap_add=self._capabilities_for(devices),
@@ -4486,7 +4488,7 @@ class DockerService:
 
                 # Must run before the cleanup below removes the container it inspects.
                 current_step = "volume_encryption_precheck"
-                await self._assert_no_encryption_downgrade(
+                existing_volume_encrypted = await self._assert_no_encryption_downgrade(
                     ssh_client=ssh_client,
                     payload=payload,
                     local_volume=local_volume,
@@ -4607,6 +4609,17 @@ class DockerService:
                         ssh_client,
                         payload.docker_image,
                     ):
+                        if existing_volume_encrypted:
+                            # DAH-2780: falling back to plain here would mount the ciphertext of a
+                            # volume the renter already has data in straight at their path. An edit
+                            # that swaps in an unlabelled image is the one way to reach it, and by
+                            # now the container that proved the volume encrypted is gone -- so the
+                            # answer taken before the cleanup is the only one left.
+                            raise RentalDockerOperationError(
+                                f"{container_name} runs an encrypted volume but "
+                                f"{payload.docker_image} carries no {_ENCRYPTED_VOLUME_IMAGE_LABEL} "
+                                "label; refusing rather than mounting it as plain"
+                            )
                         use_encrypted_volume = False
                         volume_encryption_status = VolumeEncryptionStatus.UNSUPPORTED_IMAGE
                         await self.stream_log(
@@ -5725,21 +5738,26 @@ class DockerService:
         local_volume: str | None,
         container_name: str,
         log_extra: dict[str, Any],
-    ) -> None:
-        # a recreate that would mount an already-encrypted rental volume as plaintext
+    ) -> bool:
+        # whether the pod being replaced already holds an encrypted volume; raises when this
+        # request would mount that volume as plaintext
         #
         # DAH-2780: the payload decides encryption from `is_sysbox`, which the backend reads off
         # executor specs -- and specs freeze whenever a validation cycle scores zero. A pod created
         # while the machine was known to be sysbox can therefore be recreated with is_sysbox=False,
         # and the ciphertext tree would be mounted straight at the renter's path: they would see
         # gocryptfs internals and write plaintext next to them. The container being replaced is the
-        # only place that knows the truth, so ask it before it is removed. A first create has no
-        # such container, the inspect answers UNKNOWN, and nothing here fires.
+        # only place that knows the truth, so ask it before it is removed, and carry the answer to
+        # the image-label check further down, which is the other way this deploy can end up plain.
         #
         # A rental whose volume this validator is about to create has no history to lose, so it
         # skips the inspect rather than paying for it on every deploy.
         if not local_volume:
-            return
+            return False
+
+        state = await self._local_volume_encryption_state(ssh_client, container_name)
+        if state is not _VolumeEncryptionState.ENCRYPTED:
+            return False
 
         if _should_encrypt_local_volume(
             local_volume,
@@ -5747,11 +5765,7 @@ class DockerService:
             payload.is_sysbox,
             payload.enable_volume_encryption,
         ):
-            return
-
-        state = await self._local_volume_encryption_state(ssh_client, container_name)
-        if state is not _VolumeEncryptionState.ENCRYPTED:
-            return
+            return True
 
         logger.error(
             _m(
@@ -5761,12 +5775,17 @@ class DockerService:
                     "container_name": container_name,
                     "is_sysbox": payload.is_sysbox,
                     "enable_volume_encryption": payload.enable_volume_encryption,
+                    "volume_encryption_enabled_setting": settings.ENABLE_VOLUME_ENCRYPTION,
                 }),
             )
         )
+        # name both causes -- a payload that lost `is_sysbox` to frozen specs, and the global kill
+        # switch being off. They need opposite answers, and this line is all the operator gets.
         raise RentalDockerOperationError(
             f"{container_name} runs an encrypted volume but this request would mount it as plain "
-            f"(is_sysbox={payload.is_sysbox}, enable_volume_encryption={payload.enable_volume_encryption}); "
+            f"(workload_kind={payload.workload_kind.value}, is_sysbox={payload.is_sysbox}, "
+            f"enable_volume_encryption={payload.enable_volume_encryption}, "
+            f"ENABLE_VOLUME_ENCRYPTION={settings.ENABLE_VOLUME_ENCRYPTION}); "
             "refusing rather than exposing the renter's data"
         )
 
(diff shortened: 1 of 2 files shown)


neurons/validators/src/services/docker_service.py
```diff
@@ -5702,6 +5730,65 @@ async def _local_volume_encryption_state(
...
+        # only place that knows the truth, so ask it before it is removed, and carry the answer to
+        # the image-label check further down, which is the other way this deploy can end up plain.
+        #
+        # A rental whose volume this validator is about to create has no history to lose, so it
+        # skips the inspect rather than paying for it on every deploy.
+        if not local_volume:
+            return False
+
+        state = await self._local_volume_encryption_state(ssh_client, container_name)
+        if state is not _VolumeEncryptionState.ENCRYPTED:
```
taiberium, line 5759: **NIT:** this treats `_VolumeEncryptionState.UNKNOWN` — what any failed `docker inspect` yields — as "no risk" and lets the plaintext recreate through, while `recover_pod_after_stale_vloopback_mount:5593` treats the same value as a stop. A docker hiccup during a recreate then mounts the ciphertext plain at the renter's path. "No such object" is distinguishable from a failed inspect.

neurons/validators/src/services/docker_service.py
```diff
@@ -5702,6 +5730,65 @@ async def _local_volume_encryption_state(
             return _VolumeEncryptionState.ENCRYPTED
         return _VolumeEncryptionState.PLAIN
 
+    async def _assert_no_encryption_downgrade(
```
taiberium, line 5733: **NIT:** `_assert_no_encryption_downgrade` also returns the encryption state the caller depends on, which the `assert_` name does not promise. Rename to say both jobs, or split them.

neurons/validators/src/services/docker_service.py
```diff
@@ -1049,7 +1049,14 @@ def _build_rental_container_run_spec(
...
             volumes=tuple(volumes),
-            restart_policy="unless-stopped",
+            # DAH-2780: an encrypted pod must never be restarted by docker. The gocryptfs mount
+            # lives inside the container and dies with it, and nothing remounts it on a docker-side
+            # restore -- so the container comes back with the plaintext path as an ordinary
+            # directory and the renter's writes land unencrypted on the miner's disk. The cost is
+            # real: DAH-2306 recovery only fires on the State.Error docker writes after a restart it
+            # attempted itself, so an encrypted pod that exits now stays down until it is recreated
+            # -- a host reboot included.
+            restart_policy=None if encrypted_local_volume else "unless-stopped",
```
taiberium, line 1059: **NIT:** dropping the policy also removes the host-reboot recovery that did work, and existing encrypted pods keep `unless-stopped` until their rental churns, so the exposure this closes stays open on the current fleet. `core/config.py:460` already records the end state as accepted policy, so this is a trade worth stating in full rather than a blocker.

taiberium (dismissed):
Findings inline.

neurons/validators/src/services/docker_service.py
surcyf123, line 5759: Rustam's point still holds at this head, and I would call it medium: `UNKNOWN` (any failed `docker inspect`, `:5727-5728`) falls through to `return False`, so a timed-out inspect during a recreate lets the cleanup remove the only witness and mounts the ciphertext plain. Fail closed here and treat only a missing container (`No such object` / `No such container` on stderr) as "gone"; `recover_pod_after_stale_vloopback_mount` at `:5621` already stops on the same value.

neurons/validators/tests/test_docker_service.py
```diff
@@ -5185,6 +5185,239 @@ def test_should_encrypt_local_volume_requires_local_sysbox_customer_volume(monke
...
+    # Assert
+    assert result.failure_step == "encrypted_volume_image_inspect"
+
+
+@pytest.mark.asyncio
+async def test_encryption_downgrade_allows_a_recreate_whose_container_is_gone(docker_service):
+    # Arrange: the payload asks for encryption on a machine that cannot do it -- the ordinary shape
+    # of most of the fleet -- and no container is left to contradict it
+    ssh_client = AsyncMock()
+    ssh_client.run = AsyncMock(return_value=_make_ssh_command_result(exit_status=1))
```
surcyf123, line 5370: This pins the fail-open path: `exit_status=1` with no stderr is what a docker hiccup looks like too. Make the fixture answer `No such object` on stderr, and add a sibling test where a bare `exit_status=1` raises.

neurons/validators/src/services/docker_service.py
surcyf123, line 5733: Rustam's naming point still holds: this both refuses and returns the state the image-label check depends on. Rename to say both (for example `_existing_volume_is_encrypted_or_refuse`) or split it.

surcyf123 (changes requested):
Not approving yet: this needs a rebase on main first, and the new precheck fails open on the one error path that matters. Rustam's three inline findings from 7 Sep are still true at this head (the branch has not moved since 4 Sep); I would raise the first of them from nit to medium.

1. **[medium]** `neurons/validators/src/services/docker_service.py:5759`: Rustam's point at :5759 still holds at this head. `_local_volume_encryption_state` returns `UNKNOWN` for every non-zero `docker inspect` (`:5727-5728`), and this line treats `UNKNOWN` like `PLAIN`: a timed-out or failing inspect during a recreate lets the request through, the cleanup removes the only witness, and the ciphertext is mounted plain at the renter's path. `recover_pod_after_stale_vloopback_mount` (`:5621`) stops on the same value. Fix: fail closed, and treat only a missing container as "gone" (`docker inspect` says `No such object`; `docker container inspect` says `No such container`):
```python
if state is _VolumeEncryptionState.UNKNOWN:
    # a missing container is a first create or a pod already removed: nothing to protect.
    # Anything else is an inspect that did not answer, and the volume may be encrypted.
    if not any(p in (inspect_stderr or "") for p in ("No such object", _DOCKER_NO_SUCH_CONTAINER_PHRASE)):
        raise RentalDockerOperationError(
            f"{container_name}: could not read the volume encryption state; refusing to recreate blind"
        )
    return False
if state is _VolumeEncryptionState.PLAIN:
    return False
```
This needs `_local_volume_encryption_state` to expose the stderr (or a fourth state, `GONE`).
2. **[low]** `neurons/validators/tests/test_docker_service.py:5370`: this test pins the fail-open behaviour. `exit_status=1` with no stderr is indistinguishable from a docker hiccup, and the test says it must pass. Make the fixture answer `No such object` on stderr, and add a sibling where a bare `exit_status=1` raises.
3. **[low]** `neurons/validators/src/services/docker_service.py:5733`: Rustam's naming point still holds. `_assert_no_encryption_downgrade` also returns the state the image-label check depends on. Rename to say both jobs (for example `_existing_volume_is_encrypted_or_refuse`) or split it.
4. **[info]** `neurons/validators/src/services/docker_service.py:1059`: Rustam's point still holds and the body states the trade in full: a crashed or rebooted encrypted pod now stays down until recreated, and existing encrypted pods keep `unless-stopped`. On the second half, `docker update --restart=no <pod>` is a restart-policy mutation the validator could run on encrypted pods at its next cycle; the "Not in this PR" note says no such path exists, but that one-liner is it. Worth a follow-up ticket rather than this PR.

Checked: `_restart_policy` in the SDK maps `None` to no policy, the precheck runs before `clean_existing_containers` and the test asserts that order, and fillers are excluded by `_should_encrypt_local_volume`.

PR #1266, opened 2026-09-02
DAH-2780: gate the CVM boot on sysbox actually serving, and stop rentals before the swap

arhangel66:
### Task Link

---

Part 3 of 3 for DAH-2780, and the expensive one. **This file is measured**, so every byte here changes `compose_hash` → RTMR3 → the guest's disk-encryption key: a provider taking this change must destroy and recreate the VM, and the renter's data on `hda.img` is gone. Read it line by line; merging it changes nothing on its own, and a mistake found later costs the fleet another recreate round.

Merge after #1264 (provider docs) and #1265 (validator).

## Why

On 26 August a provider restarted a CVM guest. Two things went wrong at once, both inside this script's blast radius:

1. dockerd, which systemd starts **before** `app-compose.service` (`Wants=docker.service` / `After=docker.service` in dstack's `basefiles/app-compose.service`), restored the customer's `pod_526856dd-b65d-407d-af89-7287321ac691` with its `unless-stopped` policy. Then this script stopped the sysbox daemons out from under it. The container kept handles on a runtime that no longer existed and died with exit 128.
2. sysbox came back only half-way, and nothing checked. The executor came up, reported the node healthy, and the rental stayed down for nine hours until the customer complained. $695.036512022889 in penalties, the whole miner fleet demoted to SPOT at 39.60133910497886% penalty coverage against a 20% threshold.

## What changes

**Stop rentals before the swap.** Only `pod_*` and `filler_*` are rentals (`services/const.py:238-239`, `docker_service.py:692-695`); `container_{hotkey}_*` and `health_check_*` are the validator's own throwaway probes and `executor-runner` is the executor. Stop grace is `-t 30`, matching the validator's `CONTAINER_STOP_GRACE_SECONDS` — killing GPU workloads fast wedges containerd/sysbox (DAH-2364), which is the last thing anyone wants immediately before a daemon swap. A container stopped here is not brought back by the `systemctl restart docker` below; the validator recreates it.

**Gate on the daemons actually serving.** None of the obvious signals is worth anything here, which is the whole lesson of the incident:
- the units are `Type=simple`, so `systemctl is-active` proves only that a process exists;
- sysbox-fs calls `SdNotify(Ready)` and logs `Ready` *before* `ipcService.Init()` starts listening;
- a socket file left behind by a dead daemon stays on disk, so `[ -S ... ]` passes while `connect()` gives ECONNREFUSED — that is the 26 August shape exactly.

`/proc/net/unix` answers the real question: flags `00010000` is `SO_ACCEPTCON`, a socket that has reached `listen()`, not merely `bind()`. The socket inode is also matched against the unit's `MainPID`, because the stop of the baked 0.6.7 daemons above is best-effort (`|| true`) — if it silently failed, the old GPU-incapable daemon would still be listening on the same path and a path-only gate would go green on it.

**On failure: restart with `--no-block`, then fail loudly and keep booting.** `--no-block` is not decoration: the daemon being restarted is the one that may be wedged in uninterruptible sleep, a blocking `systemctl restart` on it never returns, there is no `timeout` in this guest to bound it, and a boot script that never finishes leaves a box whose only door — the executor — never opens. Restarting by hand is what fixed the machine in one second on 26 August. If the second poll also fails, the script writes `/dstack/sysbox-not-ready` and prints a loud line, then lets the boot continue: a node that comes up visibly broken is worth more than one that never comes up. A healthy boot clears the marker, so the first reader ever added does not report a failure fixed weeks ago.

The loud line is readable. The guest writes with `StandardOutput=journal+console`, the console is the serial port, and `dstack.py:711-716` hands QEMU `-chardev stdio,id=ser0 -serial chardev:ser0` with QEMU in the foreground of `lium-cvm.sh run` — so it lands in the host journal for the CVM unit that #1264 asks providers to create. Surfacing the marker through executor specs would be a DAH-2514-class three-repo change and is deliberately not in this wave.

**Sourced, not executed.** `basefiles/app-compose.sh` runs this file as `source <(jq -r '.pre_launch_script' app-compose.json)` under bash. So `exit` here would kill the boot before the compose ever starts, and any `set -e` would change how the rest of `app-compose.sh` behaves. This script does neither, and unsets the names it defines. There is no deadline to fit inside: `app-compose.service` is `Type=oneshot` with no `TimeoutStartSec`, and systemd disables the start timeout by default for oneshot units; the host does not supervise boot progress either (`dstack.py` uses vsock only for the shutdown RPC).

One comment added on the existing `sed` stub: it exits 0 when it replaces nothing, so any future installer-digest bump must re-verify the anchor or the baked daemons would silently survive.

## Verification

We own no TDX hardware and there is no staging CVM, so the gate was staged off-guest with the real code and fixture `/proc` trees: `untracked/scripts/dah2780_gate_harness.sh`. 22 checks over 10 scenarios — healthy boot; the 26 August shape (socket file present, nobody listening) recovering after the restart; never-ready writing the marker and continuing; a stale marker cleared by a healthy boot; only sysbox-fs serving; a foreign daemon holding the socket; both sockets bound but not listening; dockerd not answering; the script sourced twice in a row; and a `systemctl stop` of the baked 0.6.7 daemons that silently failed, leaving them listening under the same unit name. Every scenario also asserts the parent shell reaches the line after the `source` and that no shell options or `lium_*` names leak into it.

The checks were then mutation-tested, because a test that certifies the wrong behaviour is worse than none — the earlier draft of this change used the `container_` prefix and its test used the same wrong name, so it would have gone green. Nine mutations, all caught (`untracked/scripts/dah2780_gate_mutations.sh`): revision-1's `container_` selector; dropping the `SO_ACCEPTCON` flag check; dropping `--no-block`; dropping the `MainPID` owner check; trusting a daemon that survived the stop; polling only sysbox-fs; not clearing a stale marker; `exit` instead of continuing the boot; and `docker stop` without the 30 s grace.

What this does **not** prove: that the recreated guest attests, and that the script behaves identically inside a measured TDX guest. Only a canary on a free machine proves those.

`compose_hash` moves, as expected — computed against a placeholder runner digest, since `lium-cvm.sh:436-457` substitutes `EXECUTOR_RUNNER_IMAGE_DIGEST` into the measured compose before it is measured, so the value is only defined per (tree × runner digest):

```
runner digest : sha256:0000…0000  (placeholder)
old           : df7893451cd832b03eeffd9b890dd2d9272216d6e55d0448e3cd08a57d8fe1cf
new           : 9be75cb21373848af315ec1d1bcc9911ef5b4e3abcb30fdfd8e15de52c5ae93e
```

Re-render with the released digest at release time to get the value to whitelist (`untracked/scripts/dah2780_compose_hash.py`). `ENABLE_ATTESTATION_WHITELIST` is `"false"` in production, so old and new hashes coexist and machines can be recreated one at a time.

## Rollout

Dormant on merge. It reaches a machine only when a provider manually runs `git checkout <tag>` + `lium-cvm.sh new`, which destroys that guest's data disk — so this goes out one machine at a time, starting with a free one, and #1264 should be in providers' hands first.

PR #1151 touches this same file (sysbox 0.7.0) and is deliberately **not** in this wave: upstream 0.7.0 does not carry the sysbox-fs FUSE deadlock fix (nestybox/sysbox-fs PR#121), and nothing records whether that installer image was built with the patch. Shipping it later costs a second recreate round; that is the accepted price.

### DIFF
--- a/neurons/executor/dstacktee/app/pre_launch_script.sh
+++ b/neurons/executor/dstacktee/app/pre_launch_script.sh
@@ -1,4 +1,39 @@
 chmod +x /usr/bin/containerd-shim-runc-v2
+
+# This file is sourced by /bin/app-compose.sh (bash) before it starts the compose. Sourced means
+# `exit` here would kill the boot before the executor ever starts, and any `set -e` would change
+# how the rest of app-compose.sh behaves -- so this script never does either, and cleans up the
+# names it defines at the end.
+
+lium_not_ready_marker=/dstack/sysbox-not-ready
+lium_say() {
+    # both the guest journal and the serial console, which is the host's only window in
+    echo "lium: $*" >&2
+}
+
+# DAH-2780: docker.service is ordered before app-compose.service, so dockerd has already restored
+# every `unless-stopped` container by the time this runs -- including customer rentals. Pulling the
+# sysbox daemons out from under a live rental is what left a customer's container dead for nine
+# hours on 26 August: it keeps handles on the old runtime and dies with exit 128. Stop rentals
+# first: a container stopped here is not brought back by the docker restart below, and the
+# validator will not start it either -- its only auto-recovery covers the DAH-2306 stale-mount
+# signature -- so the pod stays down until the customer starts it and the node takes
+# POD_NOT_RUNNING. That is the same end state as letting the swap wedge the container, minus the
+# wedged containerd and the nine hours. The grace matches the validator's
+# CONTAINER_STOP_GRACE_SECONDS -- killing GPU workloads fast wedges containerd/sysbox (DAH-2364),
+# which is the worst thing that could happen immediately before a daemon swap. Only pod_/filler_
+# are rentals; container_*/health_check_* are the validator's own throwaway probes and
+# executor-runner is the executor itself.
+if ! lium_rentals=$(docker ps -q \
+    --filter status=running --filter status=restarting \
+    --filter name=^pod_ --filter name=^filler_ 2>/dev/null); then
+    lium_say "could not list containers -- dockerd not answering, rentals may survive the sysbox swap"
+elif [ -n "$lium_rentals" ]; then
+    lium_say "stopping rental containers before the sysbox swap"
+    docker stop -t 30 $lium_rentals >/dev/null 2>&1 || \
+        lium_say "some rental containers did not stop cleanly"
+fi
+
 # dstack-nvidia-0.5.11 bakes upstream sysbox CE 0.6.7, which rejects --gpus
 # (docker device requests) and so fails the SN51 sysbox/GPU compatibility
 # check. Stop the baked daemons and force-install the Datura sysbox build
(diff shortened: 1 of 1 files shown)


neurons/executor/dstacktee/app/pre_launch_script.sh
```diff
@@ -1,4 +1,39 @@
...
+# which is the worst thing that could happen immediately before a daemon swap. Only pod_/filler_
+# are rentals; container_*/health_check_* are the validator's own throwaway probes and
+# executor-runner is the executor itself.
+if ! lium_rentals=$(docker ps -q \
+    --filter status=running --filter status=restarting \
+    --filter name=^pod_ --filter name=^filler_ 2>/dev/null); then
+    lium_say "could not list containers -- dockerd not answering, rentals may survive the sysbox swap"
+elif [ -n "$lium_rentals" ]; then
+    lium_say "stopping rental containers before the sysbox swap"
+    docker stop -t 30 $lium_rentals >/dev/null 2>&1 || \
```
taiberium, line 33: **Critical:** on the healthy path this makes a guest reboot permanently kill every rental. Today `systemctl restart docker` at `:114` restores `unless-stopped` pods once the new sysbox serves (no `live-restore` anywhere in the dstacktee tree), but a container stopped with `docker stop` is never restored, and the only validator recovery is the DAH-2306 `State.Error` matcher (`rented_machine.py:296`), which cannot fire on a clean stop. Every reboot then costs POD_NOT_RUNNING plus penalties even when the swap works. Add `docker start $lium_rentals` after the gate and the docker restart.

neurons/executor/dstacktee/app/pre_launch_script.sh
```diff
@@ -9,10 +44,78 @@ chmod +x /usr/bin/containerd-shim-runc-v2
...
+# A boot that gets this far owns the verdict, so a marker from an earlier failure goes now.
+rm -f "$lium_not_ready_marker"
+if ! lium_await_sysbox; then
+    # --no-block is not optional: the daemon being restarted is the one that may be wedged in
+    # uninterruptible sleep, and a blocking `systemctl restart` on it never returns. There is no
+    # way to bound that from here, and a boot script that never finishes leaves a box whose only
+    # door -- the executor -- never opens. Restarting by hand is what fixed the machine in one
+    # second on 26 August.
+    lium_say "sysbox is not serving; restarting the daemons"
+    systemctl --no-block restart sysbox-mgr sysbox-fs 2>/dev/null || true
```
taiberium, line 104: **NIT:** if the installer `docker run` fails, the retry branch restarts the still-installed baked 0.6.7 units with fresh PIDs, so all three gate checks pass and the boot reports healthy on a GPU-incapable sysbox.

neurons/executor/dstacktee/app/pre_launch_script.sh
```diff
@@ -9,10 +44,78 @@ chmod +x /usr/bin/containerd-shim-runc-v2
...
+    if ! lium_await_sysbox; then
+        # Keep booting on purpose. The executor is the only way anyone reaches this guest, so a
+        # node that comes up visibly broken is worth more than one that never comes up. The line
+        # below lands on the serial console, i.e. in the host's journal for the CVM unit.
```
taiberium, line 108: **NIT:** `/dstack/sysbox-not-ready` is written and cleared but nothing in this wave reads it, and the console line already carries the same signal. The 30 s grace also applies to fillers, while the validator uses `FILLER_CONTAINER_STOP_GRACE_SECONDS = 15`.

taiberium (dismissed):
Findings inline.

neurons/executor/dstacktee/app/pre_launch_script.sh
surcyf123, line 33: Rustam's point still holds at this head: a `docker stop`ped container is not restored by the docker restart below, so every guest reboot takes every plain rental down. After `systemctl restart docker`, `docker start` the stopped rentals whose mounts do not include `/lium-cipher`; encrypted pods must stay down on purpose (#1265). The body says the validator recreates it; the comment at :18-21 says it will not, and the comment is right. Also: fillers use a 15 s grace in the validator (`FILLER_CONTAINER_STOP_GRACE_SECONDS`), not 30.

neurons/executor/dstacktee/app/pre_launch_script.sh
```diff
@@ -9,10 +44,78 @@ chmod +x /usr/bin/containerd-shim-runc-v2
...
+# The stop above is best-effort, so record whatever survived it: the gate below has no other way to
+# tell the daemon we are about to install from a baked 0.6.7 one that never died. Comparing the
+# unit's MainPID alone cannot -- the unit name is the same either way.
+lium_stale_mgr_pid=$(systemctl show -p MainPID --value sysbox-mgr 2>/dev/null)
+lium_stale_fs_pid=$(systemctl show -p MainPID --value sysbox-fs 2>/dev/null)
 # TODO: pin digest
+# Whenever this digest moves, re-check that the sed anchor below still matches the installer's
+# check_existing definition: sed exits 0 when it replaces nothing, so a renamed function would
+# silently leave the gate in place and the baked 0.6.7 daemons would survive.
 docker run --rm --privileged --pid=host --net=host -v /:/host dstacktee/dstack-sysbox-installer:1.0.0@sha256:2f5dbea99176f3ea0362b85346b31b1160bfb70c1d98d1c8d375d57782127dd1 bash -c "sed -i '/^check_existing() {/,/^}/c\\check_existing() { return 0; }' /usr/local/bin/install-sysbox-complete.sh && /usr/local/bin/install-sysbox-complete.sh"
```
surcyf123, line 56: Rustam's point at :104 still holds: this exit status is ignored, so if the installer fails the restart branch revives the baked 0.6.7 units with fresh PIDs and the gate goes green on a runtime that rejects `--gpus`. Capture `$?` here and on failure skip the restart, write the marker and print the loud line.

neurons/executor/dstacktee/app/pre_launch_script.sh
```diff
@@ -9,10 +44,78 @@ chmod +x /usr/bin/containerd-shim-runc-v2
...
+    # way to bound that from here, and a boot script that never finishes leaves a box whose only
+    # door -- the executor -- never opens. Restarting by hand is what fixed the machine in one
+    # second on 26 August.
+    lium_say "sysbox is not serving; restarting the daemons"
+    systemctl --no-block restart sysbox-mgr sysbox-fs 2>/dev/null || true
+    if ! lium_await_sysbox; then
+        # Keep booting on purpose. The executor is the only way anyone reaches this guest, so a
+        # node that comes up visibly broken is worth more than one that never comes up. The line
+        # below lands on the serial console, i.e. in the host's journal for the CVM unit.
+        date -u '+%Y-%m-%dT%H:%M:%SZ sysbox daemons never started serving' > "$lium_not_ready_marker" 2>/dev/null
```
surcyf123, line 109: Rustam's point still holds: nothing in this wave reads this marker and the console line carries the same signal. Name the reader that is coming, or drop the marker.

surcyf123 (changes requested):
Not approving yet: one high finding on the rental stop, and this is a measured executor boot file, so it needs Rustam's approval and a canary, not the loop's. Rustam's three inline findings from 7 Sep are all still true at this head (the branch has not moved since 4 Sep). The gate itself reads correctly: `/proc/net/unix` column 4 is Flags and `00010000` is `SO_ACCEPTCON`, column 7 is the inode and column 8 the path, and the MainPID plus survivor-pid comparison does what the comment says.

1. **[high]** `neurons/executor/dstacktee/app/pre_launch_script.sh:33`: Rustam's point at :33 still holds at this head. A container you `docker stop` is not restored by `systemctl restart docker`, and nothing else starts it, so every guest reboot now takes every plain rental down until the customer acts, even when the swap works. The body says "the validator recreates it"; the comment at :18-21 says it will not, and the comment is right. Fix: after `systemctl restart docker` at :114, `docker start` the stopped rentals whose mounts do not include `/lium-cipher`. Encrypted pods must stay down on purpose (#1265: a docker-side restore comes back with a plaintext path), so filter on the mount, not the name:
```bash
for c in $lium_rentals; do
    docker inspect --format '{{range .Mounts}}{{println .Destination}}{{end}}' "$c" 2>/dev/null | grep -qx /lium-cipher \
        || docker start "$c" >/dev/null 2>&1 || lium_say "could not restart $c"
done
```
Then fix the body to match the comment.
2. **[medium]** `neurons/executor/dstacktee/app/pre_launch_script.sh:56`: Rustam's point at :104 still holds. The installer's exit status is ignored; if this `docker run` fails, the restart branch brings the baked 0.6.7 units back with fresh PIDs, all three gate checks pass, and the boot reports healthy on a runtime that rejects `--gpus`. Fix: capture the status (`lium_installer_ok=$?`) and on failure skip the restart, write the marker and print the loud line.
3. **[low]** `neurons/executor/dstacktee/app/pre_launch_script.sh:33`: `-t 30` also applies to `filler_*`, where the validator uses `FILLER_CONTAINER_STOP_GRACE_SECONDS = 15` (`docker_service.py:156`). Stop the two groups with their own grace, or say why fillers get 30 s here.
4. **[low]** `neurons/executor/dstacktee/app/pre_launch_script.sh:109`: Rustam's point still holds. Nothing in this wave reads `/dstack/sysbox-not-ready`. Name the reader that is coming, or drop the marker and keep the console line.
5. **[low]** body: the 22-check harness and the 9 mutations live in `untracked/scripts/` and are not in the diff, so nothing in CI guards a measured file whose mistakes cost a fleet recreate round. Commit them (for example under `neurons/executor/tests/dstacktee/`) and run them in the executor test job.

Hold for Rustam (taiberium): measured CVM boot script on the executor verification path. A canary on a free machine is the bar, as the body says.

PR #1268, opened 2026-09-03
DAH-2774: VerifyX-only network, drop scrape speedtests

jam6099:
## Summary
- Replace the vendored VerifyX library on executor and validator with the DAH-2774 build.
- Use VerifyX as the only network speed source. Cloudflare download is what we store as VerifyX download; package download stays as integrity plus a speed floor in network stats.
- Remove the scrape speedtest chain (`speedtest-cli` / Cloudflare curl / netmeasure / speedcheck), scrape network EMAs, and related executor image deps.
- Keep `verifyx_network_validation` off for now. Slow or failed network does not reject VerifyX in this PR. Turn that on in a follow-up once the new library is proven in prod.

Companion backend PR: Datura-ai/lium-io-backend (same branch name). VerifyX probe source: Datura-ai/celium-gpu-verifier#14.

## Test plan
- [ ] Targeted validator tests: `test_verifyx_check`, `test_verifyx_validation_service`, `test_machine_spec_scrape_check`
- [ ] Confirm executor and validator `libverifyx.so` checksums match
- [ ] Deploy to a canary validator/executor pair and confirm VerifyX network fields populate (`verifyx_*` / `ema_verifyx_*`)
- [ ] Confirm scrape payloads no longer carry download/upload from the old tools
- [ ] Confirm machines still validate while `verifyx_network_validation` is false


arhangel66 (changes requested):
**neurons/validators/src/miner_jobs/machine_scrape.py:1421**

Dropping `network.upload_speed` / `download_speed` zeroes `provider_stats.py:105-130` in compute-app, which averages those raw keys with no feature flag. Please land lium-io-backend#1019 first and note that ordering in the description.

_Posted by Mikhail's review agent; the verdict stands until he reads the reply._

neurons/validators/src/services/verifyx_validation_service.py
```diff
@@ -366,18 +370,31 @@ def _verify_network_test(challenge_data: dict, response_data: dict) -> Tuple[dic
         errors.append(f"Integrity check failed for {download_result['pkg']}")
         success = False
 
-    download_speed = network_execution["download"]["speed_mbps"]
-    upload_speed = network_execution.get("speedtest", {}).get("upload_mbps")
+    speedtest = network_execution["speedtest"]
```
surcyf123, line 373: Missing `speedtest` or a `null` `download_mbps` raises here, the outer `except Exception` turns it into `challenge verification failed`, and the fatal check rejects the machine even with `verifyx_network_validation` off. Use `.get()` and return `success=False` with "Network performance data unavailable" so the flag decides.

neurons/validators/tests/test_machine_spec_scrape_check.py
```diff
@@ -24,6 +26,26 @@
 UNREADABLE_TOKEN = "gAAAAABnot-ours"
 
 
+def test_machine_scrape_has_no_speedtest_helpers():
```
surcyf123, line 29: This asserts five function names are absent from the source. Nothing realistic regresses it and it breaks on any unrelated helper with the same name. Suggest deleting it; the scrape test already pins `data_network == {}`.

neurons/validators/src/services/verifyx_validation_service.py
```diff
@@ -366,18 +370,31 @@ def _verify_network_test(challenge_data: dict, response_data: dict) -> Tuple[dic
...
+        success = False
 
     if download_speed < settings.verifyx.NETWORK_MIN_DOWNLOAD_SPEED_MBPS:
         errors.append(
-            f"Network download speed inadequate: {download_speed:.2f} Mbps achieved, {settings.verifyx.NETWORK_MIN_DOWNLOAD_SPEED_MBPS:.0f} Mbps required"
+            f"Cloudflare download speed inadequate: {download_speed:.2f} Mbps achieved, {settings.verifyx.NETWORK_MIN_DOWNLOAD_SPEED_MBPS:.0f} Mbps required"
+        )
+        success = False
+
+    if package_download_speed < settings.verifyx.NETWORK_MIN_DOWNLOAD_SPEED_MBPS:
```
surcyf123, line 388: Same floor for the package download as for Cloudflare. How large is the challenge package? A small transfer is mostly latency and will fail healthy links once the flag is on; consider a lower floor or a minimum size.

surcyf123 (changes requested):
Not approving yet: this needs a rebase on main first, and one hard failure path to fix.

1. **[high]** `neurons/validators/src/services/verifyx_validation_service.py:373`: `network_execution["speedtest"]` and `speedtest["download_mbps"]` replaced the old `.get("speedtest", {})`. If the probe returns no `speedtest` block, or `download_mbps` is `null` (Cloudflare unreachable from the executor), this raises KeyError or TypeError inside `_perform_verification_checks`; `validate_verifyx_and_process_job` catches it as `challenge verification failed` and the fatal VerifyX check rejects the machine. That happens with `verifyx_network_validation` off, which the body says must not reject. One Cloudflare outage then fails the fleet. Fix:
```python
speedtest = network_execution.get("speedtest") or {}
upload_speed = speedtest.get("upload_mbps")
download_speed = speedtest.get("download_mbps")
package_download_speed = download_result.get("speed_mbps")
if not all(isinstance(v, (int, float)) and v > 0 for v in (upload_speed, download_speed, package_download_speed)):
    errors.append("Network performance data unavailable")
    return {
        "success": False,
        "download_speed": download_speed,
        "upload_speed": upload_speed,
        "package_download_speed": package_download_speed,
        "execution_time_ms": network_execution.get("execution_time_ms"),
    }, errors
```
plus a test with `speedtest` missing that asserts `_perform_verification_checks` still returns `success=True` while the flag is off.
2. Mikhail's point about `provider_stats.py` in compute-app still holds at this head: `machine_scrape.py` still drops `network.download_speed` / `upload_speed`, so lium-io-backend#1019 has to land first and the ordering belongs in the description.
3. **[low]** `neurons/validators/src/services/task/checks/network_ema.py`: the import was removed but the module and `rented_data.network_ema` stay in the tree with no reader. Fix: delete the module and its test, or say why it stays.
4. **[low]** `neurons/validators/tests/test_machine_spec_scrape_check.py:29`: `test_machine_scrape_has_no_speedtest_helpers` parses the source and asserts five function names are absent. No realistic regression trips it other than re-adding a function with the same name, and it breaks if someone adds a `cloudflare_speed` helper for another purpose. Fix: delete it; `data_network == {}` is already covered by the scrape test.
5. **[low]** `neurons/validators/src/services/verifyx_validation_service.py:388`: the same `NETWORK_MIN_DOWNLOAD_SPEED_MBPS` floor is applied to the package download. What is the package size in the challenge? A small package measures mostly latency, so this floor will trip on healthy links once the flag is turned on. Consider a separate, lower floor or a minimum size before the speed reading counts.

This is a validator verification path, so once the rebase and item 1 are done it needs Rustam's approval, not the loop's.

neurons/validators/src/services/verifyx_validation_service.py
surcyf123: Done in lium-io#1419 (takes over this PR; jam6099's commits kept, rebased on main) at 5a899f20: every speed field is read with `.get()`, a missing/null/zero/non-numeric reading returns `success=False` with "Network performance data unavailable" and `verifyx_network_validation` decides; tests cover the missing block, each bad value, and the whole verification with the flag off (passes) and on (rejects).

neurons/validators/tests/test_machine_spec_scrape_check.py
surcyf123: Done in lium-io#1419 at 1a4041bc: the AST test is deleted; `"network" not in updated_state.specs` in the scrape test stays.

neurons/validators/src/services/verifyx_validation_service.py
surcyf123: Answered in lium-io#1419 at d9661d18. The package is not small — every catalog object is 251–478 MB (celium-gpu-verifier `pkg_verify/verified_pkg_list.json` size_stats), so latency does not dominate; but its speed follows the CDN edge (229–3410 Mbps on one host in Muhammad's 15 runs). The floor is now its own setting `VERIFYX_NETWORK_MIN_PACKAGE_DOWNLOAD_SPEED_MBPS`, default unchanged at 50 so the number stays the author's; lowering it is a config change.

surcyf123:
Taken over in lium-io#1419 per Muhammad (21 Sep 05:28Z): your two commits rebased on main (one conflict — the log call moved into `evaluate_verifyx_capture`, which DAH-2834 introduced), the three inline threads answered above, and `libverifyx.so` rebuilt from celium-gpu-verifier#25 (the takeover of #14). Not closed here: Mikhail's point about `provider_stats` — the backend moved to lium-platform and lium-io-backend#1019 was closed unmerged; on lium-platform main `provider_stats._get_average_network_speed` still averages the raw `specs.network.download_speed/upload_speed`, which this change stops sending (the platform preserves the last value for known nodes; `effective_*` is unaffected because the VerifyX EMA comes first). That needs a lium-platform PR on DAH-2774 — named in #1419's Risks. This PR can be closed once #1419 is reviewed; the loop does not close a human's PR.

PR #1269, opened 2026-09-03
DAH-2847 /temp-stg PR comment trigger

taiberium:
## What

Part of DAH-2847 (a temporary dev copy of the platform per pull request; the work lives in `Datura-ai/lium-infra`, see its `TEMP_DEV.md`).

**Superseded — do not merge as it stands.** This PR adds a workflow that hands a `/temp-dev` comment to lium-infra. That shape is out: the job needed a token, and a secret in a public repository can be read by a workflow on any branch of it. lium-infra now finds those comments itself (`temp-dev-request.yml` in Datura-ai/lium-infra#31), so this repository holds no token and needs no workflow at all.

Nothing here is needed any more. Waiting on Rustam's word to close it.


.github/workflows/temp_dev.yml
```diff
@@ -0,0 +1,22 @@
...
+# branch. The work happens in Datura-ai/lium-infra; see its TEMP_DEV.md.
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
```
surcyf123: Nit: `pull-requests: write` and `issues: write` on the default `GITHUB_TOKEN` are unused — every write the reusable workflow makes (reactions, comments, the dispatch) goes through the PAT it receives as `github-token`. `contents: read` alone is enough here, and a smaller default token is what a public repository's `issue_comment` workflow wants.

.github/workflows/temp_dev.yml
```diff
@@ -0,0 +1,22 @@
+name: Temp staging
```
surcyf123: Nit: the PR title still says `/temp-stg` and this `name:` says "Temp staging"; the command is `/temp-dev` and the hosts are `*.dev.lium.io`. Worth renaming both before merge so the Actions tab and the ticket agree.

surcyf123 (commented):
Read 88cb913 together with the workflow it calls (lium-infra#31 `temp-dev-request.yml` at 7b55aad). The 22 lines themselves are right: `issue_comment: created`, PR-only, bot-marker guard, the reusable workflow does the access check (write/maintain/admin on the commenter) and refuses fork PRs before it writes anything, so a non-member commenting `/temp-dev` on a lium-io PR gets a 👎 and nothing else.

**Comment rather than approve, for two reasons outside this diff:**
1. It cannot run yet — `temp-dev-request.yml@main` exists only once #31 merges, and `TEMP_DEV_GITHUB_TOKEN` is not among the org secrets today (`gh api orgs/Datura-ai/actions/secrets`: `DEV_ENV_SALT` and the CI webhook only).
2. lium-io is the one **public** repository of the four. As #31 stands, the result comment it posts back here carries the copy's admin sign-in password (`temp-dev.yml:313` there, reviewed on #31). Until that comment stops carrying the password, a `/temp-dev` from this repository publishes it. Happy to approve once #31 is updated.

Two nits inline. Not run: anything — the caller has no side of its own to exercise.

.github/workflows/temp_dev.yml
taiberium: Both gone with the job: this repository no longer holds a token or a workflow. lium-infra sweeps the comments itself (#31), because a secret in a public repository can be read by a workflow on any branch of it. This PR is superseded — the description says so, and closing it is Rustam's call.

.github/workflows/temp_dev.yml
taiberium: Same answer: the workflow goes away entirely, so the name and the command no longer disagree. lium-infra owns the command now.

PR #1270, opened 2026-09-03
DAH-2748, hide a node the validator cannot reach on one availability error

taiberium:
### Task Link

[DAH-2748](https://app.notion.com/p/3c3b8bfdbde9815eb424e1ed25f9799e)

---

## Problem

When the validator cannot open SSH to a node, the cycle falls into the generic handler, scores 0 and logs "Pipeline validation error" with no reason code.

The backend then cannot tell two different things apart: the node failed a check, and nobody could reach the node. The second one is what breaks a customer. The node keeps its positive state, stays on the market, and the next renter gets a machine the platform cannot talk to. That is the 2026-08-21 incident in the task.

## Solution

A validation error that means "someone could not reach something" is its own class. One such error is enough to take the node off the market for that cycle. The next cycle that reaches the node clears it.

- The class is a property of the event, so a new reachability check joins it by naming who could not reach what. Today it has one member: the validator cannot open SSH to the node. Planned members: the container cannot reach Docker Hub or Hugging Face, the validator cannot reach the container.
- The score is still 0 for that cycle. No counters and no thresholds.
- The cycle sends the errors it raised to the backend, each one a full object: the code, who tried, what could not be reached, what we saw. An empty list puts the node back on the market. `None` means the cycle never got to the check, so the stored errors stay as they are and a hidden node is not re-listed by a cycle that tested nothing.
- When more than half of one cycle fails at the connect, the cycle sends `None` for every result and logs an error. Our own outage must not empty the market.

## Changes

- `services/task/availability.py` (new): the availability error class, the names for who tried and what could not be reached, and the SSH event with its impact and its fix for the provider.
- `services/task/models.py`: `AVAILABILITY_CATEGORY`, `ValidationEvent.is_availability_error`, `JobResult.availability_errors`.
- `services/task/service.py`: a transport failure raised while the SSH connection is opening builds the availability event. Every other failure keeps today's log. A cycle that opened the shell reports an empty list, and the whole event list of a complete cycle is read, so a check that fails early still hides the node.
- `miner_service.py`, `clients/compute_client.py`, `protocol/vc_protocol/validator_requests.py`: the errors travel with the cycle result.
- `core/validator.py`: the guard that silences a cycle we cannot trust, before the results are published.
- `tests/test_availability_errors.py`: 13 tests.

## Deployment Steps

Use GitHub Action. Deploy together with lium-platform#25, which stores the code and hides the node.

## Review Request

- [x] Just code review

## Other PRs

- lium-platform https://github.com/Datura-ai/lium-platform/pull/25 — the backend column, the market filter and what the provider sees.

## Risks

- Only a failure while opening the SSH connection counts. The attestation verifier raises the same error types over HTTP, so the flag is true for that one step and nothing else.
- A cycle that fails before the connect sends `None`. A node an earlier cycle hid stays hidden until a cycle reaches it, which is the safe direction.
- A node that answers but fails an ordinary check is not an availability error and keeps today's behaviour.
- The text a node's sshd sends us is kept to 500 characters, because the provider portal renders it.

### Test Case 1

**Actions**

`pytest tests/test_availability_errors.py`

**Expected Output**

13 pass: the SSH failure raises the code, a pipeline failure after the shell opened reports an empty list, a failure before the connect reports `None`, a complete cycle collects the codes of its own events, duplicates are dropped, and the event carries the reason code, the impact and the fix, one bad node among many is still hidden, a cycle that reached almost nothing reports nothing, and the peer text is capped.

## Checklist

- [x] Risks are documented
- [x] Tests are described


arhangel66 (changes requested):
as I get, we are waiting for update here, right?

surcyf123 (commented):
Read the full diff and the tests (CI is green here). Comments only. One thing worth a decision, one small one; the rest holds.

**1. One bad cycle on the validator's side hides every node it checked** (`neurons/validators/src/services/task/service.py:189-215`). `_is_ssh_transport_failure` is right about the kind of error, and `is_opening_ssh_connection` is right about where it came from — but neither says whose fault it is. When the validator's own egress, DNS or key handling is down, `asyncssh.connect` raises `OSError`/`asyncssh.Error` for every executor in the batch, each result carries `EXECUTOR_SSH_UNREACHABLE`, and lium-platform#25 takes the whole fleet off the market until the next cycle that connects. The backend's hourly sweep grew exactly this guard after 12 Aug 2026 (DAH-2658: 33 min off-chain, 7 min short of deactivating all 387 executors). Cheapest place is `miner_service.publish_machine_specs` (`miner_service.py:818`): the batch is in hand there, so when more than a share of it (say ½) failed at the connect, publish `"availability_errors": None` for the batch and log at error — `None` already means "leave the stored codes alone" on the backend side. I left the same note on #25 in case you prefer the guard there.

**2. `what_we_saw.error` is uncapped and ends up in a JSONB column the portal renders** (`availability.py:102-121`, `build_ssh_unreachable_event`). `str(e)` from asyncssh can carry the remote banner (`asyncssh.DisconnectError` includes the peer's reason string), so a node can put arbitrary text of any length into `executor.availability_errors` and onto the provider's page. Low: the portal shows it to the node's own provider and the column is JSONB, so it is length and not injection. A `[:500]` on `error` (and on `event_text` if a future check derives it from a peer string) closes it.

Checked and fine: `is_opening_ssh_connection` flips false as the first statement inside the `async with`, so an `__aexit__` error or a pipeline exception is never blamed on sshd; `has_reached_the_node` makes a later failure report `[]` (clears) and a pre-connect failure report `None` (leaves alone), and `test_a_network_error_before_the_ssh_connect_is_not_blamed_on_the_node` / `test_a_failure_after_the_shell_opened_reports_the_node_as_reachable` pin both; every result of a cycle gets the same `scored_at` after scoring (`core/validator.py:570-573`), so the backend's not-older-than guard has a value on failed results too; `availability_errors()` reads the whole event list and dedups by code; `compute_client.py` passes the field through; results with `spec=None` are still published, so a failed cycle reaches the backend.

One thought for later, not for this PR: the code list is an enum here (`AvailabilityErrorCode`) but a plain string on the backend/portal side, where the readable names live in `_READABLE_AVAILABILITY_ERRORS`. A code added here without a line there renders as its raw name in the portal — a shared `lium_core` list would keep the two in step.

arhangel66 (approved):

surcyf123 (approved):
Re-read the full diff at 250440d6 against the two points from the 08:58Z review, plus the commit after @arhangel66's approval (250440d6, the small-fleet exemption).

Checked at this head:
- **Point 1 (one bad cycle on our side hides the fleet):** `silence_availability_errors_on_our_own_outage` runs in `core/validator.py` after scoring and before `publish_machine_specs`, over the flattened `incentive.job_results` — the same objects `publish_machine_specs` reads, so the in-place `None` reaches the wire. It counts only results that checked (`availability_errors is not None`), trips at > ½ unreachable of ≥ 5 checked, and logs at error; a validator with < 5 checked nodes reports what it found. Pinned by `test_a_cycle_that_reached_almost_nothing_reports_nothing`, `test_one_unreachable_node_among_many_is_still_reported`, `test_results_that_never_checked_are_not_counted`, `test_a_validator_with_a_few_nodes_still_reports_what_it_found`.
- **Point 2 (peer text uncapped):** `what_we_saw.error` is `error[:500]` in `build_ssh_unreachable_event`; `event_text`, `impact` and `remediation` are constants. Pinned by `test_the_peer_text_of_an_ssh_failure_is_capped`.
- Unchanged and still fine: `is_opening_ssh_connection` flips false as the first statement inside the `async with`; `has_reached_the_node` → `[]`, pre-connect failure → `None`; `scored_at` is stamped on every result before the guard runs; `compute_client` / `miner_service` pass the field through.
- Validators suite at 250440d6 on a sandbox box: 1927 passed, 15 skipped, 2 failed — `test_scrape_infiniband::test_an_unreadable_sysfs_tree_is_not_reported_as_no_devices` and `test_sshd_bootstrap_script::test_adopt_path_hardens_config_for_key_only_auth`, both environmental (the box runs as root with an sshd up) and failing identically in every lium-io run on that box today, including branches without this change. `tests/test_availability_errors.py` alone: 14 passed. CI at this sha: Validators Tests, ruff, e2e gate (5/14/2) green.

The "Invalid executor id" follow-up goes to its own ticket as you asked; nothing of it is in this review.

Approving; the merge is yours.

PR #1271, opened 2026-09-03
DAH-2850, raise matmul VRAM headroom to 4 GB for confidential-computing GPUs

arhangel66:
### Task Link

[DAH-2850](https://app.notion.com/p/Validator-GPU-work-proof-OOMs-on-confidential-computing-GPUs-raise-matmul-VRAM-headroom-3d0b8bfdbde981cc8159f616b012d34b)

---

## Problem

`gpu.validate.capability` sizes its matmul challenge to the card's reported capacity minus a fixed 2 GB (`get_max_matrix_dimensions` in `matrix_validation_service.py`). Nothing accounts for memory the card already holds. A GPU in NVIDIA confidential-computing mode reserves ~1.8 GB at idle (B200: 1766 MB used vs ~730 MB out of CC mode), so on the executor `DMCompVerify.cu` allocates `d_A`, then fails with `Failed to allocate d_B: out of memory`. The executor answers `UUID mismatch: expected ..., got 'None'`, the check fails with `GPU_VERIFY_FAILED`, and because `CapabilityCheck` is fatal the node scores 0 and never gets an `executor` row.

Seen in prod on 2026-09-03 11:15 UTC on an 8x B200 CVM node (`146.88.195.16:8001`) right after it passed TDX attestation; the customer rental on it is blocked. Every CC-mode CVM node hits this, including the B300 CVM node expected on 2026-09-05.

## Solution

Reserve a quarter of the card, clamped between `MATMUL_VRAM_RESERVE_MIN_MB = 2048` and `MATMUL_VRAM_RESERVE_MAX_MB = 4096`. Cards of 16 GB and up get 4 GB; 3-8 GB consumer cards keep the 2 GB they always had (reviewer's point: an uncapped 4 GB would have given the 3 GB GTX 1060 variant a negative `dim_k`, a bare quarter would have given it less than before). It stays a constant on purpose: host-reported `memory_used_mb` is executor-controlled and must not be able to shrink the challenge. 4 GB is ~2% of a B200, the proof still touches ~97% of the card.

Both call sites go through the same function. The all-cards shadow probe already sizes from 0.90 × registry nominal and was not failing; it just gets 2 GB smaller.

## Changes

- `neurons/validators/src/services/matrix_validation_service.py`: `MATMUL_VRAM_RESERVE_MIN_MB` / `MATMUL_VRAM_RESERVE_MAX_MB`, `get_max_matrix_dimensions` reserves `min(4096, max(2048, capacity // 4))` instead of the literal `2 * 1024`.
- `neurons/validators/tests/test_matmul_vram_reserve.py`: with `dim_n` at both ends of its random range, `d_A + d_B + d_C` for a B200 fits beside the observed CC-mode reserve plus a CUDA context, and provably did not with the old 2048 MB; the reserve clears `GPU_HELD_VRAM_MB_LIMIT` plus the probe's CUDA context; every VRAM size in the registry keeps a positive `dim_k` with a reserve between 2048 MB and the 4096 MB ceiling.

## Deployment Steps

Validator release. The `validator-v*` tag only builds and pushes the validator and runner images; the rollout itself is the separate Pulumi deploy (reloader). No executor change: the executor side only receives `dim_k`. After the rollout the next central-miner cycle (every ~15 min) re-validates the affected node; the two fatal gates after capability (cached template verification, rental verification) have never run on this node and must be watched on that first cycle.

## Review Request

- [x] Just code review

## Other PRs

None.

## Risks

- The single-card work-proof shrinks by 2 GB on cards of 16 GB and up (~1% on a B200, ~2.5% on an H100 80 GB, ~8% on a 24 GB card); cards under 16 GB are unchanged. Timing of the proof is bounded by `MATRIX_VERIFY_TIMEOUT_SECONDS` from above only, so a slightly smaller matrix cannot fail an honest host.
- Cards whose CC-mode reserve plus CUDA context exceeds 4 GB would still OOM. Measured only on a B200 (~2.4 GB); the H100/H200 CC-mode reserve has not been measured.
- `GPU_HELD_VRAM_MB_LIMIT` (2048 MB, `gpu_usage.py`) is a separate idle-VRAM gate; the CC-mode B200 sits under it at 1766 MB, not touched here.

### Test Case 1

**Actions**

`cd neurons/validators && pdm run pytest tests/test_matmul_vram_reserve.py tests/test_matmul_allcards.py tests/test_matrix_validation_service_precheck.py tests/test_capability_check.py`

**Expected Output**

58 passed. The regression test pins that with 2048 MB headroom A+B+C exceeds capacity − 1766 − 600 MB for both `dim_n` values. Full validator suite: 1938 passed, the 3 failures in `test_sshd_bootstrap_script.py` fail identically on `main` (macOS harness, unrelated).

### Test Case 2

**Actions**

After the validator deploy, watch the next central-miner cycle for `146.88.195.16:8001` in Loki (`gpu.validate.capability`).

**Expected Output**

No `Failed to allocate d_B`, check passes, an `executor` row appears.

## Checklist

- [ ] Proper labels added (`ready-review`, `require-qa`, `breaking-changes` do not exist in this repo)
- [x] Risks are documented
- [x] Tests are described


neurons/validators/src/services/matrix_validation_service.py
```diff
@@ -226,7 +234,7 @@ def get_gpu_memory(self, machine_spec: dict) -> bool:
         return 0
 
     def get_max_matrix_dimensions(self, gpu_memory, dim_n):
-        gpu_memory = gpu_memory - 2 * 1024
+        gpu_memory = gpu_memory - min(MATMUL_VRAM_HEADROOM_MB, gpu_memory // 4)
```
taiberium: NIT: on 3 GB and 6 GB cards the quarter cap gives less headroom than the old 2048 MB.

```suggestion
        gpu_memory = gpu_memory - min(MATMUL_VRAM_HEADROOM_MB, max(2048, gpu_memory // 4))
```

taiberium (approved):

neurons/validators/src/services/matrix_validation_service.py
arhangel66: Landed in 80f134b1 as `min(MATMUL_VRAM_RESERVE_MAX_MB, max(MATMUL_VRAM_RESERVE_MIN_MB, gpu_memory // 4))`, with the registry-wide test asserting the reserve stays within [2048, 4096] MB for every card. Thanks.

PR #1272, opened 2026-09-04
DAH-2862: recognise TDX guests reporting the cpu as raw family/model

arhangel66:
### Task Link

[DAH-2862](https://app.notion.com/p/host-detect-tdx-does-not-recognise-CVM-guests-whose-cpu-model-is-raw-family-model-06-cf-06-ad-3d1b8bfdbde98190922af2e1065f8e71)

---

## Problem

The `host.detect.tdx` check matches the CPU by its marketing name (`Intel(R) Xeon(R) Platinum 8592+`
and friends). Inside a TDX guest the host CPU name is not passed through, so `lscpu` in the machine
scrape falls back to the raw CPUID and `specs["cpu"]["model"]` becomes a hex `family/model` string —
`06/cf` on every TEEmo H200 CVM (Emerald Rapids) and `06/ad` on the B200 CVM at 146.88.195.16
(Granite Rapids).

Result: every real CVM logs "Intel TDX CPU capability not detected" and records
`tdx_host_supported=False`. Prod validator log, 2026-09-03 13:17:25 UTC, node 146.88.195.16. The
check is non-fatal, so nothing is scored down, but the recorded boolean is wrong for exactly the
machines the check exists for.

## Solution

Keep the marketing-name regexes and additionally accept the raw CPUID form for family 6 with the
TDX-capable model ids only. Model ids follow the Linux kernel header
`arch/x86/include/asm/intel-family.h`: `SAPPHIRERAPIDS_X 0x8F`, `EMERALDRAPIDS_X 0xCF`,
`GRANITERAPIDS_X 0xAD`, `GRANITERAPIDS_D 0xAE`. Arbitrary family 6 models are still rejected.

## Changes

- `neurons/validators/src/services/task/checks/tdx_host.py`: new `_TDX_CAPABLE_RAW_CPUID` pattern,
  checked first in `TdxHostCheck.is_tdx_capable`.
- `neurons/validators/tests/test_tdx_host_check.py`: cases for `06/cf`, `06/ad`, `06/8f`, `06/AE`
  (detected) and `06/97` (Alder Lake client — not detected).

## Deployment Steps

Use GitHub Action.

## Review Request

- [x] Just code review

## Other PRs

None.

## Risks

- Low. The check is non-fatal (`fatal = False`, always `passed=True`); the only effect is the value
  of `tdx_host_supported` written into specs and the log line that goes with it.
- The raw form is matched against the whole string (anchored), so it cannot fire on a substring of a
  marketing name. Existing marketing-name behaviour is untouched.
- If a non-TDX host ever reported exactly `06/8f|cf|ad|ae`, it would be marked supported — but that
  string only appears when the CPU name cannot be resolved, which on our fleet means a CVM guest on
  that very silicon.

### Test Case 1

**Actions**

`pdm run pytest tests/test_tdx_host_check.py -q` in `neurons/validators`.

**Expected Output**

`30 passed` — including the new raw-CPUID cases.

### Test Case 2

**Actions**

`pdm run pytest tests/ -q` in `neurons/validators`.

**Expected Output**

`3 failed, 1934 passed, 15 skipped, 9 errors` — the 3 failures are the pre-existing
`tests/test_sshd_bootstrap_script.py` ones on clean main, the 9 errors are `test_port_mapping_dao.py`
setup failing on a missing local `greenlet`/DB. Neither touches this check.

## Checklist

- [x] Risks are documented
- [x] Tests are described

### DIFF
--- a/neurons/validators/src/services/task/checks/tdx_host.py
+++ b/neurons/validators/src/services/task/checks/tdx_host.py
@@ -33,9 +33,9 @@ _TDX_CAPABLE_PATTERNS: list[re.Pattern] = [
 ]
 
 # In a TDX guest lscpu has no marketing name and prints the raw CPUID as "family/model" in hex
-# ("06/cf" on the H200 CVMs, "06/ad" on the B200). Only TDX-capable family 6 models pass; ids from
-# the kernel's arch/x86/include/asm/intel-family.h (SPR 0x8F, EMR 0xCF, GNR 0xAD, GNR-D 0xAE).
-_TDX_CAPABLE_RAW_CPUID = re.compile(r"^\s*0*6/(?:8f|cf|ad|ae)\s*$", re.IGNORECASE)
+# ("06/cf" on the H200 CVMs, "06/ad" on the B200). One model id per generation the name patterns
+# above accept, from arch/x86/include/asm/intel-family.h: SPR 8f, EMR cf, GNR ad, GNR-D ae, SRF af.
+_TDX_CAPABLE_RAW_CPUID = re.compile(r"^\s*0*6/(?:8f|cf|ad|ae|af)\s*$", re.IGNORECASE)
 
 
 class TdxHostCheck:
@@ -56,7 +56,7 @@ class TdxHostCheck:
 
     @staticmethod
     def is_tdx_capable(cpu_model: str) -> bool:
-        """Return True if the CPU model name matches a known TDX-capable Intel processor."""
+        """Return True if the model name or raw CPUID belongs to a TDX-capable Intel processor."""
         if _TDX_CAPABLE_RAW_CPUID.match(cpu_model):
             return True
         return any(p.search(cpu_model) for p in _TDX_CAPABLE_PATTERNS)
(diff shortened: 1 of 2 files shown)


neurons/validators/src/services/task/checks/tdx_host.py
```diff
@@ -32,6 +32,11 @@
     re.compile(r"Xeon" + _SEP + r"W[579]-[23][4-9]\d{2}", re.IGNORECASE),
 ]
 
+# In a TDX guest lscpu has no marketing name and prints the raw CPUID as "family/model" in hex
+# ("06/cf" on the H200 CVMs, "06/ad" on the B200). One model id per generation the name patterns
+# above accept, from arch/x86/include/asm/intel-family.h: SPR 8f, EMR cf, GNR ad, GNR-D ae, SRF af.
+_TDX_CAPABLE_RAW_CPUID = re.compile(r"^\s*0*6/(?:8f|cf|ad|ae|af)\s*$", re.IGNORECASE)
```
taiberium, line 38: **NIT:** the regex also accepts `06/af` (Sierra Forest) while the PR body lists only `8f/cf/ad/ae`. The code is consistent with the existing `Xeon 6\d{3}` pattern and its `6766E` test, so the description is what is stale.

taiberium (approved):
Findings inline.

neurons/validators/src/services/task/checks/tdx_host.py
arhangel66: Confirmed: Sierra Forest support is intentional and covered by the code and test; only the PR description is stale. No runtime defect. Resolving for release.

PR #1274, opened 2026-09-04
DAH-2859: preflight the CVM host gates in lium-cvm.sh check and guard stop against zombie VMs

arhangel66:
## Summary

`lium-cvm.sh check` passed on a host that could not possibly run an attested CVM, and `lium-cvm.sh stop` reported success while the VM kept running. This PR makes `check` a real preflight for the four gates a provider host hit in Aug/Sep 2026, makes `stop` prove the VM is gone, and writes the missing host requirements into `docs/host-setup.md`.

### Task Link

[CVM host preflight: `lium-cvm.sh check` misses every gate](https://app.notion.com/p/CVM-host-preflight-lium-cvm-sh-check-misses-every-gate-TEEmo-s-B200-hit-stop-leaves-zombie-VMs-ho-3d1b8bfdbde981329cedc87b61f14b03)

---

## Problem

Onboarding one B200 host cost several days and four CVM redeploys. Every blocker was checkable in seconds before the first launch, and `check` reported none of them:

1. **QEMU.** `check` only asked whether *some* `qemu-system-x86_64` is on PATH. The launcher takes its binary from `client.conf`, and only the dstack 9.2.1 build reproduces RTMR0 against the prod verifier's oracle — a host on distro QEMU 10.x can never pass attestation (`is_valid=false`, RTMR0 mismatch). That verdict is enforced unconditionally, independent of the compose whitelist.
2. **TDX module TCB.** Build 786 (2.0.02, minor SVN 3) is below Intel's minimum for Granite Rapids (2.0.08, build 882, minor SVN 5), so DCAP verification of the CVM's quote answers `No matching TCB level found` — for our key provider, our validator and Phala's KMS alike.
3. **NUMA.** Firmware advertising more possible than online nodes (Xeon 6767P: `0-19` vs `0-1`) kills the pinned Gramine 1.5 key provider with `load_enclave … EINVAL` (Gramine issue #1452, fixed in 1.6). Hardware-dependent; an OS reprovision changes nothing.
4. **PCCS / platform registration.** Phala's public PCCS cannot serve PCK certificates for a platform Intel does not know about → AESM error 44, `load_enclave … EPERM`. `docs/host-setup.md` claimed the default "works out of the box".

Separately, `dstack.py stop` returns success when there is no `runtime.json`. Doing `rm -rf run/vms/<name>` before `stop` (the upgrade sequence, in the wrong order) therefore stops nothing while `lium-cvm.sh` prints `[SUCCESS] CVM stopped` — the VM keeps holding all 8 GPUs and its ports, and the next `new`+`run` fails with resources busy. Seen on the provider host 2026-09-03 10:25–10:35 UTC.

## Solution

`check` resolves and verifies the same QEMU the launcher will use, reads the TDX module build out of `dmesg`, compares possible vs online NUMA nodes, and probes the configured PCCS — each with a pointer to the section of `docs/host-setup.md` that fixes it. `stop` no longer trusts the exit code: after `dstack.py stop` it looks for the launcher and QEMU processes that carry the VM directory on their command line and fails loudly if any survive. The docs gain the TDX module minimum with the vendor-independent update path, the NUMA note, platform registration, the GCC 15 build note, and six troubleshooting rows.

## Changes

- `lium-cvm.sh`: new `resolve_qemu_path` (reads `[qemu] path` from the `client.conf` chain, falls back to PATH) and `cvm_pids` helpers.
- `check_requirements`: QEMU 9.2.1 required (distro 10.x → `log_error`), TDX module build ≥ 882 required, NUMA possible vs online warning, PCCS reachability + platform-registration reminder. Existing checks untouched.
- `stop_cvm`: verifies no `dstack.py run` / QEMU process still holds the VM directory; returns 1 with the surviving pids instead of `log_success`.
- `docs/host-setup.md` §1: TDX module minimum table, how to check on 6.x vs 7.0 kernels, the EFI update path from Intel's signed package.
- `docs/host-setup.md` §1/§4: NUMA possible-vs-online note and the Gramine 1.9 workaround; §4 corrects the "Phala's PCCS works out of the box" claim and adds registration.
- `docs/host-setup.md` §3: GCC 15 / bundled `dtc` build failure and how to avoid the subproject; §6: the runner digest comment; 6 new troubleshooting rows.

## Deployment Steps

Nothing to deploy — a provider-side script and its docs. Providers get it with the next `git pull` of the release tag; no CVM redeploy, no measurement change.

## Review Request

- [x] Just code review

## Other PRs

None. Related follow-ups tracked separately: DAH-2860 (re-pin the key provider to Gramine 1.9), PR #1134 (QEMU 10 verifier).

## Risks

- `check` can now fail (`return 1`) on hosts it used to pass: distro QEMU ≥ 10 and TDX module build < 882 are hard errors. That is the intent — both make attestation impossible — but a provider who ignored `check` before will now see it red.
- QEMU is resolved from `client.conf` the way `dstack.py` does it, with one simplification: the config walk-up starts at the script directory instead of the launcher's cwd. A `.dstack/client.conf` placed above the repo would be seen by the launcher and not by `check`.
- The new checks are read-only (`dmesg`, `/sys`, `curl`, `pgrep`); none of them touch a running VM.
- The `stop` guard matches processes by the VM directory on their command line. If it ever produced a false positive, `stop` would report failure for a VM that is actually down; the boundary case (`my-executor` vs `my-executor-2`) is covered by the pattern and tested below.
- No change to `dstack.py`, the measured compose, or anything inside the attestation boundary.

## Test Cases

`shellcheck` is not installed on this machine, so it was not run; `bash -n` passes.

### Test Case 1 — syntax and a full `check` run

**Actions** `bash -n neurons/executor/dstacktee/lium-cvm.sh`, then `./lium-cvm.sh check` on macOS (no `/sys`, no readable `dmesg`, no `/dev/kvm`).

**Expected Output** No syntax errors; `check` runs to the end under `set -e` and every new check takes its "cannot determine" branch instead of crashing:

```
[WARNING] QEMU 9.0.2 at /usr/local/bin/qemu-system-x86_64 is not the dstack 9.2.1 build attestation expects (docs/host-setup.md section 3)
[WARNING] No TDX module build number in dmesg (kernel 7.0 prints none, and dmesg may need root); the proof is then the quote's tee_tcb_svn — see docs/host-setup.md section 1
[WARNING] Cannot read /sys/devices/system/node/{possible,online}; skipping the NUMA check
[WARNING] PCCS https://pccs.phala.network/sgx/certification/v4/ does not answer; the key provider will fail DCAP collateral fetch
[INFO] A reachable PCCS is not enough: the platform must be registered with Intel …
```

### Test Case 2 — QEMU resolution and verdicts (stub binaries)

**Actions** A stub `qemu-system-x86_64` printing `QEMU emulator version 10.2.1 (Debian 1:10.2.1+ds-1ubuntu1)` and another printing `9.2.1`, selected through `[qemu] path` in a `client.conf` that also contains an `[image]` section; then a run with no config and no QEMU on PATH.

**Expected Output**

```
[ERROR] QEMU 10.2.1 at /tmp/…/q10/qemu-system-x86_64 is a distro build: the verifier's RTMR0 oracle is built from dstack QEMU 9.2.1, so attestation can never pass
[INFO] See docs/host-setup.md section 3 — build kvinwang/qemu-tdx branch dstack-qemu-9.2.1 and point [qemu] path in client.conf at it
[SUCCESS] QEMU is the dstack 9.2.1 build: /tmp/…/q921/qemu-system-x86_64
[ERROR] QEMU (qemu-system-x86_64) is not installed or not executable
```

The `[qemu] path` value wins over the PATH binary, and the `path` key is read only inside the `[qemu]` section.

### Test Case 3 — TDX module build and NUMA parsing

**Actions** Feed the parser real `virt/tdx` dmesg lines (builds 786 / 882 / 1013, plus a kernel-7.0 line with no `build_num`) and the NUMA pairs `0-1|0-1` and `0-19|0-1`.

**Expected Output** `786 → ERROR (below Intel minimum)`, `882 → SUCCESS`, `1013 → SUCCESS`, no `build_num` → WARNING; `0-1/0-1 → SUCCESS`, `0-19/0-1 → WARNING (Gramine 1.5 EINVAL)`.

### Test Case 4 — PCCS probe

**Actions** Point `pccs_url` at an unreachable host, then at a reachable one.

**Expected Output** `[WARNING] PCCS … does not answer` / `[SUCCESS] PCCS answers: …`, followed in both cases by the registration reminder.

### Test Case 5 — `stop` against a zombie VM

**Actions** Five cases with a stubbed `dstack.py stop` that always exits 0 (the real no-`runtime.json` behaviour): (1) nothing running; (2) a `dstack.py run` of `my-executor-2` while stopping `my-executor`; (3) a live `dstack.py run …/vms/my-executor`; (4) only a leftover `qemu-system-x86_64 … -virtfs local,path=…/vms/my-executor/shared`; (5) `dstack.py stop` itself exiting non-zero.

**Expected Output**

```
### case 1: nothing running -> stop succeeds
[SUCCESS] CVM 'my-executor' stopped
  rc=0
### case 2: dstack.py run of ANOTHER vm (my-executor-2) must not trip my-executor
[SUCCESS] CVM 'my-executor' stopped
  rc=0
### case 3: zombie launcher for my-executor still alive -> stop must fail
[ERROR] CVM 'my-executor' is still running (pids: 27226) and keeps holding its GPUs and ports
  rc=1
### case 4: only a leftover QEMU holding the vm dir -> stop must fail
[ERROR] CVM 'my-executor' is still running (pids: 27245) and keeps holding its GPUs and ports
  rc=1
### case 5: dstack.py stop itself fails -> old behaviour kept
[ERROR] Failed to stop CVM 'my-executor'
  rc=1
```

### Not verified

- Nothing was run on a real TDX host: the TDX-module, NUMA and QEMU-9.2.1 success branches were exercised with stubs and real dmesg/sysfs strings, not on hardware.
- The GCC 15 workaround in §3 (`--disable-fdt` / `--enable-fdt=system`) is verified against the pinned QEMU tree (`configs/targets/x86_64-softmmu.mak` declares no `TARGET_NEED_FDT`; `scripts/meson-buildoptions.sh` maps the flags to `-Dfdt=disabled` / `-Dfdt=system`; `meson.build` builds `subprojects/dtc` only for `-Dfdt=internal`) but not by an actual build on GCC 15.

## Checklist

- [x] Risks are documented
- [x] Tests are described

## Review pass (Mikhail)
- The build-882 floor now applies only to the 2.0 module series (Granite Rapids); Sapphire/Emerald Rapids hosts on the 1.5 series are reported, not failed.
- `stop` also reports surviving VM processes when the VM directory is already gone (the exact `rm` then `stop` sequence that left the zombie on 2026-09-03), and re-checks for up to 5 s so the launcher's own exit after QEMU is not mistaken for a survivor.

### DIFF
--- a/neurons/executor/dstacktee/lium-cvm.sh
+++ b/neurons/executor/dstacktee/lium-cvm.sh
@@ -60,6 +60,34 @@ container_running() {
     docker ps --format "table {{.Names}}" | grep -q "$1" 2>/dev/null
 }
 
+# Resolve the QEMU binary the launcher will actually use: dstack.py reads
+# [qemu] path from the client.conf chain (later files override earlier ones)
+# and only falls back to PATH when no config sets it.
+resolve_qemu_path() {
+    local resolved=""
+    local conf
+    for conf in /etc/dstack/client.conf "$HOME/.config/dstack/client.conf" "$THIS_DIR/.dstack/client.conf"; do
+        [ -f "$conf" ] || continue
+        local configured
+        configured=$(awk '/^[[:space:]]*\[/ { section = $0 }
+            section ~ /\[qemu\]/ && /^[[:space:]]*path[[:space:]]*=/ { sub(/^[^=]*=[[:space:]]*/, ""); print }' "$conf" | tail -n1)
+        [ -n "$configured" ] && resolved="$configured"
+    done
+    if [ -n "$resolved" ]; then
+        echo "$resolved"
+    else
+        command -v qemu-system-x86_64 2>/dev/null
+    fi
+}
+
+# List host processes still holding a VM directory: the dstack.py launcher and
+# the QEMU it runs as a child (both carry the VM directory on their command line).
+cvm_pids() {
+    local vm_dir="$1"
+    pgrep -f "dstack\.py run ${vm_dir}(/|[[:space:]]|\$)" 2>/dev/null || true
+    pgrep -f "qemu-system-x86_64.*${vm_dir}/" 2>/dev/null || true
+}
+
 # Function to download and extract OS image
 download_os_image() {
     local temp_file="/tmp/dstack-os-image-$$.tar.gz"
@@ -142,10 +170,29 @@ check_requirements() {
 
     # Check QEMU
     log_info "Checking QEMU installation..."
-    if command_exists qemu-system-x86_64; then
-        log_success "QEMU is installed: $(qemu-system-x86_64 --version | head -n1)"
+    local qemu_bin
+    qemu_bin=$(resolve_qemu_path || true)
+    if [ -n "$qemu_bin" ] && "$qemu_bin" --version >/dev/null 2>&1; then
+        local qemu_version
+        qemu_version=$("$qemu_bin" --version 2>/dev/null | sed -n 's/.*version \([0-9][0-9.]*\).*/\1/p' | head -n1)
+        case "$qemu_version" in
+        9.2.1)
+            log_success "QEMU is the dstack 9.2.1 build: $qemu_bin"
+            ;;
+        1[0-9].*)
+            log_error "QEMU $qemu_version at $qemu_bin is a distro build: the verifier's RTMR0 oracle is built from dstack QEMU 9.2.1, so attestation can never pass"
+            log_info "See docs/host-setup.md section 3 — build kvinwang/qemu-tdx branch dstack-qemu-9.2.1 and point [qemu] path in client.conf at it"
+            all_good=false
+            ;;
+        "")
+            log_warning "Could not parse the QEMU version of $qemu_bin; attestation requires the dstack 9.2.1 build (docs/host-setup.md section 3)"
+            ;;
+        *)
+            log_warning "QEMU $qemu_version at $qemu_bin is not the dstack 9.2.1 build attestation expects (docs/host-setup.md section 3)"
+            ;;
+        esac
     else
-        log_error "QEMU (qemu-system-x86_64) is not installed"
+        log_error "QEMU (qemu-system-x86_64) is not installed or not executable${qemu_bin:+ at $qemu_bin}"
         all_good=false
     fi
 
@@ -166,12 +213,49 @@ check_requirements() {
         log_warning "Intel TDX support not detected in CPU info"
     fi
 
+    # Check the TDX module TCB. The 882 floor (2.0.08, minor SVN 5) is Intel's minimum for
+    # the 2.0 series on Granite Rapids; the 1.5 series on Sapphire/Emerald Rapids numbers its
+    # builds differently, so it only gets reported. Too old = "No matching TCB level found".
+    log_info "Checking Intel TDX module build..."
+    local tdx_line tdx_major tdx_build
+    tdx_line=$(dmesg 2>/dev/null | grep -i 'virt/tdx' | grep -i 'build_num' | tail -n1)
+    tdx_major=$(printf '%s' "$tdx_line" | sed -n 's/.*major_version[^0-9]*\([0-9][0-9]*\).*/\1/p')
+    tdx_build=$(printf '%s' "$tdx_line" | sed -n 's/.*build_num[^0-9]*\([0-9][0-9]*\).*/\1/p')
+    if [ -z "$tdx_build" ]; then
+        log_warning "No TDX module build number in dmesg (kernel 7.0 prints none, and dmesg may need root); the proof is then the quote's tee_tcb_svn — see docs/host-setup.md section 1"
+    elif [ "$tdx_major" != "2" ]; then
+        log_info "TDX module ${tdx_major:-?}.x build $tdx_build (pre-Granite Rapids series); Intel's minimum is per platform — the quote's tee_tcb_svn is the proof, see docs/host-setup.md section 1"
+    elif [ "$tdx_build" -ge 882 ]; then
+        log_success "TDX module build $tdx_build meets Intel's minimum (882 = 2.0.08)"
+    else
+        log_error "TDX module build $tdx_build is below Intel's minimum 882 (2.0.08, minor SVN 5): DCAP verification fails with 'No matching TCB level found'"
+        log_info "See docs/host-setup.md section 1 — update the module via /boot/efi/EFI/TDX/ or a vendor BIOS"
+        all_good=false
+    fi
+
     # Check SGX devices
     log_info "Checking SGX devices..."
     if [ -e /dev/sgx_enclave ] && [ -e /dev/sgx_provision ]; then
         log_success "SGX devices are available"
     else
         log_warning "SGX devices (/dev/sgx_enclave, /dev/sgx_provision) not found"
+    fi
+
+    # Check NUMA topology. The possible set is always a superset of the online one,
+    # so any difference means the firmware over-declares nodes.
+    log_info "Checking NUMA node topology..."
+    if [ -r /sys/devices/system/node/possible ] && [ -r /sys/devices/system/node/online ]; then
+        local numa_possible numa_online
+        numa_possible=$(cat /sys/devices/system/node/possible)
+        numa_online=$(cat /sys/devices/system/node/online)
+        if [ "$numa_possible" = "$numa_online" ]; then
+            log_success "NUMA nodes possible=$numa_possible online=$numa_online"
+        else
+            log_warning "NUMA nodes possible=$numa_possible but online=$numa_online: Gramine 1.5 (the pinned key-provider image) fails load_enclave with EINVAL on such firmware"
+            log_info "See docs/host-setup.md section 4 — rebuild the key provider on gramine:1.9"
+        fi
+    else
+        log_warning "Cannot read /sys/devices/system/node/{possible,online}; skipping the NUMA check"
     fi
 
     # Check key-provider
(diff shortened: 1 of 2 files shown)


neurons/executor/dstacktee/lium-cvm.sh
```diff
@@ -142,10 +170,29 @@ check_requirements() {
...
     log_info "Checking QEMU installation..."
-    if command_exists qemu-system-x86_64; then
-        log_success "QEMU is installed: $(qemu-system-x86_64 --version | head -n1)"
+    local qemu_bin
+    qemu_bin=$(resolve_qemu_path || true)
+    if [ -n "$qemu_bin" ] && "$qemu_bin" --version >/dev/null 2>&1; then
+        local qemu_version
+        qemu_version=$("$qemu_bin" --version 2>/dev/null | sed -n 's/.*version \([0-9][0-9.]*\).*/\1/p' | head -n1)
+        case "$qemu_version" in
+        9.2.1)
```
surcyf123, line 179: A distro QEMU 9.2.1 (Ubuntu 25.04 ships `9.2.1 (Debian 1:9.2.1+ds-...)`) lands here and is reported as the dstack build although it cannot attest. Keep the full `--version` line and treat a `(Debian`/`(Ubuntu`/`+ds` suffix as the same error as the 10.x case.

neurons/executor/dstacktee/lium-cvm.sh
```diff
@@ -60,6 +60,34 @@ container_running() {
...
+    else
+        command -v qemu-system-x86_64 2>/dev/null
+    fi
+}
+
+# List host processes still holding a VM directory: the dstack.py launcher and
+# the QEMU it runs as a child (both carry the VM directory on their command line).
+cvm_pids() {
+    local vm_dir="$1"
+    pgrep -f "dstack\.py run ${vm_dir}(/|[[:space:]]|\$)" 2>/dev/null || true
```
surcyf123, line 87: `$vm_dir` is interpolated into an ERE unescaped; a checkout path with `+`, `(` or `[` makes both pgrep patterns miss and `stop` reports success. Escape the path (sed over the regex metacharacters) before building the pattern.

neurons/executor/dstacktee/lium-cvm.sh
```diff
@@ -142,10 +170,29 @@ check_requirements() {
 
     # Check QEMU
     log_info "Checking QEMU installation..."
-    if command_exists qemu-system-x86_64; then
-        log_success "QEMU is installed: $(qemu-system-x86_64 --version | head -n1)"
+    local qemu_bin
+    qemu_bin=$(resolve_qemu_path || true)
+    if [ -n "$qemu_bin" ] && "$qemu_bin" --version >/dev/null 2>&1; then
```
surcyf123, line 175: If `[qemu] path` contains `~` or `$VAR`, this fails as "not executable" while the launcher may expand it. Does `dstack.py` expand the value? If yes, expand it here too.

surcyf123 (changes requested):
Not approving yet: one check to fix, and CI is not clean at this head.

1. **[medium]** `neurons/executor/dstacktee/lium-cvm.sh:179`: the `9.2.1)` case treats any QEMU that reports version 9.2.1 as the dstack build. Ubuntu 25.04 ships distro QEMU 9.2.1 (`QEMU emulator version 9.2.1 (Debian 1:9.2.1+ds-...)`), which has no dstack TDX patches and a different ACPI oracle, so `check` prints `[SUCCESS] QEMU is the dstack 9.2.1 build` on a host that cannot attest. That is the same false pass this PR sets out to remove. Fix: keep the full `--version` line and treat a distro suffix (`(Debian`, `(Ubuntu`, `+ds`) as the same `log_error` as the 10.x case, or accept 9.2.1 only when the binary lives outside `/usr/bin`.
2. **[low]** `neurons/executor/dstacktee/lium-cvm.sh:87`: `pgrep -f` reads `$vm_dir` as an extended regex. A checkout path containing `+`, `(` or `[` makes both patterns miss, so `stop` reports success with the VM still up. Fix: escape the path before interpolating it, for example `re=$(printf '%s' "$vm_dir" | sed 's/[][\.*^$+?(){}|]/\\&/g')`.
3. **[low]** `neurons/executor/dstacktee/lium-cvm.sh:175`: a `[qemu] path` value with `~` or `$VAR` is used verbatim, so `check` reports "not executable" while the launcher may expand it. Does `dstack.py` expand that value? If yes, expand it here too (`${resolved/#\~/$HOME}`); if no, this is fine as is.
4. **[info]** CI at head `56b68239`: run 34936018144 ended with `tests-ok` FAILURE (job 104274338019) plus cancelled `e2e-gate`, `Executor Tests` and `Analyze (python)`; the rerun 34936113539 is green. The failed run is what blocks an approve from the loop. If it was superseded, a re-run or a one-line note here is enough.

What I checked: the `stop` guard against `dstack.py shutdown_instance` at this head, which returns 0 with no `runtime.json`, when the vsock shutdown request fails, and after the 30 s wait times out without `--force`, so the pgrep check is the only thing that catches all three. The `my-executor` vs `my-executor-2` boundary holds because of the trailing `/` and `(/|[[:space:]]|$)`. The script has no `pipefail`, so the `dmesg | grep | tail` chain stays safe under `set -e`. The build-882 floor applies to `major_version 2` only, matching the review-pass note.

PR #1279, opened 2026-09-06
DAH-2740 - [P1] Keep the pod's container until an edit's replacement runs

surcyf123:
Implements [DAH-2740](https://app.notion.com/p/A-pod-edit-force-removes-the-pod-s-own-container-before-recreating-it-and-cannot-roll-back-so-15-o-3c3b8bfdbde981aa9ebad4d93a73dbd7) · reviewer: @jam6099

**TL;DR** — 8 of 88 pod edits in a week force-removed the pod's own container and then failed, leaving the customer with nothing. The edit now parks the container (`rename` + `stop`), builds the replacement, and on any failure puts the parked one back the way `start_container` does; both sweepers protect `pod_<id>__prev` while the pod is rented. Risk: validator create path; fresh creates unchanged.

**What changed**
- `_EditSwap` (`services/docker_service.py`): `park()` renames and stops the current container; a parked container left by a crashed edit is recovered, never removed
- `restore()` brings the parked container back through `_bring_up_existing_container`: SDK start, gocryptfs remount (`allow_init=False`), sshd
- `EDIT_PARKED_SUFFIX` (`services/const.py`); both sweepers protect `<pod>__prev` for every rented/active pod
- `retry_ssh_command` passes `reraise=True`: callers see the last attempt's exit code and stderr, not `RetryError[…]`
- `tests/test_dah2740_edit_park_then_swap.py`: 11 tests — park/sweep/restore, a failed edit undone, a crashed edit recovered, the swap's exit never changes the create's outcome

**How to verify**
- `cd neurons/validators && pdm run pytest tests/test_dah2740_edit_park_then_swap.py -q` → 11 passed
- `pdm run pytest tests/ -q` → the suite (the two `sshd_bootstrap` failures are main's own)

**Verified by the loop:** 7 Sep 2026 · sandbox EC2 · validators 1945 passed · Result: PASS 7e395c08 · Gate: PASS eaff9e3

**Ran it:** `tools/aws_verify.sh lium-io --app io-all` on EC2 (20260910T084806Z-17ae, at dbef3e5, the earlier head of this same patch; carried to eaff9e3 unchanged) → validators… · … +2 under Details

**Self-review:** clean at eaff9e3 (carried from dbef3e5: rebase-only, patch unchanged; 15 checks, 7 low)

**Parts:** backend n/a — validator only (backend delete-path wedge: lium-platform#64) · migration n/a — no schema · frontend n/a — nothing shown changes · CLI/SDK n/a — API contract unchanged · docs n/a — nothing a provider sets · tests ✓ 11 in `test_dah2740_edit_park_then_swap.py` · dashboards n/a — no metric

**Docs:** none changed in this PR — validator internals; nothing a provider sees or sets changes.

[DAH-2740](https://app.notion.com/p/A-pod-edit-force-removes-the-pod-s-own-container-before-recreating-it-and-cannot-roll-back-so-15-o-3c3b8bfdbde981aa9ebad4d93a73dbd7).

## Problem

On the edit path (`local_volume` set) `DockerService.create_container` runs the stale-container sweep before creating the replacement. The pod's own `pod_<id>` container is never in the backend's `active_container_names`, so `clean_existing_containers` force-removes it (`docker rm -fv`, five attempts × 10 s) to make room — the edit destroys before it creates. Whatever aborts the recreate after that point leaves the customer with neither pod: in one week 8 of 88 edits died in `container_cleanup` alone (an unkillable container: `tried to kill container, but did not receive an exit event`), 5 more in `encrypted_volume_setup` (DAH-2736), the rental closed `user_initiated` as if the customer had finished, and the dashboard showed only `RetryError[<Future … raised Exception>]` because `retry_ssh_command` let tenacity wrap the real stderr.

Root cause, verified in code on `main`: `docker_service.py` `current_step = "container_cleanup"` → `clean_existing_containers(..., active_container_names=payload.active_container_names)` → `retry_ssh_command(ssh_client, 'docker rm -fv <names>')`; `core/utils.py` `retry_ssh_command` uses `@retry(stop=…, wait=…)` without `reraise`, so callers get `RetryError` and the except block logs `str(e)`.

## Change

Validator only (`neurons/validators`). The edit becomes park → build → swap, with an undo.

- **`_EditSwap`** (new, next to `_InflightCreate`): entered in the same `async with` as the SSH connection, so its exit still has a live session.
  - `park()` — called at the cleanup step only when `local_volume` is set and the container exists on the host: `docker rename pod_<id> pod_<id>__prev` (instant, kills nothing) then `docker stop -t 10 pod_<id>__prev` to free the ports and GPUs. **If the stop fails, the name is renamed back and the edit fails with `failure_step=park_current_container` and the real stderr — nothing was destroyed and the customer's pod is as it was.** This is exactly the wedge that used to arrive after a destructive `rm -fv`.
  - The parked name is added to the `active_container_names` the sweep receives, so the sweep does not treat it as stale.
  - `__aexit__` on success (or on `_CreateCancelledByDelete`): `docker rm -fv pod_<id>__prev`, best effort — a wedged remove is logged and left to the stale-container sweep, which already matches `pod_*` names. The customer's new container is up either way.
  - `__aexit__` on any other exception: `restore()` — `docker rm -fv pod_<id>` (the half-built replacement), `docker rename pod_<id>__prev pod_<id>`, then the container is brought up the way `start_existing_container` does it (SDK start, gocryptfs remount with `allow_init=False`, sshd bootstrap — `_bring_up_existing_container`, bound by the create path to the pod's volume path); logged as "previous container was restored" or, if that fails, as an error. The original exception still propagates and the backend gets its `FailedContainerRequest`.
  - A leftover `pod_<id>__prev` next to the pod (an earlier edit's replacement is the pod now) is removed before the rename. A `pod_<id>__prev` with no `pod_<id>` beside it is the customer's only copy — an edit crashed between park and restore — so `park()` renames it back and parks it again, never removes it.
  - `pod_<id>__prev` is a `pod_*` name, so both sweepers see it: the cycle cleanup (`container_cleanup._get_rented_containers`) protects the `__prev` twin of every rented pod name, and `clean_existing_containers` protects it for every active name a sibling create passes. Once the pod is no longer rented the leftover ages out like any stale `pod_*` container.
- **Real error text.** `retry_ssh_command` now passes `reraise=True`, so every caller — and the `Failed create_container` log/`detail` — sees `[clean_existing_containers] command: … exit_code 1, stderr: …` instead of `RetryError[…]`. The except block also joins `_exception_texts(e)` for any RetryError still raised elsewhere.

Fresh creates are unchanged: nothing is parked, the sweep runs as before. Volumes are untouched on both paths (`clear_volume=False` on edit, as today); the named `volume_<id>` is shared by the old and the new container.

Not in this PR (from the ticket): why a container becomes unkillable (needs executor-side state captured during a live wedge), the delete path's own wedge (68 `Unknown Error delete_container` in the same week — the backend side is lium-platform#64), and DAH-2736's gocryptfs EPERM — with this change that failure now rolls the pod back instead of destroying it, which is the shared target the ticket names.

## How to test

```
cd neurons/validators
pdm run pytest tests/test_dah2740_edit_park_then_swap.py tests/test_deploy_optimizations.py -q    # 6 + 33
pdm run pytest tests/ --strict-markers -q                                                           # full validator suite
```

Covered, with the SSH session mocked and every docker command recorded: an edit renames and stops the current container **before** the sweep and the replacement, passes the parked name to the sweep as protected, and removes only the parked container after success; a failure after parking (`docker run` raising) restores in order — rm replacement, rename back, start — and the `FailedContainerRequest` carries the real error text; an unstoppable container fails the edit at `park_current_container`, renames back, never calls the sweep or `docker run`, and issues no destructive rm; an edit whose container is not on the host behaves as a create; a fresh create parks nothing; `retry_ssh_command` raises the last attempt's exception with the stderr, not `RetryError`.

Docker semantics checked on a real daemon (a Docker-in-Docker pod): `rename` of a running container, `stop`, the replacement binding the same host port, and the rollback `rm` → `rename` → `start` bringing the original back with its port mapping.

## Verification

**Verified by the loop on 7 Sep 2026:** checked on sandbox EC2 (the CI shape, then the #1312 subnet-loop e2e stack on a clean worktree) — branch head `7e395c08` merged onto `main` `07ec64cd` (clean merge), then: pytest: validators (CI env) + ruff format report; e2e: the #1312 subnet-loop gate (protocol, cycle, rental); Validators 1945 passed, 15 skipped, 149 warnings — 2 failed are main's own (root-only sysfs/sshd artefacts, same as pass 1); e2e cycle 5 passed, 0 failed, 0 skipped; e2e protocol 13 passed, 0 failed, 0 skipped; e2e rental 2 passed, 0 failed, 0 skipped. Run time 5 min. Result: PASS. Not proven here: a real chain and real GPU hosts (the #1312 stack runs the executor without a GPU).

### Review round at `de13cb9` (jam6099, 8 Sep 2026)

- `restore()` did a bare `docker start`; park had stopped the container, so the gocryptfs FUSE mount was gone and an encrypted-volume pod came back with ciphertext in the workspace. The start now goes through the same steps as `start_existing_container`: `_restore_mount_and_sshd_after_start` is shared by both; `_EditSwap.bring_up` is set by the create path once it knows the pod's volume path. Test: the restore awaits the bring-up with the container name, pod id and volume path, and no bare `docker start` runs.
- `pod_<id>__prev` was sweeper bait: the cycle cleanup keeps only `rented_data` names and ages from Created, and a sibling create's `clean_existing_containers` reaps anything not in its active list. Both now protect the `__prev` twin of a rented/active pod name (`EDIT_PARKED_SUFFIX` moved to `services/const.py`). `park()` never `rm -fv`s a parked container whose original is gone. Tests: `test_a_parked_container_left_by_a_crashed_edit_is_the_pod_and_is_never_removed`, `test_a_leftover_parked_container_next_to_the_pod_is_removed_before_parking`, `test_the_cycle_cleanup_protects_the_parked_twin_of_every_rented_pod`, `test_a_sibling_create_sweep_keeps_the_parked_twin_of_an_active_pod`.
- `start_existing_container` keeps its order: SDK start first, the SSH session opened only after it succeeded (`test_start_container_sdk_failure_returns_failed_request_without_shell_fallback`).

**Verified by the loop on 8 Sep 2026 at head `de13cb9` (sandbox EC2 c6i.2xlarge, aws_run `20260908T094846Z-gtd2`, `tools/aws_run_jobs/reverify_io.sh` on the unpushed bundle):** the head merged onto `main` `120272c7` (clean) · validators `pdm run pytest tests/`: 2007 passed, 15 skipped, 2 failed — both `main`'s own (`sshd_bootstrap` env tests; `main` fails the same two on the box), `io-validators-new_failures.txt` empty. Result: PASS.

### Fix round at `8f829df` (self-review round r12, 8 Sep 2026)

- `_EditSwap.__aexit__` never changes the create's outcome: the success-path removal and the whole undo run under one `try/except` that logs and returns, so an SSH session that died (the very thing that fails an edit) cannot turn the create's error into the undo's `ChannelOpenError`. Test `test_an_undo_whose_ssh_session_died_keeps_the_creates_own_error`.
- **Cross-PR.** lium-io#1265 (DAH-2780, encryption-downgrade precheck) inserts at the same anchor — after `protected_volume_names.add(local_volume)`, before `current_step = "container_cleanup"`. Order: the precheck reads the container under its original name, so it goes **before** `edit_swap.park()`; whichever lands second re-resolves that block, and #1265's "by now the container is gone" comment stops being true once this merges (the container is parked, not removed). lium-io#1295 (DAH-3004) is stacked on this PR and was rebased onto this head (`104e9e4`); it lands after.

**Verified by the loop on 8 Sep 2026 at head `8f829df` (sandbox EC2 c6i.2xlarge, aws_run `20260908T102218Z-kg2r`, `tools/aws_run_jobs/reverify_io.sh` on the unpushed bundle):** merged onto `main` `9c75530e` (clean) · validators `pdm run pytest tests/`: 2009 passed, 15 skipped, 2 failed — both `main`'s own (`sshd_bootstrap` env tests), `io-validators-new_failures.txt` empty. Result: PASS.

## Notes for reviewers

- **Why not create under a temporary name and swap.** Two containers cannot hold the same host ports, and the ports are the pod's identity to the backend; the old one must be stopped before the new one can bind them. Renaming the old one aside and stopping it is the smallest step that keeps it recoverable; the only window in which the customer is without a running container is the create itself, which is the same window as today minus the destructive `rm`.
- **`docker stop -t 10`** sends SIGTERM and SIGKILL after 10 s — gentler than today's immediate `rm -f` SIGKILL, and a stop that fails is the signal we want (edit refused, pod intact) rather than a corpse.
- **`reraise=True`** changes the exception type callers of `retry_ssh_command` receive from `RetryError` to the wrapped `Exception`; the only code that inspected `RetryError` is `_exception_texts`, which handles both. The message text callers already matched on is unchanged.
- Files: `services/docker_service.py` (+~110, the helper and four short hooks), `core/utils.py` (one decorator argument), new test module.

### Self-review

Verdict `clean` at `eaff9e3` · 15 checks · by subagent-lium-io-1279-r1 · 2026-09-12 14:45Z (tools/pr_self_review.py) · carried from `dbef3e5`: rebase-only push, patch vs `main` unchanged (`git patch-id --stable`), reviewed 2026-09-10 09:01Z

|---|---|---|---|---|
| low | STALE-BODY | `PR body (Details):131` | Property: every sha the body cites is reachable on the branch — `de13cb9`, `8f829df`, `dd4c39b`, `104e9e4` and `7e395c08` (Review round / Fix round / Verified-by-the-loop / Self-review lines) all fail `git merge-base --is-ancestor <sha> HEAD` after the rebase onto `067ec87`; only the 7e395c08 one was recorded last round. | Write each as 'the earlier head' (PR_PROCESS §7) or drop the sha; `--record` replaces the Self-review block anyway. |
| low | STALE-BODY | `PR body (Details):119` | Still not re-rendered from the diff: 'How to test … # 6 + 33' (11 + 33 now, 44 pass locally), 'Files: docker_service.py (+~110 …), core/utils.py, new test module' omits services/const.py and services/container_cleanup.py and the +287 in docker_service.py, 'restores in order — rm replacement, rename back, start' is the bring-up now; the test module docstring (tests/test_dah2740_edit_park_then_swap.py:6) says 'renames the old one back and starts it' for the same reason. | Update the three Details sentences and the module docstring to '… renamed back and brought up (start, gocryptfs remount, sshd)'. |
| low | CORRECTNESS | `neurons/validators/src/services/docker_service.py:526` | Property: one edit at a time per pod — `_InflightCreate` documents that a second create for the same pod can run while the first is in flight, and `park()`'s 'crashed earlier edit' branch reads the sibling's live parked container as a crash: B renames `__prev` back and re-parks it, the `docker run --name pod_<id>` loser's `cleanup_failed_container_creation`/`restore()` `rm -fv pod_<id>` removes the winner's replacement, and the winner's `rm -fv pod_<id>__prev` races the loser's rename-back (the executor Redis lock covers only port allocation, line 1559). Not worse than main, where the sweep had already removed the original. | Before `park()`, refuse the edit at `park_current_container` (or wait on `inflight_creates.wait_until_done`) when `inflight_creates.is_running(payload.pod_id)` reports another create on this pod, with a test. |
| low | CORRECTNESS | `neurons/validators/src/services/docker_service.py:542` | Property: `park()` never removes the customer's only good container — the both-names-exist branch treats `pod_<id>` as the live pod and `rm -fv`s `__prev`; after a `restore()` whose `rm -fv pod_<id>` wedged (the ticket's own 'did not receive an exit event' case) `pod_<id>` is the dead replacement and `__prev` the customer's container, so the retry destroys it (volume survives; overlay does not — the pre-PR baseline). Recorded last round; still on this head. | Inspect `State.Status` of both before removing the leftover and refuse the edit naming both when `pod_<id>` is not running. |
| low | CORRECTNESS | `neurons/validators/src/services/docker_service.py:570` | Property: a pod deleted mid-edit leaves no named volume behind — on `_CreateCancelledByDelete` the delete's own `docker rm` found no `pod_<id>` (it was parked) and its `volume rm volume_<id>` is refused while `__prev` still references it; `__aexit__` then removes `__prev` with `-v` (anonymous volumes only), so the named `volume_<id>` outlives the pod, and both sweepers derive `volume_<id>__prev` from the parked name (container_cleanup.py:537, docker_service.py:2034-2040) so no later sweep frees it either. | On the `_CreateCancelledByDelete` branch also run `docker volume rm volume_<id> 2>/dev/null // true` after the parked container is gone, and strip `EDIT_PARKED_SUFFIX` before deriving the volume name in both sweepers, with a test. |
| low | TEST-GAP | `neurons/validators/tests/test_dah2740_edit_park_then_swap.py:266` | Measured on this head: `test_an_undo_whose_ssh_session_died_keeps_the_creates_own_error` takes 40.02 s — the raising `run` is first hit by `cleanup_failed_container_creation` (docker_service.py:5377) under `retry_ssh_command`'s `wait_fixed(10)` × 5, not by the undo as the comment says; the assertion still discriminates the guard. Recorded last round. | Stub `cleanup_failed_container_creation` with `AsyncMock(return_value=False)` in this test so only the undo's `run` raises. |
| low | STACK-HYGIENE | `neurons/validators/src/services/docker_service.py:1` | `origin/main` is now `6c16a470` (#1300 DAH-3113, #1321 DAH-3199), two commits past the brief's merge-base `067ec87`; `git merge-tree --write-tree origin/main HEAD` is clean and neither commit touches the edit/park path, `clean_existing_containers`, `start_existing_container` or `container_cleanup`. | Push as is (merges clean) or restack; a rebase-only push carries this verdict per PR_PROCESS §7. |

### Verified

Gate: PASS (eaff9e3) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1279)

### Ran it

`tools/aws_verify.sh lium-io --app io-all` on EC2 (20260910T084806Z-17ae, at dbef3e5, the earlier head of this same patch; carried to eaff9e3 unchanged) → validators 2118 passed, 15 skipped, 2 box-only fails (root reads sysfs; live sshd pidfile); executor 137; miners 27 · `ruff check --select F,ASYNC210,ASYNC251` clean · verified on Linux only (validator code, nothing runs on a laptop)

</details>


neurons/validators/src/services/docker_service.py
```diff
@@ -383,6 +383,102 @@ class _CreateCancelledByDelete(Exception):
...
+            return False
+        await self.restore()
+        return False
+
+    async def restore(self) -> None:
+        """Undo the edit: drop the half-built replacement, give the parked container its name back, start it."""
+        q_name, q_parked = shlex.quote(self.container_name), shlex.quote(self.parked_name)
+        await self.ssh_client.run(f"/usr/bin/docker rm -fv {q_name} 2>/dev/null || true")
+        renamed = await self.ssh_client.run(f"/usr/bin/docker rename {q_parked} {q_name}")
+        started = await self.ssh_client.run(f"/usr/bin/docker start {q_name}") if renamed.exit_status == 0 else renamed
```
jam6099: `restore()` only `docker start`s. `start_existing_container` exists because a bare start drops the gocryptfs plaintext mount and sshd. Park already `docker stop`d the old container, so the FUSE mount is gone. The encrypted-volume path this PR says it rolls back then comes up with ciphertext in the workspace. Remount with `allow_init=False` after rename+start. The new tests only assert the three docker verbs.

neurons/validators/src/services/docker_service.py
```diff
@@ -383,6 +383,102 @@ class _CreateCancelledByDelete(Exception):
...
+        return self
+
+    async def park(self) -> str | None:
+        """Rename the running container aside and stop it. None when there is nothing to park."""
+        listed = await self.ssh_client.run(
+            f'/usr/bin/docker ps -a --format "{{{{.Names}}}}" --filter name=^{shlex.quote(self.container_name)}$'
+        )
+        if self.container_name not in (listed.stdout or "").split():
+            return None
+        parked = f"{self.container_name}{EDIT_PARKED_SUFFIX}"
```
jam6099: `pod_<id>__prev` is still a `pod_*` name. Cycle cleanup keeps only names in `rented_data` (`pod_<id>`) and ages from Created, so a rental older than 15 min is stale the moment it is renamed. That check runs every cycle on every validator, on another SSH session, while this create is still in docker run. A sibling create's `clean_existing_containers` reaps it with no grace. `park()` also `rm -fv`s leftover `__prev` and does not look for `__prev` when the original name is gone, so a retry after a crash or a sweep hit deletes the only copy. Protect `*__prev` while the rental exists, or do not use a `pod_*` name.

jam6099 (changes requested):
Park-then-swap is the right shape. Reject until restore remounts gocryptfs and `pod_*__prev` is not sweeper bait.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `de13cb9`: `restore()` now brings the container back the way `start_existing_container` does: SDK start, gocryptfs remount with `allow_init=False`, sshd bootstrap. The steps live in `_restore_mount_and_sshd_after_start`, shared by both; the create path hands the swap a bring-up bound to the pod's volume path. The restore test asserts that call and that no bare `docker start` runs.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `de13cb9`: the name stays `pod_<id>__prev`, and both sweepers protect it while the pod is rented. The cycle cleanup adds the `__prev` twin of every rented pod name. `clean_existing_containers` adds it for every active name a sibling create passes. `park()` no longer removes a parked container when the original is gone: that is the only copy, so it gets its name back and is parked again. Four tests cover the crash retry, the leftover next to the pod, and both sweepers.

surcyf123: Done in `8f829df`: one more from the loop's own review — the swap's exit is wrapped, so an undo on a dead SSH session logs and the create's own error stands (test added). lium-io#1295, stacked on this PR, was rebased onto this head.

jam6099 (approved):
Tested on staging, works as expected.

PR #1297, opened 2026-09-06
DAH-3035 - [P1] validator: a kernel-fault probe after the GPU matmul

surcyf123:
Implements [DAH-3035](https://app.notion.com/p/Validator-a-kernel-fault-probe-after-the-GPU-matmul-so-a-card-that-faults-under-indexed-access-Ble-3d3b8bfdbde9810cb41dda954630561a) · reviewer: @taiberium

**TL;DR** — A rented 1× RTX 4090 failed every Blender render with `Illegal address in CUDA queue` after passing validation. `GpuFaultProbeCheck` runs a random-access kernel probe right after the capability matmul; both flags default off (shadow first).

**What changed**
- `miner_jobs/gpu_fault_probe.py`: standard-library CUDA probe (`ctypes`, PTX) — gather/scatter/chase over random permutations, a pinned copy, NVML before/after
- A fault (`_worker`, `xid_faults`, `nvml_faults`): a `CUresult` once the kernels run, a mismatch, a hung worker, an NVML delta, a hardware Xid on a probed card
- `GPU_FAULT_PROBE_CHECK_ENABLED` reports `GPU_FAULT_PROBE_FAILED` as a warning; `GPU_FAULT_PROBE_ENFORCEMENT_ENABLED` fails the fatal check (`checks/gpu_fault_probe.py`, `config.py`)
- `_cap()` in `checks/gpu_fault_probe.py` bounds every value copied from the report; `test_the_ssh_cap_sits_above_the_probes_own_largest_deadline` proves the 150 s cap

**How to verify**
- `cd neurons/validators && pdm run pytest -q tests/test_gpu_fault_probe_check.py` → 26 passed

**Verified by the loop:** 7 Sep 2026 · sandbox EC2 · validators 1958 passed · Result: PASS bff447d1 · Gate: PASS 3530649

**Ran it:** `python3 -I - < gpu_fault_probe.py` on a healthy rented 1× RTX 4090 → `status ok`, 5.34 s; 4 injected faults each caught · pod run at 95339c6; later commits covered by the 26 tests (26 passed at 3530649, EC2 lcv1), not re-run on a GPU

**Self-review:** clean at 3530649 (15 checks, 2 low)

**Parts:** backend ✓ validator check + probe · migration n/a — no schema · frontend n/a — existing reason codes · CLI/SDK n/a — nothing renter-facing · docs ✓ `.env.template`; lium-docs n/a until DAH-2994 (lium-docs#217) lands · tests ✓ 26 · dashboards n/a — existing panels

**Docs:** `neurons/validators/.env.template` (GPU_FAULT_PROBE_*) here · lium-docs: none yet — the reason-code page (DAH-2994, now in lium-docs#217) is not on main; the five codes go in when it lands

[DAH-3035](https://app.notion.com/p/Validator-a-kernel-fault-probe-after-the-GPU-matmul-so-a-card-that-faults-under-indexed-access-Ble-3d3b8bfdbde9810cb41dda954630561a).

## Origin
A rented 1× RTX 4090 (6 Sep, ~18:20Z) failed every Blender Cycles render with `Illegal address in CUDA queue` — on the OptiX backend and on the CUDA backend — while PyTorch on the same card ran fine. The node had passed validation that cycle and went back into the pool after the rental; the renter paid for the discovery. The validator's only GPU workload, `gpu.validate.capability`, is a cuBLAS matmul: sequential reads and writes of two operands and one result, verified by one number. A memory subsystem that faults under indexed access — the access pattern of a BVH traversal, of an embedding lookup, of any sparse workload — passes it.

## Change
`GpuFaultProbeCheck` (`gpu.validate.fault_probe`, fatal), placed right after `CapabilityCheck` in `build_checks` and `build_dry_run_checks`: same idle, capability-verified population, same active-filler skip (`_get_filler_only_container` reused). It pipes `miner_jobs/gpu_fault_probe.py` to the executor's own interpreter over stdin (`python -I - --seconds 4`, the DAH-2794 delivery: nothing to upload, nothing the miner has to update) and reads one `GPU_FAULT_PROBE_JSON:` line back.

The probe is standard-library Python: the CUDA driver API through `ctypes` (`libcuda.so.1` is in every container the NVIDIA runtime prepares, and the matmul already proves it loads there), kernels as PTX the driver JIT-compiles (`.target sm_50`, 79 ms), NVML through `nvidia-ml-py` when importable (it is, in the executor image). One forked worker per GPU, all concurrently, so a driver crash or hang takes the worker and becomes the verdict rather than a missing report. Per GPU, on a 2 GB working set (four 512 MB `uint32` buffers, or less when the card is small — `min(2 GB, free − 1 GB)`), rounds until 4 s of work:

| step | what runs | verified by |
|---|---|---|
| `k_perm` | a bijective fmix32-style permutation of `[0, n)` built on the device, new seed per round | host replays the same function |
| `k_gather` | `out[i] = in[perm[i]]` — 128 M random reads | `k_count_neq(out, perm) == 0` on the device |
| `k_scatter_add` | `cnt[perm[i]] += 1` — 128 M random atomics | every slot `== 1` |
| `k_scatter` | `inv[perm[i]] = i` — 128 M random writes | `inv[perm[i]] == i` for all i |
| `k_chase` | 32 dependent random loads per thread (4 G loads) | 64 chains replayed on the host |
| `cuMemcpyHtoDAsync` / `DtoHAsync` | 32 MB pinned-memory round trip through both copy engines | byte compare on the host |
| NVML before/after | uncorrected ECC (volatile), remapped rows pending/failure, GPU recovery action | any increase / pending / non-zero |
| `dmesg` before/after | `NVRM: Xid` lines, when the container may read the kernel log (best effort) | new lines whose PCI id is a probed card, whose Xid number is readable and is not an application type |

Any non-zero `CUresult` once the kernels run (`CUDA_ERROR_ILLEGAL_ADDRESS` is what Blender prints as "Illegal address in CUDA queue"; the setup calls up to the first launch are errors, not faults), a mismatch counter, a worker that dies or does not answer within the shared budget of `--seconds` + 30 s (+ 5 s per extra GPU) once its kernels run, an NVML delta, or a new `NVRM: Xid` line on one of the probed cards of a hardware type (application Xids 13/31/43/45, other cards' lines and lines whose Xid number cannot be read are kept as evidence only) is a **fault**. A probe that cannot start (no `libcuda`, `cuInit`, PTX JIT rejected) is an **error** and passes as `GPU_FAULT_PROBE_UNKNOWN` with the reason: inability to measure is not a fault, the same rule `CPU_TRUTH_UNKNOWN` follows.

Flags (both default **False**), the shadow-first pattern of every score-zeroing gate here:
- `GPU_FAULT_PROBE_CHECK_ENABLED` — run the probe and emit the verdict; a fault is `GPU_FAULT_PROBE_FAILED` as a *warning*, `passed=True`, "Shadow observation only: score was NOT changed".
- `GPU_FAULT_PROBE_ENFORCEMENT_ENABLED` — a fault is `GPU_FAULT_PROBE_FAILED` as an *error* and fails the fatal check, like `GPU_VERIFY_FAILED`.

Off, the check emits `GPU_FAULT_PROBE_DISABLED` and runs nothing; every scored verification is byte-for-byte today's. Reason codes: `GPU_FAULT_PROBE_OK`, `GPU_FAULT_PROBE_FAILED`, `GPU_FAULT_PROBE_UNKNOWN`, `GPU_FAULT_PROBE_DISABLED`, `GPU_FAULT_PROBE_SKIPPED_ACTIVE_FILLER`. The event's `what` carries the probe summary (per-device status, rounds, work seconds, JIT ms, faults, NVML after) and, on a fault, the Xid lines.

Cost per idle node per cycle when the check flag is on: ≈ 6 s of wall clock and one SSH session; 2 GB of VRAM and 64 MB of pinned host memory for those seconds; nothing is written to disk.

## Verification
- **Healthy pod, real run** (1× RTX 4090, driver 580.173.02, Python 3.12, `python3 -I -` with the file on stdin, exactly the check's command): `status ok`, **5.34 s in-script, 6.6 s over SSH**, 7 rounds in 4.55 s of work on 2 048 MB, JIT 79 ms (2–6 ms warm), 32 MB copy; same on three repeats. Fault injection on the same card, each caught and named: an out-of-range index (`mask = 4n−1`) → `cuStreamSynchronize -> CUDA_ERROR_INVALID_ADDRESS_SPACE (717)`, `status fault`, exit 1, 1.2 s; a deliberately wrong gather comparison → `gather: 134217727 of 134217728 elements wrong`; a corrupted copy comparison → `async memcpy round trip: data differs`; a 200 000-step chase → `hung: no result after 31s`, worker killed, verdict still printed, exit 1, 31.3 s. With `nvidia-ml-py` installed the NVML block reads `remapped_rows [0,0,0,0]`, `ecc_uncorrected null` (consumer card), `recovery_action null`; without it, `available: false` and the CUDA verdict stands alone. Pod removed after the runs (≈ 15 min, ≈ $0.08).
- `cd neurons/validators && pdm run pytest -q tests/test_gpu_fault_probe_check.py` — 26 tests with a fake runner: disabled runs nothing; the command, stdin source, 150 s cap and `retryable=False` are what the check sends; the cap is proven against the drain budget plus the parent's tail (reap grace, two forked NVML snapshots, two dmesg reads); a new Xid counts only on a probed card and only for hardware types; a hostile report (5 000-char strings, dicts and ints where strings belong) is capped on the fault and the UNKNOWN paths without an exception; the NVML snapshot fork comes back in time when NVML hangs; the exit sequence prints the verdict and closes stdout with a child the driver still holds; ok → `PROBE_OK` with the summary; fault in shadow → warning, passes; fault under enforcement → error, fails the fatal check; an NVML-only fault fails; an SSH timeout is `UNKNOWN` (no verdict — the transport cannot tell a stall from a hung host; a hung card is caught inside the probe by its own budget); `status error` (JIT refused) passes as `UNKNOWN`; no marker line passes as `UNKNOWN` with stdout/stderr tails; a malformed marker line is `UNKNOWN`, not an exception; an active filler skips; the check follows `gpu.validate.capability` in both pipelines; the probe imports only the standard library (so the `python:3.11-slim` executor image runs it); the permutation is a bijection for three sizes × three seeds (a collision would blame the GPU for a host bug); NVML snapshots are paired by UUID and a card gone from the after-snapshot is a fault; a forked NVML child that dies without a snapshot is an unavailable snapshot, not a crash.
- Full validator suite as CI runs it: **1934 passed, 15 skipped** (Python 3.11; `test_sshd_bootstrap_script.py` / `test_port_mapping_dao.py` excluded — Linux-only tooling / DB fixture, identical on `origin/main`). `ruff check` clean on the three new files.

**Verified by the loop on 7 Sep 2026:** checked on sandbox EC2 (the CI shape, then the #1312 subnet-loop e2e stack on a clean worktree) — branch head `bff447d1` merged onto `main` `07ec64cd` (clean merge), then: pytest: validators (CI env) + ruff format report; e2e: the #1312 subnet-loop gate (protocol, cycle, rental); Validators 1958 passed, 15 skipped, 151 warnings — 2 failed are main's own (root-only sysfs/sshd artefacts, same as pass 1); e2e cycle 5 passed, 0 failed, 0 skipped; e2e protocol 13 passed, 0 failed, 0 skipped; e2e rental 2 passed, 0 failed, 0 skipped. Run time 5 min. Result: PASS. Not proven here: a real chain and real GPU hosts (the #1312 stack runs the executor without a GPU).

## Notes for reviewers
- Why a new check and not a bigger matmul: the matmul is a *capacity and identity* proof (fills VRAM, seals a UUID); the failure here is a *pattern* the matmul never produces. The probe deliberately touches only 2 GB — the matmul already sweeps the card — but with 128 M random accesses per kernel across it.
- Why PTX and not a compiled binary or `compute-sanitizer`: the executor image is `python:3.11-slim` plus the driver libraries the runtime mounts; no toolkit, no `nvcc`, no sanitizer. PTX is forward-compatible and JIT-compiled by whatever driver the host has (proved on 580 / Ada). `compute-sanitizer` would also multiply the runtime ~10–100×.
- Trust model: the probe's output is executor-controlled, as the matmul's plaintext `UUID:` line was before sealing. A miner who forges `status ok` gets exactly what he has today (an unprobed node); the probe cannot be worse than its absence and is not an anti-fraud proof. Sealing it like the matmul is possible later (the `.so` is not touched here).
- Coordination with lium-io#1291 (DAH-3011, `FIRST_PASS_FAST_PATH_ENABLED`): independent. The probe reads no first-pass state and runs on both the express first pass and scored cycles when its flag is on — 6 s is the one thing a never-validated node should prove before it is listed. If #1291 lands first nothing here changes; if the team prefers to skip the probe on the first pass, it is one `ctx.config.first_pass` line in `run()`.
- Coordination with lium#120 (DAH-2877, `lium up --verify-gpus`): the CLI counts GPUs over SSH after `up`. The probe file is standard-library and stdin-deliverable, so the CLI can pipe the same file to a pod's `python3` for a renter-side health check; not done here (separate repo, separate ticket).
Docs: neurons/validators/.env.template (GPU_FAULT_PROBE_*) (this PR) · lium-docs#176 (companion)
- `build_dry_run_checks` gets the check too, next to the `CapabilityCheck` it already runs; a dry-run validator is where the shadow verdict is cheapest to read.
- Not proposed: a per-check setting for `--seconds`/working set — the numbers are measured once here and the same everywhere; make them settings when a second card class needs different ones.

**Verified by the loop (7 Sep 2026, sandbox EC2 Linux, run `20260907T143030Z-dsxg`):** head `bff447d1` — ruff check + format on the three files PASS; `tests/test_gpu_fault_probe_check.py` 19 passed (incl. the five-worker drain test: hang-in-kernels, fault, ok, die, hang-in-setup → five verdicts in one 3 s budget); Validators 1958 passed, 15 skipped, 2 failed as root-on-the-box artefacts (`test_scrape_infiniband` unreadable sysfs, `test_sshd_bootstrap_script` adopt path), unrelated to the probe. Result: PASS. Gate: not run.

### Self-review

Verdict `findings` at `3530649` · 15 checks · by subagent-responder-r58-1297 · 2026-09-10 10:17Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | STALE-BODY | `PR body:117` | The Details fault definition still reads 'a new `NVRM: Xid` line on one of the probed cards of a hardware type (application Xids 13/31/43/45, other cards' lines and lines whose Xid number cannot be read are kept as evidence only)' and the dmesg table row 'new lines whose PCI id is a probed card and whose Xid is not an application type'; at this head `xid_faults` (gpu_fault_probe.py:928) also requires a readable Xid number (`xid is not None`), so a probed-card line whose number the regex misses is evidence only, which the body does not say. | In the override JSON append to the evidence-only list 'and lines whose Xid number cannot be read'; the sentence 'Self-review: clean at 62de2f5' is re-recorded by tools/pr_self_review.py --record at this head. |
| low | STACK-HYGIENE | `neurons/validators/src/services/task/pipeline_factory.py:321` | Carried forward from the 62de2f5 verdict: the branch still sits on merge-base 55be0e9, now 25 commits behind origin/main (1a815a00); `git merge-tree --write-tree origin/main HEAD` still merges clean, so nothing blocks, but the reviewed tree is not the tree the merge builds. | Rebase onto origin/main before or after the push (a rebase-only push carries this verdict via --carry), or accept the clean merge. |

Mechanical notes dismissed: `neurons/validators/src/miner_jobs/gpu_fault_probe.py:505` Property (a) verified: `_setup_device` lines 508-589 are, statement for statement and in the same order, old probe_device lines 480-559 (HEAD~1) with `fault=` renamed and `d_in, d_idx, d_out, d_aux = device_buffers` moved to probe_device:621, before `on_phase("kernels")`; the only new statements are the `return ProbeSetup(...)` and the unpack at 610-620. · `neurons/validators/src/miner_jobs/gpu_fault_probe.py:483` Property (a) verified: the kernel phase and the free-up section (624-683) reference report, cuda, context, n, mask, funcs, stream, device_buffers, d_ctr, h_ctr, copy_bytes, h_src, h_dst, pattern, h_sample and the four unpacked buffer names plus the parameter `seconds` and module constants; all 15 are ProbeSetup fields and all are unpacked; `lib`, `module`, `device`, `error_log`, `free`, `total`, `budget`, `t_jit` are used only in setup (old and new) and are correctly not carried. · `neurons/validators/src/miner_jobs/gpu_fault_probe.py:31` Property (b) verified: `from __future__ import annotations` is at line 19 so the ProbeSetup annotations are strings; the class is nonetheless defined at 483, after Cuda (373) and DevPtr (402); dataclasses is a python:3.11 standard-library module and is in the stdlib allow-list of test_probe_source_is_standard_library_only (test:413); `_probe_namespace` (test:426) lifts only FunctionDef/Assign nodes so the ClassDef is skipped, and no test lifts probe_device, _setup_device or ProbeSetup, so nothing exec'd references the skipped class. · `neurons/validators/src/miner_jobs/gpu_fault_probe.py:704` Property (c) verified: `_worker` is unchanged (`report = probe_device(index, seconds, vram_mb, on_phase)`); the same `report` dict object built in _setup_device is carried in ProbeSetup and returned, with keys index, name, pci_bus_id (when readable), vram_free_mb, elements, working_set_mb, jit_ms, copy_mb from setup and rounds, work_s from the kernel phase — the same set as HEAD~1; status/elapsed_s are added by _worker as before. · `neurons/validators/tests/test_gpu_fault_probe_check.py:638` Property (d) verified: setup_calls is still exactly {cuCtxCreate_v2, cuMemGetInfo_v2, cuModuleGetFunction, cuMemAlloc_v2, cuMemAllocHost_v2}; the test now asserts every `cuda.call` in _setup_device (12 sites) carries raise_as_fault=False (stronger than before, which checked only the five), that `seen == setup_calls` over _setup_device, and that probe_device's `cuda.call` names are disjoint from setup_calls (they are: cuMemsetD32Async, cuMemcpyHtoDAsync_v2, cuMemcpyDtoHAsync_v2, cuStreamSynchronize, cuCtxSynchronize, cuMemFree_v2, cuMemFreeHost, cuCtxDestroy_v2); `on_phase("kernels")` is at gpu_fault_probe.py:623 inside probe_device and the literal assertion against PROBE_SOURCE holds; 14 keyword call sites renamed (12 + 2 in _count_worker), `rg '\bfault='` finds none left. · `neurons/validators/tests/test_gpu_fault_probe_check.py:711` Property (e) verified: new_lines is the after-list sliced past len(before) so after["lines"][1:] are the five new lines in order; the garbled line matches the PCI regex on a probed card but `Xid \([^)]*\): (\d+)` fails so xid is None and the new `xid is not None` guard sends it to `other`; faults == [line 1] and other == lines 2..5 in order, so `other == after["lines"][2:]` holds with the appended line; at HEAD~1 this line would have been scored as a hardware fault (None not in SOFTWARE_XIDS), so the test fails without the change. · `neurons/validators/src/services/task/checks/gpu_fault_probe.py:132` Property (f) verified: the one caller at 125 is `self._fault_verdict(`, the def at 132; `rg '\b_fault\b'` over neurons/validators/src and tests returns nothing, so no test or caller references the old name. · `neurons/validators/src/miner_jobs/gpu_fault_probe.py:49` Property (g) verified: ruff 0.16.6 `format --check` reports the three files already formatted and `ruff check` passes; WORKER_REAP_AFTER_KILL_SECONDS has one reader (drain_workers:990) and the two test lifts (test:221, 608) were renamed, `rg WORKER_REAP` finds no old name; the lcv1 tree_hash.txt equals `git rev-parse HEAD^{tree}` (a9e05cd8) and its summary is 26 passed.

### Verified

Gate: PASS (3530649) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1297)

EC2 verify at 3530649: `20260910T101325Z-lcv1` (`tests/test_gpu_fault_probe_check.py` 26 passed; full suites at the previous amend `20260910T100956Z-o83d`: validators 1964 passed + the two box-bound failures every lium-io run on the box shows, executor 145 passed). Gate: re-run at 3530649 after the review nits.

</details>

### DIFF
--- a/neurons/validators/src/miner_jobs/gpu_fault_probe.py
+++ b/neurons/validators/src/miner_jobs/gpu_fault_probe.py
@@ -1,0 +1,793 @@
+"""GPU kernel-fault probe, run on the executor by GpuFaultProbeCheck (`python -I -`, this file on stdin).
+
+Standard library only. The CUDA driver API is reached through ctypes (`libcuda.so.1` is in every
+container the NVIDIA runtime prepares), the kernels are PTX the driver JIT-compiles, NVML is read
+through nvidia-ml-py when it is importable. One forked worker per GPU runs, for a few seconds, the
+access patterns a cuBLAS matmul never exercises: a random permutation built on the device, a gather
+through it, a scatter back (plain stores and atomics), a dependent pointer chase, and a pinned-memory
+async H2D/D2H round-trip. Every result is verified on the device (mismatch counters) and a sample of
+the chase on the host. A fault is any non-zero CUDA return code once the context exists
+(CUDA_ERROR_ILLEGAL_ADDRESS is what Blender prints as "Illegal address in CUDA queue"), a data
+mismatch, a worker that crashes or hangs, or an uncorrected-ECC / remapped-row / recovery-action
+change in NVML across the run. A probe that cannot start (no libcuda, cuInit, PTX JIT) is an error,
+not a fault: the validator does not penalise what it could not measure.
+
+Prints one line `GPU_FAULT_PROBE_JSON: {...}`. Exit 0 ok, 1 fault, 2 could not run.
+"""
+
+from __future__ import annotations
+
+import argparse
+import ctypes
+import json
+import multiprocessing
+import os
+import subprocess
+import sys
+import time
+
+MB = 1024 * 1024
+JSON_MARKER = "GPU_FAULT_PROBE_JSON:"
+BLOCK = 256
+CHASE_STEPS = 32
+CHASE_SAMPLES = 64
+MIN_LOG2_N = 20  # 1 Mi elements = 4 MB per buffer
+MAX_LOG2_N = 28  # 256 Mi elements = 1 GB per buffer
+BUFFERS = 4  # in, idx, out, aux
+VRAM_RESERVE_MB = (
+    1024  # left free on the card: the runtime's own context and whatever else idles there
+)
+WORKER_GRACE_SECONDS = 30  # on top of --seconds: JIT, allocations, copies, host verification
+
+# fmix32-style mixer whose every step is a bijection on [0, 2**k): (x + seed) mod 2**k, odd multiplier
+# mod 2**k, xorshift within k bits. Mirrored in PTX by k_perm; the host replays it to check the chase.
+PERM_MUL = (0x9E3779B1, 0x85EBCA6B, 0xC2B2AE35)
+PERM_SHIFT = (15, 13, 16)
+
+PTX = r"""
+.version 6.0
+.target sm_50
+.address_size 64
+
+// idx[i] = perm(i): the permutation every other kernel gathers and scatters through
+.visible .entry k_perm(.param .u64 p_idx, .param .u32 p_n, .param .u32 p_seed, .param .u32 p_mask)
+{
+    .reg .pred %p<2>;
+    .reg .b32 %r<16>;
+    .reg .b64 %rd<8>;
+    ld.param.u64 %rd1, [p_idx];
+    ld.param.u32 %r1, [p_n];
+    ld.param.u32 %r2, [p_seed];
+    ld.param.u32 %r3, [p_mask];
+    mov.u32 %r4, %ctaid.x;
+    mov.u32 %r5, %ntid.x;
+    mov.u32 %r6, %tid.x;
+    mad.lo.s32 %r7, %r4, %r5, %r6;
+    setp.ge.u32 %p1, %r7, %r1;
+    @%p1 bra L_end;
+    add.u32 %r8, %r7, %r2;
+    and.b32 %r8, %r8, %r3;
+    mov.u32 %r10, 0x9E3779B1;
+    mul.lo.u32 %r8, %r8, %r10;
+    and.b32 %r8, %r8, %r3;
+    shr.u32 %r9, %r8, 15;
+    xor.b32 %r8, %r8, %r9;
+    mov.u32 %r10, 0x85EBCA6B;
+    mul.lo.u32 %r8, %r8, %r10;
+    and.b32 %r8, %r8, %r3;
+    shr.u32 %r9, %r8, 13;
+    xor.b32 %r8, %r8, %r9;
+    mov.u32 %r10, 0xC2B2AE35;
+    mul.lo.u32 %r8, %r8, %r10;
+    and.b32 %r8, %r8, %r3;
+    shr.u32 %r9, %r8, 16;
+    xor.b32 %r8, %r8, %r9;
+    cvta.to.global.u64 %rd2, %rd1;
+    mul.wide.u32 %rd3, %r7, 4;
+    add.s64 %rd4, %rd2, %rd3;
+    st.global.u32 [%rd4], %r8;
+L_end:
+    ret;
+}
+
+// out[i] = i
+.visible .entry k_iota(.param .u64 p_out, .param .u32 p_n)
+{
+    .reg .pred %p<2>;
+    .reg .b32 %r<8>;
+    .reg .b64 %rd<8>;
+    ld.param.u64 %rd1, [p_out];
+    ld.param.u32 %r1, [p_n];
+    mov.u32 %r2, %ctaid.x;
+    mov.u32 %r3, %ntid.x;
+    mov.u32 %r4, %tid.x;
+    mad.lo.s32 %r5, %r2, %r3, %r4;
+    setp.ge.u32 %p1, %r5, %r1;
+    @%p1 bra L_end;
+    cvta.to.global.u64 %rd2, %rd1;
+    mul.wide.u32 %rd3, %r5, 4;
+    add.s64 %rd4, %rd2, %rd3;
+    st.global.u32 [%rd4], %r5;
+L_end:
+    ret;
+}
+
+// out[i] = in[idx[i]]  (random reads)
+.visible .entry k_gather(.param .u64 p_out, .param .u64 p_in, .param .u64 p_idx, .param .u32 p_n)
+{
+    .reg .pred %p<2>;
+    .reg .b32 %r<8>;
+    .reg .b64 %rd<12>;
+    ld.param.u64 %rd1, [p_out];
+    ld.param.u64 %rd2, [p_in];
+    ld.param.u64 %rd3, [p_idx];
+    ld.param.u32 %r1, [p_n];
+    mov.u32 %r2, %ctaid.x;
+    mov.u32 %r3, %ntid.x;
+    mov.u32 %r4, %tid.x;
+    mad.lo.s32 %r5, %r2, %r3, %r4;
+    setp.ge.u32 %p1, %r5, %r1;
+    @%p1 bra L_end;
+    cvta.to.global.u64 %rd4, %rd1;
+    cvta.to.global.u64 %rd5, %rd2;
+    cvta.to.global.u64 %rd6, %rd3;
+    mul.wide.u32 %rd7, %r5, 4;
+    add.s64 %rd8, %rd6, %rd7;
+    ld.global.u32 %r6, [%rd8];
+    mul.wide.u32 %rd9, %r6, 4;
+    add.s64 %rd10, %rd5, %rd9;
+    ld.global.u32 %r7, [%rd10];
+    add.s64 %rd11, %rd4, %rd7;
+    st.global.u32 [%rd11], %r7;
+L_end:
+    ret;
+}
+
+// out[idx[i]] = i  (random writes; out becomes the inverse permutation)
+.visible .entry k_scatter(.param .u64 p_out, .param .u64 p_idx, .param .u32 p_n)
+{
+    .reg .pred %p<2>;
+    .reg .b32 %r<8>;
+    .reg .b64 %rd<12>;
+    ld.param.u64 %rd1, [p_out];
+    ld.param.u64 %rd3, [p_idx];
+    ld.param.u32 %r1, [p_n];
+    mov.u32 %r2, %ctaid.x;
+    mov.u32 %r3, %ntid.x;
+    mov.u32 %r4, %tid.x;
+    mad.lo.s32 %r5, %r2, %r3, %r4;
+    setp.ge.u32 %p1, %r5, %r1;
+    @%p1 bra L_end;
+    cvta.to.global.u64 %rd4, %rd1;
+    cvta.to.global.u64 %rd6, %rd3;
+    mul.wide.u32 %rd7, %r5, 4;
+    add.s64 %rd8, %rd6, %rd7;
+    ld.global.u32 %r6, [%rd8];
+    mul.wide.u32 %rd9, %r6, 4;
+    add.s64 %rd10, %rd4, %rd9;
+    st.global.u32 [%rd10], %r5;
+L_end:
+    ret;
+}
+
+// cnt[idx[i]] += 1  (random atomics; every slot ends at exactly 1)
+.visible .entry k_scatter_add(.param .u64 p_cnt, .param .u64 p_idx, .param .u32 p_n)
+{
+    .reg .pred %p<2>;
+    .reg .b32 %r<8>;
+    .reg .b64 %rd<12>;
+    ld.param.u64 %rd1, [p_cnt];
+    ld.param.u64 %rd3, [p_idx];
+    ld.param.u32 %r1, [p_n];
+    mov.u32 %r2, %ctaid.x;
+    mov.u32 %r3, %ntid.x;
+    mov.u32 %r4, %tid.x;
+    mad.lo.s32 %r5, %r2, %r3, %r4;
+    setp.ge.u32 %p1, %r5, %r1;
+    @%p1 bra L_end;
+    cvta.to.global.u64 %rd4, %rd1;
+    cvta.to.global.u64 %rd6, %rd3;
+    mul.wide.u32 %rd7, %r5, 4;
+    add.s64 %rd8, %rd6, %rd7;
+    ld.global.u32 %r6, [%rd8];
+    mul.wide.u32 %rd9, %r6, 4;
+    add.s64 %rd10, %rd4, %rd9;
+    red.global.add.u32 [%rd10], 1;
+L_end:
+    ret;
+}
+
+// j = i; steps times: j = idx[j]; out[i] = j  (dependent random reads)
+.visible .entry k_chase(.param .u64 p_out, .param .u64 p_idx, .param .u32 p_n, .param .u32 p_steps)
+{
+    .reg .pred %p<3>;
+    .reg .b32 %r<12>;
+    .reg .b64 %rd<12>;
+    ld.param.u64 %rd1, [p_out];
+    ld.param.u64 %rd3, [p_idx];
+    ld.param.u32 %r1, [p_n];
+    ld.param.u32 %r10, [p_steps];
+    mov.u32 %r2, %ctaid.x;
+    mov.u32 %r3, %ntid.x;
+    mov.u32 %r4, %tid.x;
+    mad.lo.s32 %r5, %r2, %r3, %r4;
+    setp.ge.u32 %p1, %r5, %r1;
+    @%p1 bra L_end;
+    cvta.to.global.u64 %rd4, %rd1;
+    cvta.to.global.u64 %rd6, %rd3;
+    mov.u32 %r8, %r5;
+    mov.u32 %r9, 0;
+    setp.ge.u32 %p2, %r9, %r10;
+    @%p2 bra L_store;
+L_loop:
+    mul.wide.u32 %rd7, %r8, 4;
+    add.s64 %rd8, %rd6, %rd7;
+    ld.global.u32 %r8, [%rd8];
+    add.u32 %r9, %r9, 1;
+    setp.lt.u32 %p2, %r9, %r10;
+    @%p2 bra L_loop;
+L_store:
+    mul.wide.u32 %rd9, %r5, 4;
+    add.s64 %rd10, %rd4, %rd9;
+    st.global.u32 [%rd10], %r8;
+L_end:
+    ret;
+}
+
+// ctr += (a[i] != b[i])
+.visible .entry k_count_neq(.param .u64 p_a, .param .u64 p_b, .param .u32 p_n, .param .u64 p_ctr)
+{
+    .reg .pred %p<3>;
+    .reg .b32 %r<8>;
+    .reg .b64 %rd<12>;
+    ld.param.u64 %rd1, [p_a];
+    ld.param.u64 %rd2, [p_b];
+    ld.param.u32 %r1, [p_n];
+    ld.param.u64 %rd3, [p_ctr];
+    mov.u32 %r2, %ctaid.x;
+    mov.u32 %r3, %ntid.x;
+    mov.u32 %r4, %tid.x;
+    mad.lo.s32 %r5, %r2, %r3, %r4;
+    setp.ge.u32 %p1, %r5, %r1;
+    @%p1 bra L_end;
+    cvta.to.global.u64 %rd4, %rd1;
+    cvta.to.global.u64 %rd5, %rd2;
+    mul.wide.u32 %rd7, %r5, 4;
+    add.s64 %rd8, %rd4, %rd7;
+    ld.global.u32 %r6, [%rd8];
+    add.s64 %rd9, %rd5, %rd7;
+    ld.global.u32 %r7, [%rd9];
+    setp.eq.u32 %p2, %r6, %r7;
+    @%p2 bra L_end;
+    cvta.to.global.u64 %rd10, %rd3;
+    red.global.add.u32 [%rd10], 1;
+L_end:
+    ret;
+}
+
+// ctr += (a[i] != v)
+.visible .entry k_count_neq_const(.param .u64 p_a, .param .u32 p_n, .param .u32 p_v, .param .u64 p_ctr)
+{
+    .reg .pred %p<3>;
+    .reg .b32 %r<8>;
+    .reg .b64 %rd<12>;
+    ld.param.u64 %rd1, [p_a];
+    ld.param.u32 %r1, [p_n];
+    ld.param.u32 %r7, [p_v];
+    ld.param.u64 %rd3, [p_ctr];
+    mov.u32 %r2, %ctaid.x;
+    mov.u32 %r3, %ntid.x;
+    mov.u32 %r4, %tid.x;
+    mad.lo.s32 %r5, %r2, %r3, %r4;
+    setp.ge.u32 %p1, %r5, %r1;
+    @%p1 bra L_end;
+    cvta.to.global.u64 %rd4, %rd1;
+    mul.wide.u32 %rd7, %r5, 4;
+    add.s64 %rd8, %rd4, %rd7;
+    ld.global.u32 %r6, [%rd8];
+    setp.eq.u32 %p2, %r6, %r7;
+    @%p2 bra L_end;
+    cvta.to.global.u64 %rd10, %rd3;
+    red.global.add.u32 [%rd10], 1;
+L_end:
+    ret;
+}
+
+// ctr += (inv[idx[i]] != i)
+.visible .entry k_check_inverse(.param .u64 p_inv, .param .u64 p_idx, .param .u32 p_n, .param .u64 p_ctr)
+{
+    .reg .pred %p<3>;
+    .reg .b32 %r<8>;
+    .reg .b64 %rd<12>;
+    ld.param.u64 %rd1, [p_inv];
+    ld.param.u64 %rd2, [p_idx];
+    ld.param.u32 %r1, [p_n];
+    ld.param.u64 %rd3, [p_ctr];
+    mov.u32 %r2, %ctaid.x;
+    mov.u32 %r3, %ntid.x;
+    mov.u32 %r4, %tid.x;
+    mad.lo.s32 %r5, %r2, %r3, %r4;
+    setp.ge.u32 %p1, %r5, %r1;
+    @%p1 bra L_end;
+    cvta.to.global.u64 %rd4, %rd1;
+    cvta.to.global.u64 %rd5, %rd2;
+    mul.wide.u32 %rd7, %r5, 4;
+    add.s64 %rd8, %rd5, %rd7;
+    ld.global.u32 %r6, [%rd8];
+    mul.wide.u32 %rd9, %r6, 4;
+    add.s64 %rd10, %rd4, %rd9;
+    ld.global.u32 %r7, [%rd10];
+    setp.eq.u32 %p2, %r7, %r5;
+    @%p2 bra L_end;
+    cvta.to.global.u64 %rd11, %rd3;
+    red.global.add.u32 [%rd11], 1;
+L_end:
+    ret;
+}
+"""
+
+KERNELS = (
+    "k_perm",
+    "k_iota",
+    "k_gather",
+    "k_scatter",
+    "k_scatter_add",
+    "k_chase",
+    "k_count_neq",
+    "k_count_neq_const",
+    "k_check_inverse",
+)
+
+
+def perm(i: int, seed: int, mask: int) -> int:
+    x = (i + seed) & mask
+    for mul, shift in zip(PERM_MUL, PERM_SHIFT):
+        x = (x * mul) & mask
+        x ^= x >> shift
+    return x
+
+
+class ProbeError(Exception):
+    """The probe could not start on this device (library, cuInit, JIT): reported as an error, not a fault."""
+
+
+class CudaFault(Exception):
+    """A CUDA call failed once the context existed, or a result did not verify: the device is faulty."""
+
+
+class Cuda:
+    """The slice of the driver API the probe uses, through ctypes."""
+
+    def __init__(self) -> None:
+        self.lib = None
+        for name in ("libcuda.so.1", "libcuda.so"):
+            try:
+                self.lib = ctypes.CDLL(name)
+                break
+            except OSError:
+                continue
+        if self.lib is None:
+            raise ProbeError("libcuda.so.1 not found")
+        self.lib.cuGetErrorName.argtypes = [ctypes.c_int, ctypes.POINTER(ctypes.c_char_p)]
+        self.lib.cuGetErrorName.restype = ctypes.c_int
+
+    def error_name(self, code: int) -> str:
+        name = ctypes.c_char_p()
+        if self.lib.cuGetErrorName(code, ctypes.byref(name)) == 0 and name.value:
+            return name.value.decode()
+        return f"CUDA_ERROR_{code}"
+
+    def call(self, fn: str, *args, fault: bool = True) -> None:
+        code = getattr(self.lib, fn)(*args)
+        if code != 0:
+            message = f"{fn} -> {self.error_name(code)} ({code})"
+            raise CudaFault(message) if fault else ProbeError(message)
+
+
+class DevPtr(int):
+    """A CUdeviceptr; distinguishes 64-bit pointer arguments from 32-bit scalars in _launch."""
+
+
+def _launch(cuda: Cuda, func, stream, n: int, *args) -> None:
+    # kernel arguments travel as an array of pointers to their ctypes values, which must outlive the call
+    values = [ctypes.c_uint64(a) if isinstance(a, DevPtr) else ctypes.c_uint32(a) for a in args]
+    params = (ctypes.c_void_p * len(values))(*[ctypes.addressof(v) for v in values])
+    grid = (n + BLOCK - 1) // BLOCK
+    cuda.call("cuLaunchKernel", func, grid, 1, 1, BLOCK, 1, 1, 0, stream, params, None)
+
+
+def _read_counter(cuda: Cuda, d_ctr: int, h_ctr, stream) -> int:
+    cuda.call("cuMemcpyDtoHAsync_v2", h_ctr, ctypes.c_uint64(d_ctr), ctypes.c_size_t(4), stream)
+    cuda.call("cuStreamSynchronize", stream)
+    return ctypes.c_uint32.from_address(h_ctr.value).value
+
+
+def probe_device(index: int, seconds: float, vram_mb: int) -> dict:
+    report: dict = {"index": index}
+    cuda = Cuda()
+    lib = cuda.lib
+    lib.cuLaunchKernel.argtypes = [
+        ctypes.c_void_p,
+        ctypes.c_uint,
+        ctypes.c_uint,
+        ctypes.c_uint,
+        ctypes.c_uint,
+        ctypes.c_uint,
+        ctypes.c_uint,
+        ctypes.c_uint,
+        ctypes.c_void_p,
+        ctypes.POINTER(ctypes.c_void_p),
+        ctypes.POINTER(ctypes.c_void_p),
+    ]
+    lib.cuMemcpyHtoDAsync_v2.argtypes = [
+        ctypes.c_uint64,
+        ctypes.c_void_p,
+        ctypes.c_size_t,
+        ctypes.c_void_p,
+    ]
+    lib.cuMemcpyDtoHAsync_v2.argtypes = [
+        ctypes.c_void_p,
+        ctypes.c_uint64,
+        ctypes.c_size_t,
+        ctypes.c_void_p,
+    ]
+    lib.cuMemsetD32Async.argtypes = [
+        ctypes.c_uint64,
+        ctypes.c_uint,
+        ctypes.c_size_t,
+        ctypes.c_void_p,
+    ]
+    lib.cuMemAlloc_v2.argtypes = [ctypes.POINTER(ctypes.c_uint64), ctypes.c_size_t]
+    lib.cuMemFree_v2.argtypes = [ctypes.c_uint64]
+    lib.cuMemAllocHost_v2.argtypes = [ctypes.POINTER(ctypes.c_void_p), ctypes.c_size_t]
+    lib.cuMemFreeHost.argtypes = [ctypes.c_void_p]
+    lib.cuMemGetInfo_v2.argtypes = [
+        ctypes.POINTER(ctypes.c_size_t),
+        ctypes.POINTER(ctypes.c_size_t),
+    ]
+    lib.cuCtxCreate_v2.argtypes = [ctypes.POINTER(ctypes.c_void_p), ctypes.c_uint, ctypes.c_int]
+    lib.cuCtxDestroy_v2.argtypes = [ctypes.c_void_p]
+    lib.cuStreamCreate.argtypes = [ctypes.POINTER(ctypes.c_void_p), ctypes.c_uint]
+    lib.cuStreamSynchronize.argtypes = [ctypes.c_void_p]
+    lib.cuModuleLoadDataEx.argtypes = [
+        ctypes.POINTER(ctypes.c_void_p),
+        ctypes.c_void_p,
+        ctypes.c_uint,
+        ctypes.POINTER(ctypes.c_int),
+        ctypes.POINTER(ctypes.c_void_p),
+    ]
+    lib.cuModuleGetFunction.argtypes = [
+        ctypes.POINTER(ctypes.c_void_p),
+        ctypes.c_void_p,
+        ctypes.c_char_p,
+    ]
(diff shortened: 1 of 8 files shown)


taiberium (changes requested):
Reviewed with a second independent pass. Both reviewers hit the first three. Every one of them turns a healthy node into a faulty one once enforcement is on.

**Critical**
- `miner_jobs/gpu_fault_probe.py:474,512,522` — `cuCtxCreate_v2`, `cuMemAlloc_v2` and `cuMemAllocHost_v2` use the default `fault=True`, so an out-of-memory card, a busy card or a container memlock cap raises `CudaFault` and is reported as broken hardware. The file's own docstring says a probe that cannot start is an error, not a fault. The VRAM budget is clamped to 0 at `:479` while `:480` still allocates.
- `miner_jobs/gpu_fault_probe.py:619` — a hung enumeration child is never killed. `poll(30)` fails, `join(5)` returns with the child alive, and multiprocessing's exit handler then blocks on an untimed join, so the interpreter never exits, the verdict is never flushed and the check times out into a fault. The worker loop at `:753` does kill; this path does not.
- `checks/gpu_fault_probe.py:72` — `error_type == "timeout"` is the only branch routed to `_fault`, and `runner.py:73` sets that string for any `asyncio.TimeoutError` around `ssh.run`, a stalled channel included. With `retryable=False` a transport stall is scored as a GPU fault, while the sibling no-report path passes as UNKNOWN.
- `miner_jobs/gpu_fault_probe.py:741,757` — one fixed deadline of `--seconds + 30` covers all GPUs at once, and a worker that misses it is written up as `status="fault"`, not as an error. An 8-GPU node under host IO load, forking eight interpreters that each JIT and allocate, is scored as faulty hardware.

**NIT**
- `miner_jobs/gpu_fault_probe.py:681` — a rise in the corrected or uncorrected remapped-row counts is ignored unless `isPending` or `failureOccurred` is set.
- `checks/gpu_fault_probe.py:110` — `_summary` and `_fault` copy executor-controlled `faults`, `nvml_after` and `xid` into the event with no size cap, while `stdout_tail` is capped at 800 chars.
- `checks/gpu_fault_probe.py:106` — `what: dict` and `report: dict` are unparameterised.
- `core/config.py:273` — the comment says "NVML ECC/row-remap deltas", but `remapped_rows` and `recovery_action` are absolute checks.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,793 @@
...
+
+    cuda.call("cuInit", 0, fault=False)
+    device = ctypes.c_int()
+    cuda.call("cuDeviceGet", ctypes.byref(device), index, fault=False)
+    name = ctypes.create_string_buffer(256)
+    lib.cuDeviceGetName(name, 256, device)
+    report["name"] = name.value.decode(errors="replace")
+
+    context = ctypes.c_void_p()
+    cuda.call("cuCtxCreate_v2", ctypes.byref(context), 0, device)
```
taiberium: **Critical:** `cuCtxCreate_v2`, `cuMemAlloc_v2` and `cuMemAllocHost_v2` use the default `fault=True`, so an out-of-memory card, a busy card or a container memlock cap raises `CudaFault` and is reported as broken hardware. The file's own docstring says a probe that cannot start is an error, not a fault. The VRAM budget is also clamped to 0 at `:479` while `:480` still allocates.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,793 @@
...
+        cuda.call("cuInit", 0, fault=False)
+        count = ctypes.c_int()
+        cuda.call("cuDeviceGetCount", ctypes.byref(count), fault=False)
+        conn.send({"count": count.value})
+    except Exception as exc:  # noqa: BLE001
+        conn.send({"error": str(exc)})
+    conn.close()
+
+
+def _device_count(mp) -> int:
```
taiberium, line 730: **Critical:** a hung enumeration child is never killed. `poll(30)` fails, `join(5)` returns with the child alive, and multiprocessing's exit handler then blocks on an untimed join. The interpreter never exits, the verdict is never flushed, and the check times out into a fault. The worker loop at `:753` does kill; this path does not.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,793 @@
...
+
+    workers = []
+    for index in indices:
+        parent_conn, child_conn = mp.Pipe(duplex=False)
+        process = mp.Process(target=_worker, args=(index, args.seconds, args.vram_mb, child_conn))
+        process.start()
+        child_conn.close()
+        workers.append((index, process, parent_conn))
+
+    deadline = time.perf_counter() + args.seconds + WORKER_GRACE_SECONDS
```
taiberium: **Critical:** one fixed deadline of `--seconds + 30` covers all GPUs at once, and a worker that misses it is written up as `status="fault"`, not as an error. An 8-GPU node under host IO load, forking eight interpreters that each JIT and allocate, is scored as faulty hardware.

neurons/validators/src/services/task/checks/gpu_fault_probe.py
```diff
@@ -0,0 +1,163 @@
...
+            command, timeout=PROBE_TIMEOUT_SECONDS, retryable=False, stdin_text=PROBE_SOURCE
+        )
+        report = _parse_report(run.stdout)
+        what = {
+            "executor_uuid": ctx.executor.uuid,
+            "duration_ms": run.duration_ms,
+            "exit_code": run.exit_code,
+        }
+
+        if run.error_type == "timeout":
```
taiberium, line 86: **Critical:** `error_type == "timeout"` is the only branch routed to `_fault`, and `runner.py:73` sets that string for any `asyncio.TimeoutError` around `ssh.run`, a stalled channel included. With `retryable=False` a transport stall is scored as a GPU fault, while the sibling no-report path passes as UNKNOWN.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,793 @@
...
+
+def nvml_faults(before: dict, after: dict) -> list[str]:
+    faults = []
+    for b, a in zip(before.get("gpus", []), after.get("gpus", [])):
+        if a.get("ecc_uncorrected") is not None and b.get("ecc_uncorrected") is not None:
+            if a["ecc_uncorrected"] > b["ecc_uncorrected"]:
+                faults.append(
+                    f"gpu {a['index']}: uncorrected ECC errors {b['ecc_uncorrected']} -> {a['ecc_uncorrected']}"
+                )
+        rows = a.get("remapped_rows")
```
taiberium: **NIT:** a rise in the corrected or uncorrected remapped-row counts is ignored unless `isPending` or `failureOccurred` is set.

neurons/validators/src/services/task/checks/gpu_fault_probe.py
```diff
@@ -0,0 +1,163 @@
...
+        return CheckResult(
+            passed=True,
+            event=render_message(Msg.UNKNOWN, ctx=ctx, check_id=self.check_id, what=what),
+        )
+
+    def _fault(self, ctx: Context, what: dict, *, error: str, report: dict | None) -> CheckResult:
+        enforce = settings.GPU_FAULT_PROBE_ENFORCEMENT_ENABLED
+        what["error"] = error
+        if report is not None:
+            what["xid"] = report.get("xid")
```
taiberium: **NIT:** executor-controlled `faults`, `nvml_after` and `xid` are copied into the event with no size cap, while `stdout_tail` is capped at 800 chars.

neurons/validators/src/services/task/checks/gpu_fault_probe.py
```diff
@@ -0,0 +1,163 @@
...
+            return self._fault(
+                ctx, what, error="; ".join(report.get("faults") or []) or "fault", report=report
+            )
+        what["error"] = report.get("error") or f"probe status {status!r}"
+        return CheckResult(
+            passed=True,
+            event=render_message(Msg.UNKNOWN, ctx=ctx, check_id=self.check_id, what=what),
+        )
+
+    def _fault(self, ctx: Context, what: dict, *, error: str, report: dict | None) -> CheckResult:
```
taiberium: **NIT:** `what: dict` and `report: dict` are unparameterised.

taiberium (commented):
Findings inline.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `c548166`: `cuCtxCreate_v2`, `cuMemGetInfo_v2`, `cuStreamCreate`, `cuMemAlloc_v2` and `cuMemAllocHost_v2` are `fault=False` now — a refused context, an out-of-memory or busy card, a memlock cap on pinned memory all raise `ProbeError` and come back as `status=error`; and a budget below the smallest working set (16 MB) is a `ProbeError` with the free/reserved figures instead of an allocation against 0. An AST test asserts every one of those calls in `probe_device` carries `fault=False`.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `c548166`: `_device_count` kills the enumeration child that is still alive after `join(5)` (and joins it again), so multiprocessing's exit handler has nothing to block on and the `ProbeError("device enumeration hung")` verdict is printed.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `c548166`: Two changes. The worker reports the phase it reached (`{"phase": "kernels"}` over the pipe once setup — JIT, allocations, pinned copies — is done), and the parent classifies by it: a worker that hangs or dies **in setup** is `status=error` (the host, the driver or the container's limits kept it from measuring), one that does so **once its kernels run** is a fault. And the shared deadline grows by `WORKER_GRACE_PER_GPU_SECONDS` (5 s) per extra worker, so eight interpreters forking, JIT-compiling and allocating at once get 30 + 35 s on top of `--seconds`, not 30.

neurons/validators/src/services/task/checks/gpu_fault_probe.py
surcyf123: Done in `c548166`: Agreed — the runner's "timeout" cannot tell a stalled channel from a hung host, and a hung card is already reported by the probe's own per-worker deadline as a verdict. The SSH timeout is now `UNKNOWN` (passed, with the stdout/stderr tails) like the other no-verdict paths, and `PROBE_TIMEOUT_SECONDS` is 120, above the probe's largest deadline (8 GPUs: 4 + 30 + 35 s) — a test pins that ordering so the cap can never fire before the probe's own.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `c548166`: `nvml_faults` compares `remapped_rows[0]` (corrected) and `[1]` (uncorrected) before/after and flags a rise as `remapped rows corrected a -> b, uncorrected c -> d`, independent of `isPending`/`failureOccurred` (which keep their own line). Test with the four shapes.

neurons/validators/src/services/task/checks/gpu_fault_probe.py
surcyf123: Done in `c548166`: Bounded before they enter the event: `faults` → at most 16 lines of 300 chars, `nvml_after` → at most 16 GPUs with only `index/uuid/ecc_uncorrected/remapped_rows/recovery_action` (strings cut at 300), `xid` → `available/count` and the last 5 lines cut at 200, the error headline cut too. A test feeds a 100 × 5 000-char report and checks the caps and the event size.

neurons/validators/src/services/task/checks/gpu_fault_probe.py
surcyf123: Done in `c548166`: `dict[str, Any]` on `what`, `report`, `_parse_report`, `_summary` and the two new cap helpers.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,841 @@
...
+        parent_conn, child_conn = mp.Pipe(duplex=False)
+        process = mp.Process(target=_worker, args=(index, args.seconds, args.vram_mb, child_conn))
+        process.start()
+        child_conn.close()
+        workers.append((index, process, parent_conn))
+
+    # one deadline for the concurrent workers, with grace that grows with their number: eight interpreters
+    # forking, JIT-compiling and allocating at once on a loaded host take longer than one
+    grace = WORKER_GRACE_SECONDS + WORKER_GRACE_PER_GPU_SECONDS * (len(workers) - 1)
+    deadline = time.perf_counter() + args.seconds + grace
```
taiberium: **Critical:** one shared deadline, drained worker by worker. Once it passes, the `while` at `:786` runs zero iterations, `report` stays `None`, `phase` stays `"setup"`, and `:810` writes "worker died in setup with exit code 0". That is bad for a healthy worker, but the real problem is the other direction: a genuine `status="fault"` from a later GPU is discarded the same way. A hang on GPU 0 masks a real fault on GPU 3 and the whole node comes back UNKNOWN. `deadline` appears only at `:782`, `:786` and `:799` — there is no per-worker budget anywhere. Give each worker its own deadline, or drain the polls concurrently.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,841 @@
...
+            if (
+                rows_before
+                and len(rows_before) >= 2
+                and (rows[0] > rows_before[0] or rows[1] > rows_before[1])
+            ):
+                faults.append(
+                    f"gpu {a['index']}: remapped rows corrected {rows_before[0]} -> {rows[0]}, "
+                    f"uncorrected {rows_before[1]} -> {rows[1]}"
+                )
+            if rows[2] or rows[3]:
```
taiberium: **NIT:** `rows[0]` and `rows[1]` are compared against the before-snapshot, but `rows[2] or rows[3]` is tested on the after-snapshot alone, and `rows_before[2]`/`[3]` are never read. `isPending` stays set until a GPU reset, so a card that was already pending a remap before the probe faults on every cycle. The test at `tests/test_gpu_fault_probe_check.py:390` only covers the 0-to-1 transition, so this case is uncovered.

neurons/validators/src/services/task/checks/gpu_fault_probe.py
```diff
@@ -0,0 +1,229 @@
...
+from ..pipeline import CheckResult, Context
+from .capability import _get_filler_only_container
+
+# The probe runs on the executor's own interpreter, standard library only, and ships over stdin the way
+# the scrape source does (DAH-2794): nothing to upload, nothing the miner has to update.
+PROBE_SOURCE_PATH = Path(__file__).resolve().parents[3] / "miner_jobs" / "gpu_fault_probe.py"
+PROBE_SOURCE = PROBE_SOURCE_PATH.read_text()
+PROBE_JSON_MARKER = "GPU_FAULT_PROBE_JSON:"
+# Measured on a healthy 1x RTX 4090 (driver 580): 5.3 s in-script, 6.6 s over SSH, 7 rounds on a 2 GB
+# working set. A hung card is caught INSIDE the probe: every worker has its own deadline (--seconds + 30 s
```
taiberium: **NIT:** stale twice over — it says "every worker has its own deadline" (there is one shared one) and that a hang "is reported as a fault", which the setup/kernels split made untrue: a setup hang is `error`. The same wording is repeated in `tests/test_gpu_fault_probe_check.py:200`.

taiberium (changes requested):
Everything from the last round is genuinely fixed — setup failures raise `ProbeError`, the hung enumeration child is killed, grace scales per GPU, an SSH stall passes as UNKNOWN, and the event fields are capped. Thank you.

One thing I under-called last time turns out to be worse than a NIT, so I am holding on it. Details inline.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `bff447d`: `drain_workers()` polls every pipe together (`multiprocessing.connection.wait`) until the one deadline, so each worker has the whole budget and a report already sent is read; test: hang-in-kernels + fault + ok + die + hang-in-setup → five right verdicts in one budget.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `bff447d`: `isPending`/`failureOccurred` fault only on the 0 → 1 transition against the before-snapshot; test covers 1 → 1 (quiet) and 1 → failed.

neurons/validators/src/services/task/checks/gpu_fault_probe.py
surcyf123: Done in `bff447d`: both comments now say one shared budget drained concurrently, setup hang = error, kernel hang = fault.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,979 @@
...
+    process.join(1)
+    if process.is_alive():
+        process.kill()
+        process.join(5)
+    return snapshot
+
+
+def nvml_faults(before: dict, after: dict) -> list[str]:
+    faults = []
+    for b, a in zip(before.get("gpus", []), after.get("gpus", [])):
```
taiberium: NIT: This pairs the two NVML snapshots by position, so a card that disappears mid-run lines up with the wrong one. Match on `uuid`.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,979 @@
...
+
+def nvml_snapshot_forked(mp) -> dict:
+    # in its own fork with a deadline: nvmlInit and the per-GPU reads block on a card that has wedged the
+    # driver, and this parent must always print its verdict
+    parent_conn, child_conn = mp.Pipe(duplex=False)
+    process = mp.Process(target=_nvml_worker, args=(child_conn,))
+    process.start()
+    child_conn.close()
+    snapshot = (
+        parent_conn.recv()
```
taiberium: NIT: If the forked child dies, `recv()` raises EOFError and the probe prints no verdict. Catch it, as `drain_workers` does.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,979 @@
...
+    cuda.call("cuMemcpyDtoHAsync_v2", h_ctr, ctypes.c_uint64(d_ctr), ctypes.c_size_t(4), stream)
+    cuda.call("cuStreamSynchronize", stream)
+    return ctypes.c_uint32.from_address(h_ctr.value).value
+
+
+def probe_device(index: int, seconds: float, vram_mb: int, on_phase=None) -> dict:
+    report: dict = {"index": index}
+    cuda = Cuda()
+    lib = cuda.lib
+    lib.cuLaunchKernel.argtypes = [
```
taiberium, line 423: NIT: `probe_device` is 197 lines. Move this argtypes block into a `_declare_driver_signatures(lib)` helper.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,979 @@
...
+            "index": index,
+            "status": "fault" if worker["phase"] == "kernels" else "error",
+            "error": describe(worker["phase"]),
+        }
+
+    def died(index: int) -> None:
+        code = _exit_code(pending[index]["process"])
+        no_verdict(index, lambda phase: f"worker died in {phase} with exit code {code}")
+
+    def sweep(timeout: float) -> None:
```
taiberium: NIT: `sweep` does not say what it sweeps. Rename it `poll_worker_pipes`.

taiberium (approved):

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `62de2f5`: snapshots are paired by UUID (index when a binding has none); a card gone after the run is a fault.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `62de2f5`: EOFError from a dead NVML child is an unavailable snapshot; the same guard is on the device-count fork.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `62de2f5`: the argtypes block, PCI bus id included, lives in `_declare_driver_signatures(lib)`.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `62de2f5`: renamed to `poll_worker_pipes`.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,1008 @@
...
+    """
+    if not (before.get("available") and after.get("available")):
+        return [], []
+    new_lines = (after.get("lines") or [])[len(before.get("lines") or []) :]
+    faults, other = [], []
+    for line in new_lines:
+        pci = re.search(r"Xid \(PCI:([0-9a-fA-F:.]+)\)", line)
+        code = re.search(r"Xid \([^)]*\): (\d+)", line)
+        xid = int(code.group(1)) if code else None
+        if pci and pci_key(pci.group(1)) in bus_ids and xid not in SOFTWARE_XIDS:
```
taiberium: NIT: when the regex misses the Xid number, `xid` is `None`, and `None not in SOFTWARE_XIDS` scores a hardware fault. Add `xid is not None and`.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,1008 @@
...
+    lib.cuModuleGetFunction.argtypes = [
+        ctypes.POINTER(ctypes.c_void_p),
+        ctypes.c_void_p,
+        ctypes.c_char_p,
+    ]
+    lib.cuDeviceGetName.argtypes = [ctypes.c_char_p, ctypes.c_int, ctypes.c_int]
+    lib.cuDeviceGetPCIBusId.argtypes = [ctypes.c_char_p, ctypes.c_int, ctypes.c_int]
+
+
+def probe_device(index: int, seconds: float, vram_mb: int, on_phase=None) -> dict:
```
taiberium: NIT: `probe_device` is 141 lines with a seam at `on_phase("kernels")`. Move the allocation phase into its own function.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,1008 @@
...
+        self.lib.cuGetErrorName.argtypes = [ctypes.c_int, ctypes.POINTER(ctypes.c_char_p)]
+        self.lib.cuGetErrorName.restype = ctypes.c_int
+
+    def error_name(self, code: int) -> str:
+        name = ctypes.c_char_p()
+        if self.lib.cuGetErrorName(code, ctypes.byref(name)) == 0 and name.value:
+            return name.value.decode()
+        return f"CUDA_ERROR_{code}"
+
+    def call(self, fn: str, *args, fault: bool = True) -> None:
```
taiberium: NIT: rename `fault: bool = True` to `raise_as_fault`. All 14 call sites already pass it by keyword.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
```diff
@@ -0,0 +1,1008 @@
...
+BUFFERS = 4  # in, idx, out, aux
+VRAM_RESERVE_MB = (
+    1024  # left free on the card: the runtime's own context and whatever else idles there
+)
+WORKER_GRACE_SECONDS = 30  # on top of --seconds: JIT, allocations, copies, host verification
+WORKER_GRACE_PER_GPU_SECONDS = (
+    5  # more of it per extra worker: they fork, JIT and allocate at the same time
+)
+NVML_GRACE_SECONDS = 10  # an NVML snapshot runs in its own fork: nvmlInit hangs on a wedged card
+WORKER_REAP_SECONDS = 5  # one shared wait for killed workers to be reaped, whatever their number
```
taiberium: NIT: the name hides that this is the wait after the kill. Rename it `WORKER_REAP_AFTER_KILL_SECONDS`.

neurons/validators/src/services/task/checks/gpu_fault_probe.py
```diff
@@ -0,0 +1,255 @@
...
+                else ""
+            )
+            return self._fault(ctx, what, error=error or "fault", report=report)
+        what["error"] = _cap(report.get("error")) or f"probe status {_cap(status)!r}"
+        return CheckResult(
+            passed=True,
+            event=render_message(Msg.UNKNOWN, ctx=ctx, check_id=self.check_id, what=what),
+        )
+
+    def _fault(
```
taiberium: NIT: `_fault` returns `passed=True` in shadow mode, the shipping default. Rename it `_fault_verdict`.

taiberium (approved):

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `3530649`: `xid is not None and xid not in SOFTWARE_XIDS`. A probed-card line whose number the regex misses now lands in the evidence list, not in the faults. Test line added: such a line used to score as hardware.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `3530649`: renamed to `raise_as_fault` at the definition and the 14 keyword call sites.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `3530649`: renamed to `WORKER_REAP_AFTER_KILL_SECONDS` in the probe and the two tests that lift it.

neurons/validators/src/services/task/checks/gpu_fault_probe.py
surcyf123: Done in `3530649`: renamed to `_fault_verdict`. One commit on top of your approval, so a re-approval click is needed. EC2: the 26 probe tests pass at this head; the full validator suite passed at the previous amend, save the two box-bound failures every run on that box shows.

neurons/validators/src/miner_jobs/gpu_fault_probe.py
surcyf123: Done in `3530649`: `_setup_device(index, vram_mb)` now holds the allocation phase and returns a `ProbeSetup` dataclass. `probe_device` unpacks it, calls on_phase(kernels) and runs the kernels and the free-up as before, statement for statement. The AST test walks both: every call in the setup function is raise_as_fault=False, and the kernel phase makes none of the five setup calls.

taiberium (approved):

taiberium (commented):
Second pass. Both flags are new in this PR and default to `false`, so no deployment can already set them and every scored verification is unchanged today — that is why the first NIT is not a blocker. It does have to be fixed before the enforcement flag is ever turned on.

NIT (fix before enabling enforcement): `src/miner_jobs/gpu_fault_probe.py:831` — `nvml_faults` has no `available` guard, unlike `xid_faults:916`. When the after-snapshot comes back `{"available": false, ...}` — the fork died, `nvmlInit` failed, or the 10 s poll expired — `after_gpus` is empty and every card is written up as "missing from the NVML snapshot after the run". Eight healthy GPUs in, eight fabricated faults out, `status` is `fault` with zero CUDA errors. `test_a_dead_nvml_child_is_an_unavailable_snapshot_not_a_crash` builds exactly that value and no test feeds it to `nvml_faults`.

NIT: `tests/test_gpu_fault_probe_check.py:213` — the deadline test uses 8 GPUs, but `services/const.py:106` allows 14. At 14 the same arithmetic gives 151 s against the 150 s cap, before the `_device_count` fork adds more. It fails open, so no node loses score, but the largest nodes would never get probed and would spend 150 s of SSH per cycle.

NIT: `src/services/task/checks/gpu_fault_probe.py:18` — `PROBE_SOURCE` is read at import of `services.task.checks`. `Dockerfile.preflight` copies `src/services` without `src/miner_jobs`, so the first preflight import of this package dies with `FileNotFoundError`. A lazy read removes the trap.

PR #1298, opened 2026-09-07
DAH-1932 - [P1] a first-seen executor UUID gets one cycle before stale cleanup

surcyf123:
Fixes [DAH-1932](https://app.notion.com/p/Fix-case-node-re-register-failed-customer-container-341b8bfdbde9819aa39fccb3798d18e3) · reviewer: @jam6099

**TL;DR** — A re-added executor registers under a new UUID for the same IP:port; the validator's first cycle sees an empty `rented_data` for it and the stale-container cleanup removes the paying tenant's pod. A UUID this validator has never cleaned now gets one cycle of grace (a Redis set). It is a bridge: the backend never remaps on a zero-score first publish, so closing the race is open on the thread.

**What changed**
- `StaleContainerCleanupCheck`: `SADD cleanup_seen_executors <uuid>` first; 1 → skip removal this cycle, 0 → cleanup as before; Redis trouble counts as first sight (`stale_container_cleanup.py`)
- `CLEANUP_SEEN_EXECUTORS_SET`; `RedisService.sadd` returns Redis's added-count instead of discarding it (`redis_service.py`)
- Tests: +5 in `test_stale_container_cleanup_check.py` — first sight skips, second cleans, known UUID after a restart cleans, Redis trouble stays non-fatal, `sadd` returns 1 then 0

**Risks**
- None beyond the diff: no migration, no pod restart, no config change, no data rewrite.

**How to verify**
- `cd neurons/validators && pdm run pytest tests/test_stale_container_cleanup_check.py -q` → 14 passed
- `pdm run pytest tests/ -q` → the validators suite; two failures are the warm box's baseline (root reads a chmod-000 sysfs tree; sshd adopt path), the same at the earlier head

**Verified by the loop:** 7 Sep 2026 · sandbox EC2 · validators 1943 passed · Result: PASS 07ec64cd · Gate: PASS 447f522

**Ran it:** `pdm run pytest tests/ -q` → 2307 passed, 15 skipped, 2 known box env failures (root/sysfs, sshd adopt) · warm EC2 20260914T035456Z-gnme at 447f522

**Self-review:** clean at 447f522 (17 checks, 5 fixed in body)

**Parts:** backend n/a — validator neuron only · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — no client · docs n/a — validator internals, nothing a provider sets · tests ✓ 5 · dashboards n/a — no metric

**Docs:** none needed because only validator internals change; nothing a provider sees or sets.

[DAH-1932](https://app.notion.com/p/Fix-case-node-re-register-failed-customer-container-341b8bfdbde9819aa39fccb3798d18e3).
## Origin

Notion DAH-1932 (Oxygen, "Fix case node re-register → failed customer container"), section **P0 — skip cleanup for first-seen executor UUIDs**, status note of 2026-07-20: "the core race is live and unprotected — a subnet_id rotation mid-rental deterministically kills the pod container in the first validation cycle of the new UUID." The ticket names the fix shape; this PR is that fix and nothing else (P1/P2 of the ticket are separate work).

## Problem

A miner re-adds an executor and it registers under a new subnet UUID (B) for the same IP:port while a customer's pod is still running on it. In the validator's first cycle for B:

1. `rented_data` is fetched once per cycle and keyed by UUID — empty for B; the tenant is filed under A.
2. `StaleContainerCleanupCheck` (early in the pipeline, before the port checks — DAH-2164) removes every `pod_*`/filler container older than 15 minutes that is not in `rented_data` → the customer's container is removed.
3. The backend's A→B remap (`ExecutorDao.upsert`, IP:port fallback) happens only after a spec publish with a positive score — and the first cycle for B publishes 0 (the tenant's pod holds the port range, so `PortCountCheck` is fatal), so the remap does not run after cycle 1 either (jam6099's review).

Live cleanup has been on since PR #991 (`CONTAINER_CLEANUP_DRY_RUN=False`); there is no grace period in the code today.

## Change

`neurons/validators/src/services/task/checks/stale_container_cleanup.py` — before calling `container_cleanup.cleanup(...)`, the check does `SADD cleanup_seen_executors <executor uuid>` in the validator's Redis. `1` (never seen) → container removal is skipped for this cycle and the event carries `first_sight_grace: true`; `0` → cleanup runs exactly as before. The DPHN cache reclaim and the download-temporary sweep (DAH-2475 / DAH-2805) still run on a first sight — they do not depend on rented state. A Redis error is treated as first sight and logged: skipping one GC cycle costs one cycle of delayed GC, removing a live tenant's container cannot be undone. The check stays `fatal = False`.

`neurons/validators/src/services/redis_service.py` — `CLEANUP_SEEN_EXECUTORS_SET` constant; `RedisService.sadd` now returns Redis's added-count (it discarded it before; the callers in `compute_client.py` and `attestation_service.py` ignore the return value). Review round (`a9ecf9f`): the comment in the check says the grace is a bridge, and `test_redis_service_sadd_returns_one_only_for_a_new_member` pins the return against a fake Redis.

Why Redis and not the check instance: the check is a per-process singleton, so an in-memory set would give every executor a fresh grace cycle on each validator restart/deploy; the seen-set survives restarts. It grows by one member per UUID the miner ever presented (no TTL, as the ticket's rejected-alternatives note prefers over timestamps; the same miner-supplied `uuid` already goes into `TDX_ATTESTED_EXECUTOR_SET` and `EXPRESS_LANE_VALIDATED_SET` the same way). First cycle after deploy: every UUID is unseen, so the grace applies fleet-wide once (one cycle of no removal, orphans not handed to `PortCountCheck` that cycle), then only to genuinely new UUIDs.

Tests, `neurons/validators/tests/test_stale_container_cleanup_check.py` (+5 — the fifth, from the review round, pins `RedisService.sadd`'s return; existing ones now build the context with a `SeenExecutorsRedis` that already knows every UUID, so they are unchanged in meaning): first sight skips removal but keeps the sweeps and records the UUID; the second visit of the same UUID cleans up; a *new check instance* does not re-grant grace (Redis, not memory); a Redis error skips removal and stays non-fatal.

## Verification

```
cd neurons/validators && pdm run pytest tests/test_stale_container_cleanup_check.py tests/test_pipeline_default_scenarios.py tests/test_container_cleanup.py -q
# 37 passed (14 + 2 + 21) at 447f522 (warm EC2 20260914T035456Z-gnme: validators 2307 passed, 15 skipped, 2 box-env failures)
pdm run pytest tests/ -q --strict-markers      # whole validator suite, CI env vars, python 3.11, fresh pdm install on Linux
# 1943 passed, 15 skipped, 2 failed: test_scrape_infiniband::test_an_unreadable_sysfs_tree_is_not_reported_as_no_devices
#   (the run was root, which can read a chmod-000 tree) and test_sshd_bootstrap_script::test_fallback_adopt_after_bind_race_…
#   (passes on re-run); neither touches the cleanup path
```

**Verified by the loop on 7 Sep 2026:** checked on sandbox EC2 (the CI shape, then the #1312 subnet-loop e2e stack on a clean worktree) — the earlier head merged onto `main` `07ec64cd` (clean merge), then: pytest: validators (CI env) + ruff format report; e2e: the #1312 subnet-loop gate (protocol, cycle, rental); Validators 1943 passed, 15 skipped, 149 warnings — 2 failed are main's own (root-only sysfs/sshd artefacts, same as pass 1); e2e cycle 5 passed, 0 failed, 0 skipped; e2e protocol 13 passed, 0 failed, 0 skipped; e2e rental 2 passed, 0 failed, 0 skipped. Run time 5 min. Result: PASS. Not proven here: a real chain and real GPU hosts (the #1312 stack runs the executor without a GPU).

## Notes for reviewers

- Exactly the "simplest sufficient fix" the ticket describes (seen-UUIDs set, skip for one cycle); `first_seen_at` timestamps and keying `rented_data` by machine address were rejected there.
- Scope: the grace is a bridge, not the close of the kill race. jam6099 showed on the PR that cycle 1 publishes a zero score, so the backend never remaps A→B and cycle 2 still removes the pod; the two ways to close it (the validator aliases the first-seen UUID to the tenant by address where the context is built, or the backend remaps on the zero-score publish) are on the thread, waiting for his pick — no design change in this round. P1 (stale-id command re-resolve for deploy/redeploy/undeploy with a stale subnet_id) and P2 (force-finish after prolonged inactivity) are backend/orchestration work and stay on the ticket.
- Until the remap question above is settled, the second cycle still removes the container in the re-register case; the ticket's own fallback is the machine-address keying.

---
Opened by the Lium automation account; commits are anonymous by policy. Ticket: DAH-1932.

### Self-review

Verdict `clean` at `447f522` · 17 checks · by subagent-drain-r133-fresh · 2026-09-14 04:05Z (tools/pr_self_review.py)

Fixed in the body: 5 body-only finding(s) answered by re-rendering the body, no code change, no new verdict (PR_PROCESS §7). `neurons/validators/src/services/task/checks/stale_container_cleanup.py:77` CROSS-PR (medium) — The body's Cross-PR paragraph says lium-io#1302 lands after this PR and restacks onto it, but #1302 is on main (696bdeaf8) and this head is rebased on it; the open PRs that rewrite the same two hunks (the cleanup() call… → body-only (Details 'Cross-PR'): rewritten by tools/pr_body_edit.py on 2026-09-14 04:05Z — #1302 merged first (696bdea), this head is rebased on it; #1351 and #1345 named as the open PRs on the same two hunks. No code changed. · `neurons/validators/tests/test_stale_container_cleanup_check.py:204` STALE-BODY (low) — Test counts predate the rebase: the file now has 14 tests (9 on main incl. #1302's test_unremovable_orphan_is_recorded_in_state_for_the_port_check, +5 here), so 'How to verify → 13 passed', 'Ran it → 13 passed' and the… → body-only: How to verify / Ran it / Verification now say 14 passed and 37 (14 + 2 + 21) at 447f522 from the warm EC2 run 20260914T035456Z-gnme; the Self-review line is this fresh verdict, not a carry (2026-09-14 04:05Z). · `neurons/validators/tests/test_pipeline_default_scenarios.py:365` STALE-BODY (low) — The earlier self-review table carried in the body has a TEST-GAP row saying both pipeline scenarios run this check and hit the Redis-error branch because DummyRedisService has no sadd; the scenarios' explicit check list… → body-only: the earlier self-review table (with the TEST-GAP row about DummyRedisService) is replaced by this verdict's block on --record (2026-09-14 04:05Z). · `neurons/validators/src/services/task/checks/stale_container_cleanup.py:69` CORRECTNESS (low) — Rollout behaviour the body does not state: on the first cycle after this deploys, every executor UUID in the fleet is unseen, so the whole fleet gets one cycle with first_sight_grace=true and no container removal, and a… → body-only (Details 'Change'): one sentence added on 2026-09-14 04:05Z — first cycle after deploy the grace applies fleet-wide once, then only to new UUIDs. Accepted behaviour, no code changed. · `neurons/validators/src/services/task/checks/stale_container_cleanup.py:133` SECURITY (low) — ctx.executor.uuid is a miner-supplied string (ExecutorSSHInfo.uuid: str in datura/requests/miner_requests.py, no length or format check) written into a Redis set with no TTL, so the body's '36-byte member per UUID' is a… → body-only (Details 'Change'): '36-byte member' → 'one member per UUID the miner ever presented', with the TDX_ATTESTED_EXECUTOR_SET / EXPRESS_LANE_VALIDATED_SET precedent named (2026-09-14 04:05Z). No code changed.

Mechanical notes dismissed: `neurons/validators/src/services/task/checks/stale_container_cleanup.py:134` The suppression carries its reason on the same line ('best-effort check, never fatal'), BLE001 is not in the repo's ruff select (pyproject.toml: E, F, I, UP) so the tag is inert, and the sibling checks custom_build_orphan_sweep.py and gpu_usage.py use the identical `except Exception as exc:  # noqa: BLE001` form.

### Verified

Gate: PASS (447f522) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1298)

**Cross-PR:** lium-io#1302 (DAH-2991) merged first (`696bdea`); this head is rebased on it — `container_cleanup.cleanup(...)` returns three values, the first-sight branch sets all three to empty, and #1302's `updates` (orphans handed to `PortCountCheck`) is kept. Open PRs that rewrite the same two hunks: lium-io#1351 (DAH-3338, adds reaped pod states after the unpack) and lium-io#1345 (DAH-2834, adds `host_facts=` to the call); whichever lands second re-resolves against the `if first_sight / else` shape here.

### Ran it

`pdm run pytest tests/test_stale_container_cleanup_check.py -q` → 13 passed on the warm EC2 box at a9ecf9f (aws_verify 20260909T033850Z-chwg); full validators suite at the earlier head o5gu: 1952 passed, 2 baseline failures (root/sysfs, sshd adopt)

</details>


neurons/validators/src/services/task/checks/stale_container_cleanup.py
```diff
@@ -50,11 +56,21 @@ def __init__(self) -> None:
...
-            executor_uuid=ctx.executor.uuid,
-        )
+        # DAH-1932: when a miner re-adds an executor it gets a new subnet UUID for the same
+        # IP:port. `rented_data` is keyed by UUID, so the tenant that is still running on the
+        # box is listed under the OLD uuid and this executor looks unrented; the backend only
+        # remaps the row after this cycle's spec publish. Removing "orphans" now would kill a
+        # paying customer's pod. A UUID this validator has never run the cleanup for gets one
+        # cycle of grace; the set lives in Redis so a validator restart does not reopen the race.
+        first_sight = await self._first_sight(ctx)
+        if first_sight:
```
jam6099, line 69: SADD burns the only grace cycle on a publish that cannot remap. `save_executor_into_db` returns before upsert when score and synthetic_job_score are both 0 (`lium-io-backend` `apps/server/src/worker.py:291`). Re-register UUID B is absent from `rented_data`, so the tenant still looks unrented, the live pod still holds the ports, `PortCountCheck` is fatal, `ScoreCheck` never runs, and both scores stay 0. Cycle 2 sees the UUID, `rented_data` is still empty, and cleanup removes the pod. Ship the IP:port remap on that zero-score publish. A one-SADD skip cannot close this race by itself.

jam6099 (changes requested):
One SADD grace does not close the kill. The first publish scores 0, so the IP:port remap never runs, and cycle 2 still removes the pod.

neurons/validators/src/services/task/checks/stale_container_cleanup.py
surcyf123: Two options. A: the validator aliases the first-seen UUID to the tenant by address, once, where the context is built. `RentedExecutor` carries `executor_ip_address` and `executor_ip_port`, so a rental listed under the old UUID at this address becomes this executor's rental for the whole pipeline. Cleanup and `PortCountCheck` then see the paying tenant in cycle 1, the score is positive, and the backend's `upsert` remaps the row by address as it does today. B: the backend remaps on the zero-score publish, in lium-io-backend `worker.py` before the score gate, and this grace stays as the one-cycle bridge. I'd take A because it closes the kill inside the validator in the same cycle and needs no second repo. B leaves cycle 1 scored 0 and the customer's pod one cycle from removal. Say which and it goes on this branch with a test that replays your sequence.

neurons/validators/src/services/task/checks/stale_container_cleanup.py
```diff
@@ -50,11 +56,24 @@ def __init__(self) -> None:
...
+        # DAH-1932: when a miner re-adds an executor it gets a new subnet UUID for the same
+        # IP:port. `rented_data` is keyed by UUID, so the tenant that is still running on the
+        # box is listed under the OLD uuid and this executor looks unrented. Removing "orphans"
+        # now would kill a paying customer's pod. A UUID this validator has never run the cleanup
+        # for gets one cycle of grace; the set lives in Redis so a validator restart does not
+        # reopen the race. The grace is a bridge, not the fix: the backend remaps the row to the
+        # new UUID only on a publish with a positive score, and a first cycle that fails
+        # PortCountCheck publishes 0 — closing that (in the validator by address, or in the
+        # backend's zero-score publish) is the open question on the PR.
+        first_sight = await self._first_sight(ctx)
```
taiberium, line 68: NIT: #1302 changes this same call to return three values. Whichever lands second needs the rebase.

taiberium (approved):

neurons/validators/src/services/task/checks/stale_container_cleanup.py
surcyf123: Not changing this because the code is right either way: #1298 lands first, #1302 rebases onto it; both bodies now say so under Cross-PR.

neurons/validators/src/services/task/checks/stale_container_cleanup.py
surcyf123: Done in `9a8566b` on lium-platform#328: the backend remaps the row by IP:port on the zero-score publish, before the score gate. Same hotkey, same address, not reclaimed, a customer pod in an active state on the row. Only `executor_id` moves; `updated_at` is kept so the stale sweep does not read the cycle as contact. Five tests replay your sequence; on `main` the first one fails with the row still under uuid A. This grace stays as the bridge for the cycle in which that publish lands. Same ticket, same reviewer on both.

jam6099 (approved):

PR #1309, opened 2026-09-07
DAH-3137 - [P2] signed on-host GPU hardware-signature challenge

surcyf123:
Implements DAH-3137 · reviewer: @jam6099

**TL;DR** — Signed on-host GPU hardware-signature challenge (liumd building block). Touches validator/executor (`neurons/executor/Dockerfile`) — read that file first.

**What changed**
- `neurons/validators/src/services/gpu_signature.py` (+287 −0)
- `neurons/executor/gpu_sig/gpu_sig.cu` (+279 −0)
- `neurons/validators/tests/test_gpu_signature.py` (+145 −0)
- `neurons/validators/src/services/task/checks/gpu_signature.py` (+144 −0)
- `neurons/executor/gpu_sig/sha256.h` (+128 −0)
- `neurons/executor/gpu_sig/README.md` (+62 −0)

**How to verify**
- `gh pr checks 1309 -R Datura-ai/lium-io` → all green

**Verified by the loop:** 7 Sep 2026 · sandbox EC2 · validators 1955 passed · executor 128 passed · Result: PASS 0fcdaff8 · Gate: not run

**Docs:** neurons/executor/gpu_sig/README.md · neurons/validators/.env.template (GPU_SIGNATURE_*) + neurons/executor/gpu_sig/README.md (this PR) · public: none changed in this PR — off by default; absence of the binary skips the check

## DAH-3137 — Signed on-host GPU hardware-signature challenge (a building block for liumd / DAH-2834)

First shippable increment of "make provider-GPU verification a compiled binary that already lives on the GPU" (owner ask, 7 Sep). Full design + threat model: `~/lium-ads/verification-binary/DESIGN.md`. Child of @jam6099's **DAH-2834 (liumd v1)** — the loop is taking that work over; see the DAH-2834 comment.

### The gap this closes
Idle incentive is paid per advertised GPU-hour by class, so the core spoof is presenting **1×B300 as 8×B300**. Today count/UUID/model come from **NVML**, which a root provider shims (DAH-2662: `/etc/ld.so.preload` → a `.so` overriding `nvmlDeviceGetCount/GetHandleByIndex/GetUUID`). The all-cards work-proof (DAH-2671) is the count defense, but its own code notes it lacks (a) a **per-card cryptographic seal** (`_run_one_card` trusts exit-0 + non-empty stdout) and (b) a second independent physical signal, and it binds identity to NVML UUIDs.

### What this adds (additive, flag-off, observe-only)
A nonce-bound, sealed per-GPU hardware-signature check whose prober **ships pre-built in the executor image** (not uploaded per check):

- **`neurons/executor/gpu_sig/gpu_sig.cu`** — CUDA prober: `--nonce`, `--device`; reads kernel identity from `/proc/driver/nvidia/gpus/*/information` (outside NVML); seeded **FP32 tiled-SGEMM → sustained TFLOPS**; **device-to-device VRAM bandwidth → GB/s**; `HMAC-SHA256` seal (key = `HMAC(master, kernel_uuid)`) over the exact string `nonce|device|kernel_uuid|pci|tflops|gbps|ms`. Self-contained SHA-256/HMAC in `sha256.h` (no OpenSSL). The signed `msg` is authoritative, so C/Python float formatting can never diverge.
- **Validator `GpuSignatureCheck`** (`gpu.validate.signature`, `fatal=False`, behind `ENABLE_GPU_SIGNATURE_CHECK`, default off) — issues one nonce, runs the binary per claimed card via `CUDA_VISIBLE_DEVICES`, verifies each seal + nonce freshness + per-class envelope, aggregates for count. **Count-spoof signals:** device-selection failure on a missing ordinal, **duplicate kernel UUID** across cards (one physical card answering for many), a blown aggregate wall-clock, and a **kernel-vs-NVML UUID mismatch** (a userspace shim). Runs right after the capability matmul on the same idle, non-filler population.
- **`services/gpu_signature.py`** — pure-stdlib verify / envelope / aggregate, unit-tested without a GPU.
- **`GPU_SIGNATURE_ENVELOPE`** — per-class ground-truth floors, marked **uncalibrated** (gate nothing) until calibrated on real hardware with this exact binary.
- **Packaging** — optional multi-stage Dockerfile builder, **default off**: a normal executor image build pulls no CUDA and installs nothing (the check then skips `binary_absent`). Opt in with `--build-arg GPUSIG_BUILDER=nvidia/cuda:*-devel --build-arg INSTALL_GPU_SIGNATURE=true --build-arg LIUM_SIG_KEY=<hex>`.

### Safety / non-breaking
- Default flags off + binary absent from the default image ⇒ the check emits a `SKIPPED` event and **changes nothing**. Non-fatal, so even a failing verdict never halts the pipeline or zeroes score.
- `GPU_SIGNATURE_ENFORCEMENT_ENABLED` currently only raises event severity; the score gate (a `Context` flag consumed by `calculate_scores`, mirroring `cpu_truth_passed`) is a deliberate follow-up once the envelope is calibrated — same shadow→warn→enforce rollout MATMUL_ALLCARDS used.

### Honest scope
A root provider who fully reverse-engineers the binary can still return fabricated fast numbers (the master seal key is baked in the image). This closes the **cheap/scalable** spoofs (NVML shims, canned/replayed answers, wrong-class numbers, count serialisation) and builds the challenge-response + envelope machinery liumd will carry. Full unforgeability needs the derived-key seal folded into liumd and/or **NVIDIA CC attestation** (already wired: `attestation_service.py` / `gpu_attestation_service.py`). Kernel-mode tampering and relay attacks are only defeated by the CC/TDX tier — see DESIGN.md §8.

### Tests
- `tests/test_gpu_signature.py` (16 cases, no GPU): seal good / tampered-numbers / wrong-key / missing; replayed-nonce; prober-error; envelope (uncalibrated no-gate + calibrated flags underperformer); count-lie via device-selection; **duplicate-UUID**; over-wall-clock; stdout parsing. All green.
- **Crypto interop**: `gpu_sig/test_seal.c` reproduces the seal with no CUDA; verified byte-identical to Python's `hmac` (`639e3b7d…`).
- **`nvcc` compile** is exercised by the Dockerfile's `make gpu_sig` when built with `INSTALL_GPU_SIGNATURE=true` (host `sha256.h` compiles clean via `cc`; the `.cu` is standard runtime-API CUDA).
- **GPU-pod calibration is the immediate next step and is required before enforcement**: run `gpu_sig` on one real card per class, feed `tflops`/`gbps` floors into `GPU_SIGNATURE_ENVELOPE`, and set `calibrated=True` only for measured classes. Until then the envelope gates nothing and the check is observe-only — so this PR is safe to merge without the numbers.

Docs: neurons/executor/gpu_sig/README.md · neurons/validators/.env.template (GPU_SIGNATURE_*) + neurons/executor/gpu_sig/README.md (this PR) · public: none changed in this PR — off by default; absence of the binary skips the check

## Verification

**Verified by the loop on 7 Sep 2026:** checked on sandbox EC2 (the CI shape, then the #1312 subnet-loop e2e stack on a clean worktree) — branch head `0fcdaff8` merged onto `main` `07ec64cd` (clean merge), then: pytest: validators, executor (CI env) + ruff format report; e2e: the #1312 subnet-loop gate (protocol, cycle, rental); Validators 1955 passed, 15 skipped, 155 warnings — 2 failed are main's own (root-only sysfs/sshd artefacts, same as pass 1); Executor 128 passed, 18 warnings; e2e cycle 5 passed, 0 failed, 0 skipped; e2e protocol 13 passed, 0 failed, 0 skipped; e2e rental 2 passed, 0 failed, 0 skipped. Run time 5 min. Result: PASS. Not proven here: a real chain and real GPU hosts (the #1312 stack runs the executor without a GPU).

</details>

### DIFF
--- a/neurons/executor/gpu_sig/Makefile
+++ b/neurons/executor/gpu_sig/Makefile
@@ -1,0 +1,37 @@
+# Build the gpu_sig hardware-signature prober (DAH-3137).
+#
+# The GPU binary needs a CUDA toolkit (nvcc). The seal-interop test builds with
+# a plain C compiler (no CUDA), so CI/dev can verify the HMAC matches Python.
+#
+#   make gpu_sig        # needs nvcc + a CUDA-capable build host
+#   make test_seal      # host-only crypto interop harness (cc)
+#   make LIUM_SIG_KEY=<hex> gpu_sig   # bake a release seal key
+
+NVCC ?= nvcc
+CC ?= cc
+LIUM_SIG_KEY ?= lium-gpu-sig-dev-key-do-not-use-in-prod
+# Target the datacenter + consumer arches the fleet runs (Ampere..Blackwell).
+GPU_ARCHS ?= -gencode arch=compute_80,code=sm_80 \
+             -gencode arch=compute_86,code=sm_86 \
+             -gencode arch=compute_89,code=sm_89 \
+             -gencode arch=compute_90,code=sm_90
+
+gpu_sig: gpu_sig.cu sha256.h
+	$(NVCC) -O3 $(GPU_ARCHS) \
+	  -DLIUM_SIG_MASTER_KEY="\"$(LIUM_SIG_KEY)\"" \
+	  -o gpu_sig gpu_sig.cu
+
+# Portable single-arch build for a pod whose exact SM you know (faster to build):
+#   make ARCH=sm_89 gpu_sig_single
+gpu_sig_single: gpu_sig.cu sha256.h
+	$(NVCC) -O3 -arch=$(ARCH) \
+	  -DLIUM_SIG_MASTER_KEY="\"$(LIUM_SIG_KEY)\"" \
+	  -o gpu_sig gpu_sig.cu
+
+test_seal: test_seal.c sha256.h
+	$(CC) -O2 -o test_seal test_seal.c
+
+clean:
+	rm -f gpu_sig gpu_sig_single test_seal
+
+.PHONY: clean
--- a/neurons/executor/Dockerfile
+++ b/neurons/executor/Dockerfile
@@ -1,3 +1,26 @@
+# ---- optional gpu_sig builder (DAH-3137) --------------------------------------
+# Compiles the on-host GPU hardware-signature prober (neurons/executor/gpu_sig).
+# The builder base defaults to the SAME slim image as the final stage, so a
+# DEFAULT build pulls no CUDA and produces an EMPTY binary — the runtime install
+# below then no-ops and the validator check skips (binary_absent). Nothing about
+# the current image changes unless you opt in:
+#   docker build \
+#     --build-arg GPUSIG_BUILDER=nvidia/cuda:12.6.2-devel-ubuntu22.04 \
+#     --build-arg INSTALL_GPU_SIGNATURE=true \
+#     --build-arg LIUM_SIG_KEY=<hex-seal-key-matching-the-validator> ...
+ARG GPUSIG_BUILDER=python:3.11-slim
+FROM ${GPUSIG_BUILDER} AS gpusig
+ARG INSTALL_GPU_SIGNATURE=false
+ARG LIUM_SIG_KEY=lium-gpu-sig-dev-key-do-not-use-in-prod
+COPY gpu_sig /gpu_sig
+RUN set -e; \
+    if [ "$INSTALL_GPU_SIGNATURE" = "true" ]; then \
+      apt-get update && apt-get install -y --no-install-recommends make build-essential && \
+      cd /gpu_sig && make gpu_sig LIUM_SIG_KEY="$LIUM_SIG_KEY" && cp gpu_sig /gpu_sig.bin; \
+    else \
+      : > /gpu_sig.bin; \
+    fi
+
 FROM python:3.11-slim
 
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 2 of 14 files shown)


neurons/executor/gpu_sig/Makefile
```diff
@@ -0,0 +1,37 @@
...
+CC ?= cc
+LIUM_SIG_KEY ?= lium-gpu-sig-dev-key-do-not-use-in-prod
+# Target the datacenter + consumer arches the fleet runs (Ampere..Blackwell).
+GPU_ARCHS ?= -gencode arch=compute_80,code=sm_80 \
+             -gencode arch=compute_86,code=sm_86 \
+             -gencode arch=compute_89,code=sm_89 \
+             -gencode arch=compute_90,code=sm_90
+
+gpu_sig: gpu_sig.cu sha256.h
+	$(NVCC) -O3 $(GPU_ARCHS) \
```
jam6099, line 20: `nvcc` links cudart shared by default. The final executor image is `python:3.11-slim`. The NVIDIA runtime injects `libcuda`, not `libcudart`. After COPY, `test -x` is true and every card returns `no_result`. README already says `-cudart static`. Add that flag here. The gencode list is SASS-only and stops at sm_90; B200/B300 in `GPU_SIGNATURE_ENVELOPE` cannot load this binary.

jam6099 (changes requested):
`nvcc` links cudart shared. The slim executor image will not start this binary. Score is untouched while the flag stays off.

neurons/executor/gpu_sig/Makefile
surcyf123: Not changing this here because the PR is closed on your 09:56Z call and the challenge moves to celium-gpu-verifier. Its Makefile gets `-cudart static` and the sm_100 and sm_120 gencodes with a PTX fallback, so the slim image starts it and B200 and B300 load it.

PR #1332, opened 2026-09-08
DAH-3240 - [P1] rent: one host probe for the volume stage

surcyf123:
Implements [DAH-3240](https://app.notion.com/p/Rent-volume-creation-stage-is-5-serial-SSH-commands-incl-a-Docker-Hub-plugin-install-on-every-rent-3d5b8bfdbde98120b20adbbf1179e825) · reviewer: @taiberium

**TL;DR** — The volume-creation stage of a rent is p50 3.3 s (cached, no filler, 7 d prod) for a 61 ms `docker volume create`: five serial SSH commands, one a `docker plugin install` that asks Docker Hub and fails "already exists" every rent (0.9 s measured). Behind `RENTAL_VOLUME_FAST_PATH_ENABLED` (off) one probe returns root dir, df, volumes and plugin state; the install is skipped when the plugin is enabled. A probe failure takes the old path.

**What changed**
- `RENTAL_VOLUME_FAST_PATH_ENABLED` (False) + `.env.template` line (`neurons/validators/src/core/config.py`)
- `probe_volume_host`: one SSH command, `ROOT`/`DF`/`VOL`/`VOLS`/`PLUGIN` lines; None on any error (`services/docker_service.py`)
- `resolve_volume_sizing(host_probe=)`: reads root/df/names from the probe; only `volume inspect` still runs (`docker_service.py`)
- `create_local_volume(host_probe=)`: reuses the root dir, skips `plugin install` when `Enabled=true` (`docker_service.py`)
- `create_container` probes once, only when something reads it; passes the probe to both calls (`docker_service.py`)
- `df_command` / `parse_df_available_bytes` split out so the probe measures df the same way (`core/docker_utils.py`)

**How to verify**
- `cd neurons/validators && pytest tests/test_volume_fast_path.py -q` → 37 passed
- `pytest tests/ -q` → 2 failed (main's own), 2036 passed
- flag on, one rent → profiling board "Docker volume creation" ~3.3 s → ~1.3 s; log `volume_host_probe`

**Verified by the loop:** 9 Sep 2026 · sandbox EC2 · Gate: PASS 742e877

**Ran it:** exact probe command on a live RTX 4090 executor host → `ROOT … VOL volume_demo vloopback:latest · VOLS 0 · PLUGIN true` in 0.87–0.97 s vs 1.86–1.90 s for the five commands it replaces · pod zesty-matrix-f0 8 Sep 22:53Z · EC2 20260909T140738Z-ejro at 742e877: 2036 passed

**Self-review:** clean at 742e877 (15 checks, 5 low)

**Parts:** backend n/a — validator-only change · migration n/a — no schema change · frontend n/a — nothing renter-visible changes · CLI/SDK n/a — no API or CLI surface · docs n/a — flag in `.env.template`, nothing renter-facing · tests ✓ 37 · dashboards n/a — existing profiler step

**Docs:** `neurons/validators/.env.template` (the flag); no renter docs — nothing a renter sees changes.

### Why this stage costs 3.3 s for a 61 ms create

Prod `rental_history.profilers`, 7 d to 8 Sep 22:00Z (`~/lium-ads/speed/stage_query.sql`): "Docker volume creation step finished" p50 3.27 s / p90 7.75 s on cached rentals with no filler to stop (n=273), 1.70 / 6.24 with a filler (n=1,437). The stage, before `docker volume create` runs, is:

1. `docker info --format {{.DockerRootDir}}` (`resolve_volume_sizing` → `get_docker_root_dir`)
2. `docker run --rm -v <root>:/hostfs:ro alpine df -P -B1 /hostfs` (`_get_fs_available_bytes`)
3. `docker volume ls --format …` (`_get_existing_vloopback_bytes`)
4. `docker volume inspect <vloopback volumes>` (same; only when any exist)
5. `docker info --format {{.DockerRootDir}}` again (`create_local_volume`)
6. `docker plugin install ashald/docker-volume-loopback --alias vloopback --grant-all-permissions DATA_DIR=…` — unconditional; the CLI asks the registry for the plugin's privileges first, then the daemon answers "already exists" (exit 1, ignored)

Each is an SSH channel round trip to the executor; on the consumer-GPU half of the fleet (3090/4090/5090 at 26–28 s p50 pending→RUNNING) a round trip is 0.2–0.6 s.

Measured on a live RTX 4090 executor-class host (pod swift-raven-9e, Ukraine, Docker 27.3.1, 8 Sep 22:10Z, `~/lium-ads/speed/proof/host_bench_out.txt`), each command host-local, three runs:

```
docker info --format {{.DockerRootDir}}                                   61–67 ms
docker run --rm -v /var/lib/docker:/hostfs:ro alpine:3.20 df -P -B1      728–741 ms
docker volume ls --format {{.Name}} {{.Driver}}                           21–22 ms
docker plugin inspect --format {{.Enabled}} vloopback                     19–20 ms
docker plugin install ashald/docker-volume-loopback --alias vloopback …  874–928 ms rc=1  (already installed)
docker volume create -d vloopback -o size=10g                             61 ms
```

### What the flag changes

`probe_volume_host` runs `_volume_host_probe_command(with_df=…)`:

```
root="$(/usr/bin/docker info --format '{{.DockerRootDir}}')"; printf 'ROOT\t%s\n' "$root"; printf 'DF\t%s\n' "$(/usr/bin/docker run --rm -v "$root":/hostfs:ro docker.io/library/alpine:3.19 df -P -B1 /hostfs | tr '\n' '\r')"; /usr/bin/docker volume ls --format 'VOL\t{{.Name}}\t{{.Driver}}'; printf 'VOLS\t%s\n' "$?"; printf 'PLUGIN\t%s\n' "$( (/usr/bin/docker plugin inspect --format '{{.Enabled}}' vloopback 2>/dev/null || echo absent) | tail -n 1)"
```

- The probe runs only when something reads it: `measures_host_for_volume_sizing(payload)` (the predicate `resolve_volume_sizing` itself now uses for its two passthrough returns) or a limited volume (`volume_limit_gb` set, so `create_local_volume` needs the root dir and the plugin state). `with_df` is the fresh-sizing predicate; an unlimited volume on a passthrough contract pays for no command at all.
- The df record folds its two lines onto one (`\n` → `\r`); the parser splits records on `\n` only and unfolds before `parse_df_available_bytes` (the same parser `df_available_bytes` uses).
- `VOL` lines are filtered by `_is_vloopback_driver` and `_is_safe_docker_volume_name` exactly as `_get_existing_vloopback_bytes` filters them; the sizes still come from `docker volume inspect` (`_inspect_vloopback_volumes_bytes`, shared by both paths). An empty volume list is legitimate, so the list is followed by `VOLS\t<exit status of docker volume ls>`; anything but `0` makes the parser raise (the per-command path raises on that exit status too), so a failed listing can never read as "no existing volumes".
- `PLUGIN` is `true` / `false` / `absent`; only `true` skips the install. A disabled or absent plugin runs the same `docker plugin install` command as before, with `DATA_DIR` from the probe's root dir.
- Parser errors (a bad df record included) echo at most 512 characters of the host's output (`_PROBE_OUTPUT_LOG_CAP`) into the log line; two tests feed 5,000-character records and assert the message stays short.
- A missing section, a non-numeric df column, a non-zero `VOLS`, an SSH error: `probe_volume_host` logs `volume_host_probe_fallback` and returns None; `resolve_volume_sizing` and `create_local_volume` then run the per-command path unchanged. The probe is one extra command in that case. Like every other command on this path it has no per-command timeout; the connection's keepalive (30 s × 4, `_CREATE_CONTAINER_SSH_KEEPALIVE_*`) catches a dead link.
- Sizing math (`pool` / `request_cap` / `df_guard`, the 1 GB floor, `min_volume_gb`), driver options, `sparse` gating and the create timeout are untouched: `test_resolve_volume_sizing_with_probe_matches_per_command_result` asserts the probe path returns the same `VolumeSizingResult` as the per-command path for the same host facts.

Commands per rent on this stage: 5–6 → 1–2 when the plugin is already enabled (probe, plus inspect when vloopback volumes exist; 2–3 when the plugin is disabled or absent — the install still runs), and no Docker Hub call on the enabled path.

### Ran it

Pod `zesty-matrix-f0` (RTX 4090, Ukraine, Lium template `e03e4d64-fec3-483b-9de9-b6e8a86b404b` (a template id, not a commit) = the dind image, Docker 27.3.1 in the pod; `/usr/bin/docker` symlinked to the pod's docker binary), 8 Sep 22:53Z, `~/lium-ads/speed/proof/ran_it_probe_v2.txt`:

```
== probe command (with df), exact text from _volume_host_probe_command(with_df=True):
root="$(/usr/bin/docker info --format '{{.DockerRootDir}}')"; printf 'ROOT\t%s\n' "$root"; printf 'DF\t%s\n' "$(/usr/bin/docker run --rm -v "$root":/hostfs:ro docker.io/library/alpine:3.19 df -P -B1 /hostfs | tr '\n' '\r')"; /usr/bin/docker volume ls --format 'VOL\t{{.Name}}\t{{.Driver}}'; printf 'VOLS\t%s\n' "$?"; printf 'PLUGIN\t%s\n' "$( (/usr/bin/docker plugin inspect --format '{{.Enabled}}' vloopback 2>/dev/null || echo absent) | tail -n 1)"
== output (tabs shown as \t, \r as \r):
ROOT\t/var/lib/docker
DF\tFilesystem           1-blocks       Used Available Capacity Mounted on\r/dev/nvme0n1p3       490574913536 55282839552 410296958976  12% /hostfs\r
VOL\tvolume_demo\tvloopback:latest
VOLS\t0
PLUGIN\ttrue
== probe wall: 970 ms
== today, the same facts over 5 commands (run 1, host-local, no ssh RTT): 1855 ms
   probe wall (run 1): 868 ms
== today, the same facts over 5 commands (run 2, host-local, no ssh RTT): 1900 ms
   probe wall (run 2): 875 ms
== _parse_volume_host_probe(output, with_df=True), parser from the branch, run on the Mac:
   VolumeHostProbe(docker_root_dir='/var/lib/docker', df_avail_bytes=410296958976, vloopback_volume_names=['volume_demo'], loopback_plugin_enabled=True)
== negative control: same output with `docker volume ls` exiting 127 (no VOL lines, VOLS 127):
   raises: volume host probe: docker volume ls exit status '127' in 'ROOT\t/var/lib/docker\nDF\tFiles …
```

An earlier run of the command on the same host before `/usr/bin/docker` existed there printed `ROOT\t` / `DF\t` / `PLUGIN\tabsent` — the parser raised `no DockerRootDir` on it, which is the fallback path.

Validator suite on EC2 (`tools/aws_run_jobs/reverify_io.sh`, run `20260909T083313Z-agtn`, head `bafc4f6a` on `main@b05764a7`, Python 3.11.11): `2 failed, 2036 passed, 15 skipped` — the two failures are main's own (`test_scrape_infiniband::test_an_unreadable_sysfs_tree_is_not_reported_as_no_devices`, `test_sshd_bootstrap_script::test_adopt_path_hardens_config_for_key_only_auth`; main baseline `2 failed, 1999 passed`). `ruff format --check` under the CI pin (`uvx ruff@0.5.1`): the new test module is formatted; the touched existing files were not ruff-formatted on main and are left as they are (report-only in CI). Not run on a real validator against a real executor: the loop has no validator it may point at prod nodes; the staging validator (testnet 37) is where the flag goes first.

### Tests (`tests/test_volume_fast_path.py`, 37)

- probe command: one line, names every section, reuses `df_command`, contains no `plugin install`; `with_df=False` has no helper container
- parser: root / df / vloopback names (non-vloopback and unsafe names dropped) / plugin; `false` / `absent` / empty → not enabled; no df when not requested; missing root / df / plugin state, a garbage df, a missing `VOLS` and `VOLS\t1` each raise
- `probe_volume_host`: one command with the exact text; SSH error → None; garbage → None
- the command and the parser together through `sh -c` with a docker stub: volumes parsed, empty list → zero volumes, failed `volume ls` → `VOLS\t1` → raise, absent plugin → not enabled
- `measures_host_for_volume_sizing` is the predicate `resolve_volume_sizing` uses: `path == "fresh"` iff the predicate, and only then are commands run
- `resolve_volume_sizing`: probe result == per-command result for the same host facts (4 commands → 1 inspect); no vloopback volumes → no command; a probe without df → the per-command measurement
- `create_local_volume`: enabled plugin → no SSH command, identical volume spec; plugin absent → the install command with the probe's `DATA_DIR` (negative control); no probe → `docker info` + install as before
- `create_container`: flag off → never probes, `host_probe=None` to both callees; flag on → one probe, `with_df=True` for fresh sizing, the probe object reaches sizing and create; either passthrough contract (legacy, storage-opt-unsupported) with a limited volume → `with_df=False`; passthrough with no volume limit → never probes; probe None → both callees receive `host_probe=None`

### Risk

- The plugin can be enabled in `plugin inspect` and still fail a create (plugin process wedged): unchanged from today, where the install's "already exists" is ignored and the create fails the same way.
- Docker volume names cannot contain tabs or spaces (`[a-zA-Z0-9][a-zA-Z0-9_.-]*`), so a `VOL` record always has three fields; a record that does not parse is dropped (as the current path drops a name with a space via `split(maxsplit=1)`) and the pool is bounded by `df_guard` either way.
- Rollback: the flag.

### Cross-PR

Cross-PR: lium-io#1295 (DAH-3004, concurrent SSH + Docker connect) touches `create_container` ~250 lines above this hunk; no shared lines. Cross-PR: lium-io#1316 (DAH-2971, dead-code sweep) deletes unused `/usr/bin/docker …` command strings in `miner_jobs/backup_storage.py` and re-indents four unused locals in `docker_service.py`; this PR keeps `/usr/bin/docker` in its own new commands like every live command on this path — either order merges clean (`git merge-tree` clean against both). lium-io#1326 (DAH-3217) reads DockerRootDir for the stale-mount repair (`_docker_root_dir_or_default`); different functions, no shared lines. lium-io#1283 (DAH-2977) is executor-side. None of them lands first by necessity.

### Self-review

Verdict `findings` at `742e877` · 15 checks · by subagent-r39q1332 · 2026-09-09 14:23Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | PROSE-VS-CODE | `PR body (Details > Ran it, the '== probe command (with df), exact text from _volume_host_probe_command(with_df=True):' block):149` | The block is labelled 'exact text' and now shows the head's command (PLUGIN through `$( (...) / tail -n 1)`), but the host run it reports (`speed/proof/ran_it_probe_v2.txt` line 2, pod zesty-matrix-f0 8 Sep 22:53Z) executed the earlier form without `tail -n 1`; the output shown (`PLUGIN\ttrue`) is identical for an enabled plugin and the substitution is disclosed in the Verified paragraph, so nothing a reviewer would act on changes. | Label the block 'command at the earlier head; the head adds `/ tail -n 1` around the PLUGIN substitution, which changes nothing when the plugin is enabled', or re-run the head's text on a host and paste that output. |
| low | PROSE-VS-CODE | `PR body (Details > Ran it, last paragraph: 'An earlier run ... printed ROOT\t / DF\t / PLUGIN\tabsent — the parser raised no DockerRootDir'):168` | With a `DF\t` line present the head parser raises 'volume host probe: unexpected df output' before it reaches the root check: `_parse_volume_host_probe` parses DF inside the loop (docker_service.py:415-420) and tests `docker_root_dir` only after the loop (line 429); reproduced on the head's parser with 'ROOT\t\nDF\t\nVOLS\t127\nPLUGIN\tabsent\n' -> 'unexpected df output', and 'no DockerRootDir' only when the DF line is absent. Either way the result is the fallback path, so the behaviour claim holds; only the named error differs. | Say 'the parser raised `unexpected df output` on it' (or drop `DF\t` from the list if that run had no DF line); check by feeding the quoted three lines to `_parse_volume_host_probe(..., with_df=True)`. |
| low | STALE-BODY | `PR body (above the fold: 'Ran it: ... suite EC2 20260909T083313Z-agtn at bafc4f6' and 'Verified by the loop: ... Gate: PASS bafc4f6'):90` | Both above-fold lines still cite the earlier head bafc4f6 while a suite run on the head tree exists (20260909T140738Z-ejro, tree 0f535695 = 742e8773, 2036 passed / 2 env-dependent failures / 15 skipped); the gate line is regenerated on push and `--record` rewrites Self-review, but the Ran-it suite token is authored text. | Point the Ran-it suite token at `20260909T140738Z-ejro at 742e8773` when the body is recorded at this head. |
| low | CODE-QUALITY | `neurons/validators/src/services/docker_service.py:380` | Carried from the aab58e5/bafc4f6 verdicts, unchanged at head: the `_volume_host_probe_command` docstring says 'a failing section leaves its tag out', but a failed `docker info` still prints `ROOT\t` (empty, line 391), a failed df prints `DF\t` (line 385) and a missing plugin prints `PLUGIN\tabsent` (lines 397-398, now via `tail -n 1`); the parser raises on the empty/unparsable value, not on a missing tag. | Reword to 'a failing section prints its tag with an empty value (`ROOT\t`, `DF\t`) or `absent` (PLUGIN), and the volume list is followed by `VOLS\t<exit status>`; the parser raises on each'. |
| low | TEST-GAP | `neurons/validators/tests/test_volume_fast_path.py:540` | Carried from the earlier verdicts, unchanged at head: the 'storage_opt_unsupported passthrough' parameter leaves `disk_share` at `_payload`'s None default (test_deploy_optimizations.py:136-153 sets no disk_share), so `measures_host_for_volume_sizing` short-circuits on `disk_share is None` and the create_container gate is never exercised with the shape the backend sends for that contract (`disk_share` set, `storage_limit_gb=None`); the sizing-level test at line 304 covers that pair, so the property holds. | Change the second parameter to `dict(disk_share=0.5, storage_limit_gb=None, volume_limit_gb=2)`. |

Mechanical notes dismissed: `neurons/validators/tests/test_volume_fast_path.py:61` `_probe_stdout(root="/var/lib/docker")` is the fixture string the parser reads as the `ROOT` line (what `docker info --format {{.DockerRootDir}}` prints); no mount is built from it. · `neurons/validators/tests/test_volume_fast_path.py:133` Assertion on `probe.docker_root_dir`, a value parsed from the fixture string; nothing is mounted. · `neurons/validators/tests/test_volume_fast_path.py:150` Same assertion in the with_df=False parser test; `_parse_volume_host_probe` only reads text. · `neurons/validators/tests/test_volume_fast_path.py:232` Assertion after the `sh -c` run against the stub: the stub's `info --format` branch echoes the path and its `run --rm` branch prints canned df text, so `-v "$root":/hostfs:ro` never reaches a daemon; the only real mount on the branch is `df_command` (core/docker_utils.py), the same read-only docker-root bind into the alpine df helper that `df_available_bytes` on main already performs. · `neurons/validators/tests/test_volume_fast_path.py:335` `VolumeHostProbe(docker_root_dir=...)` literal handed to `resolve_volume_sizing`, which uses the root dir for log fields only. · `neurons/validators/tests/test_volume_fast_path.py:363` Same dataclass literal for the no-vloopback-volumes sizing test; no command is built from it (`ssh_client.run.assert_not_called()`). · `neurons/validators/tests/test_volume_fast_path.py:387` Same dataclass literal for the probe-without-df test; the per-command path then runs its own `docker info`. · `neurons/validators/tests/test_volume_fast_path.py:431` Dataclass literal for `create_local_volume` with an enabled plugin; the test asserts no SSH command runs at all.

Cross-PR: lium-io#1316 (dead-code sweep) removes the dead s3fs `/usr/bin/docker` plugin commands in `backup_storage.py`; this PR's `test_volume_fast_path.py` line naming `/usr/bin/docker` is unrelated to that path. This PR lands first; #1316 is rebased right before its merge.

### Verified

Gate: PASS (742e877) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1332)

**Verified by the loop on 9 Sep 2026, taiberium's two nits at `742e8773`:** `_LOOPBACK_PLUGIN_ALIAS = _VLOOPBACK_DRIVER_PREFIX`, and the `PLUGIN` substitution goes through `tail -n 1` — a missing plugin makes `docker plugin inspect` print a blank stdout line before it exits 1, so the old `$(… || echo absent)` carried `\nabsent` and the field parsed as `""` (still "not enabled", never `absent`). Shown with a stub: old form `PLUGIN\t\nabsent`, new form `PLUGIN\tabsent`; the test's docker stub prints the blank line like docker does and `test_probe_command_through_a_shell_absent_plugin_is_not_enabled` asserts `PLUGIN\tabsent` in the output. The two probe-command excerpts above are updated to the new text. Warm EC2 `20260909T140738Z-ejro`: validators 2036 passed, 2 failed (the warm host's own sysfs/sshd tests, failing on every branch's run today); only the validators suite runs for this tree (the only neuron touched).

</details>


neurons/validators/src/services/docker_service.py
```diff
@@ -3368,15 +3480,70 @@ async def _get_existing_vloopback_bytes(
                 )
         return total_bytes
 
+    @staticmethod
+    def uses_fresh_volume_sizing(payload: ContainerCreateRequest) -> bool:
```
taiberium: NIT: `fresh` reads two ways here. `measures_host_for_volume_sizing` says what the True branch means.

taiberium (approved):

neurons/validators/src/services/docker_service.py
surcyf123: Done in `bafc4f6`: Renamed in bafc4f6: `measures_host_for_volume_sizing`, and the local in `create_container` is `measures_host`. The `vloopback_fresh_sizing_*` log keys are unchanged. Validators suite at bafc4f6 on EC2: 2036 passed; the two failures are the sysfs/sshd box-only ones main shares.

neurons/validators/src/services/docker_service.py
```diff
@@ -354,6 +356,92 @@ class VolumeSizingResult:
     existing_volumes_bytes: int | None = None
 
 
+_LOOPBACK_PLUGIN_ALIAS = "vloopback"
```
taiberium: NIT: this repeats `_VLOOPBACK_DRIVER_PREFIX`, and the two must stay equal or `_is_vloopback_driver` stops matching the volumes we create.

```suggestion
_LOOPBACK_PLUGIN_ALIAS = _VLOOPBACK_DRIVER_PREFIX
```

neurons/validators/src/services/docker_service.py
```diff
@@ -354,6 +356,92 @@ class VolumeSizingResult:
...
+        if with_df
+        else ""
+    )
+    return (
+        "root=\"$(/usr/bin/docker info --format '{{.DockerRootDir}}')\"; "
+        "printf 'ROOT\\t%s\\n' \"$root\"; "
+        f"{df_part}"
+        "/usr/bin/docker volume ls --format 'VOL\\t{{.Name}}\\t{{.Driver}}'; "
+        "printf 'VOLS\\t%s\\n' \"$?\"; "
+        "printf 'PLUGIN\\t%s\\n' \"$(/usr/bin/docker plugin inspect --format '{{.Enabled}}' "
```
taiberium: NIT: for a missing plugin `docker plugin inspect` writes a blank stdout line, so `absent` never reaches this field. Pipe the substitution through `tail -n 1`.

taiberium (approved):

neurons/validators/src/services/docker_service.py
surcyf123: Done in `742e877`: Done in `742e8773`: applied as suggested.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `742e877`: Done in `742e8773`: `$( (… || echo absent) | tail -n 1)`. Confirmed with a stub — the old form carried `PLUGIN\t\nabsent`, so the field was ""; the test's docker stub now prints the blank line like docker does and asserts `PLUGIN\tabsent`.

PR #1333, opened 2026-09-09
DAH-3257 - [P2] rent: one host probe for the pre-run listings, removals unchanged

surcyf123:
Implements [DAH-3257](https://app.notion.com/p/Rent-pre-run-host-stage-is-8-serial-SSH-commands-cleaning-GPU-probe-power-limit-check-port-che-3d6b8bfdbde981f39961f67991083d2c) · reviewer: @taiberium

**TL;DR** — Between the pull and `docker run` a rent asks the host eight read-only listings over eight serial SSH commands: 0.9 s cleaning + 1.1 s GPU probe p50 on a far node, for ~80 ms of host work. Behind `RENTAL_PRERUN_HOST_PROBE_ENABLED` (off) one tagged command returns them all; each step reads its section, removals and writes unchanged. A failed section, or an unparseable probe, puts that step, or every step, back on its own command.

**What changed**
- `RENTAL_PRERUN_HOST_PROBE_ENABLED` (False) + `.env.template` line (`src/core/config.py`)
- `services/prerun_host_probe.py` (new): the one-line probe command (nine tagged sections, `<TAG>_RC` each), its parser
- `probe_prerun_host`: runs it once after the pull, bounded at 30 s (the nvidia-smi bound); None on any error → per-command path
- The cleanup, the three volume sweeps and the image-label check take `host_probe=`; the sweeps return what they removed
- `build_gpu_docker_config_for_executor(host_probe=)`, `shared_device_nodes_command`; `_query_power_state(host_probe=)` behind the two restore helpers — the last-resort raise always queries live
- `create_container` probes once with the flag on; withdraws the docker listings after any removal

**How to verify**
- `pytest tests/test_prerun_host_probe.py -q` → 58 passed · `pytest tests/ -q` → 1 failed (main's own), 2058 passed
- flag on, one rent → profiling "Container cleaning" + "GPU device probe" shrink to the probe's wall; log `prerun_host_probe`

**Verified by the loop:** 9 Sep 2026 · sandbox EC2 · pytest 2058 passed · Result: PASS 37ee708 · Gate: PASS 37ee708

**Ran it:** probe vs the 8 commands: RTX 4090 (far) 6.0 s → 0.95 s, L40S (near) 3.3 s → 0.62 s · two pods, 9 Sep, the earlier head · validators suite at `37ee708`: 2058 passed, 1 failed (main's own) · EC2 warm box 20260909T131205Z-43b8

**Self-review:** clean at 37ee708 (15 checks, 5 low)

[...]

PR #1334, opened 2026-09-09
DAH-3258 - [P2] rent: inspector start alongside the encrypted-volume mount

surcyf123:
Implements [DAH-3258](https://app.notion.com/p/Rent-post-run-stage-runs-the-inspector-start-the-keys-exec-and-the-env-exec-serially-after-the-enc-3d6b8bfdbde98113b280f219e9e396df) · reviewer: @taiberium

**TL;DR** — After `docker run` a rent runs serially the gocryptfs mount (4 SSH commands), the keys exec, the bootstrap / Jupyter (skipped on a cached template), the `/etc/environment` exec and the inspector collector start: 2.3 + 2.4 + 0.45 s p50 on a far node. Behind `RENTAL_POSTRUN_CONCURRENT_ENABLED` (off) the collector starts alongside the mount (a host-side process) and keys + `/etc/environment` are ONE exec after the mount. Nothing moves past RUNNING.

**What changed**
- `RENTAL_POSTRUN_CONCURRENT_ENABLED` (False) + `.env.template` line (`src/core/config.py`)
- `build_authorized_keys_and_environment_exec_spec`: keys on stdin, `/etc/environment` lines in one exec variable — one exec, no renter value in argv or logs
- `add_ssh_public_keys_with_rental_docker(environment=)` runs the combined spec (one builder; keys-only without lines)
- `create_container`: inspector start as a task before the mount, awaited before `ContainerCreated`; > 64 KiB environments keep their own exec
- `_settle_inspector_after_failed_create`: a post-run failure awaits the start, stops the collector when no other pod is rented, then cleans up

**How to verify**
- `pytest tests/test_postrun_concurrent.py -q` → 17 passed · `pytest tests/ -q` → 2 failed (main's own), 2016 passed
- flag on, one rent → profiling "Inspector collector start" ≈ 0, "Adding public keys" (the environment step) skipped; log `exec_add_authorized_keys_and_environment`

**Verified by the loop:** Gate: PASS eeb6d28

**Ran it:** keys exec + env exec + inspector start (3 serial channels) vs 1 combined exec ‖ inspector start, one open connection: RTX 4090 (far)… · EC2 `20260911T072224Z-v0w5` · … +2 under Details

**Self-review:** clean at eeb6d28 (15 checks, 6 low)

**Parts:** backend n/a — validator-only change · migration n/a — no schema change · frontend n/a — nothing renter-visible changes · CLI/SDK n/a — no API or CLI surface · docs n/a — operator flag in `.env.template` · tests ✓ 17 · dashboards n/a — existing profiler steps show it

**Docs:** `neurons/validators/.env.template` (the flag and what it reorders); no renter docs — a renter sees only a shorter stage.

### Where the time goes

Prod `rental_history.profilers`, 7 d to 8 Sep 22:00Z (`~/lium-ads/speed/DEPLOY_UNDER_10S.md` §1), cached rental, no filler to stop: "encrypted volume" p50 2.29 s (86 % of pods), "ssh keys exec + bootstrap-skip + jupyter" 2.36 s, "env vars exec" 0 s p50 / 1.97 s p90 (most rentals carry no custom environment; an image-managed Jupyter pod carries `JUPYTER_PASSWORD`), "inspector collector start" 0.48 s. The steps are serial, and each `docker exec` over the rental Docker SDK is `exec_create` + `exec_start` (stdin socket) + `exec_inspect` over the SSH-tunnelled daemon socket — three round trips per exec on top of the exec itself.

### What moves, what stays, and why

| step | today | flag on | why it may / may not move |
|---|---|---|---|
| gocryptfs mount (upload, run, wipe, verify) | first | first, unchanged | the keys go to `/root/.ssh` and `/root` IS the mount point — written before the mount they would be hidden under it. Stays before the keys. |
| inspector collector start | last | **alongside the mount**, awaited before the return | `nohup <python> inspector_executor.py --start-collector &` on the host; reads the container from outside, needs neither the mount, the keys nor sshd. It does not gate SSH readiness, but it is still awaited before `ContainerCreated` — nothing runs after RUNNING, so no trust-window question arises. |
| authorized_keys exec | after the mount | after the mount, **carries `/etc/environment` too** | the environment lines are plain data; `/etc/environment` is read by PAM at login, so writing them before sshd is up (they were written after the bootstrap) changes nothing a renter sees. The bootstrap itself does not read the file. |
| sshd bootstrap / Jupyter | after the keys | after the keys, unchanged | DAH-2341: whichever sshd comes up first must already find the keys. |
| `/etc/environment` exec | after Jupyter | folded into the keys exec | one exec instead of two; `ADDING_PUBLIC_KEYS` (the step that timed it) is recorded `skipped`. |
| `_abort_if_cancelled_by_delete`, rented-pod cache, `ContainerCreated` | last | last, unchanged | the delete-cancel check still runs after every write. |

Everything that affects security — the mount before the keys, the keys before sshd, the delete-cancel check, the container cleanup on failure — keeps its order. What changes is only that a host-side process start no longer waits for the container's filesystem, and two writes into the container ride one exec.

### The combined exec

`build_authorized_keys_and_environment_exec_spec(container_name, public_keys, environment)`:

```
argv  = sh -c 'mkdir -p /root/.ssh && chmod 700 /root/.ssh && cat >> /root/.ssh/authorized_keys && printf '\''%s'\'' "$LIUM_ENVIRONMENT_LINES" >> /etc/environment'
stdin = <one public key per line>           (exactly build_authorized_keys_exec_spec's stdin)
env   = LIUM_ENVIRONMENT_LINES=<KEY=VALUE\n…>  (exactly build_environment_exec_spec's stdin)
```

- The keys stay on stdin, as today. The environment lines travel as an exec-process variable: the SDK's `exec_create(environment=…)` carries them in the request body, `rental_exec_spec_log_fields` logs `environment_keys` (the variable's NAME) and `stdin_bytes` — a renter's `HF_TOKEN` value is in neither argv nor any log line (`test_combined_spec_no_renter_value_in_argv_or_log_fields`). `printf '%s' "$VAR"` writes the value verbatim (`%` in a value is literal — `test_combined_script_percent_in_a_value_is_literal`).
- `_environment_file_text` is the one filter both `build_environment_exec_spec` and the combined spec use (blank key or value dropped), so the bytes appended to `/etc/environment` are the same on both paths (`test_combined_script_through_sh_writes_both_files_like_the_two_execs`).
- No environment lines → the combined spec IS `build_authorized_keys_exec_spec`'s spec, byte for byte.
- Linux caps one environment string at `MAX_ARG_STRLEN` (128 KiB) and `CustomOptions` caps nothing, so `environment_fits_exec_variable` (≤ `MAX_ENVIRONMENT_EXEC_VAR_BYTES` = 64 KiB of lines) gates the merge: a larger environment keeps its own stdin-based exec exactly as with the flag off (`test_flag_on_oversized_environment_keeps_its_own_exec`).
- The exec runs as uid 0 as every rental-bootstrap exec does (DAH-2534, `_exec_in_container_sync`).

### Failure paths

- The combined exec exits non-zero → `Failed to add SSH public keys and environment: exit_status=…` → the post-run `except` → `cleanup_failed_container_creation` as today (`failure_step=add_public_keys`).
- The mount fails with the collector already started → `_settle_inspector_after_failed_create`: await the start task (it never raises — `_run_inspector_collector_lifecycle` logs its own failures), then, when `_has_rented_containers` is False, `--stop-collector` so the host is left as a delete would leave it; another rented container on the executor keeps the collector, as on delete. Best-effort: a Redis or SSH error there is logged and the create's own failure is what is reported. Then the existing cleanup runs. With the flag off nothing precedes the mount, so this path does not exist there (`test_flag_off_failed_mount_never_started_the_inspector`).
- `ENABLE_INSPECTOR=false` → no task, `INSPECTOR_START` recorded `skipped` as today.
- The whole create being cancelled (`CancelledError`) while the task runs is the one path not covered by the settle: the connection closes and the task's `ssh_client.run` fails, which `_run_inspector_collector_lifecycle` logs as `Inspector collector start failed` — the same line a lost connection produces today.

### Ran it

Two live pods (dind template, `pod_bench` = an alpine container standing in for the rental container), 9 Sep 02:25–02:28Z, `~/lium-ads/speed/proof/lever45/{bench.sh,far.txt,near.txt}`. One ssh connection, a new channel per command (as asyncssh does), three runs each, median. TODAY = `docker exec -i … sh -c '<keys script>' < keys`, `docker exec -i … sh -c 'cat >> /etc/environment' < env`, `nohup sh -c 'sleep 0' >/dev/null 2>&1 &` (the inspector start's shape), serial; FAST = one `docker exec -i -e LIUM_ENVIRONMENT_LINES=… sh -c '<combined script>' < keys` with the nohup channel launched concurrently:

```
RTX 4090, Ukraine (far):  new channel RTT 635 ms
  TODAY  3 serial channels:  2931 / 2401 ms   (two runs of the script)
  FAST   1 exec ‖ inspector:  949 /  919 ms
L40S, United States (near): new channel RTT 427 ms
  TODAY  3 serial channels:  1965 ms
  FAST   1 exec ‖ inspector:   630 ms
written by the FAST exec (on the pod): authorized_keys 12 lines (3 runs × 2 keys + earlier runs) · /etc/environment tail: HF_TOKEN=hf_bench MODEL=bench
```

This is a lower bound for the prod effect: a `docker exec` through the ssh CLI is one channel, while the validator's SDK exec is three tunnelled HTTP round trips — the exec that disappears costs more there than here. The measurement does not include the gocryptfs mount (the alpine stand-in has no gocryptfs); the mount's own four commands are unchanged by this PR, and the inspector's 0.45 s hides under them.

Validator suite on EC2 (`tools/aws_run_jobs/io_validators_pytest.sh`, run `20260911T072224Z-v0w5` at `eeb6d28c`, Python 3.11.11): `2 failed, 2124 passed, 15 skipped` — the two failures are main's own (`test_scrape_infiniband::test_an_unreadable_sysfs_tree_is_not_reported_as_no_devices`, `test_sshd_bootstrap_script::test_adopt_path_hardens_config_for_key_only_auth`; main baseline `2 failed, 1999 passed`). `ruff format --check` (`uvx ruff@0.5.1`): the new test module is formatted. Not run on a real validator against a real executor; the staging validator is where the flag goes first.

### Tests (`tests/test_postrun_concurrent.py`, 17)

- the combined spec: without environment it equals the keys spec (blank-only environment too); keys on stdin, lines in the one variable, argv is the keys script + the printf; no renter value in argv or log fields; through `sh -c` both files get the same bytes the two specs wrote and `.ssh` is 0700; `%` literal
- `add_ssh_public_keys_with_rental_docker`: with environment → one combined exec; without / `{}` → the keys spec as before; the failure names both / names keys only
- `create_container`, with a trace whose mount waits for the inspector start and whose inspector start then takes 50 ms (so it can only be complete at the return when `create_container` awaited it): flag off → `mount → keys → env → inspector`, two execs, `ADDING_PUBLIC_KEYS` not skipped; flag on → the inspector starts before the mount finishes and is the last event before the return, the mount finishes before the keys exec, one exec equal to the combined spec, `ADDING_PUBLIC_KEYS` skipped; no environment → the plain keys exec; an environment above 64 KiB → the keys exec + its own environment exec; an unencrypted pod: the inspector start runs during the keys exec (before `keys_exec_done`); a failed mount → `start`, `stop`, then the cleanup, in that order, no exec ran, `failure_step=encrypted_volume_setup`; a failed mount with another rented container → `start` only; `ENABLE_INSPECTOR=false` → no lifecycle call, step skipped; flag off + failed mount → no lifecycle call

### Risk

- The collector observes the container ~2 s earlier than today (during the mount). It is a metrics collector for rented containers; it does not gate anything in the create.
- A post-run failure after the collector started now issues a `--stop-collector` when no other rented container is on the executor — the delete path's exact rule; with another tenant the collector stays.
- Rollback: the flag.

### Cross-PR

Cross-PR: lium-io#1332 (DAH-3240) and lium-io#1333 (DAH-3257) are merged and under this branch's merge-base (the restack onto main kept all three flags); the branch merges clean onto `origin/main`. The only open sibling touching `docker_service.py` is lium-io#1295 (DAH-3004, SSH connect), which shares no line with this PR.

### Self-review

Verdict `findings` at `eeb6d28` · 15 checks · by subagent-r105 · 2026-09-11 07:21Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | STALE-BODY | `neurons/validators/src/core/config.py:233` | The body's Cross-PR paragraph (and the folded self-review row at config.py:216) describes lium-io#1332 and lium-io#1333 as open siblings whose flag hunks conflict at the same anchor, but both are merged and are ancestors of this branch's merge-base (067ec878 IS #1332's merge commit; #1333 merged as 3d326758 before it), and `git merge-tree --write-tree HEAD origin/main` at dd27dc39 merges clean — the property 'every PR named as open in Cross-PR is open' does not hold. | Re-render the Cross-PR paragraph: '#1332 and #1333 are on main and under this branch; the branch merges clean onto main; only lium-io#1295 (DAH-3004, SSH connect) is still open and shares no line.' and drop the obsolete CROSS-PR row from the folded self-review table. |
| low | STALE-BODY | `neurons/validators/tests/test_postrun_concurrent.py:1` | The body cites `fece0d5` three times (Verified by the loop, Self-review verdict, Verified/Gate) and a suite count measured on `main@b05764a7` (2 failed / 2016 passed); `fece0d5` is not reachable from HEAD after the restack (`git merge-base --is-ancestor fece0d5 HEAD` fails), so the property 'every sha in the body is reachable on the branch' does not hold until the gate and self-review are re-recorded at eeb6d28. | After the push, re-run the gate and `tools/pr_self_review.py lium-io 1334 --record <this verdict> --apply` so every sha in the body is eeb6d28 or an ancestor; re-run `pytest tests/ -q` on the restacked head and update the counts. |
| low | TEST-GAP | `neurons/validators/src/services/docker_service.py:900` | The body's Failure-paths section states 'a Redis or SSH error there is logged and the create's own failure is what is reported', but no test drives the `except Exception` branch of `_settle_inspector_after_failed_create`: both failed-mount tests wire `_has_rented_containers=AsyncMock(return_value=...)` and never raise (unchanged since fece0d5). | Add one test where `_has_rented_containers` raises RuntimeError with the flag on and a failing mount; assert the result is still `FailedContainerRequest(failure_step='encrypted_volume_setup')`, cleanup ran once and `lifecycle_actions == ['start']`. |
| low | PROSE-VS-CODE | `neurons/validators/src/services/docker_service.py:3010` | The 'What changed' bullet reads as if `add_ssh_public_keys_with_rental_docker(environment=)` itself keeps a >64 KiB environment on its own exec; the gate (`environment_fits_exec_variable`) lives in `create_container` at line 5332, and a direct caller passing an oversized dict would get one combined spec with an oversized exec variable. | Move the '> 64 KiB keeps its own exec' clause to the `create_container` bullet, or add the size check (fall back to the keys spec and let the caller run the env exec) inside `add_ssh_public_keys_with_rental_docker`. |
| low | CODE-QUALITY | `neurons/validators/src/services/docker_service.py:120` | `environment_fits_exec_variable` sits between `build_environment_exec_spec` and `build_remove_authorized_keys_exec_spec`, breaking the alphabetical order the rest of the `services.rental_docker_sdk` import block keeps (known nit from the fece0d5 review, still present). | Move `environment_fits_exec_variable` below `build_remove_authorized_keys_exec_spec`. |
| low | CODE-QUALITY | `neurons/validators/tests/test_postrun_concurrent.py:231` | `_Trace.cleanup` is defined above `__init__`, unlike every other class in the tests directory (known nit from the fece0d5 review, still present). | Move `cleanup` below `__init__`, next to `mount` and `lifecycle`. |

Mechanical notes dismissed: `neurons/validators/tests/test_postrun_concurrent.py:123` `0o777` is the permission MASK in `stat().st_mode & 0o777`, used to assert the `.ssh` directory the combined script created is exactly `0o700`; nothing is chmod'ed 0o777 and the script under test runs `chmod 700`.

### Verified

Gate: PASS (eeb6d28) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1334)

</details>

### DIFF
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -115,7 +115,6 @@ from services.rental_docker_sdk import (
     RentalDockerSdkClientFactory,
     VolumeMount,
     build_authorized_keys_and_environment_exec_spec,
-    build_authorized_keys_exec_spec,
     build_container_command_argv,
     build_environment_exec_spec,
     environment_fits_exec_variable,
@@ -3020,19 +3019,14 @@ class DockerService:
     ) -> None:
         """Append the renter's keys to authorized_keys — and, with ``environment`` (DAH-3258), the
         renter's `/etc/environment` lines in the same exec. Raises on a non-zero exit."""
-        if environment:
-            exec_spec = build_authorized_keys_and_environment_exec_spec(
-                container_name=container_name,
-                public_keys=public_keys,
-                environment=environment,
-            )
-        else:
-            exec_spec = build_authorized_keys_exec_spec(
-                container_name=container_name,
-                public_keys=public_keys,
-            )
+        # falls back to the keys-only spec when there are no environment lines
+        exec_spec = build_authorized_keys_and_environment_exec_spec(
+            container_name=container_name,
+            public_keys=public_keys,
+            environment=environment,
+        )
         with_environment = bool(exec_spec.environment)
-        what = "SSH public keys and environment" if with_environment else "SSH public keys"
+        added_items = "SSH public keys and environment" if with_environment else "SSH public keys"
         result = await exec_logged_rental_docker_sdk_operation(
             docker_client=docker_client,
             operation=(
@@ -3043,13 +3037,13 @@ class DockerService:
         )
         if result.exit_status != 0:
             await self.stream_log(
-                result.stderr or result.stdout or f"Failed to add {what}",
+                result.stderr or result.stdout or f"Failed to add {added_items}",
                 "error",
                 log_tag,
             )
             logger.warning(
                 _m(
-                    f"Failed to add {what}",
+                    f"Failed to add {added_items}",
                     extra=get_extra_info({
                         **log_extra,
                         "container_name": container_name,
@@ -3060,7 +3054,7 @@ class DockerService:
                 )
             )
             raise Exception(
-                f"Failed to add {what}: "
+                f"Failed to add {added_items}: "
                 f"exit_status={result.exit_status}; "
                 f"stderr={result.stderr}; stdout={result.stdout}"
             )
--- a/neurons/validators/src/services/docker_service.py
+++ b/neurons/validators/src/services/docker_service.py
@@ -114,9 +114,11 @@ from services.rental_docker_sdk import (
     RentalDockerSdkClient,
     RentalDockerSdkClientFactory,
     VolumeMount,
+    build_authorized_keys_and_environment_exec_spec,
     build_authorized_keys_exec_spec,
     build_container_command_argv,
     build_environment_exec_spec,
+    environment_fits_exec_variable,
     build_remove_authorized_keys_exec_spec,
     require_rental_docker_ssh_host_key,
 )
@@ -870,6 +872,39 @@ class DockerService:
     ) -> bool:
         rented_machine = await self.redis_service.get_rented_machine(executor_info)
         return bool(rented_machine and rented_machine.get("containers"))
+
+    async def _settle_inspector_after_failed_create(
+        self,
+        inspector_task: asyncio.Task,
+        *,
+        ssh_client: asyncssh.SSHClientConnection,
+        executor_info: ExecutorSSHInfo,
+        default_extra: dict,
+    ) -> None:
+        """DAH-3258: the collector was started alongside the mount and the create then failed.
+
+        Let the start finish (it never raises), then leave the host as a delete would: stop the
+        collector when no other rented container keeps it needed. Best-effort — the create's
+        own failure is what gets reported.
+        """
+        try:
+            await inspector_task
+            if not await self._has_rented_containers(executor_info):
+                await self._run_inspector_collector_lifecycle(
+                    ssh_client=ssh_client,
+                    executor_info=executor_info,
+                    action="stop",
+                    default_extra=default_extra,
+                )
+        except asyncio.CancelledError:
+            raise
+        except Exception as exc:
+            logger.warning(
+                _m(
+                    "Inspector collector could not be settled after a failed create",
+                    extra=get_extra_info({**default_extra, "error": str(exc)}),
+                ),
+            )
 
     def _ssh_bootstrap_script_path(self) -> Path:
         return Path(__file__).resolve().parent / "assets" / "sshd_bootstrap.sh"
@@ -2981,26 +3016,40 @@ class DockerService:
         public_keys: list[str] | tuple[str, ...],
         log_tag: str,
         log_extra: dict,
+        environment: dict[str, str] | None = None,
     ) -> None:
-        exec_spec = build_authorized_keys_exec_spec(
-            container_name=container_name,
-            public_keys=public_keys,
-        )
+        """Append the renter's keys to authorized_keys — and, with ``environment`` (DAH-3258), the
+        renter's `/etc/environment` lines in the same exec. Raises on a non-zero exit."""
+        if environment:
+            exec_spec = build_authorized_keys_and_environment_exec_spec(
+                container_name=container_name,
+                public_keys=public_keys,
+                environment=environment,
+            )
+        else:
+            exec_spec = build_authorized_keys_exec_spec(
+                container_name=container_name,
+                public_keys=public_keys,
+            )
+        with_environment = bool(exec_spec.environment)
+        what = "SSH public keys and environment" if with_environment else "SSH public keys"
         result = await exec_logged_rental_docker_sdk_operation(
             docker_client=docker_client,
-            operation="exec_add_authorized_keys",
+            operation=(
+                "exec_add_authorized_keys_and_environment" if with_environment else "exec_add_authorized_keys"
+            ),
             exec_spec=exec_spec,
             log_extra=log_extra,
         )
         if result.exit_status != 0:
             await self.stream_log(
-                result.stderr or result.stdout or "Failed to add SSH public keys",
+                result.stderr or result.stdout or f"Failed to add {what}",
                 "error",
                 log_tag,
             )
             logger.warning(
                 _m(
-                    "Failed to add SSH public keys",
+                    f"Failed to add {what}",
                     extra=get_extra_info({
                         **log_extra,
                         "container_name": container_name,
@@ -3011,7 +3060,7 @@ class DockerService:
                 )
             )
             raise Exception(
-                "Failed to add SSH public keys: "
+                f"Failed to add {what}: "
                 f"exit_status={result.exit_status}; "
                 f"stderr={result.stderr}; stdout={result.stdout}"
             )
@@ -5241,6 +5290,24 @@ class DockerService:
 
                 await self.stream_log("Created Docker Container", "success", log_tag)
 
+                # DAH-3258: the inspector collector is a host-side process that reads the container
+                # from outside; it needs neither the encrypted mount nor the keys, so with the flag on
+                # it starts now and runs alongside them. It is awaited before ContainerCreated is
+                # returned (nothing here moves past RUNNING), and a create that fails after this
+                # point settles it the way a delete does (_settle_inspector_after_failed_create).
+                postrun_concurrent = settings.RENTAL_POSTRUN_CONCURRENT_ENABLED
+                inspector_extra = {**default_extra, "container_name": container_name}
+                inspector_task: asyncio.Task | None = None
+                if postrun_concurrent and settings.ENABLE_INSPECTOR:
+                    inspector_task = asyncio.create_task(
+                        self._run_inspector_collector_lifecycle(
+                            ssh_client=ssh_client,
+                            executor_info=executor_info,
+                            action="start",
+                            default_extra=inspector_extra,
+                        )
+                    )
+
                 try:
                     if use_encrypted_volume:
                         current_step = "encrypted_volume_setup"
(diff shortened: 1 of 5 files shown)


taiberium (approved):

neurons/validators/src/services/docker_service.py
```diff
@@ -2981,26 +3016,40 @@ async def add_ssh_public_keys_with_rental_docker(
...
         log_extra: dict,
+        environment: dict[str, str] | None = None,
     ) -> None:
-        exec_spec = build_authorized_keys_exec_spec(
-            container_name=container_name,
-            public_keys=public_keys,
-        )
+        """Append the renter's keys to authorized_keys — and, with ``environment`` (DAH-3258), the
+        renter's `/etc/environment` lines in the same exec. Raises on a non-zero exit."""
+        if environment:
```
taiberium: NIT: `build_authorized_keys_and_environment_exec_spec` already falls back to the keys spec when there are no environment lines. Call it once and drop the branch.

neurons/validators/src/services/docker_service.py
```diff
@@ -2981,26 +3016,40 @@ async def add_ssh_public_keys_with_rental_docker(
...
+                public_keys=public_keys,
+                environment=environment,
+            )
+        else:
+            exec_spec = build_authorized_keys_exec_spec(
+                container_name=container_name,
+                public_keys=public_keys,
+            )
+        with_environment = bool(exec_spec.environment)
+        what = "SSH public keys and environment" if with_environment else "SSH public keys"
```
taiberium: NIT: `what` hides what it holds. Rename it to `added_items`.

taiberium (approved):

taiberium (commented):
Second pass. `RENTAL_POSTRUN_CONCURRENT_ENABLED` is `False` at `core/config.py:229`, written `=false` in `.env.template`, and set in no compose file, entrypoint or deployment repo. Its only reader is `create_container`, so with the flag off the new code is unreachable and the rest of the diff is behaviour-identical. Nothing blocks. Three NITs, all for before the flip.

NIT: `docker_service.py:5464` — `except Exception` does not catch `CancelledError`, so a cancelled create never settles the start task: the collector keeps running on an executor with no pod and the task is orphaned. The body names this failure mode and no test covers it.

NIT: `docker_service.py:5302` — the start now runs before `_cache_rented_pod_best_effort`, so the collector begins before the pod is recorded as rented. `collector_start` lives in `libinspector.so` and cannot be read from this repo, so "it only reads the container from outside" is unverified. Please confirm on staging that a flag-on pod still shows up in inspector data.

NIT: `docker_service.py:5341` — with the flag on, a failed `/etc/environment` write is reported as `failure_step=add_public_keys`, so anything keyed on `failure_step=set_environment` stops seeing those failures.

NIT: `docker_service.py:5338` — `environment_fits_exec_variable(None)` is `True`, so `ADDING_PUBLIC_KEYS` is recorded skipped on every flag-on rent, including rents that carry no environment at all.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `eeb6d28`: one call to `build_authorized_keys_and_environment_exec_spec`, branch and the keys-only import gone; both paths still covered in `test_postrun_concurrent.py` (17 passed).

neurons/validators/src/services/docker_service.py
surcyf123: Done in `eeb6d28`: renamed `what` to `added_items` (three uses).

PR #1338, opened 2026-09-09
DAH-3272 - [P2] keep the gocryptfs passphrase off the provider's disk and argv

surcyf123:
Implements [DAH-3272](https://app.notion.com/p/Encrypted-volumes-gocryptfs-passphrase-reaches-the-provider-s-argv-and-disk-3d6b8bfdbde981d5b445e55b0d0f55b4) · reviewer: @taiberium

**TL;DR** — The gocryptfs setup script (pad + wrapped passphrase) went into the SSH command string, i.e. the remote shell's argv, and it and the passfile were written under `/tmp`, the container's writable layer on the provider's disk. The script now travels on the SSH channel's stdin and both files live on `/dev/shm`. Risk: the `docker exec -i … cat >` upload now depends on asyncssh `input=` delivering EOF, the same path `docker build` already uses.

**What changed**
- Setup script uploaded with `ssh_client.run(cmd, input=script)`; the command string is `docker exec -u 0 -i <pod> sh -c 'cat > <path>'` (`neurons/validators/src/services/docker_service.py`)
- Script and passfile paths moved from `/tmp/.x…` to `/dev/shm/.x…` (`_VOLUME_SETUP_TMPFS`) (`neurons/validators/src/services/docker_service.py`)
- `test_setup_encrypted_local_volume_does_not_log_key` asserts one stdin call, no heredoc, no key material or `/tmp` in any command (`neurons/validators/tests/test_docker_service.py`)
- Scripted SSH fake accepts `input=` (`neurons/validators/tests/test_encrypted_volume_wiring.py`)
- setup script created 0600 (`umask 077` before `cat`), like the passfile is chmod'd (`docker_service.py`)

**How to verify**
- `cd neurons/validators && pdm run pytest tests/test_docker_service.py tests/test_encrypted_volume_wiring.py` → 195 passed, 1 skipped (bwrap sandbox)
- On a rented encrypted pod: `tr '\0' ' ' < /proc/$(pgrep -x gocryptfs)/cmdline` → `-passfile /dev/shm/.x…`; `ls -a /dev/shm /tmp` → no `.x*` left

**Verified by the loop:** sandbox EC2 · Gate: PASS 023eef7

**Ran it:** `aws_verify.sh lium-io 023eef7 --only …/test_docker_service.py` → 192 passed, 1 skipped · EC2 warm box, run 20260911T075259Z-msu5

**Self-review:** clean at 023eef7 (9 checks)

**Parts:** backend ✓ (validator) · executor n/a — never sees the key · migration n/a — no schema change · frontend n/a — validator-internal · CLI/SDK n/a — no API or wire change · docs n/a — no renter-visible behaviour · tests ✓ 2 files · dashboards n/a — no new metric or log field

**Docs:** none changed in this PR — no renter- or provider-visible behaviour changes; the threat model is summarised under Details.

### Why

`setup_encrypted_local_volume` derives the per-pod passphrase on the validator (`VolumeKeyDeriver`, HKDF over `VOLUME_MASTER_SECRET` and the pod id) and mounts gocryptfs inside the rental container over SSH to the executor. Two steps handed the passphrase to the provider's host with no Inspector signal:

1. **argv.** The setup script was appended to the SSH command as a heredoc. sshd runs an exec request as `$SHELL -c '<command>'`, so the whole command — including `pad_hex` and `wrapped_hex`, from which one line of shell recovers the passphrase — is argv[2] of the remote shell while the exec runs. Every process on the host can read `/proc/<pid>/cmdline`; `execsnoop`, auditd `execve` records and any eBPF exec tracer capture it deterministically. The XOR wrap was designed to keep the raw passphrase out of `ps`; it does not help when both halves are in the same argv.
2. **disk.** Script and passfile were written to `/tmp/.x…` in the rental container. `/tmp` is the container's writable layer — on the node inspected, `upperdir=/var/lib/containerd/io.containerd.snapshotter.v1.overlayfs/snapshots/<n>/fs` on the provider's disk. `rm -f` unlinks; the blocks stay until reused (`debugfs`, `xfs_undelete`). This happens at create and at every remount (stale-mount recovery, `start_container`).

Observed on a live rental (9 Sep 2026 04:14Z, RTX 4090, $0.30/h, killed after 8 min): `gocryptfs -fg -notifypid=432 /lium-cipher /root -passfile /tmp/.xd15728af -o allow_other -nonempty`; `/dev/shm` mounted as a 14 GB tmpfs in the same container; `uid_map 0 165536 65536` (sysbox userns).

### The change

- `_VOLUME_SETUP_TMPFS = "/dev/shm"`: Docker and sysbox mount a tmpfs there in every container that does not use `--ipc=host|none` (Lium sets neither). The setup script's own `trap` still removes `$0` and `$_pf`; `wipe_tmp_files` still runs `rm -f` on both after the exec.
- `ssh_client.run(upload_cmd, input=setup_script)`: asyncssh writes the data to the channel and sends EOF, so `cat > <path>` inside `docker exec -i` terminates. Precedent in the same file: `docker build` gets its Dockerfile the same way (`docker_service.py` `input=dockerfile_content`).
- The 30 Jul design comment "No stdin" referred to the *script* not reading the passphrase from stdin (`read -r`); the script is unchanged and still reads nothing. The transport of the script is what moved.

### What this does not fix (the loop's renter-data-privacy write-up covers these)

- The sysbox user namespace (`uid_map 0 165536 65536`) is what keeps host uid 0 out of the plaintext: on a userns-remapped Docker host (EC2, kernel 6.18) `cat /proc/<gocryptfs pid>/root/<mount>/secret` is `Permission denied` for both the `allow_other` and the plain mount, while `docker exec` and `nsenter -U -m` (both traced by the Inspector) read it. The FUSE module parameter `allow_sys_admin_access=1` disables that check on the host; nothing reports it today.
- The passfile and the setup script on `/dev/shm` are still readable by host root for the second they exist: Docker keeps the container's shm as a host-side tmpfs at `/var/lib/docker/containers/<id>/mounts/shm`, so the provider can watch that directory directly, without `/proc/<pid>/root` (taiberium, 11 Sep). This change removes the on-disk copy and the argv copy, not the live window.
- Bootstrap restore (`_run_bootstrap_restore`): #1341 (DAH-3274) takes `volume_passphrase` out of the operation spec and deletes `encrypted_bootstrap`. The spec itself is still SFTP'd to `/root/app/storage-operations/<id>.json` on the provider's disk with the bucket's access key, secret and session token — DAH-3397 (filed 11 Sep from taiberium's second pass).
- The passphrase is in the gocryptfs process memory for the life of the pod; `/proc/<pid>/mem`, ptrace and `process_vm_readv` are traced by the Inspector, `/proc/kcore`, `/dev/mem` and kernel modules are not.
- Host swap: `/dev/shm` pages and gocryptfs memory (no mlock) can be swapped out on a node with swap enabled — the same exposure the process memory already has, not a new one.

### Verification log

- EC2 warm box (c6i.2xlarge), run `20260909T043719Z-zd5x` at the pushed head: `tests/test_docker_service.py tests/test_encrypted_volume_wiring.py` → 195 passed, 1 skipped (`test_gocryptfs_setup_script_xor_material_no_stdin`, needs bwrap), ruff format report-only.
- Not run: a create of an encrypted pod through this validator code (needs a staging validator with a sysbox executor). The two behaviours the change relies on were checked separately: `/dev/shm` tmpfs present in a live pod; `input=` transport already in production for `docker build` (asyncssh `create_process` writes the input and sends EOF).
- Self-review round 1 (pre-push) returned six low findings; the two code nits (stale "No stdin" comment, `-i` not pinned by the test) are folded into this commit, the body nits above.

### Self-review

Verdict `clean` at `023eef7` · 9 checks · by subagent-r106b · 2026-09-11 08:03Z (tools/pr_self_review.py)

Mechanical notes dismissed: `neurons/validators/src/services/docker_service.py:2561` Round-r106 low finding resolved: the comment now says the script 'carries the same material as the passfile the script chmods 600', which is what lines 755-756 do (`printf > "$_pf"` then `chmod 600 "$_pf"`); it no longer claims the passfile is 0600 from its first byte. `git diff 7a738803 023eef7a` is this one line only. · `neurons/validators/src/services/docker_service.py:2564` `shlex.quote` yields one single-quoted argument `'umask 077 && cat > /dev/shm/.x<hex>'`, so `sh -c` runs both in one shell and the umask applies to the redirection (verified in a scratch shell: file created -rw-------); nothing else runs in that exec, so the umask affects nothing else. · `neurons/validators/src/services/docker_service.py:2488` A pre-existing file at the target would keep its old mode (umask only governs creation, verified), but the path is `/dev/shm/.x` + 8 random hex from uuid4 in a container whose root is the data owner and whose previous run's files are removed by the script's trap and `wipe_tmp_files`; a 2^-32 collision with a leftover is not a real risk (optionally `set -C` would fail closed instead). · `neurons/validators/src/services/docker_service.py:2588` The mount step runs `docker exec -u 0 <pod> sh <script_path>` as uid 0 inside the container, the same identity that created the 0600 file, so it reads it; the script's `rm -f "$0"` also runs as root. · `neurons/validators/src/services/docker_service.py:2497` `wipe_tmp_files` runs `docker exec -u 0 <pod> rm -f <passfile> <script>` with check=False; unlink needs directory write permission on /dev/shm, not file mode, and root inside the container has it — unaffected by the umask. · `neurons/validators/tests/test_docker_service.py:5821` Not a tautology: dropping `umask 077 &&`, changing the value, or reordering it after `cat` fails the `startswith`; with a mocked ssh client the command string is the only observable, and the sandbox test for the script body cannot see the upload command, so string pinning is the right level. · `neurons/validators/tests/test_docker_service.py:2357` The only other `sh -c` / `cat >` prefix assertion in the test suite is for the SSH bootstrap path (`docker exec -i pod_test sh -c` writing `/tmp/lium-ssh-bootstrap.sh`), which this commit does not touch; `test_encrypted_volume_wiring.py` asserts no command prefix, and the `/tmp/.x` occurrences at 5848/5914 are sandbox-local paths, not ssh commands. · `neurons/validators/src/services/docker_service.py:2564` The commit's `umask` runs in the container's `sh` (dash/busybox/bash all have the POSIX builtin) and changes no other exec, no timeout, no retry or state — nothing new for CORRECTNESS/PERF beyond the one-shell question above.

### Verified

Gate: PASS (023eef7) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1338)

</details>


taiberium (approved):

taiberium (approved):

taiberium (commented):
Second pass. `input=` is a real asyncssh kwarg, the "docker build already does this" precedent is a live path, and nothing sets `--ipc`, so `/dev/shm` really is a per-container tmpfs. Failure branches are unchanged and `wipe_tmp_files()` still runs before every `fail_step`. Nothing blocks.

NIT: `storage_operations.py:91` with `docker_service.py:5539` — the bootstrap-restore spec carries `volume_passphrase` and the S3 keys and is written over SFTP to `/root/app/storage-operations/<id>.json` on the provider's filesystem. The title says "off the provider's disk"; this sibling path is not covered. Worth a follow-up ticket.

NIT: the body's "What this does not fix" understates the residual window. Docker keeps the container's shm as a host-side tmpfs at `/var/lib/docker/containers/<id>/mounts/shm`, so the provider can watch that directory directly and does not need `/proc/<pid>/root`.

NIT: `docker_service.py:2571` — the uploaded script holds the same key material as the passfile but is written with the default umask and never chmod'd, while the passfile gets 600.

PR #1323, opened 2026-09-08
DAH-3207 - [P2] executor /pod_logs: the signed string must be the container name

surcyf123:
Fixes [DAH-3207](https://app.notion.com/p/lium-io-executor-POST-pod_logs-accepts-only-a-miner-signature-made-over-the-requested-container-na-3d5b8bfdbde981fb9cc8f9558cf75c36) · reviewer: @jam6099

**TL;DR** — `POST /pod_logs` let `MinerMiddleware` verify the miner's signature over `data_to_sign` and then read `container_name` without comparing the two, so any string the executor was ever sent a valid signature for read any container's lifecycle events. The route now requires `data_to_sign` to be the requested container name, as `/upload_ssh_key` and `/remove_ssh_key` require `public_key == data_to_sign` (issue #744). Risk: none for the miner — it already signs exactly the container name.

**What changed**
- `_validate_pod_logs_consistency`: 400 `Container name mismatch` unless `data_to_sign.strip() == container_name.strip()`, before the store is read (`neurons/executor/src/routes/apis.py`)
- 6 tests through the real middleware and route: own request 200, SSH key or other name 400, stranger 401 (`neurons/executor/tests/test_pod_logs_signed_name.py`)

**How to verify**
- `cd neurons/executor && pdm run pytest tests/test_pod_logs_signed_name.py -v` → 6 passed; on `main` the three that expect 400 get 200
- e2e stack (lium-io#1312): `POST /pod_logs` signed by the stack's miner over an SSH public key, naming a pod → 400 (main: 200 `[]`); signed over the name → 200
- `pdm run pytest tests/ -q` → 134 passed

**Verified by the loop:** 8 Sep 2026 · sandbox EC2 · e2e podlogs probe 200 → 400 · targeted 19/19 · Result: PASS 0b54a451 · Gate: not run

**Ran it:** e2e stack, `POST /pod_logs` with the miner's signature over an SSH key naming a pod container: 200 `[]` → 400 `{"detail":"Container name mismatch"}`; signed over the name: 200 both times · EC2 20260908T050308Z-sdnk

**Self-review:** clean at 0b54a45 (15 checks, 3 low item(s) listed; subagent-r3b)

**Parts:** backend ✓ (executor) · migration n/a — no schema change · frontend n/a — no user surface · CLI/SDK n/a — executor API only · docs n/a — see Docs · tests ✓ 6 · dashboards n/a — a WARNING in the executor log

[...]

PR #1339, opened 2026-09-09
DAH-2834 - [P0] executor POST /verify: the verification suite from one…

surcyf123:
Implements [DAH-2834](https://app.notion.com/p/3cfb8bfdbde98140b3aefa0c2d657720) · reviewer: @taiberium

**TL;DR** — liumd phase 1, executor side, flag `EXECUTOR_LOCAL_VERIFY_ENABLED=false`: `POST /verify` runs the SSH checks' scripts from ONE validator-signed intent and answers ONE document. The intent arrives through the validator's own SSH session at the loopback port; a network peer is 403. Off → 404. Consumer: #1340 (tunnel client: its next round).

**What changed**
- `POST /verify` + `/version.capabilities`: signature over the body as sent; window 401, busy 409, replay 409, off 404
- Loopback peers only: a network peer is 403 before the body is read, nonce kept; `/version.local_verify_port` names the tunnel's port; uvicorn `proxy_headers=False`
- `LocalVerifyService`: the SSH path's scripts, each step its own result, deadline with child kill + reap
- Wire types `payloads/verify.py`: `extra="forbid"` (unknown field → 422); signing blob, schema id, capability defined ONCE in datura
- The signed `miner_hotkey` binds the intent to this miner's executors (another miner's → 401, nonce kept)
- CVM: the three settings are on `dstacktee/app/init_script.sh`'s whitelist and in `dstacktee/.env.example`

**Risks**
- #1340 as pushed posts over the network: with both flags on it is 403 and takes its SSH path; nothing fails. Both flags are off on prod.
- `config.py` settings change on the next deploy — rollback: revert and redeploy. uvicorn `proxy_headers=False`: nothing read forwarded headers before.

**How to verify**
- `cd neurons/executor && python -m pytest tests/test_local_verify.py -q` → 32 passed (network peer 403 and nonce kept, `/version` port, auth matrix, per-card challenges, deadline)

**Verified by the loop:** not yet run for this head.

**Ran it:** `tools/aws_verify.sh lium-io <head> --app io` → executor 263, miners 43, validators 2,284 passed (+2 box-baseline) · EC2 20260914T073938Z-r2ur · not run on macOS · … +1 under Details

**Self-review:** stale — head moved f366ef5 → 6a2e22b by pr_restack; re-run tools/pr_self_review.py

**Parts:** backend ✓ · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — consumer is #1340 · docs ✓ · tests ✓ 32 · dashboards n/a — metric in #1340

**Docs:** `neurons/executor/README.md` ✓ · `dstacktee/README.md` + `.env.example` ✓ · public: none needed because the flag is off by default and nothing changes for a provider.

Implements [DAH-2834](https://app.notion.com/p/liumd-v1-Rust-CLI-on-executor-answering-validator-commands-specs-verifyx-3cfb8bfdbde98140b3aefa0c2d657720) (liumd phase 1, part A — the executor side; part B is #1340) · reviewer: @taiberium (backup @jam6099)

### Origin
`design/LIUMD_FAST_PATH.md` (approved default): Python-first inside the existing executor image; one validator-signed, nonce'd intent → the executor runs the verification suite locally, in parallel where independent, and answers with ONE result document; the validator calls it once; the agent holds no key; the SSH path stays as the fallback. `speed/SWEEP_provider_verify.md` §1: the verification cycle is 25–40 SSH commands, of which the VerifyX run (80 s p50, 45 s at first-pass sizes) and the capability matmul (26.5 s, 10 s first-pass) are the two longest stages and are serial today. Standing priorities P90/P92/P93.3: sub-2-min verification, no more SSH round-trips.

### What this PR adds (flag `EXECUTOR_LOCAL_VERIFY_ENABLED=false`)
- `POST /verify` (`neurons/executor/src/routes/apis.py::local_verify`). Flag off → 404 (an image without the route answers 422 from `MinerMiddleware`; the validator treats every non-200 as "use SSH"). Auth is the route auth every validator-facing route here uses — `dependencies.auth.verify_signature` against the pinned `VALIDATOR_HOTKEY_SS58` — over the canonical JSON of the request body **as sent** (`canonical_intent_message(raw)`: sorted keys, no whitespace, minus `signature`), plus `check_intent_window` (issued_at within ±`LOCAL_VERIFY_INTENT_WINDOW_SECONDS`=120 s, expiry ≤ 2 windows + 60 s out) → 401, then busy → 409 (checked BEFORE the nonce is claimed, so a busy refusal does not burn the intent), then `NonceCache.claim` (single-use nonce until its expiry) → 409. The signed message is `datura.requests.validator_requests.local_verify_signing_blob`, the one definition the validator client (#1340) signs with. `/verify` is on `MinerMiddleware`'s `SELF_AUTHENTICATED_POST_PATHS` (main's #1323 allowlist) beside `/ping` and `/hardware_utilization` — all three carry their own signature check. The intent also names the miner it is for (`miner_hotkey`, under the signature): an executor whose `MINER_HOTKEY_SS58_ADDRESS` is another miner answers 401 before the nonce is claimed (`check_intent_target`), so an intent relayed to another provider's executor runs nothing there; the shared portal hotkey is not accepted in that field because every executor trusts it.
- `GET /version` now also returns `capabilities: ["local_verify/1"]` and `local_verify_port` (this process's `INTERNAL_PORT`, the loopback port the tunnel targets; the miner's `EXTERNAL_PORT` may differ) when the flag is on (neither otherwise) — the validator's version check. `/version` is plain HTTP, so a proxy can change the port: a wrong one only fails the tunnel's connect, and the validator runs its SSH checks as before.
- CVM executors: `dstacktee/app/init_script.sh` rebuilds `.env` from a whitelist at CVM start; the three settings (`EXECUTOR_LOCAL_VERIFY_ENABLED`, `LOCAL_VERIFY_MAX_DEADLINE_SECONDS`, `LOCAL_VERIFY_INTENT_WINDOW_SECONDS`) are on it and in `dstacktee/.env.example`, so a CVM keeps them.
- `services/local_verify_service.py::LocalVerifyService.run(intent)`: the steps the intent names, under one deadline (`min(deadline_s, LOCAL_VERIFY_MAX_DEADLINE_SECONDS)`), one suite at a time (`BusyError` → 409):
  - `matmul` — `decrypt_challenge.py <dim_n> <dim_k> <seed> <cipher_text>`, the exact script and arguments the validator's SSH path runs (`matrix_validation_service.py`); with `devices=[...]` one pinned run per CUDA device, each with its own `DeviceChallenge` cipher text (the all-cards proof; a shared or repeated one is 422), otherwise one run.
  - `verifyx` — `verifyx_executor.py <seed> <cipher_text>`, as `verifyx_validation_service.py` runs it; the answer carries `data.lib_sha256` of `libverifyx.so` so the validator can apply its outdated-library refusal.
  - `docker` / `ports` / `inspector` — facts: docker info + runtimes (sysbox) + disk usage + published ports + the number of ports in `RENTING_PORT_RANGE`/`RENTING_PORT_MAPPINGS` not published by docker + `libinspector.so` / `inspector_executor.py` presence and sha256. Blocking clients run in their own small thread pool (`_facts_executor`, not asyncio's default one) with a 20 s cap (`FAST_STEP_TIMEOUT_SECONDS`).
  - Independent steps run concurrently; VerifyX then matmul in the pipeline's order unless the intent says `parallel_gpu` (the validator sets it only at first-pass sizes — full-size VerifyX beside the matmul OOMs 64–128 GB hosts, sweep §3 #6). A step past the deadline is cancelled and its child process killed (`_kill`), and is reported as `timeout`; the other steps' evidence is still returned.
- The answer (`payloads/verify.py::VerifyResult`): schema id, echoed nonce + executor_uuid, executor version, `started_at`, `elapsed_ms`, `deadline_hit`, and per step `status` (ok | failed | timeout | skipped), `ms`, `exit_status`, raw `stdout` (capped), `stderr_tail`, `stdout_sha256`, `data`. **Executor signature: none** — `signer: "none"`, `signature: null`. No signing key exists on an executor today, and the design (§3) says one would prove only that the adversary's host signed; the field is there for a future key. What protects the unsigned answer is the channel: the validator posts the intent through a direct-tcpip channel of the SSH session it already holds (host key pinned to the TDX quote on a CVM) to the executor's loopback port, and `/verify` refuses any peer that is not loopback (403, before the body is read, no nonce claimed; `_is_loopback_client`). The miner's port-forward from the network therefore never carries the intent or the result, and cannot rewrite `lib_sha256` or a step's output. `executor.py` starts uvicorn with `proxy_headers=False`, so no `X-Forwarded-For` header can stand in for the peer. The document is evidence, not a verdict: the validator unseals and scores the outputs with the functions it already has.

### How to verify
- `cd neurons/executor && python -m pytest tests/test_local_verify.py -q` → 32 passed (route through FastAPI's `TestClient` behind a wrapper that sets the TCP peer, scripts and docker faked at the subprocess / client boundary — no GPU needed).
- Transport: a peer that is not loopback (docker gateway `172.18.0.1`, the CVM's slirp `10.0.2.2`, IPv4-mapped, a non-address) is 403 before the body is read and its nonce stays usable; `::1` is accepted (`test_a_network_peer_is_403_before_the_body_is_read_and_burns_nothing`). `GET /version` carries `local_verify_port` only while the flag is on (`test_version_advertises_the_capability_and_tunnel_port_only_when_enabled`).
- With the flag off: `GET /version` has `"capabilities": []` and `POST /verify` answers 404 (`test_version_advertises_the_capability_and_tunnel_port_only_when_enabled`, `test_flag_off_is_404`).
- Auth: wrong key or a tampered field → 401, an intent for another miner's executor → 401 with the nonce still usable, replayed nonce → 409, expired / skewed → 401, malformed → 422, a signature that is not hex → 400 as on `/ping` (`test_wrong_key_or_tampered_field_is_401`, `test_an_intent_for_another_miners_executor_is_401_and_not_burnt`, `test_replayed_nonce_is_refused`, `test_expired_or_skewed_intent_is_401`, `test_malformed_intent_is_422_before_any_signature_check`).
- Busy (409) is answered before the nonce is claimed, so the same signed intent stays valid; #1340 today takes the SSH path on any 409 (`busy_or_replay`) rather than waiting — a busy executor is already running a validator's suite, so that fallback loads the GPU twice; wait-and-resend is a #1340 follow-up, not phase 1.

### Ran it
`python -m pytest neurons/executor/tests/test_local_verify.py -q` on the sandbox EC2 warm box (aws_verify 20260914T073938Z-r2ur, whole `--app io`; 20260914T075714Z-ywwx, the executor tests at f366ef5 → 263 passed) → 32 passed; the same file runs the signed intent end to end through `TestClient` (`test_signed_intent_runs_the_suite`): flag on, `GET /version` → `capabilities: ["local_verify/1"]`, `POST /verify` with a validator-signed intent → 200, five steps, `matmul`/`verifyx` `status: ok` with the fake scripts' stdout and sha256, `deadline_hit: false`, `signer: "none"`; the same intent from a network peer → 403 and the nonce is still accepted from loopback afterwards (`test_a_network_peer_is_403_before_the_body_is_read_and_burns_nothing`).
Not run on a GPU host: the executor image is built and released by the fleet pipeline; a real node cannot run this branch without a fleet release (said in the brief — "probably not"; confirmed: no).

### Stage budget (design ≈ 70–95 s to AVAILABLE) vs what phase 1 gives
Sweep §2 express-lane budget today: 162 s p50 to AVAILABLE, of which VerifyX 45 s + matmul 10 s serial. With this route + #1340: the two run side by side at first-pass sizes → max(45, 10) = 45 s, and the SSH commands of those two stages (script upload + run + kill; 3–4 of the cycle's 25–40) are one HTTP call: **≈ −10 to −12 s p50 → ≈ 150 s**. The design's 70–95 s needs phase 2 — the docker/ports/inspector facts this route already returns being consumed by the port-connectivity / scrape checks and the rental probe running beside the GPU steps — which is not in this PR (flag-gated follow-ups). Measured on the stub (#1340 `test_measure_one_call_against_the_ssh_sequence_on_the_stub`, 150 ms RTT, 0.4 s per step): SSH sequence 1.26 s → one call 0.40 s.

### Not in this PR
- An executor signing key (design §3: no).
- The validator consumer (#1340) and any check consuming the docker/ports/inspector facts (phase 2). #1340's pushed head still posts `/verify` to the executor's network address: with both flags on it is answered 403 on every cycle and takes its SSH path (any non-200 other than 404 is `http_error` → fallback). Opening the direct-tcpip channel on the session `InteractiveShellService` already holds, to `local_verify_port`, is #1340's next round.
- Checking the intent's `executor_uuid` against this executor: the executor has no setting for its own uuid; the validator refuses an answer with the wrong uuid. The binding to this host is the signed `miner_hotkey` (above); an executor configured with a miner other than the one the validator lists it under answers 401 and the validator takes SSH — a misconfiguration that shows up as `refused` in #1340's outcome log, not as a failed verification.
- The kernel-fault probe (main's #1297, `gpu_fault_probe.py` shipped over SSH stdin after the matmul) stays on SSH; a stdin-shipped script is a phase-2 question.

### Self-review

Verdict `findings` at `f366ef5` · 9 checks · by subagent-fresh-r126 · 2026-09-14 07:57Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | STALE-BODY | `neurons/executor/tests/test_local_verify.py:44` | Property: a run id in the body was made at the head it is attributed to. The Details `### Ran it` block says run 20260914T075011Z-qd5u ran 'executor tests at the pushed head', but qd5u ran at bbfb69b; the only change since is this one-line removal of an unused import, which cannot alter collection or outcome, so 263 / 32 still hold — the attribution is what is off. | Write 'at bbfb69b (this head differs by one removed unused import)' or replace it with the f366ef5 executor run once it exists. |

Mechanical notes dismissed: `neurons/executor/tests/test_local_verify.py:44` Delta bbfb69b..f366ef5 is exactly the removal of the unused `check_intent_target` import (1 file, -1); no other reference to the name remains in the test module, so the round-2 CODE-QUALITY low is resolved and nothing else changed. · `neurons/executor/tests/test_local_verify.py:530` Body names hold at this head: `test_version_advertises_the_capability_and_tunnel_port_only_when_enabled` (line 530) and `test_a_network_peer_is_403_before_the_body_is_read_and_burns_nothing` (line 539) exist; the old `_only_when_enabled` name is gone from the body. · `neurons/executor/tests/test_local_verify.py:1` Counts hold: 30 test functions + a 3-way parametrize = 32 collected; fold and Details both say 32 passed; executor 263 at r2ur (5647037) and qd5u (bbfb69b), unchanged by an import removal. · `neurons/executor/README.md:1` Run ids: the fold's Verified and Ran-it lines cite r2ur (whole --app io: executor 263 / miners 43 / validators 2,284 + 2 box-baseline fails, verified in RECOVERY/aws_run/20260914T073938Z-r2ur/job.log); Details cites r2ur and qd5u (RECOVERY/aws_run/20260914T075011Z-qd5u/job.log: executor 263 passed); the Ran-it command uses `<head>` rather than a sha literal, as the brief asks. · `neurons/executor/src/routes/apis.py:427` Self-review / Gate lines still read 65d8744: rewritten by --record and the gate at the pushed head (PR_PROCESS §7), not an author claim.

### Verified

Gate: not run at 6a2e22b — head moved; re-run the quality gate

</details>


neurons/executor/src/payloads/verify.py
```diff
@@ -0,0 +1,101 @@
...
+
+class VerifyIntentBody(BaseModel):
+    """What the validator signs. `signature` covers the canonical JSON of these fields."""
+
+    # Only this schema is understood: another version is refused (422) rather than run as v1.
+    schema_id: Literal["lium.local_verify/1"] = Field(alias="schema", default=SCHEMA)
+    nonce: str = Field(min_length=16, max_length=128)
+    issued_at: int  # unix seconds, validator clock
+    expires_at: int  # unix seconds; refused after this
+    executor_uuid: str = Field(min_length=1, max_length=128)
```
taiberium, line 91: NIT: The executor never compares this uuid with its own, so a signed intent also runs on another host.

taiberium (approved):

neurons/executor/src/payloads/verify.py
surcyf123: Right, and it cannot: the executor does not hold its uuid. The miner assigns it in its own DB; the executor process never learns it. A uuid set through the executor env would come from the same party a check would guard against. A miner could relay an intent to another host either way. The field is an echo: the result carries it back, so the validator can match the answer to the executor it addressed. Binding a run to a machine is what the attestation flags are for (TDX, GPU attestation). Code left unchanged; I can add a comment on the field if you prefer that in the code.

neurons/executor/src/payloads/verify.py
```diff
@@ -0,0 +1,101 @@
...
+
+class MatmulStep(BaseModel):
+    """The capability matmul challenge, exactly the arguments `decrypt_challenge.py` takes."""
+
+    dim_n: int
+    dim_k: int
+    seed: int
+    cipher_text: str = Field(min_length=1, max_length=4096)
+    # CUDA device indexes to pin one run to each (the all-cards work-proof); None = one unpinned run.
+    devices: list[Annotated[int, Field(ge=0)]] | None = Field(default=None, max_length=MAX_DEVICES)
```
taiberium: **Critical:** Every device gets the same `cipher_text`, so one real run can fake all cards. Give each device its own challenge.

neurons/executor/src/services/local_verify_service.py
```diff
@@ -0,0 +1,475 @@
...
+def parse_port_range(port_range: str | None, port_mappings: str | None) -> list[tuple[int, int]]:
+    """(internal, external) pairs the way the validator's `port_utils.get_all_ports` reads them."""
+    if port_mappings:
+        return sorted((int(i), int(e)) for i, e in json.loads(port_mappings))
+    if port_range:
+        if "-" in port_range:
+            lo, hi = (int(p.strip()) for p in port_range.split("-"))
+            return [(p, p) for p in range(lo, hi + 1)]
+        return sorted((int(p.strip()), int(p.strip())) for p in port_range.split(","))
+    return []
```
taiberium: **Critical:** When both settings are unset the validator counts all ports 20000-65536, not none. Return that default here too.

neurons/executor/src/services/local_verify_service.py
```diff
@@ -0,0 +1,475 @@
...
+            data={"per_card": cards},
+        )
+    return await run_script(matmul_argv(step, python), timeout=MATMUL_TIMEOUT_SECONDS)
+
+
+async def run_verifyx(step: VerifyXStep, *, python: str = sys.executable) -> StepResult:
+    result = await run_script(verifyx_argv(step, python), timeout=VERIFYX_TIMEOUT_SECONDS)
+    # The validator compares the library digest before it trusts a response (core/checksums);
+    # over SSH that is one more command, here it rides along.
+    result.data = {"lib_sha256": sha256_of_file(LIBVERIFYX_PATH)}
```
taiberium: NIT: This hashes a 10 MB file on the event loop. Move it into `_facts_executor`, like `_inspector_facts` does.

neurons/executor/src/payloads/verify.py
```diff
@@ -0,0 +1,101 @@
...
+    """One step's evidence. `status` is about the run, not the verdict: `ok` means the step produced
+    output the validator can judge; the judging happens on the validator."""
+
+    status: str  # ok | failed | timeout | skipped
+    ms: int = 0
+    exit_status: int | None = None
+    stdout: str | None = None
+    stderr_tail: str | None = None
+    stdout_sha256: str | None = None
+    data: dict[str, Any] | None = None
```
taiberium: NIT: `dict[str, Any]` hides the shape of a field two repos read. Give each step a typed model.

taiberium (changes requested):
Requested changes.

neurons/executor/src/payloads/verify.py
surcyf123: Fixed in 0b49363: `devices` is now a list of `DeviceChallenge(index, seed, cipher_text)` — each pinned run passes its own seed and cipher text (`matmul_argv(step, seed=…, cipher_text=…)`), so one real run unseals for one card only. A shared cipher text (between cards or with the step's own), or a repeated index, is refused at validation and answers a JSON 422 on the route (`test_a_shared_or_repeated_card_challenge_is_refused`, `test_a_shared_card_challenge_is_a_422_on_the_route`); `test_all_cards_pins_one_run_per_device_with_its_own_challenge` asserts each card's sealed output carries its own cipher text and seed. The validator side of the derivation (domain-separated from the intent nonce) lands with the phase that first sends `devices`.

neurons/executor/src/services/local_verify_service.py
surcyf123: Fixed in 0b49363: `parse_port_range(None, None)` returns `DEFAULT_PORT_RANGE = (20000, 65536)` as pairs — the same default `port_utils.get_all_ports` uses; `_port_facts` excludes the ssh port on both sides of the pair as the validator does. `test_no_configured_ports_means_the_validators_default_range` loads the validator's `port_utils.py` by path and asserts the two agree.

neurons/executor/src/services/local_verify_service.py
surcyf123: Fixed in 0b49363: the digest runs on `_facts_executor` (`run_in_executor`, bounded by FAST_STEP_TIMEOUT_SECONDS), started before the script so it is done when the script is; `test_the_verifyx_library_is_hashed_off_the_event_loop` asserts the thread is the `local-verify-facts` pool. A deadline that lands in the digest await is re-raised, never swallowed (a swallowed one let the GPU group run on — pinned by `test_a_deadline_during_the_library_digest_does_not_start_the_next_gpu_step`).

neurons/executor/src/payloads/verify.py
surcyf123: Fixed in 0b49363: `StepResult.data` is `MatmulData | VerifyXData | DockerFacts | PortFacts | InspectorFacts` (`payloads/verify.py`), each collector returns its model; the wire field names are unchanged, so #1340's reader is untouched (`test_signed_intent_runs_the_suite` round-trips the result through JSON and `VerifyResult.model_validate`).

neurons/executor/src/routes/apis.py
```diff
@@ -317,15 +327,93 @@ async def ping(_: None = Depends(verify_ping_signature)):
...
+    try:
+        intent = VerifyIntent.model_validate(raw)
+    except ValidationError as exc:
+        # `exc.errors()` carries the raising ValueError object in `ctx` for a model_validator
+        # refusal (a shared card challenge): serialised here, or the 422 would be a 500.
+        raise HTTPException(status_code=422, detail=json.loads(exc.json()))
+    # Signed as sent: the validator signs the document it puts on the wire, so a field it left at
+    # its default is not re-serialised here and a field it did send cannot be altered in flight.
+    await verify_signature(SignaturePayload(signature=intent.signature), canonical_intent_message(raw))
+    body = VerifyIntentBody.model_validate({k: v for k, v in raw.items() if k != "signature"})
```
taiberium: NIT: `VerifyIntent` already is a `VerifyIntentBody`, so this line validates the same fields twice. Pass `intent` to `check_intent_window` and `service.run`.

neurons/executor/src/payloads/verify.py
```diff
@@ -0,0 +1,190 @@
...
+    containers: list[ContainerFact]
+
+
+class PortFacts(BaseModel):
+    port_range: str | None
+    port_mappings: str | None
+    configured: int
+    sampled: int
+    published_by_docker: list[int]
+    free: int
```
taiberium: NIT: `free` reads as bytes next to `free_bytes` in this file.

```suggestion
    free_ports: int
```

neurons/executor/src/payloads/verify.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class VerifyIntent(VerifyIntentBody):
+    signature: str = Field(min_length=1, max_length=1024)
+
+
+class ScriptRun(BaseModel):
+    """What one script run produced: exit status, capped stdout, the stderr tail."""
+
+    status: str  # ok | failed | timeout | skipped
```
taiberium: NIT: four fixed values, and a comment to say which. A `Literal` says it in the type.

```suggestion
    status: Literal["ok", "failed", "timeout", "skipped"]
```

neurons/executor/src/services/local_verify_service.py
```diff
@@ -0,0 +1,508 @@
...
+            gpu["verifyx"] = lambda: run_verifyx(steps.verifyx, python=self.python)
+        if steps.docker:
+            facts["docker"] = lambda: run_facts("docker", _docker_facts)
+        if steps.ports:
+            facts["ports"] = lambda: run_facts(
+                "ports", lambda: _port_facts(self.port_range, self.port_mappings, self.ssh_port)
+            )
+        if steps.inspector:
+            facts["inspector"] = lambda: run_facts("inspector", _inspector_facts)
+        gpu_order = list(gpu)
```
taiberium: NIT: `gpu_order` holds step names, but a reader takes it for GPU indexes. `gpu_step_names` says it.

taiberium (approved):

neurons/executor/src/routes/apis.py
surcyf123: Done in `a04d33e`: the route passes the validated `VerifyIntent` to `check_intent_window` and `service.run`; the second `model_validate` is gone.

neurons/executor/src/payloads/verify.py
surcyf123: Done in `a04d33e`: `free_ports` in `PortFacts`, the service and the test.

neurons/executor/src/payloads/verify.py
surcyf123: Done in `a04d33e`: `status: Literal["ok", "failed", "timeout", "skipped"]`, comment dropped.

neurons/executor/src/services/local_verify_service.py
surcyf123: Done in `a04d33e`: `gpu_step_names` everywhere the list is built and read.

neurons/executor/src/payloads/verify.py
```diff
@@ -0,0 +1,199 @@
...
+    nonce: str
+    executor_uuid: str
+    executor_version: str
+    started_at: int
+    elapsed_ms: int
+    deadline_hit: bool = False
+    steps: dict[str, StepResult]
+    # No executor signing key exists today (design §3: a key on the adversary's host proves only
+    # that the adversary's host signed). The field is here so a future key can fill it.
+    signer: str = "none"
```
taiberium, line 205: The HTTP response is unsigned, so a proxy can replace `lib_sha256` and bypass the validator's library gate. Please tunnel it through the existing pinned SSH connection.

neurons/executor/src/core/config.py
```diff
@@ -91,5 +91,15 @@ class Settings(BaseSettings):
     # host clock that runs ahead is treated like one that runs behind; wide enough for NTP drift.
     CONTAINER_SIGNATURE_MAX_AGE_SECONDS: int = Field(env="CONTAINER_SIGNATURE_MAX_AGE_SECONDS", default=300)
 
+    # liumd phase 1 (DAH-2834): `POST /verify` runs the verification suite locally from one
+    # validator-signed intent instead of 25–40 SSH commands. Off: the route answers 404 and
+    # `/version` does not advertise it, so validators keep the SSH path. On: only the signed,
+    # nonce'd, time-windowed intent of the pinned validator hotkey is accepted.
+    EXECUTOR_LOCAL_VERIFY_ENABLED: bool = Field(env="EXECUTOR_LOCAL_VERIFY_ENABLED", default=False)
```
taiberium, line 146: CVM startup drops these settings because `dstacktee/app/init_script.sh` rebuilds `.env` from a whitelist. Please add all three local-verify variables to that whitelist and `.env.example`.

taiberium (changes requested):
See the inline comments.

neurons/executor/src/payloads/verify.py
surcyf123: Done in `f366ef5`: `/verify` now answers loopback peers only, and `/version` names that port in `local_verify_port`. The validator opens a direct-tcpip channel on the SSH session it already holds and posts the intent through it, so the unsigned result never crosses the host's port-forward. A network peer gets 403 before the body is read, and its nonce stays usable. uvicorn runs with `proxy_headers=False`, so no forwarded header can pose as loopback. The client side of the tunnel is the next round on #1340; as pushed, #1340 gets 403 and runs its SSH checks.

neurons/executor/src/core/config.py
surcyf123: Done in `f366ef5`: all three settings are on the `init_script.sh` whitelist and in `dstacktee/.env.example`, with one line in the dstacktee README.

datura/datura/requests/validator_requests.py
```diff
@@ -73,6 +73,25 @@ def ssh_pubkey_signing_blob(public_key: str, nonce: str | None = None) -> str:
     return f"{public_key}\nnonce:{nonce}"
 
 
+# liumd phase 1 (DAH-2834): the validator's one-call verification intent, `POST /verify` on the
+# executor. Schema id of the wire documents and the capability string `GET /version` advertises.
+LOCAL_VERIFY_SCHEMA = "lium.local_verify/1"
+LOCAL_VERIFY_CAPABILITY = "local_verify/1"
+
+
+def local_verify_signing_blob(intent: dict) -> str:
```
taiberium: NIT: a bare `dict` on the one function both repos sign with hides the shape.

```suggestion
def local_verify_signing_blob(intent: dict[str, object]) -> str:
```

neurons/executor/src/routes/apis.py
```diff
@@ -356,15 +369,132 @@ async def ping(_: None = Depends(verify_ping_signature)):
...
+            max_deadline_s=settings.LOCAL_VERIFY_MAX_DEADLINE_SECONDS,
+            port_range=settings.RENTING_PORT_RANGE,
+            port_mappings=settings.RENTING_PORT_MAPPINGS,
+            ssh_port=settings.SSH_PORT,
+        )
+    return _local_verify_service
+
+
+@apis_router.post("/verify")
+async def local_verify(request: Request):
```
taiberium: NIT: this handler is 55 lines and does two jobs. Move the guards (flag, peer, parse, signature, window, nonce) into `_admit_verify_intent(request)`.

taiberium (approved):

taiberium (approved):
All findings fixed and verified: the CVM whitelist carries the three local-verify variables, /verify answers loopback peers only through the validator's pinned SSH channel, and the guards are split out with the signing blob typed.

PR #1335, opened 2026-09-09
DAH-3264 - [P2] validator names a GPU benchmark that could not allocate VRAM

surcyf123:
Implements [DAH-3264](https://app.notion.com/p/validator-names-a-GPU-benchmark-that-could-not-allocate-VRAM-GPU_VERIFY_VRAM_UNAVAILABLE-instead-o-3d6b8bfdbde9815a9c99c29ece75df85) · reviewer: @taiberium

**Problem** — Speed sweep P91/P92, "misreported reasons": a GPU probe that could not allocate VRAM was reported as `GPU_VERIFY_FAILED` / "UUID mismatch … got 'None'".

**Value** — not measured — the body has no Worth it line (SO §69) and the ticket gives no prod number — a human decides.

**Goal** — [DAH-3264](https://app.notion.com/p/validator-names-a-GPU-benchmark-that-could-not-allocate-VRAM-GPU_VERIFY_VRAM_UNAVAILABLE-instead-o-3d6b8bfdbde9815a9c99c29ece75df85) (P2).

**Fits** — Sits in `lium-io` — `neurons/validators`; touches backend, docs, tests.

**Flag:** no flag

**Risks**
- None beyond the diff: no migration, no pod restart, no config change, no data rewrite.

**How to verify**
- `cd neurons/validators && pytest tests/test_capability_check.py tests/test_capability_check_mid_cycle_workload.py -q` → 25 passed

**Verified by the loop:** Gate: PASS e54cae7

**Ran it:** EC2 aws_verify 20260916T090246Z-xr7q @ e54cae7 (warm lium-io box i-08da00ecbef0b1f33), `cd neurons/validators && pytest tests/ -q` → 2636 passed, 15 skipped, 2 deselected in 181.01s
the…

**Self-review:** clean at e54cae7 (15 checks, 3 low)

**Parts:** backend ✓ · migration n/a — no schema · frontend n/a — portal renders these fields · CLI/SDK n/a — validator-internal · docs ✓ lium-docs main (#217, merged) · tests ✓ 4 (8 cases) + 3 of #1368's · dashboards n/a — same log line, new value

**Docs:** `GPU_VERIFY_VRAM_UNAVAILABLE` row is in `docs/providers/troubleshooting.md` §6 on lium-docs main (lium-docs#217, merged 10 Sep); main's default `help_uri` (#1287) points every error verdict at that table.

<details><summary>Details</summary>

### What changed

[...]

PR #1315, opened 2026-09-07
DAH-2958 - [P1] provider time-to-ready — 30 s portal snapshot, express lane

surcyf123:
Implements [DAH-2958](https://app.notion.com/p/Provider-time-to-ready-a-new-node-waits-for-the-15-min-cycle-boundary-and-then-for-the-slowest-mine-3d3b8bfdbde9816f907ac5ec52096c92) · reviewer: @taiberium

**TL;DR** — A new node waited a median 21 min for its first score: the 5-min portal snapshot hid it, publishing waited for the fleet-wide cycle, the first pass ran the full probes, a cold network sample failed the gate. Four changes, three behind flags off by default: 30 s snapshot, express lane, fast first pass, one VerifyX retry. Risk: flags off, nothing changes.

**What changed**
- `SNAPSHOT_TTL_SECONDS` 300 → 30 in the central miner's portal client (`miners/src/clients/miner_portal_api.py`)
- `EXPRESS_LANE_ENABLED` (off): verifies executors registered after the first cycle since start, alone; publishes spec-only (`core/express_lane.py`)
- `FIRST_PASS_FAST_PATH_ENABLED` (off): the express verification runs with `first_pass=True` → smaller probes (`services/miner_service.py`)
- `VERIFYX_COLD_SAMPLE_RETRY_ENABLED` (off): one more sample when a new node's first is under the gate (`checks/verifyx.py`)
- job files in a fresh `temp/cycle-*` per cycle; the lane's are kept while it reads them (`services/file_encrypt_service.py`)
- a wave-verified executor stays claimed until the cycle records its publish (`services/miner_service.py`)

**How to verify**
- `pdm run pytest tests/test_express_lane.py tests/test_first_pass_fast_path.py tests/test_verifyx_cold_sample_retry.py -q` → 49 passed
- `pdm run pytest tests/ -q --strict-markers` in validators → 1988 passed + main's own 2 failures; miners → 27 passed

**Verified by the loop:** 8 Sep 2026 · sandbox EC2 · validators 1988 passed · miners 27 passed · Result: PASS 284eb153 · Gate: PASS 0b5c576

**Ran it:** validators suite at 0b5c576 on sandbox EC2 (aws_run 20260908T073232Z-s20v) → 49 targeted, 1988 passed, main's own 2 fail · e2e stack at 284eb153 (aws_run 20260908T003721Z-7etq; this round changes no wire path) · not run on macOS (server code)

**Self-review:** clean at 0b5c576 (15 checks, 2 low)

[...]

PR #1312, opened 2026-09-07
DAH-3147 - [P1] the subnet loop as a compose stack and the e2e-gate check

surcyf123:
Implements DAH-3147 · reviewer: @jam6099

**TL;DR** — Nothing in CI ever runs a validator against an executor. `e2e/` is the subnet loop on one machine (real executor, real miner, the validator's services + pytest; the chain is the only stub) behind an `e2e-gate` job routed through #1311's `route` and required through `tests-ok`; green and red seen in CI. The GPU path is selectable, not yet run on a GPU host.

**What changed**
- `e2e/` (29 files): dind, executor, miner+db, redis, val-db, tester; suites `protocol` (14), `cycle` (5), `rental` (2); `gate.sh`
- `.github/workflows/test.yml`: `e2e-gate` with `needs: route` / `if: e2e == 'true'`; `tests-ok` needs it; concurrency per PR or sha; paginated comments
- `.github/actions/changed-packages`: `e2e` output (any package, or `e2e/`)
- `e2e/docker-compose.e2e.yml`, `stack.env`, `Makefile`, `gpu.yml`: `E2E_GPU` from the environment; sshd listens on the advertised port (2222, both paths) via a `command:` drop-in

**How to verify**
- `make -C e2e e2e-full` on a docker host → build, up, cycle 5/5, protocol 14/14 (incl. the SSH banner on the advertised port), rental 2/2, `artifacts/summary.md`
- green CI [34187108820](https://github.com/Datura-ai/lium-io/actions/runs/34187108820) (`e2e-gate` 6m1s); the `e2e-gate` job at the current head is the run for 50a2dcf

**Verified by the loop:** 8 Sep 2026 · sandbox EC2 · Gate: PASS 8babbac

**Ran it:** `gate.sh` CI run 34187108820 (GitHub-hosted Linux) → build 159 s, cycle 5/5, protocol 14/14, rental 2/2, `e2e-gate` PASS 6m1s · negative control (stock sshd :22) fails the listener test, EC2 apkd · 50a2dcf/8babbac judged by reading (no Docker on this Mac); its CI `e2e-gate` is the run · not run on a GPU host or macOS

**Self-review:** clean at 8babbac (15 checks, 5 low)

**Parts:** backend n/a — no neuron code changes · migration n/a — no schema · frontend n/a — none here · CLI/SDK n/a — not this repo · docs ✓ `e2e/README.md` · tests ✓ 21 e2e tests, CI green + red, 1 negative control · dashboards n/a — no metrics

[...]

PR #1340, opened 2026-09-09
DAH-2834 - [P0] validator consumes the executor's one-call /verify answer instead of…

surcyf123:
Fixes [DAH-2834](https://app.notion.com/p/liumd-v1-Rust-CLI-on-executor-answering-validator-commands-specs-verifyx-3cfb8bfdbde98140b3aefa0c2d657720) · reviewer: @taiberium

**TL;DR** — Validator side of liumd phase 1, flag `VALIDATOR_LOCAL_VERIFY_ENABLED=false`, based on #1339. One signed `POST /verify` replaces the SSH matmul + VerifyX sequence and rides the pipeline's SSH session to the executor's loopback port. A step counts only if it passed inside the SSH path's time cap.

**What changed**
- `LocalVerifyCheck` (`checks/local_verify.py`): one signed intent, SSH-function judgement, per-step fallback
- `LocalVerifyClient` (`services/local_verify_client.py`): the intent goes through `ssh.forward_local_port` to the executor's `local_verify_port`
- Tunnel failures → SSH: no `local_verify_port` → `no_tunnel_port`; 403, channel not opened or session gone → `refused`; bind failure → `transport`
- `build_intent` names the executor's miner (`miner_hotkey`, signed); step models in `datura/…/validator_requests.py`
- `CapabilityCheck` / `VerifyXCheck` consume a passing `ctx.state.local_verify`, else SSH
- Scope: phase 1b of DAH-2834 (epic open); CLI + full `specs` = #1343–#1346

**Risks**
- Flag off (default): none beyond the diff; no migration, no pod restart, no config change.
- Flag on: one direct-tcpip channel per `/verify` on the validator's SSH session; rollback `VALIDATOR_LOCAL_VERIFY_ENABLED=false`.
- sshd with `AllowTcpForwarding no` or no `local_verify_port`: `refused` / `no_tunnel_port`, every step over SSH, visible in the `[local_verify] outcome` line.
- `FIRST_PASS_FAST_PATH_ENABLED` off (default): the call carries VerifyX alone, the matmul stays on SSH (`scored_ssh`); the matmul saving needs both flags on.

**How to verify**
- `cd neurons/validators && python -m pytest tests/test_local_verify.py tests/test_verifyx_validation_service.py tests/test_first_pass_fast_path.py -q` → 76 passed (EC2 aws_verify 20260915T095252Z-xzsz at 4f4b8be)

**Verified by the loop:** Gate: PASS c78a255

**Ran it:** `pytest tests/test_local_verify.py`… · EC2 aws_verify 20260915T095252Z-xzsz (4f4b8be; the fix round c78a255 rewrites one test, EC2 run below) · … +1 under Details

**Self-review:** clean at c78a255 (carried from 4f4b8be: fix round +23 −17 in 1 file(s) answering the reviewer's comments, human re-reads; 15 checks, 2 low)

**Parts:** backend ✓ · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — validator-internal · docs n/a — flag off · tests ✓ 55 · dashboards n/a — Loki line only

**Docs:** none needed because the flag is off by default and the change is validator-internal.

Implements [DAH-2834](https://app.notion.com/p/liumd-v1-Rust-CLI-on-executor-answering-validator-commands-specs-verifyx-3cfb8bfdbde98140b3aefa0c2d657720) (liumd phase 1, part B — the validator side; part A is #1339) · reviewer: @taiberium (backup @jam6099)

### Origin
`design/LIUMD_FAST_PATH.md`: the validator sends ONE signed intent and consumes ONE answer instead of driving the verification over SSH; the SSH path stays as the fallback. `speed/SWEEP_provider_verify.md` §2/§3 #6: VerifyX (45 s at first-pass sizes) and the capability matmul (10 s) are serial stages of the express-lane budget (162 s p50 to AVAILABLE); running them side by side needs an executor-side call because the pipeline is one ordered list over an immutable `Context`.

### What this PR adds (flag `VALIDATOR_LOCAL_VERIFY_ENABLED=false`)
- `services/task/checks/local_verify.py::LocalVerifyCheck`, in the default pipeline after `GpuUsageCheck` and before `VerifyXCheck` / `CapabilityCheck` (not in the dry-run list). Flag off → one `LOCAL_VERIFY_DISABLED` event, nothing else. On: `GET /version` on the executor's API port; if `capabilities` lacks `local_verify/1` → `LOCAL_VERIFY_NOT_ADVERTISED`, SSH path unchanged; if it names no usable `local_verify_port` (an int in 1–65535) → fallback `no_tunnel_port`, nothing built or opened. Otherwise it builds the same two challenges the SSH checks would — `ValidationService.prepare_matmul_challenge` (first-pass `vram_budget_mb` as #1315) and `VerifyXValidationService.prepare_verifyx_challenge` (first-pass sizes as #1315) — signs one intent with the validator hotkey (`local_verify_client.py::sign_intent` over `datura.local_verify_signing_blob` — the function the executor verifies with; 32-hex nonce, 120 s TTL, `parallel_gpu` only on the first pass, `deadline_s` = `LOCAL_VERIFY_TIMEOUT_SECONDS` − 30 s so an answer the executor cut at its deadline still arrives) and `POST /verify`s it once through the pipeline's SSH session (connect `LOCAL_VERIFY_CONNECT_TIMEOUT_SECONDS`=5 s, total `LOCAL_VERIFY_TIMEOUT_SECONDS`=240 s; see Transport below). While `MATMUL_ALLCARDS_CHECK_ENABLED` is on the intent carries no matmul: the DAH-2671 all-cards work-proof (`_probe_all_claimed_cards`) runs only inside the SSH matmul path, so `CapabilityCheck` keeps running it over SSH and a consumed local pass can never skip the probe or its enforcement (`fallbacks.matmul = allcards_ssh`).
- The answer is judged by the functions the SSH path ends in — `evaluate_matmul_output` (unseal + uuid match) and `evaluate_verifyx_capture` (unseal + the RAM/disk/network parsing) — after the client checked it echoes the intent's schema, nonce and executor uuid. VerifyX's outdated-`libverifyx.so` refusal applies to the answer's `lib_sha256` as the SSH checksum gate does.
- Only a judged, **passing** step is left in `ctx.state.local_verify`; `CapabilityCheck` and `VerifyXCheck` consume it (`transport: "local_verify"` in the event) and skip their SSH run — the EMA / 100 Mbps / RAM / disk thresholds after that are the unchanged code, so the first-AVAILABLE decision uses the same thresholds. Everything else — refusal (401/403/404/409), a tunnel channel sshd cannot open, transport error, timeout, malformed or mismatched answer, a step that did not run, a step that ran and FAILED the judgement, an unexpected exception inside the check — leaves that step to the SSH path (`transport: "ssh"`). The new transport can save time; it cannot change a verdict on its own. When the DAH-2959 cold-sample retry (always SSH) beats a consumed local VerifyX answer, the event says `transport: "ssh"` (taiberium, 14 Sep).
- Metric per outcome: one `[local_verify] outcome` log line with `outcome` (consumed | fallback), `step` (call | matmul | verifyx), `reason` (ok, not_advertised, no_tunnel_port, not_supported, refused, busy_or_replay, timeout, transport, http_error, malformed, schema_mismatch, nonce_mismatch, executor_mismatch, step_failed, step_timeout, step_skipped, step_malformed, step_no_stdout, step_overtime, allcards_ssh, scored_ssh, lib_mismatch, local_failed, prepare_failed, cipher_generation_failed, no_keypair, internal_error), `round_trip_ms`, `executor_elapsed_ms`, `first_pass`.
- Refactor so both transports share code: `validate_gpu_model_and_process_job` = `prepare_matmul_challenge` → SSH → `evaluate_matmul_output` (native `gen_ptr` freed in `finally` on both paths); `validate_verifyx_and_process_job` = `prepare_verifyx_challenge` → SSH → `evaluate_verifyx_capture`. Behaviour of the SSH path is unchanged; its existing suites pass as they are.

### Transport (taiberium on #1339, 14 Sep: an unsigned HTTP result could be rewritten by a proxy on the miner's port-forward)
Since #1339 (merged into main as `c102a33`) the executor answers `POST /verify` for loopback peers only (403 before the body is read, uvicorn `proxy_headers=False`) and `GET /version` names `local_verify_port` (its `INTERNAL_PORT`). This PR's client (`LocalVerifyClient.verify`) therefore sends the intent inside the SSH session the pipeline already holds to the executor (`Context.ssh`, the `InteractiveShellService` connection; host key pinned to the TDX quote on a CVM): `asyncssh.SSHClientConnection.forward_local_port("127.0.0.1", 0, "127.0.0.1", local_verify_port)` binds an ephemeral listener on the validator's loopback; aiohttp POSTs to `http://127.0.0.1:<listener port>/verify`; sshd in the executor container opens a direct-tcpip channel to `127.0.0.1:<local_verify_port>`, where uvicorn sees a loopback peer. The listener is closed in `finally` after the one call. The transport is the same trust root as the SSH checks: whoever could read or change the answer could already run the SSH commands. `/version` is still read over the executor's API port: a proxy that rewrites the port can only aim the tunnel at another loopback port of the same container, a peer the miner already controls; whatever answers there is judged as the executor's own answer would be (nonce echo, unseal) or falls back to SSH. The validator-side listener (`127.0.0.1:0`) admits any connection from the validator container's own loopback for the life of the one call and forwards it to the executor's whole `INTERNAL_PORT`, not only `/verify`; that is the validator's own trust domain (the process that holds the hotkey), and the listener is closed in `finally`.
Failure modes, each a fallback and never a verdict: no `local_verify_port` (an image before #1339, or a rewritten field) → `no_tunnel_port`; the channel cannot open (nothing on that port, `AllowTcpForwarding no`) → asyncssh closes the local end without a byte (`SSHLocalForwarder` on `ChannelOpenError`), aiohttp raises `ServerDisconnectedError` → `refused`; a 403 from the route → `refused`; `forward_local_port` itself raising (a loopback bind failure) → `transport`; a session already closed binds the listener and fails at the channel open, so it is `refused` too; the deadline/`deadline_s`, echo checks, bounded reads and `allow_redirects=False` are unchanged. `LOCAL_VERIFY_CONNECT_TIMEOUT_SECONDS` bounds `/version` and the tunnel's local bind and connect; the channel open on the executor is inside the whole-call `LOCAL_VERIFY_TIMEOUT_SECONDS`. One label `refused` covers a 401 (signature, window, miner), a 403 (peer not loopback) and a channel sshd could not open; the `detail` field tells them apart.
Tests: `FakeSSH.forward_local_port` (a loopback listener piped to the destination like the direct-tcpip channel; a dead destination closes the local end), `FakeExecutor` refuses 403 a `/verify` whose peer did not come through the tunnel — `test_the_post_rides_the_ssh_session_and_the_network_path_is_refused` shows the old network POST as the 403 and the tunnelled one as the answer; `test_an_image_that_names_no_tunnel_port_leaves_everything_to_ssh` (port absent / `"8001"` / 0 / 65536 / `True`); `test_a_tunnel_nobody_answers_leaves_everything_to_ssh`; 403 in the refusal parametrisation; the consumed-path test asserts the forward targeted `127.0.0.1:<local_verify_port>` and that no listener outlived the call.

### How to verify
- `cd neurons/validators && python -m pytest tests/test_local_verify.py -q` → 55 passed: an in-process fake executor (`aiohttp` `TestServer` implementing `/version` + `/verify` with real signature checking and the loopback-peer refusal) reached through a fake asyncssh forward, the check + both consumers run in pipeline order, every refusal/mismatch/timeout parametrised to a fallback, the flag-off case, the per-step fallback, the overtime rule (a pass slower than the SSH path's cap for that step is left to SSH), the filler-only skip, the scored-cycle wire (serial, full-size VerifyX only: the matmul stays on SSH, `scored_ssh`, because the round trip would hold the serial VerifyX too and is no measure of the matmul), executor-controlled strings capped/closed before any label or event (answer and `/version` reads bounded at 2 MiB, no redirects), the exception guard.
- Equivalence: `test_ssh_and_local_transports_of_the_same_output_give_the_same_verdict` feeds the same `decrypt_challenge.py` stdout once through the SSH mock and once through the fake executor's `/verify` — same `ValidationResult` (success, uuid, metrics).
- Neighbouring suites unchanged: `tests/test_capability_check.py tests/test_verifyx_check.py tests/test_first_pass_fast_path.py tests/test_matrix_validation_service_precheck.py tests/test_matmul_vram_reserve.py tests/test_matmul_allcards.py tests/test_verifyx_validation_service.py tests/test_pipeline_default_scenarios.py tests/test_verifyx_cold_sample_retry.py` (166 passed with the new file).

### Stage budget (design ≈ 70–95 s to AVAILABLE) vs what phase 1 gives
Express-lane budget today (sweep §2): 162 s p50, VerifyX 45 s + matmul 10 s serial + their SSH commands (upload, run, kill — 3–4 of the cycle's 25–40). With #1339 on the executor and this flag on: one HTTP call, the two GPU steps side by side at first-pass sizes → max(45, 10) = 45 s: **≈ −10 to −12 s p50 → ≈ 150 s**. Not yet: the docker/ports/inspector facts the executor already returns are left in the answer and not kept — no check consumes them yet, so the outcome carries no field for them (taiberium's NIT; phase 2 adds the field with its reader) and port connectivity, scrape, the rental probe and main's kernel-fault probe (#1297, `GpuFaultProbeCheck` right after `CapabilityCheck`, stdin-shipped over `ctx.runner`, flag-gated — it runs whatever transport the capability event carries) still go over SSH, and the rental probe does not run beside the GPU steps — the first three are phase 2 and are what the design's 70–95 s needs.

### Not in this PR
- Consuming the docker/ports/inspector facts (phase 2, own flag).
- A capability cache per executor image (one `GET /version` per executor per cycle while the flag is on; ≤ 10 s on a dead host before SSH runs as before).
- Any change to scoring thresholds or to the SSH path's behaviour.

### Suite
`tests/test_scrape_infiniband.py::test_an_unreadable_sysfs_tree_is_not_reported_as_no_devices` and `tests/test_sshd_bootstrap_script.py::test_adopt_path_hardens_config_for_key_only_auth` fail on `origin/main` without this branch (root reads a chmod-000 tree; a real sshd runs on the box) — the same two #1315 and #1335 reported. Validators suite on EC2 at the pushed head (aws_verify 20260914T142759Z-nyim, tree of 7bffafa, rebased onto main after #1339 merged): 2,510 passed, 2 failed (those), 15 skipped; executor 296 passed, miners 43 passed. `tests/test_local_verify.py` alone: 55 passed (EC2 aws_verify 20260915T095252Z-xzsz at 4f4b8be: the three touched suites, 76 passed).

Cross-PR: the six stacked children (#1343 → #1344 → #1345 → #1346 → #1347 → #1348) change `local_verify_client.py` and `checks/local_verify.py` too and still call `client.capabilities(ctx.executor)` / `client.verify(ctx.executor, intent)`; this PR lands first and each child is rebased onto it with the call sites moved to `client.advertised(ctx.executor)` / `client.verify(ctx.ssh, port, intent)` (#1347's background `gpu_early` task opens and closes its own listener through the same `verify`). #1347 also replaces this PR's `_NothingToSend` with `_PrepareFailed`; it carries the rename.

### Ran it

`python -m pytest neurons/validators/tests/test_local_verify.py -q -s -k measure` on the sandbox EC2 warm box (aws_verify job liumd-b, see the Verified line; the same line on the author's Mac) → `MEASURE stub: ssh_sequence=1.26s (3 commands, RTT 0.15s, step 0.4s serial) one_call=0.40s (1 GET /version + 1 POST /verify, steps parallel) saving=0.85s`. The stub is the fake executor above with a 150 ms RTT and a 0.4 s step; it measures transport and ordering, not GPU time.
Not run on a real node. The e2e compose stack (#1312) is on main now and in this PR's base, but its `local_verify` path needs a GPU executor (`E2E_GPU=1`), which the loop's sandbox EC2 boxes do not have. A real node also needs an executor image with #1339, a fleet release. The judge itself is covered by the equivalence test above; a dev with a GPU host can run the compose stack with both flags on.

### Self-review

Verdict `findings` at `c78a255` · 15 checks · by subagent-r185-1340 · 2026-09-15 10:11Z (tools/pr_self_review.py) · carried from `4f4b8be`: fix round on that reviewed head, +23 −17 in `neurons/validators/tests/test_verifyx_validation_service.py` answering https://github.com/Datura-ai/lium-io/pull/1340#discussion_r4013474494 (PR_PROCESS §7 fast path: ≤ 40 lines, ≤ 3 files, no sensitive path; the human who asked re-reads)

|---|---|---|---|---|
| low | STALE-BODY | `neurons/validators/tests/test_local_verify.py:1463` | The body's Details still carry the absolute wording the code no longer claims — 'the scored-cycle wire (... the matmul stays on SSH, `scored_ssh`, because the serial VerifyX puts the round trip past the matmul cap)' — while `_matmul_ssh_reason` now says the round trip 'is no measure of the matmul' and fails a passing one 'on a large share of hosts'; and the counts disagree with each other: the verify line says 55 + 5 while the Details How-to-verify, Ran it and Parts still say 54 (this file gained test_a_scored_cycle_with_verifyx_off_sends_the_matmul_alone) and the neighbouring-suite count (166) predates the new test in test_verifyx_validation_service.py. | Reword the Details sentence to the trade-off (round trip holds VerifyX + matmul, no bound on the matmul alone; with VerifyX off the matmul rides alone) and re-count 54 → 55 in Details/Ran it/Parts and the neighbouring set at the pushed head. |
| low | DEAD-CODE | `neurons/validators/src/services/task/checks/local_verify.py:320` | `matmul_ssh_reason or "not_asked"`: matmul_challenge is None only when _prepare_challenges got matmul_on_ssh=True, i.e. matmul_ssh_reason is not None, so `not_asked` is unreachable and is not in the body's reason list. | Drop the `or "not_asked"` (an assert or a plain use of matmul_ssh_reason), or accept as a defensive default. |

### Verified

Gate: PASS (c78a255) on EC2 sandbox Linux x86_64 (aws_run gate)

</details>


taiberium (approved):

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -0,0 +1,348 @@
...
+def _over_time(name: str, answer: LocalVerifyAnswer) -> bool:
+    return answer.round_trip_ms > STEP_WALL_CLOCK_CAP_MS[name]
+
+
+@dataclass
+class LocalVerifyOutcome:
+    """What the consuming checks read. A field is set only when the local step ran AND passed the
+    validator's judgement; None means "run it over SSH"."""
+
+    matmul: Any | None = None  # matrix_validation_service.ValidationResult
```
taiberium: NIT: `Any` hides what these two fields hold. Use `ValidationResult` and `VerifyXResponse` from the two services this file already imports.

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -0,0 +1,348 @@
...
+    verifyx: Any | None = None  # verifyx_validation_service.VerifyXResponse
+    facts: dict[str, Any] = field(default_factory=dict)  # docker / ports / inspector evidence
```
taiberium: NIT: nothing reads `facts`. Only one test looks at it, so remove the field until a reader needs it.

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -0,0 +1,348 @@
...
+        if not settings.VALIDATOR_LOCAL_VERIFY_ENABLED:
+            return CheckResult(
+                passed=True, event=render_message(Msg.DISABLED, ctx=ctx, check_id=self.check_id)
+            )
+        try:
+            return await self._run(ctx)
+        except Exception as exc:  # noqa: BLE001 — the pipeline has no guard; a bug here must not end the node's cycle
+            return self._fallback(ctx, "call", "internal_error", f"{type(exc).__name__}: {exc}")
+
+    async def _run(self, ctx: Context) -> CheckResult:
```
taiberium, line 118: NIT: this method is 117 lines. Split it at the seam between the challenge preparation and the call.

taiberium (approved):

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `a8bad61`: `facts` and the loop that filled it are gone; the one test assertion on it as well. Phase 2 adds the field together with its reader; the body says so.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `a8bad61`: `_run` is 47 lines now: the early returns and the capabilities call. `_prepare_challenges` builds the two challenges and raises `_NothingToSend` with the fallback reason, closing a built matmul challenge first. `_call_and_judge` builds the intent, calls the executor and judges. The close of the matmul challenge stays in one try/finally around the call.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `a8bad61`: matmul and verifyx are typed ValidationResult and VerifyXResponse, optional; the helpers' challenge parameters carry MatmulChallenge and VerifyXChallenge the same way.

neurons/validators/src/core/config.py
```diff
@@ -241,6 +241,18 @@ class Settings(BaseSettings):
...
+    # capability matmul and VerifyX with ONE signed `POST /verify` instead of the SSH commands, and
+    # judge the answer with the same unseal/verify code. Any refusal, timeout, mismatch or local
+    # failure falls back to the SSH path for that step (logged per outcome, `[local_verify]`).
+    # Off by default; the SSH path is unchanged either way.
+    VALIDATOR_LOCAL_VERIFY_ENABLED: bool = Field(env="VALIDATOR_LOCAL_VERIFY_ENABLED", default=False)
+    # Whole-call budget for `POST /verify` (seconds). Sized for the first pass; a full-size VerifyX
+    # may take longer and then falls back, so keep this above VERIFYX first-pass p90 (90 s) + matmul.
+    # The executor is told to stop 30 s earlier (`deadline_s` = this − 30, min 5) so a cut answer
+    # with finished steps inside still arrives inside the budget.
+    LOCAL_VERIFY_TIMEOUT_SECONDS: int = Field(env="LOCAL_VERIFY_TIMEOUT_SECONDS", default=240)
```
taiberium, line 264: NIT: the SSH path allows VerifyX 600s, the local budget cuts it at 210s. Those hosts run VerifyX twice, so raise the budget or keep this path first-pass only.

neurons/validators/src/services/local_verify_client.py
```diff
@@ -0,0 +1,276 @@
...
+
+# The signed bytes: the wire document minus `signature`, sorted keys, no whitespace. The executor
+# rebuilds the same string from the request body it received — with this very function.
+canonical_intent_message = local_verify_signing_blob
+
+
+def build_intent(
+    *,
+    executor_uuid: str,
+    matmul: dict[str, Any] | None,
```
taiberium: NIT: `dict[str, Any]` on the wire. The executor already declares `MatmulStep` and `VerifyXStep`, so move them to `datura` and build the intent from them.

neurons/validators/src/services/local_verify_client.py
```diff
@@ -0,0 +1,276 @@
...
+        "nonce": secrets.token_hex(16),
+        "issued_at": int(now),
+        "expires_at": int(now) + INTENT_TTL_SECONDS,
+        "executor_uuid": executor_uuid,
+        "deadline_s": deadline_s,
+        "parallel_gpu": parallel_gpu,
+        "steps": {
+            "matmul": matmul,
+            "verifyx": verifyx,
+            "docker": True,
```
taiberium, line 137: NIT: nothing on the validator reads the docker, ports and inspector steps. Ask for them in phase 2, together with the reader.

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -0,0 +1,384 @@
...
+    statuses (failed | timeout | skipped; `malformed` from the parser for anything else), and
+    `step_no_stdout` for an `ok` step that carries nothing to judge."""
+    return "step_no_stdout" if step.status == "ok" else f"step_{step.status}"
+
+
+# The SSH path fails a matmul that runs past MATRIX_VERIFY_TIMEOUT_SECONDS and a VerifyX run past
+# VERIFYX_COMMAND_TIMEOUT_SECONDS. The executor's own per-step caps and `step.ms` are its word, so
+# the same bound is applied to the one clock the validator holds: the whole call's round trip,
+# which is an upper bound on any step's wall-clock (serial or side by side).
+STEP_WALL_CLOCK_CAP_MS = {
```
taiberium: NIT: the name reads as a cap on the step's own time, but it bounds the whole round trip.

```suggestion
ROUND_TRIP_CAP_MS_BY_STEP = {
```

neurons/validators/src/services/task/pipeline.py
```diff
@@ -125,6 +125,10 @@ class ContextState:
     # (not cached / no backend digest / unreadable RepoDigest — strict fail-open).
     recommended_image_digest_match: bool | None = None
     executor_image_report: ExecutorImageReport | None = None
+    # liumd phase 1 (DAH-2834): what `POST /verify` answered this cycle, already judged
+    # (checks/local_verify.LocalVerifyOutcome). None = not attempted or fell back entirely; the
+    # capability and VerifyX checks consume a judged step when present and run over SSH otherwise.
+    local_verify: Any | None = None
```
taiberium: NIT: `Any` hides what this holds. Move `LocalVerifyOutcome` to `services/local_verify_client.py` and import it here, the way `executor_image_report` does.

taiberium (approved):

neurons/validators/src/services/local_verify_client.py
surcyf123: Done in `243b836`: `MatmulStep`, `VerifyXStep`, `DeviceChallenge` and the forbid base live in `datura`; `build_intent` takes the models and the executor imports the same ones.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `243b836`: `ROUND_TRIP_CAP_MS_BY_STEP`, in the check, the module docstring and the tests.

neurons/validators/src/services/task/pipeline.py
surcyf123: Done in `243b836`: `LocalVerifyOutcome` lives in `services/local_verify_client.py`; `ContextState.local_verify` is typed with it.

neurons/validators/src/core/config.py
surcyf123: Not changing this because #1343, next in the stack, keeps the local call to the first pass. There VerifyX writes 16 GB and 1 GB and finishes inside the 210 s deadline. A full-size cycle never takes this path, so no host runs VerifyX twice.

neurons/validators/src/services/local_verify_client.py
surcyf123: Not changing this because #1345, three PRs up the stack, reads the docker, ports and inspector facts from this call. Asking for them here saves a second wire shape. The `build_intent` docstring now says who reads them.

neurons/validators/src/services/task/checks/verifyx.py
```diff
@@ -197,6 +207,7 @@ async def run(self, ctx: Context) -> CheckResult:
                     "verifyx_success": True,
                     "verifyx_network_success": verifyx_network.get("success"),
                     "network": speedtest_network,
+                    "transport": transport,
```
taiberium, line 213: When the SSH retry wins, this event still reports `local_verify`. Please set `transport = "ssh"` with `result = retry`.

taiberium (changes requested):
This phase does not implement the ticket's `liumd` CLI or full `specs` result. Please keep DAH-2834 open as a non-blocking follow-up.

neurons/validators/src/services/task/checks/verifyx.py
surcyf123: Done in `d61562e`: `transport` becomes `ssh` when the retry wins, and a two-case test pins the winning and the losing retry.

neurons/validators/src/services/local_verify_client.py
```diff
@@ -0,0 +1,385 @@
...
+            raise LocalVerifyUnavailable(
+                "refused",
+                f"tunnel to {EXECUTOR_LOOPBACK}:{local_verify_port} closed without an answer: "
+                f"{type(exc).__name__}",
+            )
+        except Exception as exc:  # other aiohttp client errors
+            raise LocalVerifyUnavailable("transport", f"{type(exc).__name__}: {exc}")
+        finally:
+            listener.close()
+            await listener.wait_closed()
```
taiberium: Please bound `listener.wait_closed()` with `connect_timeout_s`. On Python 3.12+, `asyncio.Server.wait_closed()` can wait for accepted connections after `close()`, and this await runs outside the request timeout. A stuck tunnel can therefore block the validator cycle instead of falling back to SSH.

taiberium (approved):
The retry reports transport=ssh now, and I pushed the wait_closed() bound with a regression test.

neurons/validators/src/services/local_verify_client.py
surcyf123: Your bound is the tip of this branch now: your commit came through the rebase onto main as `7bffafa`, unchanged.

neurons/validators/src/services/verifyx_validation_service.py
```diff
@@ -141,10 +141,130 @@ class VerifyXResponse:
...
+            "cipher_text": cipher_text,
+            "challenge_input": challenge_input,
+        }
+        return VerifyXChallenge(
+            validator=verifyx_validator,
+            seed=seed,
+            cipher_text=cipher_text,
+            challenge_input=challenge_input,
+            log_extra=log_extra,
+            expected_lib_sha256=sha256_from_path(self.lib_name),
```
taiberium, line 221: NIT: `validate_verifyx_and_process_job` already hashed this file two lines up, so every SSH run reads the .so twice. Cache the digest on the service.

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -0,0 +1,385 @@
...
+# VERIFYX_COMMAND_TIMEOUT_SECONDS. The executor's own per-step caps and `step.ms` are its word, so
+# the same bound is applied to the one clock the validator holds: the whole call's round trip,
+# which is an upper bound on any step's wall-clock (serial or side by side).
+ROUND_TRIP_CAP_MS_BY_STEP = {
+    "matmul": MATRIX_VERIFY_TIMEOUT_SECONDS * 1000,
+    "verifyx": VERIFYX_COMMAND_TIMEOUT_SECONDS * 1000,
+}
+
+
+def _over_time(name: str, answer: LocalVerifyAnswer) -> bool:
```
taiberium, line 82: NIT: on a scored cycle the serial VerifyX pushes the round trip past the 120s matmul cap, so a passing matmul always falls back. Gate this on `first_pass`.

taiberium (approved):

neurons/validators/src/services/verifyx_validation_service.py
surcyf123: On it now. Fix and reply by 10:20Z.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: On it now. Fix and reply by 10:20Z.

neurons/validators/src/services/verifyx_validation_service.py
surcyf123: Done in `4f4b8be`: `lib_sha256()` reads the .so once per service; the checksum gate and the challenge share the digest. The test hashes a real file and rewrites it between two runs (c78a255).

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `4f4b8be`: past the first pass with VerifyX on, the call no longer asks for the matmul (`scored_ssh`); with VerifyX off the matmul still rides the call alone. Two tests pin both shapes.

PR #1341, opened 2026-09-09
DAH-3274 - [P1] encrypted restore via the pod's mount; passphrase never reaches executor

surcyf123:
Implements [DAH-3274](https://app.notion.com/p/3d6b8bfdbde981518274c4a4e745b178) · reviewer: @taiberium

**Problem** — `lium up --restore-backup` into an encrypted volume shipped the pod's gocryptfs passphrase to the executor (`docker run -e`, readable in `docker inspect`).

**Goal** — [DAH-3274](https://app.notion.com/p/3d6b8bfdbde981518274c4a4e745b178) (P1) · serves B-146, B-144.

**Fits** — Sits in `lium-io` — `neurons/validators`, `neurons/executor`; touches backend ((validator)), executor, tests (4 files, 15 tests).

**Flag:** no flag

**Risks**
- Renter-visible: `lium up --restore-backup` with encryption on (the CLI default) now fails at create, step `bootstrap_restore_hold`; the `lium bk restore` hint is in the failure detail. Plain volumes, online restores and backups unchanged.

**How to verify**
- `pdm run pytest tests/test_docker_service.py tests/test_storage_operations.py -q` (validators) and `pdm run pytest tests/test_storage_runner.py -q` (executor) → pass

**Verified by the loop:** Gate: PASS 55a72e3

**Ran it:** `aws_verify.sh lium-io --app io` on EC2 (run ily1) at this head: validators 2343 passed (2 box-bound failures, same on main), executor 239 passed

**Self-review:** clean at 55a72e3 (15 checks, 1 low)

**Parts:** backend ✓ (validator) · executor ✓ · migration n/a — no schema · frontend n/a — validator internal · CLI/SDK n/a — no wire change · docs n/a — encrypted-volumes.md:31 already says encrypted restores need a running pod (B-146) · tests ✓ 4 files, 15 tests · dashboards n/a — no new metric

**Docs:** none needed because `pod-users/encrypted-volumes.md:31` already says backups and restores of an encrypted volume only work while the pod is up, which is what the hold enforces.

### What changed

- Encrypted create-time restore refused in the preflight, before probes, volume and pod, while `_ENCRYPTED_BOOTSTRAP_RESTORE_ON_HOLD` is set (`services/docker_service.py`)
- Held path kept: restore after the pod's mount, before key injection, as `encrypted_running` + `bootstrap: true`, never the passphrase (`services/`)
- `encrypted_bootstrap` mode, passfile helper and `-e LIUM_VOLUME_PASSPHRASE` removed; `volume_passphrase` refused (`neurons/executor/src/storage/`)
- Tests: hold refusal with nothing created; held path order and both probes with the hold lifted; no passphrase in argv or spec (`validators/tests/`, `executor/tests/`)

### Why

`design/RENTER_DATA_PRIVACY.md` row 4 (9 Sep 2026). The validator derives one passphrase per pod (`HKDF(VOLUME_MASTER_SECRET, pod_id)`) and, for every path but one, keeps it inside the validator→pod channel. The exception was the bootstrap restore: `_run_bootstrap_restore` put `volume_passphrase` into the storage-operation spec, `start_storage_operation` SFTP'd that JSON into the executor container (a file on the provider's disk), the executor's `WorkspaceResolver` turned it into `DockerEncryptedVolumeWorkspace`, and `ResticStorageRunner` exported it as `LIUM_VOLUME_PASSPHRASE` and started the helper with `docker run … -e LIUM_VOLUME_PASSPHRASE`. Docker copies forwarded env into the container config, so `docker inspect lium-storage-*` and `/var/lib/docker/containers/<id>/config.v2.json` held the key in clear for as long as the helper existed — and the helper's `ENCRYPTED_BOOTSTRAP_SCRIPT` wrote it to `/tmp/.lium-volume-passphrase` inside the helper too. The key is static for the pod's life: one restore = the provider can decrypt that volume forever. `docker inspect` is suppressed by the Inspector as inventory, so nothing logged it.

### The restore path now

0. **On hold (14 Sep, taiberium):** steps 2 to 5 run while the image's entrypoint is already up, so the entrypoint and the restore write under `/root` at the same time. Until the workload can be held stopped for the whole restore, `create_container` refuses an encrypted `--restore-backup` in the preflight, before the probes, the volume and the pod: step `bootstrap_restore_hold`, message "an encrypted volume cannot be restored at create time yet … run `lium bk restore` once it is up". The hold is the module constant `_ENCRYPTED_BOOTSTRAP_RESTORE_ON_HOLD`, not a setting. `test_create_container_refuses_an_encrypted_restore_while_the_hold_is_on` pins the refusal and goes when the constant goes; the three tests of steps 1 to 5 lift the constant (`_lift_encrypted_bootstrap_restore_hold`) and are what the follow-up inherits unchanged. A plain volume is not affected. What the renter sees today: the create fails with the fixed headline and `failure_step: bootstrap_restore_hold`; the sentence naming `lium bk restore` is the failure detail (validator log, restore log), because lium-platform renders only the headline and the step. The docs already say restores of an encrypted volume only work while the pod is up (`apps/docs/docs/pod-users/encrypted-volumes.md:31`), so the hold matches what renters were told.
1. `create_container` creates the volume; a **plain** volume is still restored before `docker run` (`mode: plain_volume`), so the image's entrypoint finds the data.
2. For an **encrypted** volume the restore is skipped there. `docker run` starts the pod with the ciphertext volume at `/lium-cipher`; `setup_encrypted_local_volume` mounts gocryptfs over `/root` inside the pod (passphrase over the SSH channel's stdin onto `/dev/shm`, #1338).
3. `_run_bootstrap_restore(encrypted=True, container_name=…)` sends `mode: encrypted_running` + `container_name`. The executor's `_resolve_encrypted_running` does `docker inspect` (cipher volume at `/lium-cipher`, running, cgroup matches), checks `/proc/<pid>/mountinfo` for the live `fuse.gocryptfs` at `/root`, checks the target is a directory (a `bootstrap: true` restore may write over what the entrypoint left; an online restore still needs an empty target), then runs restic through `nsenter -t <pid> -U` into the pod's user namespace with `--target /proc/<pid>/root/root/…`. No key is involved; the sysbox user namespace is what lets the helper see plaintext.
4. Key injection (`add_ssh_public_keys_with_rental_docker`) and the sshd bootstrap follow, as before — after the restore, so the backup cannot clobber the customer's fresh `authorized_keys` (pinned by `test_create_container_bootstrap_restore_order_follows_the_volume_kind`, which lifts the hold of step 0). The pod's PID 1 is not paused: an image that starts Jupyter at boot may drop `.jupyter` or `.local` into the fresh mount between the mount and the restore. Nothing there is the customer's, so the create-time spec carries `workspace.bootstrap: true` and the executor only requires the target to be a directory; restic restores with its default `--overwrite`: the backup's version of any path present in both wins, entries the entrypoint wrote that are not in the snapshot are left in place (no `--delete`). An executor image without the key would ignore it and keep the strict check, so the validator asks first: before the volume and the pod exist, `supports_bootstrap_restore` runs the executor's python against its own `storage.models` (`'bootstrap' in WorkspaceSpec.__dataclass_fields__`); a `no` fails the create at step `bootstrap_restore_probe` with "the provider must update the executor image" — no fallback: the old encrypted mode would hand the executor the passphrase, and a plain-volume restore would put the backup's plaintext on a disk the renter asked to encrypt — the code comment at the probe says why the label check above downgrades and this does not.
5. Failure inside step 3 goes through the existing post-run `except` (`cleanup_failed_container_creation` removes the container and the volume; `failure_step=bootstrap_restore`).

Executor side: `WorkspaceMode.ENCRYPTED_BOOTSTRAP`, `DockerEncryptedVolumeWorkspace`, `_resolve_encrypted_bootstrap`, `ENCRYPTED_BOOTSTRAP_SCRIPT`, `_encrypted_volume_helper_command`, the `-e LIUM_VOLUME_PASSPHRASE` forward and the env/`_secret_values` plumbing are deleted. `WorkspaceSpec.from_mapping` refuses any spec that carries `volume_passphrase` before parsing the rest, so an old validator against a new executor fails loudly (`OperationSpecError`) instead of restoring anything.

### Compatibility

- New validator → old executor: the `encrypted_running` mode itself has been in the executor since restic backups shipped (used by online `lium restore` today) — that mode needs no image bump. The `bootstrap` key this PR adds does: see the next bullet and the release order.
- Old validator → new executor: the spec carries `volume_passphrase`, which `from_mapping` refuses before it reads the mode (`workspace must not carry volume_passphrase: …`, an `OperationSpecError`); the restore fails, the key never lands.
- New validator → old executor, `bootstrap` key: probed before `docker run` (step 4); an image without it fails the create cleanly, nothing is created on the box. **Release order (taiberium, 10 Sep): `executor-v*` before `validator-v*`** — the probe reads a field this PR adds to the executor's models, so a new validator against a fleet of old executors refuses every encrypted `--restore-backup` create until the executors update. The probe exits 3 for a missing field (old image) and the create says so; any other exit (no interpreter at `python_path` = 127, a broken import = 1) is reported with the exit status and the stderr tail, not read as an old image (`supports_bootstrap_restore`, `test_a_probe_that_cannot_run_is_reported_not_read_as_an_old_image`).
- No setting: the only alternative behaviour is the leak, so there is no safe "off" state to gate. `ENABLE_VOLUME_ENCRYPTION` still gates encryption itself. The 14 Sep hold is a constant in the code, not a setting, for the same reason.

### 11 Sep second pass (taiberium), what changed in that round
- The engine binary (`supports_storage_operation`) is probed next to the models field, before the volume and the pod exist; new models + no restic no longer fails with the pod built (`docker_service.py`, `test_create_container_encrypted_restore_stops_before_docker_run_without_the_engine`).
- `workspace.bootstrap` on `action: backup` is refused at parse time instead of being ignored (`models.py`, `test_bootstrap_flag_is_refused_outside_encrypted_running`).
- The `workspace.py` comment now says the winner flipped (entrypoint used to write last; the backup does now).
- Not changed: a wall-clock deadline on `wait_for_storage_operation`. A restore of a large volume has no bounded duration, so a fixed cap would kill legitimate restores; the runner-liveness check already ends a dead runner. A stall deadline (no log growth for N minutes) is the right shape and is a follow-up on the ticket, not this PR.

### 14 Sep third pass (taiberium), what changed at HEAD
- Encrypted create-time restore refused in the preflight until the pod workload can be held stopped throughout the restore (step 0 above; `docker_service.py`, `test_create_container_refuses_an_encrypted_restore_while_the_hold_is_on`). The three tests of the held path lift the constant with `_lift_encrypted_bootstrap_restore_hold`.
- Nothing else changed: the leak fix, the probes and the executor side are as reviewed on 11 Sep.

### What this does not fix

Holding the pod workload stopped during a create-time restore (so the hold can go) is a follow-up on the ticket, not this PR. Two more follow-ups on the ticket, both in lium-platform: the API still accepts `--restore-backup` with encryption on, so the renter learns of the refusal only after an executor was rented and the create failed (a pre-check at `POST /pods` keyed on `enable_volume_encryption` + `backup_id`), and `apps/docs/docs/developers/cli/reference/up.md:95` lists `--restore-backup` with encryption on by default and needs a one-line caveat for as long as the hold stands; `pod-users/encrypted-volumes.md:31` already says encrypted restores need a running pod. Rows 2/3 (argv, `/tmp`) are #1338. The writable layer (`/workspace`) stays plaintext (B-146), `fuse.allow_sys_admin_access` is unchecked (B-144), Inspector consequences are DAH-3275.

### Ran it

`tools/aws_verify.sh lium-io <head> --app io --bundle` (EC2 warm lium-io box, run `20260914T074603Z-ily1`), this head, rebased on `9972ed2`:

```
validators pytest tests/: rc=1 in 164 s — 2 failed, 2343 passed, 15 skipped
executor pytest tests/: rc=0 in 4 s — 239 passed
```

The two failures are the warm box's (root can read an unreadable sysfs tree; an sshd is already running on the box) and appear in every lium-io run on it (10 Sep: `-xi6s`, `-iwce`, `-msmw`, `-99ec`, `-bmvk`, on unrelated branches and main). The same content before the rebase (run `20260914T073319Z-kjrm`): validators 2082 passed, executor 145 passed, the same two failures. The probe's exit 0 against this checkout's executor models is asserted inside the run (`test_bootstrap_probe_finds_the_field_in_this_repo_s_executor_models`); its exit 3 against main's models follows from `WorkspaceSpec` on main having no `bootstrap` field (inspection, not run).

### Verified

Gate: PASS (55a72e3) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1341)

EC2 verify at 55a72e3 (rebased on `9972ed2`): `20260914T074603Z-ily1` (validators 2343 passed + the two box-bound failures, executor 239 passed); the same content before the rebase: `20260914T073319Z-kjrm` (validators 2082 passed, executor 145 passed).

EC2 verify of the 11 Sep round (heads rewritten by the 14 Sep rebase): `20260911T073448Z-1i78` (io suites, validators 2081 passed + the same two box-bound failures, executor 145 passed); `20260911T073700Z-dols` validators `tests/test_docker_service.py` 196 passed, 1 skipped; `20260911T073916Z-bvq7` executor `tests/test_storage_runner.py` 43 passed.

### Self-review

Verdict `findings` at `55a72e3` · 15 checks · by subagent-r130-1341 · 2026-09-14 07:52Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | STALE-BODY | `PR body:79` | 'Ran it' cites run `20260914T073319Z-kjrm` as 'this head', but kjrm ran at the pre-rebase commit (4aa4589, base 3d32675); the base has since moved to 9972ed2 (103 files, incl. #1279/#1332/#1334/#1338 in docker_service.py) and the run for 55a72e3 is still in flight. | Replace kjrm with the id and counts of the EC2 run at 55a72e3 once it finishes (or say 'pre-rebase head' for kjrm). |

Mechanical notes dismissed: `neurons/executor/tests/test_storage_runner.py:474` `requested_path="/root"` is the spec's `workspace.requested_path` string in the parametrized online-restore test; `storage.workspace.subprocess.run` is monkeypatched and no `-v` host bind mount is built from it. · `neurons/executor/tests/test_storage_runner.py:499` Same string in the bootstrap-restore payload; the path is reached as `/proc/<pid>/root/root` through nsenter in the pod's user namespace, not as a docker bind mount, and `subprocess.run` is stubbed. · `neurons/validators/tests/test_workspace_backup_restore.py:620` `local_volume_path="/root"` is the plaintext mount path inside the pod passed to `_run_bootstrap_restore`; it becomes `workspace.volume_path`/`requested_path` in the JSON spec, and `start_storage_operation` is an AsyncMock, so nothing is mounted. · `neurons/validators/tests/test_workspace_backup_restore.py:631` An assertion that the rendered spec's `requested_path` equals "/root"; a string comparison, not a mount. · `neurons/validators/tests/test_workspace_backup_restore.py:651` `local_volume_path="/root"` on the call that is expected to raise before `supports_storage_operation` or `start_storage_operation` is awaited (both asserted not awaited); no command is built. · `neurons/validators/src/services/docker_service.py:262` Round-1 STACK-HYGIENE closed: merge-base is 9972ed2, `git merge-tree --write-tree origin/main HEAD` is clean, `_VOLUME_SETUP_TMPFS` (#1338) sits at 256 and the hold constant at 262; `git range-diff` shows the only content changes across the rebase are the DAH-3240 context lines around `if not local_volume:` and the constant's comment. · `PR body:75` Round-1 PROSE-VS-CODE (renter message), DOCS-MISSING (up.md), step-0 'inherits' wording and SCOPE (backend pre-check) are folded: Risks names step `bootstrap_restore_hold` and puts the hint in the detail; step 0 says what the renter sees and which three tests the follow-up inherits; Parts/Docs mark the `up.md:95` caveat a follow-up; 'What this does not fix' names both lium-platform follow-ups. Gate/Self-review sha lines are tool-owned and not flagged. · `neurons/validators/src/services/docker_service.py:5580` Rebase interactions checked: DAH-2740 creates the edit's replacement under the original name (5010-5018), so `container_name` passed to the held encrypted restore is the running pod; DAH-3258's concurrent inspector task reads the container from outside and does not touch the mount; DAH-3240's `volume_probe` is created inside `if not local_volume:` after the hold, so the refusal still precedes every volume-stage command. `a Python syntax check` passes on the rebased files; the test helpers (`_patch_create_container_happy_path` at 1855, `_executor_info_for` at 2039) are intact.

</details>


neurons/validators/src/services/docker_service.py
```diff
@@ -5188,16 +5219,24 @@ async def _run_bootstrap_restore(
...
         if encrypted:
-            workspace["volume_passphrase"] = VolumeKeyDeriver.from_settings(settings).material(
-                payload.pod_id
-            ).passphrase
+            workspace["container_name"] = container_name
+            # The pod has been running since `docker run`; its entrypoint may already have
+            # written under the fresh mount (`.jupyter`, `.bashrc`). At create time nothing
+            # there is the customer's, so the executor lets the backup write over it instead
+            # of refusing a non-empty target as an online restore would.
+            workspace["bootstrap"] = True
```
taiberium, line 5439: **Critical:** An executor on the older image ignores `bootstrap` and refuses a non-empty target, so the create fails. Probe the executor for this key first.

taiberium (changes requested):
Requested changes.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `f168beb`: `supports_bootstrap_restore` (storage_operations.py) asks the executor's WorkspaceSpec for the bootstrap field over SSH. create_container runs it before create_local_volume, so an old image fails at `bootstrap_restore_probe` with nothing left behind. Tests: the old-executor create test and the probe test against this repo's executor models.

neurons/validators/src/services/docker_service.py
```diff
@@ -4539,6 +4540,50 @@ async def create_container(
...
+
+                if payload.bootstrap_restore and use_encrypted_volume:
+                    # The encrypted restore after `docker run` sends `workspace.bootstrap`; an
+                    # executor image from before DAH-3274 ignores the key and refuses the target
+                    # the entrypoint has already written to, so the create would fail with the
+                    # pod half-built. Stop here, before the volume and the pod exist: no
+                    # fallback keeps the passphrase off the executor.
+                    current_step = "bootstrap_restore_probe"
+                    if not await supports_bootstrap_restore(ssh_client, executor_info.python_path):
+                        raise RuntimeError(
```
taiberium, line 4750: NIT: an old executor image fails the whole create, while the image-label check downgrades to a plain volume instead. Is the hard failure intended?

neurons/validators/src/services/docker_service.py
```diff
@@ -4901,6 +4920,24 @@ async def create_container(
                         profilers.append(ProfilerStep.since(ProfilerStepName.ENCRYPTED_VOLUME_SETUP, prev_timestamp))
                         prev_timestamp = now_ms()
 
+                        if payload.bootstrap_restore:
+                            # DAH-3274: restore into the mounted plaintext through the pod's own
+                            # gocryptfs (the executor nsenters the pod's user namespace, as the
+                            # online `lium restore` does). The passphrase stays in the
+                            # validator→pod channel; the executor never sees it. Runs before the
+                            # key injection below so the restore target (/root) is still empty.
```
taiberium: NIT: this comment says `/root` is empty, but the entrypoint can write to the fresh mount. Reword it: the restore runs first so it cannot erase the new keys.

taiberium (approved):

neurons/validators/src/services/docker_service.py
surcyf123: Done in `2660701`: the comment now says the entrypoint may have written to the mount, and that the restore runs first so it cannot erase the injected keys.

neurons/validators/src/services/docker_service.py
surcyf123: Intended; the comment at the probe now says why (`2660701`). The label check downgrades because the renter's own image can never encrypt, and the status reports it. An old executor is the provider's to update. A plain restore would put the backup's plaintext on a disk the renter asked to encrypt; the old encrypted mode leaked the passphrase.

neurons/validators/src/services/docker_service.py
```diff
@@ -4701,6 +4702,55 @@ async def create_container(
...
+                    # executor image from before DAH-3274 ignores the key and refuses the target
+                    # the entrypoint has already written to, so the create would fail with the
+                    # pod half-built. Stop here, before the volume and the pod exist. This is a
+                    # hard failure on purpose, unlike the label check above: that one downgrades
+                    # because the renter's own image can never encrypt (and the status says so),
+                    # while an old executor is the provider's to update — a plain-volume restore
+                    # would put the backup's plaintext on a disk the renter asked to encrypt, and
+                    # the old encrypted mode would hand the executor the passphrase.
+                    current_step = "bootstrap_restore_probe"
+                    if not await supports_bootstrap_restore(ssh_client, executor_info.python_path):
```
taiberium, line 4749: **NIT:** The probe reads a field this same PR adds to the executor, so release `executor-v*` before `validator-v*`.

neurons/validators/src/services/storage_operations.py
```diff
@@ -51,6 +51,27 @@ async def supports_storage_operation(
...
+
+async def supports_bootstrap_restore(
+    ssh_client: asyncssh.SSHClientConnection,
+    python_path: str,
+) -> bool:
+    result = await ssh_client.run(
+        f"{shlex.quote(python_path)} -c {shlex.quote(_BOOTSTRAP_PROBE)}",
+        check=False,
+    )
+    return result.exit_status == 0
```
taiberium: **NIT:** Any non-zero exit reads as an old executor image, so exit 127 looks the same. Return the exit status and stderr tail in the `RuntimeError`.

neurons/executor/tests/test_storage_runner.py
```diff
@@ -443,6 +446,89 @@ def run_docker(command: list[str], **kwargs: object) -> SimpleNamespace:
...
+
+    monkeypatch.setattr("storage.workspace.subprocess.run", run_docker)
+
+    with pytest.raises(WorkspaceResolutionError, match="new or empty"):
+        WorkspaceResolver(
+            docker_binary="docker", proc_root=tmp_path, environ={"LIUM_STORAGE_HELPER_IMAGE": "executor:test"}
+        ).resolve(operation)
+
+
+def test_bootstrap_encrypted_restore_writes_over_what_the_entrypoint_left(
```
taiberium: **NIT:** The test only matches text in the preflight script, so rename it `test_bootstrap_encrypted_restore_skips_the_emptiness_preflight`.

taiberium (approved):

neurons/validators/src/services/docker_service.py
surcyf123: Done in `230da0d`: the body now states the release order in the fold and under Compatibility: `executor-v*` before `validator-v*`. The bullet that said no image bump is needed now says the mode needs none, the bootstrap key does.

neurons/validators/src/services/storage_operations.py
surcyf123: Done in `230da0d`: the probe now exits 3 when the models have no bootstrap field, and only that reads as an old image. Any other exit raises with the status and the last lines of stderr. So a missing interpreter (127) or a broken import (1) names its cause in the create's failure detail. Tests: exit 0 and 3 as before, plus 127 and 1 with the stderr tail asserted.

neurons/executor/tests/test_storage_runner.py
surcyf123: Done in `230da0d`: renamed to test_bootstrap_encrypted_restore_skips_the_emptiness_preflight. One commit on top of your approval, so a re-approval click is needed. EC2 run o9cm: executor 145 passed; validators 2080 passed, with the two box-bound failures every run on that box shows.

taiberium (approved):

taiberium (commented):
Second pass, read at the PR head. `bootstrap: true` is produced only by `_run_bootstrap_restore(encrypted=True)` at create time — the online restore spec never sets it — so `lium restore` into a live pod keeps its empty-target guard, and the old encrypted path ran no target check at all, so nothing regressed. Backup paths are untouched. Nothing blocks.

NIT: `neurons/validators/src/services/docker_service.py:5105` — `wait_for_storage_operation` has no wall-clock deadline and nothing wraps `create_container` in a timeout. An S3 endpoint that black-holes now stalls a create that already holds the pod's GPUs and host ports; before, it stalled with nothing created. Pre-existing shape, worse position.

NIT: `neurons/executor/src/storage/models.py:110` — `bootstrap` is refused outside `encrypted_running` but no longer outside `restore`, so `action: backup` with `bootstrap: true` now parses and is silently ignored. One more condition keeps the field meaning one thing.

NIT: `neurons/validators/src/services/storage_operations.py:68` — the probe checks the models field but not the restic binary, and `supports_storage_operation` now runs after `docker run` for encrypted volumes. An executor with new models and no restic fails with the pod already built.

NIT: `neurons/executor/src/storage/workspace.py:188` — the comment says the backup wins "as it did when the restore preceded the entrypoint", which is backwards: with the restore first the entrypoint wrote last and won. The winner really flipped; say so.

neurons/validators/src/services/docker_service.py
```diff
@@ -5071,6 +5104,26 @@ async def create_container(
...
 
+                        if payload.bootstrap_restore:
+                            # DAH-3274: restore into the mounted plaintext through the pod's own
+                            # gocryptfs (the executor nsenters the pod's user namespace, as the
+                            # online `lium restore` does). The passphrase stays in the
+                            # validator→pod channel; the executor never sees it. Runs before the
+                            # key injection below: the entrypoint may already have written to the
+                            # fresh mount, but the restore comes first, so it cannot erase the
+                            # keys the injection puts under /root.
+                            current_step = "bootstrap_restore"
```
taiberium, line 5590: Please reject encrypted bootstrap restore in the existing preflight until the pod workload can remain stopped throughout the restore.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `55a72e3`: the preflight refuses an encrypted create-time restore before the probes, the volume and the pod. The step is `bootstrap_restore_hold`. The hold is a module constant, not a setting. The follow-up that keeps the workload stopped removes it; the three tests of the held path lift it and carry over unchanged. One consequence for you to weigh: `lium up --restore-backup` with encryption on, the CLI default, now fails at create. Renters restore with `lium bk restore` once the pod is up, as the encrypted-volumes doc already says. The branch was rebased onto main because #1338 landed on the same lines. Your review request is renewed on this head.

taiberium (approved):

PR #1342, opened 2026-09-09
DAH-3275 - [P1] inspector: provider-origin verdict, renter access event, enforcement flag

surcyf123:
Implements [DAH-3275](https://app.notion.com/p/3d6b8bfdbde9816ab645cf20764b09d1) · reviewer: @taiberium

**TL;DR** — Every Inspector outcome returned `passed=True` and the validator's own execs counted as attacks (607 of 632 MALICIOUS rounds on 8 Sep). The check now separates platform-origin from provider-origin findings, records a verdict (evidence hashes, sensor attestation, action) and, behind `INSPECTOR_ENFORCE_ENABLED` (default off) when a rented pod is affected, tells the renter, fails the check and requests an `inspector_auto` quarantine.

**What changed**
- `inspector_verdict.py`: platform-origin = a Docker control-plane action (`DockerExec`, `DockerCp`, `DockerRm`, …) from the executor container; sha256 per provider finding
- `InspectorRentedCheck`: provider-origin → `MALICIOUS` + verdict; check fails, score gate and renter event only when `verdict.action` is `quarantine` (`services/task/checks/inspector.py`)
- `INSPECTOR_ENFORCE_ENABLED` (default `False`) + `inspector_passed` score gate, like `cpu_truth_passed` (`core/config.py`, `task/pipeline.py`, `task/score_calculator.py`)
- Sensor integrity: TDX-attested host with `ENABLE_ATTESTATION_WHITELIST` on → measured image vouches for `libinspector.so`, shell `sha256sum` skipped; else `shell_sha256_unattested`

**How to verify**
- `cd neurons/validators && pdm run pytest tests/test_inspector_verdict.py tests/test_inspector_check.py tests/test_inspector_validation_service.py -q` → passes
- class-A exec (`cat /root/.ssh/authorized_keys`, tag `nested_from:executor-executor-1`) → platform-origin; `host: true` + enforce → `passed=False`, `action == "quarantine"`; on a non-rented container → `action == "none"`

**Verified by the loop:** Gate: PASS b836a23

**Ran it:** `tools/aws_verify.sh lium-io b836a23 --bundle` → validators 2057 passed, 2 box-only fails (sysfs, sshd adopt) · EC2 20260911T074251Z-uy86

**Self-review:** clean at b836a23 (15 checks, 3 low)

**Parts:** backend ✓ (validator; the platform consumer of `context.verdict` is arhangel66's PR) · migration n/a — existing jsonb · frontend n/a — existing pod log view · CLI/SDK n/a — validator only · docs n/a — flag off · tests ✓ 4 files, 34 new · dashboards n/a — digest follow-up

**Docs:** none changed in this PR — default off, nothing visible until the flag is on; renter/provider pages are B-147 (3).

**Cross-PR:** independent of #1341 — different files, either order lands.

### Why

`design/RENTER_DATA_PRIVACY.md` §3–4 and `private/INSPECTOR_20260908.md`. Three things made "we log every time a provider views data" empty: (1) `checks/inspector.py` returned `passed=True` in every branch and nothing called `ProviderBanService` (`inspector_auto` had no caller); (2) the validator's own execs — `cat /root/.ssh/authorized_keys` from `rented_machine.py`, the executor's `df -k` metric, the executor's legacy `tar` restore — are `DockerExec` findings whenever the host's Tetragon loses the ancestry to sshd (10 hosts, 607 findings on 8 Sep); (3) the renter saw nothing.

### Classification

A finding is platform-origin when **all** hold: `kind` is one of the sensor's Docker control-plane kinds (the `tamper-docker-cli` and `tamper-docker-sock-write` policies: `DockerExec`, `DockerCp`, `DockerRm`, `DockerStop`, … — the executor creates, copies into, stops and removes a pod's containers through the Docker API from inside its own container, so any of them can be ours; round 3, taiberium); `host` is not `true`; a `nested_from:<name>` tag names an executor-stack container by the sensor's own rule (`executor`, `executor-*`, `*-executor-*` — so `lium-executor-executor-1` too). The sensor itself already drops docker-policy findings whose ancestry reaches sshd or the container's pid 1, so every such finding is one whose ancestry it could not trust; until DAH-3278 hardens that on the verifier, all of them count as the platform's. Everything else — the same command from the host, an `nsenter`, a memory or overlayfs read, an exec the sensor could not attribute — is provider-origin. The executor's own backup and restore (`executor/src/storage/restic.py`) never reach the verdict: the sensor judges the encrypted flow's `nsenter -t <pid> -U` benign by nstype (`CLONE_NEWUSER` only, nsenter.rs, never by a container name) and the unencrypted flow's `volume_<pod_id>` bind a runtime mount (mount.rs, runc / container-stack binaries), so neither is a finding; a name-based exemption here would let a `lium-storage-…`-named provider container borrow the platform's origin (round 6, taiberium).

The first cut also matched the payload against an exact allow-list; round 1 of the review showed it rejected what the platform really runs (`hardware_service.py` `df -k /root` on every unencrypted rental, `restore_storage.py` `tar … --strip-components=1` + `mkdir -p` + `chown`, `backup_storage.py` `du -sb` + `tar -czf`, the key-injection / sshd / gocryptfs `sh -c '<script>'` execs), i.e. it would have told renters about provider access every metrics cycle and, under enforcement, quarantined the host. The platform's exec payloads are too many and too script-shaped for an exact list to stay honest, so the classifier is the ancestry alone — what `design/RENTER_DATA_PRIVACY.md` row 16 asks for — and the payloads seen from the executor container are recorded in the verdict (`platform_exec_commands`, deduped, ≤ 20 × 160 chars) so the daily digest shows what runs. `test_the_real_platform_execs_are_recognised_from_the_executor_container` builds the argv from `restore_storage.workspace_command` / `backup_storage.workspace_command` / `VolumeAccess.docker_exec_args`, so the test and the jobs cannot drift apart silently. What keeps the tag trustworthy — a provider `docker exec`-ing into the executor container to borrow its ancestry — is the verifier's half (DAH-3278: stack-binary ancestry gating, sysbox-fs exemption); on an unattested host that is the residual risk and the verdict says `sensor: unattested`.

The verdict names the pod from `finding.container` / `docker.resolved_name` (`pod_<id>`, or `volume_<id>` as `DockerVolumeRm` carries it); the path-shaped kinds the sensor emits with `container: None` — `OverlayFsRead`/`OverlayFsWrite` (fs.rs) and `DockerVolumeMount` (mount.rs), the path in `command`, tag `rental_volume`, the 8 Sep class-D shape — name the pod through the volume in that path (`/var/lib/docker/volumes/volume_<id>/…` or `…/propagated-mount/volume_<id>`, the sensor's own `volume_name_from_path` rule). A `pod_*` that is not in the rented list (left or joined since the list was fetched) is recorded as `unmatched_containers` (≤ 20 × 128 chars + a count) and no renter is told about a container that was not theirs; only when a provider finding names neither a container nor a rental-volume path does every rented pod on the host get the event. `kind` and `tags` of any JSON type classify without raising (non-string kind → `unknown`, never platform-origin). Renter-visible `finding_kinds` come from a fixed vocabulary — the sensor's 39 `RuntimeInterferenceKind` names (celium-gpu-verifier `inspector/src/collector/analysis/types.rs`, pinned by `test_known_finding_kinds_are_the_sensors_runtime_interference_kinds`), else `unknown` — the report is produced on the provider's root when the sensor is unattested, so no raw string from it reaches a pod log; the raw finding stays in the evidence. A report whose `findings` is not a list of objects is an `ERROR` outcome (broken sensor), never a provider finding.

### What the backend receives

`InspectorEventRequest.context.verdict`:

```json
{"provider_findings": 1, "platform_findings": 1, "finding_kinds": ["DockerExec"], "affected_pod_ids": ["<pod>"], "platform_exec_commands": ["cat /root/.ssh/authorized_keys"],
 "evidence_sha256": ["<sha256 of the canonical finding JSON>"], "report_sha256": "…",
 "sensor": "unattested", "enforce": false, "action": "none", "ban_source": null, "unmatched_containers": [], "unmatched_containers_count": 0,
 "unnamed_findings": 0, "platform_kind_counts": {"DockerExec": 1}}
```

`unnamed_findings` counts the provider findings that named no container and no rental volume — each one marks every rented pod on the host (an overlay2 path has no `volumes/` segment), so the shadow digest can say how often that fallback fires; `platform_kind_counts` is the platform findings by kind, for the same shadow numbers (round 6, taiberium).

With `INSPECTOR_ENFORCE_ENABLED=true` and a provider-origin finding that names a rented pod (`affected_pod_ids` non-empty; a finding naming no container at all counts as every rented pod): `action: "quarantine"`, `ban_source: "inspector_auto"`. A provider-origin finding on a container that is not a rented pod — the provider inside their own container — is recorded under `unmatched_containers` with `action: "none"`, whatever the flag (round 5, taiberium). The backend half — `save_inspector_event` calling `ProviderBanService.ban_provider(source=inspector_auto)` (which already removes the executors from the marketplace and notifies renters; it does not delete pods) when `action == "quarantine"` — is lium-platform work for arhangel66; until it lands the validator's consequence is the score gate.

### Renter event

Published only with `INSPECTOR_ENFORCE_ENABLED=true` (round 2, taiberium: in shadow mode the classifier is still being measured — 607 of 632 rounds were the sensor's false positives on 8 Sep — and a wrong "the provider read your pod" cannot be taken back) on `STREAMING_LOG_CHANNEL` (the backend's `handle_log_stream` → `save_pod_status(pod_id, logs=…)`, the same stream the create flow uses), one entry per affected pod: `log_tag: provider_access_detected`, `log_status: error`, text "Provider-side access to this pod detected (DockerExec; unattested sensor). Lium has recorded the evidence[ and asked to take the host off the marketplace]." plus `finding_kinds`, `provider_findings` (count), `report_sha256`, at most 20 `evidence_sha256` — all from that pod's findings only (its container, its volume, or a finding that names nothing), never another renter's (round 6) — (`evidence_sha256_truncated: true` when cut — the finding count is the sensor's to choose; the full list is in the inspector event), `sensor`, `action`. A publish failure is logged and does not lose the verdict (it is in the inspector event).

### Sensor integrity

`validate_rented_executor(…, sensor_attested=ctx.tdx_attestation_passed and settings.ENABLE_ATTESTATION_WHITELIST)`. On a dstack CVM the executor image — and `libinspector.so` in it — is part of the stack `attestation_service` measures against `TDX_WHITELIST` — but only when `ENABLE_ATTESTATION_WHITELIST` is on; prod runs it off today, so a passed attestation alone says the quote is genuine, not that the image is ours, and the shell checksum stays (round 3, taiberium); a `sha256sum` answered by the provider's shell adds nothing there and is skipped (`diagnostics.sensor_integrity = tdx_measured_image`). On every other host the shell checksum still runs (it catches an outdated image) and the verdict says `sensor: unattested` / `shell_sha256_unattested`, which is the honest statement: the sensor runs on the provider's root. NVIDIA GPU attestation (`gpu_attestation_passed`) does not measure the executor image, so it does not count.

### Flag semantics

`INSPECTOR_ENFORCE_ENABLED=false` (default): the verdict and evidence are recorded in the inspector event, no renter event, `passed=True`, score untouched — shadow. `true` and a rented pod affected: `passed=False` (check stays non-fatal, the cycle finishes), `inspector_passed=False` → `calculate_scores` zeroes the score with the warning "Provider-side access to a rented pod detected by the Inspector", `action=quarantine` requested, the affected renters told. `true` and no rented pod affected (the finding names a container that is not a rented pod): `MALICIOUS` recorded with the container under `unmatched_containers`, `passed=True`, score untouched, no renter event, `action=none`. `CLEAN`, `ERROR`, `COLLECTOR_NOT_RUNNING`, `CANARY_FAILED`, `COLLECTOR_RECENTLY_STARTED` are unchanged in this PR (the ≥ 4-cycles-dark signal is §4.3's second half).

### Ran it

`tools/aws_verify.sh lium-io e22b8d5 --bundle …` (EC2 warm lium-io box, run `20260910T094523Z-cjpb`; earlier heads 3477a6f, 10f53d8, 22ae818, e2cc27c, 518b653, 7218c8d, e3bd0bd, 693a416 ran the same way with the same two failures):

```
validators pytest tests/: rc=1 in 111 s — 2 failed, 2050 passed, 15 skipped
```

The two failures are the warm box's (root can read an unreadable sysfs tree; an sshd is already running on the box) and appear in every lium-io run on it today, including main.

### Self-review

Verdict `findings` at `b836a23` · 15 checks · by subagent-r91-1342 · 2026-09-11 07:51Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | STALE-BODY | `PR body:36` | Being fixed in the body override per the author: the Classification paragraph should state, with the sensor citations now in the module docstring, why the backup's `nsenter -U` and the runc-driven volume bind never reach the validator (the answer to NITs 1–2); the code at b836a23 already matches the body's existing 'an nsenter … is provider-origin' sentence. | Add the one-sentence sensor citation to Classification before push; nothing to change in code. |
| low | PROSE-VS-CODE | `PR body:46` | Being fixed in the body override per the author: the `context.verdict` example lacks `unnamed_findings` and `platform_kind_counts`, and the Renter-event paragraph should say the kinds, count and ≤ 20 hashes are of the findings about that pod (`findings_for_pod`), never another renter's. | Add both keys to the example and the per-pod sentence, citing `test_the_renter_event_carries_only_that_pods_findings`. |
| low | STALE-BODY | `PR body:21` | Being fixed in the body override per the author: Parts ('31 new' tests), Verified, Ran-it and Self-review lines are rendered from 9bee804; b836a23 adds four tests (`…cannot_borrow_the_platforms_origin`, `…carries_only_that_pods_findings`, `…in_shadow_is_not_called_access_to_a_rented_pod`) and extends `…tells_every_renter_on_the_host`. | Re-render at the pushed head (the tooling records gate and verdict lines; update the test count by hand). |

Mechanical notes dismissed: `neurons/validators/src/services/task/inspector_verdict.py:49` `--privileged` is a member of `_EXEC_FLAGS`, the set `docker_exec_command` skips while parsing a docker-exec argv the sensor reported; nothing in the diff runs or configures a container. · `neurons/validators/tests/test_inspector_verdict.py:51` `"cwd": "/root"` is a value in the `_finding()` fixture dict (the sensor's reported cwd), not a mount. · `neurons/validators/tests/test_inspector_verdict.py:78` `VolumeAccess(volume_path="/root")` is the argument the real restore/backup job builders take to produce the argv strings under test; nothing is mounted. · `neurons/validators/tests/test_inspector_verdict.py:92` `-C /root` is part of the tar argv string built by `backup_storage.workspace_command` for the `docker_exec_command` test; a string fixture, no mount. · `neurons/validators/tests/test_inspector_verdict.py:168` `"cwd": "/root"` in the `_path_finding()` fixture dict; a dict value, not a mount. · `neurons/validators/tests/test_inspector_verdict.py:1` ruff I001 (import block unsorted) on this file is pre-existing at the reviewed head 9bee804 (same result on `git show 9bee8048:…`), not introduced by b836a23.

### Verified

Gate: PASS (b836a23) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1342)

</details>

### DIFF
--- a/neurons/validators/src/services/task/checks/inspector.py
+++ b/neurons/validators/src/services/task/checks/inspector.py
@@ -148,7 +148,11 @@ class InspectorRentedCheck:
             inspector_event = _build_inspector_event(
                 ctx, event, rented_pods, result, outcome="MALICIOUS", report=report, verdict=verdict
             )
-            await _tell_renters(ctx, verdict, when=event.when.isoformat())
+            if enforce:
+                # The renter hears about it only when the verdict acts: in shadow mode the
+                # classifier is still being measured against the sensor's false positives, and a
+                # "the provider read your pod" event on a wrong call cannot be taken back.
+                await _tell_renters(ctx, verdict, when=event.when.isoformat())
             updates: dict[str, Any] = {
                 "default_extra": extra,
                 "state": replace(ctx.state, inspector_event=inspector_event),
--- a/neurons/validators/src/services/task/checks/inspector.py
+++ b/neurons/validators/src/services/task/checks/inspector.py
@@ -91,6 +91,24 @@ class InspectorRentedCheck:
 
         report = dict(result.report or {})
         findings = _findings(report)
+        if findings is None:
+            # a report whose findings are not a list of objects is a broken sensor, not a
+            # provider caught in the act — it must not zero a score or request a quarantine
+            event = render_message(
+                Msg.VALIDATION_ERROR,
+                ctx=ctx,
+                check_id=self.check_id,
+                what={"error": "inspector report findings are not a list of objects", "findings": report.get("findings")},
+                extra=extra,
+            )
+            inspector_event = _build_inspector_event(
+                ctx, event, rented_pods, result, outcome="ERROR", report=report
+            )
+            return CheckResult(
+                passed=True,
+                event=event,
+                updates={"default_extra": extra, "state": replace(ctx.state, inspector_event=inspector_event)},
+            )
         warnings = _collector_start_warnings(report)
         enforce = settings.INSPECTOR_ENFORCE_ENABLED
         verdict = build_verdict(
@@ -309,8 +327,9 @@ def _collector_start_warnings(report: dict[str, Any]) -> list[dict[str, Any]]:
     ]
 
 
-def _findings(report: dict[str, Any]) -> list[dict[str, Any]]:
+def _findings(report: dict[str, Any]) -> list[dict[str, Any]] | None:
+    """The report's findings as a list of objects, or None when the report is malformed."""
     findings = report.get("findings") or []
-    if not isinstance(findings, list):
-        return [{"value": findings}]
-    return [item if isinstance(item, dict) else {"value": item} for item in findings]
+    if not isinstance(findings, list) or not all(isinstance(item, dict) for item in findings):
+        return None
+    return findings
--- a/neurons/validators/src/services/task/checks/inspector.py
+++ b/neurons/validators/src/services/task/checks/inspector.py
@@ -329,7 +329,9 @@ def _collector_start_warnings(report: dict[str, Any]) -> list[dict[str, Any]]:
 
 def _findings(report: dict[str, Any]) -> list[dict[str, Any]] | None:
     """The report's findings as a list of objects, or None when the report is malformed."""
-    findings = report.get("findings") or []
+    findings = report.get("findings")
+    if findings is None:
+        findings = []
     if not isinstance(findings, list) or not all(isinstance(item, dict) for item in findings):
         return None
     return findings
--- a/neurons/validators/src/services/task/checks/inspector.py
+++ b/neurons/validators/src/services/task/checks/inspector.py
@@ -98,7 +98,11 @@ class InspectorRentedCheck:
                 Msg.VALIDATION_ERROR,
                 ctx=ctx,
                 check_id=self.check_id,
-                what={"error": "inspector report findings are not a list of objects", "findings": report.get("findings")},
+                what={
+                    "error": "inspector report findings are not a list of objects",
+                    "findings_type": type(report.get("findings")).__name__,
+                    "findings_preview": repr(report.get("findings"))[:200],
+                },
                 extra=extra,
             )
             inspector_event = _build_inspector_event(
--- a/neurons/validators/src/services/task/checks/inspector.py
+++ b/neurons/validators/src/services/task/checks/inspector.py
@@ -123,55 +123,26 @@ class InspectorRentedCheck:
             enforce=enforce,
         )
         if verdict.provider_origin:
-            what: dict[str, Any] = {
-                "findings": verdict.provider_findings,
-                "platform_findings": len(verdict.platform_findings),
-                "verdict": verdict.as_payload(),
-                "summary": report.get("summary", {}),
-                "canary_ok": report.get("canary_ok"),
-            }
-            if warnings:
-                what["warnings"] = warnings
-            event = render_message(
-                Msg.MALICIOUS_FINDINGS,
-                ctx=ctx,
-                check_id=self.check_id,
-                severity="error" if enforce else None,
-                impact=(
-                    "Provider-origin access to a rented pod: score zeroed, quarantine requested"
-                    if enforce
-                    else "Provider-origin access to a rented pod recorded; score unchanged (INSPECTOR_ENFORCE_ENABLED off)"
-                ),
-                what=what,
+            return await self._act_on_provider_origin(
+                ctx,
+                verdict=verdict,
+                report=report,
+                warnings=warnings,
+                rented_pods=rented_pods,
+                result=result,
                 extra=extra,
             )
-            inspector_event = _build_inspector_event(
-                ctx, event, rented_pods, result, outcome="MALICIOUS", report=report, verdict=verdict
-            )
-            if enforce:
-                # The renter hears about it only when the verdict acts: in shadow mode the
-                # classifier is still being measured against the sensor's false positives, and a
-                # "the provider read your pod" event on a wrong call cannot be taken back.
-                await _tell_renters(ctx, verdict, when=event.when.isoformat())
-            updates: dict[str, Any] = {
-                "default_extra": extra,
-                "state": replace(ctx.state, inspector_event=inspector_event),
-            }
-            if enforce:
-                # Non-fatal check: passed=False alone changes nothing downstream, the score gate
-                # in calculate_scores reads this flag (same mechanics as cpu_truth_passed).
-                updates["inspector_passed"] = False
-            return CheckResult(passed=not enforce, event=event, updates=updates)
 
         clean_what: dict[str, Any] = {
             "summary": report.get("summary", {}),
             "canary_ok": report.get("canary_ok"),
-            "verdict": verdict.as_payload(),
+            "verdict": verdict.as_payload().model_dump(),
         }
         if verdict.platform_findings:
             # every finding was one of our own execs seen from a host whose Tetragon lost the
-            # sshd ancestry — recorded, not malicious (the 8 Sep false positives)
-            clean_what["platform_findings"] = verdict.platform_findings
+            # sshd ancestry — recorded, not malicious (the 8 Sep false positives); the findings
+            # themselves are in the inspector event's report
+            clean_what["platform_findings"] = len(verdict.platform_findings)
         if _canary_failed(report):
             outcome = "CANARY_FAILED"
             event = render_message(
@@ -235,6 +206,60 @@ class InspectorRentedCheck:
             },
         )
 
+    async def _act_on_provider_origin(
+        self,
+        ctx: Context,
+        *,
+        verdict: InspectorVerdict,
+        report: dict[str, Any],
+        warnings: list[dict[str, Any]],
+        rented_pods: list[RentedPod],
+        result: InspectorValidationResponse,
+        extra: dict[str, Any],
+    ) -> CheckResult:
+        """The act-and-publish half of the check: a provider-origin verdict becomes the MALICIOUS
+        event and, under INSPECTOR_ENFORCE_ENABLED, fails the check and tells the renters."""
+        enforce = verdict.enforce
+        what: dict[str, Any] = {
+            "findings": verdict.provider_findings,
+            "platform_findings": len(verdict.platform_findings),
+            "verdict": verdict.as_payload().model_dump(),
+            "summary": report.get("summary", {}),
+            "canary_ok": report.get("canary_ok"),
+        }
+        if warnings:
+            what["warnings"] = warnings
+        event = render_message(
+            Msg.MALICIOUS_FINDINGS,
+            ctx=ctx,
+            check_id=self.check_id,
+            severity="error" if enforce else None,
+            impact=(
+                "Provider-origin access to a rented pod: score zeroed, quarantine requested"
+                if enforce
+                else "Provider-origin access to a rented pod recorded; score unchanged (INSPECTOR_ENFORCE_ENABLED off)"
+            ),
+            what=what,
+            extra=extra,
+        )
+        inspector_event = _build_inspector_event(
+            ctx, event, rented_pods, result, outcome="MALICIOUS", report=report, verdict=verdict
+        )
+        if enforce:
+            # The renter hears about it only when the verdict acts: in shadow mode the
+            # classifier is still being measured against the sensor's false positives, and a
+            # "the provider read your pod" event on a wrong call cannot be taken back.
+            await _tell_renters(ctx, verdict, when=event.when.isoformat())
+        updates: dict[str, Any] = {
+            "default_extra": extra,
+            "state": replace(ctx.state, inspector_event=inspector_event),
+        }
+        if enforce:
+            # Non-fatal check: passed=False alone changes nothing downstream, the score gate
+            # in calculate_scores reads this flag (same mechanics as cpu_truth_passed).
+            updates["inspector_passed"] = False
+        return CheckResult(passed=not enforce, event=event, updates=updates)
+
 
 def _build_inspector_event(
     ctx: Context,
@@ -269,7 +294,7 @@ def _build_inspector_event(
         # so the consumer can act on it without re-deriving the classification.
         payload["context"] = {
             **(result.diagnostics or {}),
-            **({"verdict": verdict.as_payload()} if verdict is not None else {}),
+            **({"verdict": verdict.as_payload().model_dump()} if verdict is not None else {}),
         }
     return payload
 
@@ -294,7 +319,7 @@ async def _tell_renters(ctx: Context, verdict: InspectorVerdict, *, when: str) -
             await redis.publish(
                 STREAMING_LOG_CHANNEL,
                 {
-                    "logs": [renter_access_event(verdict, pod_id=pod_id, when=when)],
+                    "logs": [renter_access_event(verdict, pod_id=pod_id, when=when).model_dump()],
                     "miner_hotkey": ctx.miner_hotkey,
                     "executor_uuid": ctx.executor.uuid,
                     "pod_id": pod_id,
--- a/neurons/validators/src/services/task/checks/inspector.py
+++ b/neurons/validators/src/services/task/checks/inspector.py
@@ -276,10 +276,13 @@ def _build_inspector_event(
 
 def _sensor_attested(ctx: Context) -> bool:
     # The sensor (libinspector.so + Tetragon) ships inside the executor image. On a dstack CVM
-    # the image is part of the measured stack the validator checked against TDX_WHITELIST, so
-    # the report comes from a measured binary; elsewhere its checksum was read through the
-    # provider's own shell and proves nothing — say so in the verdict.
-    return bool(ctx.tdx_attestation_passed)
+    # the image is part of the measured stack — but only when ENABLE_ATTESTATION_WHITELIST is on
+    # does the validator compare that stack against TDX_WHITELIST (attestation_service._verify_tdx);
+    # with the flag off (prod today) a passed attestation says the quote is genuine, not that the
+    # image is ours, so the report is not from a measured binary and the shell checksum stays.
+    # Elsewhere the checksum was read through the provider's own shell and proves nothing — say
+    # so in the verdict.
+    return bool(ctx.tdx_attestation_passed) and settings.ENABLE_ATTESTATION_WHITELIST
 
 
 async def _tell_renters(ctx: Context, verdict: InspectorVerdict, *, when: str) -> None:
--- a/neurons/validators/src/services/task/checks/inspector.py
+++ b/neurons/validators/src/services/task/checks/inspector.py
@@ -12,7 +12,7 @@ from services.redis_service import STREAMING_LOG_CHANNEL
 from core.config import settings
 from core.utils import _m, get_extra_info
 
-from ..inspector_verdict import InspectorVerdict, build_verdict, renter_access_event
+from ..inspector_verdict import ACTION_QUARANTINE, InspectorVerdict, build_verdict, renter_access_event
 from ..messages import InspectorMessages as Msg
 from ..messages import render_message
 from ..pipeline import CheckResult, Context
@@ -218,8 +218,16 @@ class InspectorRentedCheck:
         extra: dict[str, Any],
     ) -> CheckResult:
         """The act-and-publish half of the check: a provider-origin verdict becomes the MALICIOUS
-        event and, under INSPECTOR_ENFORCE_ENABLED, fails the check and tells the renters."""
-        enforce = verdict.enforce
+        event and, when the verdict's action is quarantine (INSPECTOR_ENFORCE_ENABLED and a rented
+        pod affected), fails the check and tells the renters. A finding that names no rented pod
+        is recorded with `unmatched_containers` and acts on nobody, whatever the flag."""
+        acts = verdict.action == ACTION_QUARANTINE
+        if acts:
+            impact = "Provider-origin access to a rented pod: score zeroed, quarantine requested"
+        elif verdict.enforce:
+            impact = "Provider-origin finding on a container that is not a rented pod recorded; score unchanged"
+        else:
+            impact = "Provider-origin access to a rented pod recorded; score unchanged (INSPECTOR_ENFORCE_ENABLED off)"
         what: dict[str, Any] = {
             "findings": verdict.provider_findings,
             "platform_findings": len(verdict.platform_findings),
@@ -233,19 +241,15 @@ class InspectorRentedCheck:
             Msg.MALICIOUS_FINDINGS,
             ctx=ctx,
             check_id=self.check_id,
-            severity="error" if enforce else None,
-            impact=(
-                "Provider-origin access to a rented pod: score zeroed, quarantine requested"
-                if enforce
-                else "Provider-origin access to a rented pod recorded; score unchanged (INSPECTOR_ENFORCE_ENABLED off)"
-            ),
+            severity="error" if acts else None,
+            impact=impact,
             what=what,
             extra=extra,
         )
         inspector_event = _build_inspector_event(
             ctx, event, rented_pods, result, outcome="MALICIOUS", report=report, verdict=verdict
         )
-        if enforce:
+        if acts:
             # The renter hears about it only when the verdict acts: in shadow mode the
             # classifier is still being measured against the sensor's false positives, and a
             # "the provider read your pod" event on a wrong call cannot be taken back.
@@ -254,11 +258,11 @@ class InspectorRentedCheck:
             "default_extra": extra,
             "state": replace(ctx.state, inspector_event=inspector_event),
         }
-        if enforce:
+        if acts:
             # Non-fatal check: passed=False alone changes nothing downstream, the score gate
             # in calculate_scores reads this flag (same mechanics as cpu_truth_passed).
             updates["inspector_passed"] = False
-        return CheckResult(passed=not enforce, event=event, updates=updates)
+        return CheckResult(passed=not acts, event=event, updates=updates)
 
 
 def _build_inspector_event(
--- a/neurons/validators/src/services/task/checks/inspector.py
+++ b/neurons/validators/src/services/task/checks/inspector.py
@@ -1,19 +1,25 @@
 from __future__ import annotations
 
+import logging
 import time
 from dataclasses import replace
 from typing import Any
 
 from services.const import SECONDS_PER_BLOCK
 from services.inspector_validation_service import InspectorValidationResponse
+from services.redis_service import STREAMING_LOG_CHANNEL
 
 from core.config import settings
+from core.utils import _m, get_extra_info
 
+from ..inspector_verdict import InspectorVerdict, build_verdict, renter_access_event
 from ..messages import InspectorMessages as Msg
 from ..messages import render_message
 from ..pipeline import CheckResult, Context
 from ..models import ValidationEvent
 from protocol.vc_protocol.compute_requests import RentedPod
+
+logger = logging.getLogger(__name__)
 
 
 class InspectorRentedCheck:
@@ -50,11 +56,13 @@ class InspectorRentedCheck:
             "rented_pods": [{"name": p.container_name, "pod_id": p.pod_id} for p in rented_pods],
         }
 
+        sensor_attested = _sensor_attested(ctx)
         result = await ctx.services.inspector.validate_rented_executor(
             ctx.services.shell,
             ctx.ssh,
             ctx.executor,
             ctx.default_extra,
+            sensor_attested=sensor_attested,
         )
 
         if result.error:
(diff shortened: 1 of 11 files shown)


neurons/validators/src/services/task/checks/inspector.py
```diff
@@ -96,25 +136,38 @@ async def run(self, ctx: Context) -> CheckResult:
...
-            return CheckResult(
-                passed=True,
-                event=event,
-                updates={
-                    "default_extra": extra,
-                    "state": replace(ctx.state, inspector_event=inspector_event),
-                },
+                ctx, event, rented_pods, result, outcome="MALICIOUS", report=report, verdict=verdict
             )
+            await _tell_renters(ctx, verdict, when=event.when.isoformat())
```
taiberium: **Critical:** The renter is told the provider read the pod even with enforce off, on a classifier that called 607 of 632 rounds wrong. Gate this on the flag.

taiberium (changes requested):
Requested changes.

neurons/validators/src/services/task/checks/inspector.py
surcyf123: Done in `10f53d8`: Gated in 10f53d8: the renter event is published only under `INSPECTOR_ENFORCE_ENABLED`. In shadow mode the verdict and evidence still land in the inspector event, and no renter is told. The shadow test now asserts no publish; the publish-failure test runs under enforcement, the only mode that publishes. The flag comment and the remediation text say the same. Validators suite at 10f53d8 on EC2: 2048 passed.

neurons/validators/src/services/task/checks/inspector.py
```diff
@@ -197,9 +264,48 @@ def _build_inspector_event(
...
+        }
     return payload
 
+
+def _sensor_attested(ctx: Context) -> bool:
+    # The sensor (libinspector.so + Tetragon) ships inside the executor image. On a dstack CVM
+    # the image is part of the measured stack the validator checked against TDX_WHITELIST, so
+    # the report comes from a measured binary; elsewhere its checksum was read through the
+    # provider's own shell and proves nothing — say so in the verdict.
+    return bool(ctx.tdx_attestation_passed)
```
taiberium: **Critical:** Prod sets `ENABLE_ATTESTATION_WHITELIST=false`, so nothing compares the image against `TDX_WHITELIST` and this drops the libinspector checksum. Check that flag here too.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,325 @@
...
+POD_CONTAINER_PREFIX = "pod_"
+# the pod's volume is `volume_<pod_id>` (docker_service.py create flow; the sensor's
+# RENTAL_VOLUME_PREFIXES) — an OverlayFsRead on it is a finding about that pod
+VOLUME_CONTAINER_PREFIX = "volume_"
+
+_EXEC_OPTIONS_WITH_VALUE = {"-u", "--user", "-e", "--env", "-w", "--workdir"}
+_EXEC_FLAGS = {"-i", "-t", "-it", "-ti", "-d", "--detach", "--privileged", "--interactive", "--tty"}
+_SHELL_WRAPPERS = {"sh", "/bin/sh", "bash", "/bin/bash", "/usr/bin/sh", "/usr/bin/bash"}
+_SHELL_COMMAND_FLAGS = {"-c", "-lc", "-ec", "-lec"}
+_DOCKER_EXEC_KINDS = {"DockerExec"}
```
taiberium: **Critical:** Only `DockerExec` counts as platform-origin, so our own `docker cp` and `docker rm` on a rented pod quarantine the provider. Widen `_DOCKER_EXEC_KINDS`.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,325 @@
...
+ACTION_NONE = "none"
+BAN_SOURCE = "inspector_auto"
+RENTER_EVENT = "provider_access_detected"
+
+
+@dataclass(frozen=True)
+class InspectorVerdict:
+    provider_findings: list[dict[str, Any]]
+    platform_findings: list[dict[str, Any]]
+    evidence: list[str]
```
taiberium: NIT: `evidence` holds sha256 digests, so name it `evidence_sha256` like the payload key.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,325 @@
...
+    report_sha256: str
+    classes: list[str]
+    affected_pod_ids: list[str]
+    sensor: str
+    enforce: bool
+    action: str
+    extra: dict[str, Any] = field(default_factory=dict)
```
taiberium: NIT: `extra` hides `platform_payloads` and `unmatched_containers` in an untyped bag. Make them real fields.

taiberium (changes requested):
Requested changes.

neurons/validators/src/services/task/checks/inspector.py
surcyf123: Done in `3477a6f`: the sensor counts as attested only when `ENABLE_ATTESTATION_WHITELIST` is on; with it off the shell checksum runs. Two tests cover both.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `3477a6f`: every Docker control-plane kind the sensor's two docker policies emit is platform-origin from the executor container; nsenter, memory and overlayfs reads stay the provider's.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `3477a6f`: renamed to `evidence_sha256`, the same name as the payload key.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `3477a6f`: `platform_payloads`, `unmatched_containers` and `unmatched_containers_count` are typed fields; `extra` is gone.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,350 @@
...
+    when: str,
+) -> dict[str, Any]:
+    """One pod-log entry the renter sees in their pod's event stream."""
+    classes = ", ".join(verdict.classes) or "access"
+    sensor = "attested sensor" if verdict.sensor == SENSOR_ATTESTED else "unattested sensor"
+    return {
+        "log_text": (
+            f"Provider-side access to this pod detected ({classes}; {sensor}). "
+            "Lium has recorded the evidence"
+            + (" and removed the host from the marketplace." if verdict.action == ACTION_QUARANTINE else ".")
```
taiberium: NIT: `quarantine` only asks the backend to act, so word this as "asked to take the host off the marketplace".

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,350 @@
...
+_EXEC_OPTIONS_WITH_VALUE = {"-u", "--user", "-e", "--env", "-w", "--workdir"}
+_EXEC_FLAGS = {"-i", "-t", "-it", "-ti", "-d", "--detach", "--privileged", "--interactive", "--tty"}
+_SHELL_WRAPPERS = {"sh", "/bin/sh", "bash", "/bin/bash", "/usr/bin/sh", "/usr/bin/bash"}
+_SHELL_COMMAND_FLAGS = {"-c", "-lc", "-ec", "-lec"}
+# The kinds the sensor's two docker policies emit (`tamper-docker-cli`, `tamper-docker-sock-write`;
+# docker_cli.rs `finding_is_rental_interference`). The executor drives a pod's whole life through
+# the Docker API from inside its own container — create, start, `docker cp` for a restore, `docker
+# rm` when the rental ends — so any of these with the executor's ancestry is the platform's, not
+# only the exec. `DockerVolumeMount` is mount.rs (path-shaped), never the executor's.
+_DOCKER_CONTROL_PLANE_KINDS = frozenset(
```
taiberium: NIT: this set repeats the `Docker*` half of `KNOWN_FINDING_KINDS` by hand, so move that constant above and derive this one.

neurons/validators/src/services/task/checks/inspector.py
```diff
@@ -96,25 +136,42 @@ async def run(self, ctx: Context) -> CheckResult:
...
 
         clean_what: dict[str, Any] = {
             "summary": report.get("summary", {}),
             "canary_ok": report.get("canary_ok"),
+            "verdict": verdict.as_payload(),
         }
+        if verdict.platform_findings:
+            # every finding was one of our own execs seen from a host whose Tetragon lost the
+            # sshd ancestry — recorded, not malicious (the 8 Sep false positives)
+            clean_what["platform_findings"] = verdict.platform_findings
```
taiberium: NIT: `platform_findings` is a list here and a count on line 128, so use the count in both.

neurons/validators/src/services/task/checks/inspector.py
```diff
@@ -96,25 +136,42 @@ async def run(self, ctx: Context) -> CheckResult:
...
+                await _tell_renters(ctx, verdict, when=event.when.isoformat())
+            updates: dict[str, Any] = {
+                "default_extra": extra,
+                "state": replace(ctx.state, inspector_event=inspector_event),
+            }
+            if enforce:
+                # Non-fatal check: passed=False alone changes nothing downstream, the score gate
+                # in calculate_scores reads this flag (same mechanics as cpu_truth_passed).
+                updates["inspector_passed"] = False
+            return CheckResult(passed=not enforce, event=event, updates=updates)
```
taiberium: NIT: `run` is 207 lines and has two phases, so move this act-and-publish part into its own method.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,350 @@
...
+
+@dataclass(frozen=True)
+class InspectorVerdict:
+    provider_findings: list[dict[str, Any]]
+    platform_findings: list[dict[str, Any]]
+    evidence_sha256: list[str]
+    report_sha256: str
+    classes: list[str]
+    affected_pod_ids: list[str]
+    sensor: str
```
taiberium: NIT: `sensor` holds an attestation state, so rename the field `sensor_attestation` and keep the wire key.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,350 @@
...
+RENTER_EVENT = "provider_access_detected"
+
+
+@dataclass(frozen=True)
+class InspectorVerdict:
+    provider_findings: list[dict[str, Any]]
+    platform_findings: list[dict[str, Any]]
+    evidence_sha256: list[str]
+    report_sha256: str
+    classes: list[str]
```
taiberium: NIT: the codebase says `kind`, so rename this field `finding_kinds` before a consumer locks the wire key.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,350 @@
...
+) -> InspectorVerdict:
+    provider: list[dict[str, Any]] = []
+    platform: list[dict[str, Any]] = []
+    for finding in findings:
+        (platform if is_platform_origin(finding) else provider).append(finding)
+
+    rented = set(rented_pod_ids)
+    affected: set[str] = set()
+    unmatched: set[str] = set()
+    unnamed = False
```
taiberium: NIT: `unnamed` is a bare adjective, so rename it `has_unnamed_finding`.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,350 @@
...
+    platform_payloads: list[str] = field(default_factory=list)
+    # provider findings on a container that is not in the rented list: recorded, no renter told
+    unmatched_containers: list[str] = field(default_factory=list)
+    unmatched_containers_count: int = 0
+
+    @property
+    def provider_origin(self) -> bool:
+        return bool(self.provider_findings)
+
+    def as_payload(self) -> dict[str, Any]:
```
taiberium: NIT: `as_payload()` and `renter_access_event()` return `dict[str, Any]` with fixed fields, so model them and call `model_dump()` at the emit boundary.

taiberium (approved):

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `e22b8d5`: the renter text now reads "asked to take the host off the marketplace"; the ban itself stays the backend's call.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `e22b8d5`: `KNOWN_FINDING_KINDS` moved above and `_DOCKER_CONTROL_PLANE_KINDS` is now every `Docker*` kind from it minus the path-shaped `DockerVolumeMount`; the set is unchanged (29 kinds).

neurons/validators/src/services/task/checks/inspector.py
surcyf123: Done in `e22b8d5`: the clean path records `len(verdict.platform_findings)` like the malicious path; the findings themselves are still in the inspector event's report.

neurons/validators/src/services/task/checks/inspector.py
surcyf123: Done in `e22b8d5`: the MALICIOUS event, renter publish and score-gate flag moved to `_act_on_provider_origin`; `run` now ends at the classification and the clean outcomes.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `e22b8d5`: dataclass field renamed `sensor_attestation`; the wire key stays `sensor` in the verdict payload and the renter event.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `e22b8d5`: `classes` is `finding_kinds` in the dataclass, the verdict payload and the renter event, and `finding_class()` is `finding_kind()`; nothing on lium-platform main reads the old key yet.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `e22b8d5`: renamed `has_unnamed_finding`.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `e22b8d5`: `as_payload()` returns a `VerdictPayload` and `renter_access_event()` a `RenterAccessEvent`, both pydantic; `model_dump()` runs at the three emit sites. The unmatched-container fields are always present now, empty list and 0 when nothing was unmatched.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,387 @@
...
+        elif pod_id in rented:
+            affected.add(pod_id)
+        else:
+            # a pod that left or joined since the rented list was fetched, or not a pod at all:
+            # recorded, but no renter is told about a container that was not theirs
+            unmatched.add(name)
+    if has_unnamed_finding:
+        # the sensor named no container at all: every renter on this host is told
+        affected |= rented
+    action = ACTION_QUARANTINE if (provider and enforce) else ACTION_NONE
```
taiberium: **Critical:** A finding on a container that is not a rented pod also sets `provider_origin`, so enforcement quarantines a provider who entered their own container. Gate `action` on `affected`.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,387 @@
...
+    finding_kinds: list[str]
+    affected_pod_ids: list[str]
+    evidence_sha256: list[str]
+    report_sha256: str
+    sensor: str
+    enforce: bool
+    action: str
+    ban_source: str | None
+    # the platform execs' payloads, deduplicated, for the daily digest (never gate on them)
+    platform_payloads: list[str]
```
taiberium: NIT: `platform_payloads` and `exec_payload` are hard to read on their own. `platform_exec_commands` and `docker_exec_command` say what they hold.

taiberium (changes requested):
Requested changes.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `9bee804`: renamed to platform_exec_commands and docker_exec_command, including the wire key in VerdictPayload. Nothing on main or in lium-platform read the old name.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `9bee804`: the verdict's action is quarantine only when a rented pod is affected and the flag is on. The check now acts on the verdict's action, not on the raw flag. A finding on a container that is not a rented pod is recorded under unmatched_containers. It passes the check, zeroes no score and tells no renter. A finding that names no container at all still counts as every rented pod on the host. Tests: `test_a_finding_on_a_container_that_is_not_a_rented_pod_takes_no_action` and `test_finding_on_a_non_rented_container_under_enforcement_is_recorded_and_acts_on_nobody`. One follow-up for the lium-platform digest is named in the body. Its SQL reads the report's findings, not the verdict's action, so it lists these recorded rounds until it filters on action.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+def is_platform_origin(finding: dict[str, Any]) -> bool:
+    """A Docker control-plane action the sensor traced to inside the executor container
+    (`host=false`, a `nested_from:<executor-stack container>` tag). The sensor already drops
+    docker-policy findings whose ancestry reaches sshd or the container's pid 1, so every such
+    finding is one whose ancestry it could not trust; until DAH-3278 hardens that on the verifier,
+    all of them count as the platform's. Everything else — an nsenter, a memory or overlayfs read,
+    anything from the host — is the provider's."""
+    if _kind(finding) not in _DOCKER_CONTROL_PLANE_KINDS:
```
taiberium, line 266: NIT: our encrypted backup runs `nsenter` into the rented pod (`executor/src/storage/restic.py`), so this classifies it as provider access. Exempt it before the flag goes on.

neurons/validators/src/services/task/checks/inspector.py
```diff
@@ -169,6 +206,64 @@ async def run(self, ctx: Context) -> CheckResult:
...
+        extra: dict[str, Any],
+    ) -> CheckResult:
+        """The act-and-publish half of the check: a provider-origin verdict becomes the MALICIOUS
+        event and, when the verdict's action is quarantine (INSPECTOR_ENFORCE_ENABLED and a rented
+        pod affected), fails the check and tells the renters. A finding that names no rented pod
+        is recorded with `unmatched_containers` and acts on nobody, whatever the flag."""
+        acts = verdict.action == ACTION_QUARANTINE
+        if acts:
+            impact = "Provider-origin access to a rented pod: score zeroed, quarantine requested"
+        elif verdict.enforce:
```
taiberium: NIT: with the flag off and no rented pod named, the event still says access to a rented pod. Branch on `verdict.affected_pod_ids` instead.

neurons/validators/src/services/task/inspector_verdict.py
```diff
@@ -0,0 +1,393 @@
...
+        return name
+    if _kind(finding) in _PATH_SHAPED_KINDS or RENTAL_VOLUME_TAG in _tags(finding):
+        command = finding.get("command")
+        if isinstance(command, str):
+            return volume_name_from_path(command)
+    return None
+
+
+def _platform_exec_commands(platform: list[dict[str, Any]]) -> list[str]:
+    seen: list[str] = []
```
taiberium: NIT: `seen` holds the deduplicated commands this function returns, so a name like `commands` reads better.

taiberium (approved):

taiberium (commented):
Second pass. `INSPECTOR_ENFORCE_ENABLED` is `core/config.py:213`, default `False`, and set in no compose file, template or deploy config, so every enforcement path is shadow only. A missing, stale or unparseable report ends as `outcome="ERROR"` with `passed=True`, and `ctx.inspector_passed` is the only way the inspector reaches `calculate_scores`, so nothing today can zero a score here. Nothing blocks. Three NITs, all about the shadow numbers the flag decision will rest on.

NIT: `inspector_verdict.py:64` — for an unencrypted volume our own restore and backup run `docker run -v volume_<pod_id>:/workspace` from the executor container. If the sensor emits a mount finding for that, it reads as provider access to a rented pod. Count it in shadow before the flip.

NIT: `inspector_verdict.py:340` — one provider finding that names neither a container nor a rental-volume path marks every rented pod on the host, and an overlay2 path has no `volumes/` segment for `volume_name_from_path`. With 21k OverlayFsRead in a day this fallback looks easy to reach; please measure how often it fires.

NIT: `checks/inspector.py:230` — in shadow, a finding on a container that is not a rented pod still gets the impact text "Provider-origin access to a rented pod", so the counts you read will be wrong in the same direction.

NIT: `inspector_verdict.py:370` — the per-pod renter entry carries host-wide `finding_kinds`, `provider_findings` and evidence hashes, so a renter can be shown kinds that came from another renter's pod. Filter per pod before rendering the event.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Not changing this because the sensor never reports that nsenter: restic's helper runs `nsenter -t <pid> -U`, a CLONE_NEWUSER-only setns, which nsenter.rs judges benign by nstype, not by container name. A name-based exemption here would let any container called lium-storage-x borrow the platform's origin; a new test pins that it cannot. The docstring cites the two sensor rules. Say if you still want the exemption by 12:00Z.

neurons/validators/src/services/task/checks/inspector.py
surcyf123: Done in `b836a23`: the impact text branches on `affected_pod_ids`; a non-rented container reads "not a rented pod" with the flag off too, with a test.

neurons/validators/src/services/task/inspector_verdict.py
surcyf123: Done in `b836a23`: renamed to `commands`.

PR #1344, opened 2026-09-09
DAH-2834 - [P2] liumd 2b: rental probe runs beside the GPU steps (flag off)

surcyf123:
Implements [DAH-2834](https://app.notion.com/p/liumd-v1-Rust-CLI-on-executor-answering-validator-commands-specs-verifyx-3cfb8bfdbde98140b3aefa0c2d657720) · stacked on #1343 · reviewer: @jam6099
**Merge order:** this PR merges INTO #1343's branch, not main. Merge #1343 first; merge this one only to fold it into #1343.

**Problem** — On the first pass the backend rental probe (`check_executor_health`, ≈ 25 s) starts right before the one `POST /verify` call and overlaps the GPU steps; `RentalVerificationCheck` awaits it when its request is identical, `Pipeline.run` removes what nobody consumed.

**Value** — not measured — the body has no Worth it line (SO §69) and the ticket gives no prod number — a human decides.

**Goal** — [DAH-2834](https://app.notion.com/p/liumd-v1-Rust-CLI-on-executor-answering-validator-commands-specs-verifyx-3cfb8bfdbde98140b3aefa0c2d657720) (P2).

**Fits** — Sits in `lium-io` — `neurons/validators`; touches backend, tests (19).

**Flag:** no flag

**Risks**
- **Merge order:** this PR merges INTO #1343's branch, not main. Merge #1343 first; merge this one only to fold it into #1343.
- Changes runtime configuration (`config.py`) on the next deploy, no pod restart — noticed only if a value is wrong — rollback: revert the file and redeploy.

**How to verify**
- `cd neurons/validators && python -m pytest tests/test_local_verify_health_check_probe.py tests/test_rental_verification_check.py tests/test_local_verify.py -q` — a halt cancels the probe and removes its pod

**Verified by the loop:** Gate: PASS f9b7bf6

**Ran it:** `tools/aws_verify.sh lium-io cb9bfa1 --app io` (the stack top #1348, which carries this head f9b7bf6) → EC2 20260915T160124Z-o32b: validators 2,773 passed, 15 skipped, plus 2 as…

**Self-review:** clean at f9b7bf6 (carried from 5baed9e: rebase-only, patch unchanged; 15 checks, 3 low)

**Parts:** backend ✓ · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — validator-internal · docs n/a — flag off · tests ✓ 19 · dashboards n/a — `step=health_check_probe` on the existing line

[...]

PR #1345, opened 2026-09-09
DAH-2834 - [P2] liumd 2a: executor facts replace read-only SSH listings (flag off)

surcyf123:
Implements [DAH-2834](https://app.notion.com/p/liumd-v1-Rust-CLI-on-executor-answering-validator-commands-specs-verifyx-3cfb8bfdbde98140b3aefa0c2d657720) · stacked on #1344 · reviewer: @jam6099
**Merge order:** this PR merges INTO #1344's branch, not main. Merge #1344 first; merge this one only to fold it into #1344.

**Problem** — `speed/LIUMD_PHASE2.md` §2. #1339's `/verify` returns `docker` / `ports` / `inspector` facts and #1340 stores them in `ctx.state.local_verify.facts`, but nothing reads them.

**Goal** — [DAH-2834](https://app.notion.com/p/liumd-v1-Rust-CLI-on-executor-answering-validator-commands-specs-verifyx-3cfb8bfdbde98140b3aefa0c2d657720) (P2).

**Fits** — Sits in `lium-io` — `neurons/validators`, `neurons/executor`; touches backend, tests (28).

**Flag:** no flag

**Risks**
- **Merge order:** this PR merges INTO #1344's branch, not main. Merge #1344 first; merge this one only to fold it into #1344.

**How to verify**
- `gh pr checks 1345 -R Datura-ai/lium-io` → all green

**Verified by the loop:** Gate: PASS c945e2c

**Ran it:** `tools/aws_verify.sh lium-io cb9bfa1 --app io` (the stack top #1348, which carries this head c945e2c) → EC2 20260915T160124Z-o32b: validators 2,773 passed, 15 skipped, plus 2 as…

**Self-review:** clean at c945e2c (15 checks, 3 low)

**Parts:** backend ✓ · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — internal · docs n/a — flag off · tests ✓ 28 · dashboards n/a — Loki line

**Docs:** none needed because both flags are off by default and the change is validator/executor-internal; the never-replaced list is in `local_verify_facts.py`.

### What changed

- `services/local_verify_facts.py`: the parser — closed status set, name grammar, `created` vs the host's `now`, bounded counts; out of bounds ⇒ absent
- `checks/local_facts.py` (`LocalFactsCheck`, before the stale cleanup): one facts-only intent; any refusal or bad answer → SSH as today
- `ContainerCleanup.cleanup(host_facts=…)`: candidates and ages from the fact (`_rental_container_age_minutes_from_facts`); removal re-aged over SSH
- `StaleContainerCleanupCheck`: a removal drops the published-ports fact for the cycle (the freed ports must stay probeable); the other facts stand
- `PortConnectivityCheck` / orchestrator: docker-published ports leave the probed window after the selection (shrinks, never shifts); no verdict from the fact

### Origin
`speed/LIUMD_PHASE2.md` §2. #1339's `/verify` already returns `docker` / `ports` / `inspector` facts and #1340 stores them in `ctx.state.local_verify.facts`, but nothing reads them — and `LocalVerifyCheck` runs after the checks that could (stale cleanup, port selection, inspector). All three facts are host-reported (an executor can say anything, exactly as in the stdout of a `docker ps` over SSH), so they may stand in for *read-only SSH commands whose answer the host controlled anyway*, and never for a probe that proves something.

### What this PR adds (flag `LOCAL_VERIFY_FACTS_ENABLED=false`, under `VALIDATOR_LOCAL_VERIFY_ENABLED`)
- **`LocalFactsCheck`** (`executor.local_facts`, default pipeline only, right before `StaleContainerCleanupCheck`): `GET /version`; if `local_verify/1` is advertised, one signed intent with `matmul=None, verifyx=None`, `deadline_s` derived like the GPU call's (`facts_deadline_s()`: the budget minus the 30 s transport margin, floor 5, capped at the executor's 20 s collector timeout; 5 s at the default 25 s budget), whole-call budget `LOCAL_VERIFY_FACTS_TIMEOUT_SECONDS=25`. The parsed `LocalFacts` goes to `ctx.state.local_facts`; the capability answer is kept so `LocalVerifyCheck` does not read `/version` again. Every other outcome (not advertised, refusal, timeout, malformed, internal error) is an info event and an empty state — nothing here can fail a node.
- **Bounds** (`services/local_verify_facts.py`): `containers` a list of ≤ 512 objects, each `name` in docker's own grammar (`[A-Za-z0-9][A-Za-z0-9_.-]*`, ≤ 128, `fullmatch`) and `status` in docker's closed set (`created|restarting|running|paused|removing|exited|dead`) — one bad item drops the whole fact; `created` is parsed by the validator (RFC 3339 within 20–40 chars; docker's nine fractional digits are rewritten to six) and an unparsable one drops that container's AGE, not the fact; `now` must be an int in (0, 2**40) (the host's clock — the age is host-clock minus host-created, the same pair the SSH path reads; an unbounded int would overflow the division, so it is no clock); `published_by_docker` ≤ 4096 ints in 1..65535; `lib_sha256` exactly 64 lowercase hex. Steps that are not `ok` contribute nothing.
- **Reader 1, `ContainerCleanup.cleanup(…, host_facts)`.** The fact is aged OUTSIDE the cleanup's guarded block: any exception lands on the SSH listing, never on "removed nothing this cycle". When the fact can age containers: the candidate list is the fact's rental-prefixed names (no `docker ps -a --filter`); a candidate the fact says is younger than the threshold is skipped (no inspect+date pair); a candidate the fact says is stale — or cannot age — is **re-aged over SSH** (`inspect .Created` + `date +%s`) and removed only if SSH agrees. A lying clock removes nothing; a name the fact omits is a name the SSH listing (host-controlled too) would have hidden the same way. Rented containers and fillers are never touched, as before.
- **Reader 2, the probed window.** `verify_ports(published_ports=…)` hands the docker-published ports to the orchestrator APART from `unavailable_ports`: the selector still picks today's 300-port window from the backend-controlled set (rentals + fillers), and the published ports are removed from that window afterwards — the ports probed WITH the fact are a subset of the ports probed WITHOUT it, so an executor cannot move the sample (the count verdict and the rental probe's port are computed over it). A list that covers the whole window keeps the window: the connect-back decides. Which of the kept ports work is still proven by the connect-back; the DinD/sysbox probe runs as before. No skip and no verdict comes from the fact (round 1 removed a B-2 skip that left `verified_ports=[]` for the fatal `RentalVerificationCheck` to fail on).
- **Reader 3, observe-only.** `InspectorRentedCheck` logs `step=inspector reason=digest_match|digest_mismatch` against `InspectorValidationService.local_checksum`; the SSH `sha256_from_executor` pre-check and the run are unchanged (jam6099's row — consumption waits for the agreement rate).
- **Executor:** `_docker_facts` gains `"now": int(time.time())`.

### NEVER replaced by a fact (kept from the design)
The DinD probe and the sysbox proof (`hello-world` under sysbox, the validator's inbound connection — `docker.sysbox_runtime` seeds nothing); the connect-back (`published_by_docker` says which ports NOT to try, never which work); the filler liveness verdict and the stale removal (`docker rm` after an SSH re-read); the cached-template verdict; the inspector run.

### How to verify
28 tests (26 functions, the parametrised cases counted): the parser's closed sets and bounds (a non-string status, a clock outside (0, 2**40)) (every bad shape → fact absent, unparsable `created` → unaged container); the check's intent (no GPU step, `deadline_s=5` at the default budget), bounded facts, not advertised, http 500 / nonce mismatch / no usable fact → `LOCAL_FACTS_UNAVAILABLE` with the SSH listings in place, a bug → fallback logged as a WARNING `outcome=error` with the traceback, a slow executor → fallback inside the budget (the production client), one `/version` per cycle; the cleanup reads candidates and ages from the fact and re-ages the stale one over SSH before `rm`, a lying clock removes nothing, an unaged container falls to the SSH pair, rented ones untouched, facts that cannot age → `docker ps` as today, the check hands the state facts through; published ports handed through only when present, the executor's word alone never skips the probe, the probed-with-fact set is a subset of the probed-without set and the selector never sees the fact, a list covering the window keeps the window, a fact the cleanup cannot age falls to `docker ps -a`; the inspector digest is logged and the SSH pre-check still runs; pipeline order (facts before both readers, not in dry-run).

### Not in this PR
Consuming the inspector digest; `docker.images` (RepoDigests) for the cached-template check (a further executor key); the DinD start (2c).

### Suite
Round 0 (its pre-rewrite head, now only in the stack's history): EC2 aws_verify 20260909T082618Z-ijqv, 139 targeted validator tests passed. Stack head (this branch plus 2c, aws_verify 20260909T102826Z-vcot at a 9 Sep stack head no longer on the branch, EC2 c6i.2xlarge): validators 2120 passed / 15 skipped / 2 failed — main's own `test_scrape_infiniband::test_an_unreadable_sysfs_tree_is_not_reported_as_no_devices` and `test_sshd_bootstrap_script::test_adopt_path_hardens_config_for_key_only_auth` (both fail on origin/main, listed in lium-io#1332's body) — executor 164 passed, miners 27 passed.

### Ran it

`tools/aws_verify.sh lium-io cb9bfa1 --app io` → EC2 20260915T160124Z-o32b at the stack top #1348 cb9bfa1, which carries this head c945e2c (restacked on lium-io#1344's f9b7bf6 plus one fix commit, the `float | None` annotations in container_cleanup.py): validators 2,773 passed, 15 skipped (+ the 2 root-sensitive tests as loopverify: 2 passed), executor 374 passed, miners 43 passed; c6i.2xlarge warm box, 4 min wall. Not on a real node.

### Self-review

Verdict `clean` at `c945e2c` · 15 checks · by subagent-r226-sr1345 · 2026-09-15 14:27Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | STALE-BODY | `PR body:0` | Property: every test-evidence line in the body names a run at the reviewed head. The head is now c945e2c (a rebase onto liumd-phase2-b f9b7bf6 plus one annotation-only commit), while **Ran it**, **Verified by the loop** and **Self-review** all cite the previous head. `git range-diff` shows the 12 patch commits identical after the rebase and the new commit changes only two return/parameter annotations in container_cleanup.py (verified by reading: `_fact_ages_or_none` returns what `_rental_container_age_minutes_from_facts` returns, typed `dict[str, float / None] / None` at line 498, and `_list_rental_containers` passes the same dict through or returns None), so the annotations cannot change behaviour. What did change under the patch is the base: f9b7bf6 rewrites `checks/local_verify.py`, `pipeline.py`, `pipeline_factory.py` and `rental_verification.py` (the health-check-probe rounds), which this PR's `LocalVerifyCheck` and pipeline-order tests sit on top of; the 20260915T110230Z-p9my run was on 5baed9e's base, not this one. Not run here: pytest, per the review constraints; `a compile check` and `ruff --select F,E9` on container_cleanup.py at c945e2c report only the pre-existing F541 at line 529. | Re-run the covering suite (the 7 touched validator test files + executor tests/test_local_verify.py, or the whole validators suite as before) at c945e2c on EC2 and cite that run id and sha in Ran-it / Verified-by-the-loop; `--record` this verdict so the Self-review line names c945e2c. |
| low | TEST-GAP | `neurons/validators/tests/test_local_verify_facts.py:253` | Carried from the the previous head self-review (subagent-r206b), still true at c945e2c: the only assertion on the derived deadline is `intent["deadline_s"] == 5`, which exercises the ge=5 floor with the process-wide default budget (it holds for any budget in 5..34 and the test does not pin the setting). The cap — the reason `FACTS_DEADLINE_CAP_S = 20` exists (the executor's collector timeout) — and the budget-minus-margin slope have no test; dropping the `min(...)` in `facts_deadline_s()` keeps this test green. `git grep facts_deadline_s` on the tests at c945e2c finds nothing. | monkeypatch `LOCAL_VERIFY_FACTS_TIMEOUT_SECONDS` to 25 in that test, and add one table test on `facts_deadline_s()`: budget 25 -> 5, 40 -> 10, 60 -> 20 (the cap). |
| low | CODE-QUALITY | `neurons/validators/src/core/config.py:301` | Carried from the the previous head self-review, unchanged at c945e2c: the comment on `LOCAL_VERIFY_FACTS_TIMEOUT_SECONDS` reads 'Whole-call budget for the facts-only call; the executor caps each collector at 20 s' and does not say the intent's `deadline_s` is derived from this value (`facts_deadline_s`: budget - 30 s, floor 5, cap 20), so an operator raising it to give a slow host more time does not learn the executor deadline grows only above 35 s. | Append 'the intent's `deadline_s` is derived from it (budget - 30 s, floor 5, cap 20; see checks/local_facts.facts_deadline_s)' in config.py and one clause in .env.template. |

Mechanical notes dismissed: `neurons/validators/src/services/container_cleanup.py:174` Delta commit c945e2c verified: `_fact_ages_or_none` returns `self._rental_container_age_minutes_from_facts(host_facts)` (line 179) or None, and that static method is typed and built as `dict[str, float | None] | None` (lines 498-511: a container with `created_at is None` maps to None); `_list_rental_containers` (190-197) returns `(list(fact_ages), fact_ages)` or `(ssh listing, None)`, so `tuple[list[str], dict[str, float | None] | None]` is exact. No other caller exists (`git grep` at c945e2c: only cleanup() lines 96/98). The sole consumer of the ages, cleanup() lines 112-114, does `fact_age = fact_ages.get(name)` then `if fact_age is not None and fact_age <= threshold: continue`, so a None age falls to the SSH inspect+date pair as the docstring says; `extra['from_facts'] = fact_ages is not None` (line 99) reads only the outer Optional. Keeping the whole annotation as a string (rather than quoting only `LocalFacts`) matches line 74's existing style; harmless. `a compile check` clean; the r206b CODE-QUALITY low at 174/191 is closed. · `neurons/validators/src/services/container_cleanup.py:180` Mechanical 'suppression added': the reason for `# noqa: BLE001` is the helper's docstring two lines above (175-177: 'a fact that raises is logged and treated as absent (the fact is an optimisation, SSH is the source of truth)'); the exception is logged at WARNING with the error (181-186) and the return of None sends cleanup() to `_get_all_rental_containers` (line 197), which test_a_fact_the_cleanup_cannot_age_lands_on_the_ssh_listing_not_on_removing_nothing drives. Untouched by c945e2c except the annotation on line 174. · `neurons/validators/src/services/task/checks/local_facts.py:74` Mechanical 'suppression added': the reason follows `# noqa: BLE001` on the same line ('the pipeline has no guard; a bug here must not end the node's cycle'), and the handler logs at WARNING with `exc_info` and `outcome=error` before the `internal_error` fallback (test_a_bug_in_the_check_is_a_fallback_not_a_halt_but_a_loud_one). Identical to the previous head (range-diff `=`). · `neurons/validators/tests/test_local_verify_facts.py:426` Mechanical 'f-string without placeholders': line 426 at c945e2c is `assert any("docker rm -f" in c and "pod_old" in c for c in commands)` — the scanner matched `f" in c and "` inside the `rm -f"` literal; there is no f-string on the line. False positive. · `neurons/validators/src/services/container_cleanup.py:529` `ruff --select F,E9` on the file at c945e2c reports one F541 (`f"Error getting rental containers"` in `_get_all_rental_containers`); it is pre-existing code outside every hunk of this PR. · `neurons/validators/src/services/task/checks/inspector.py:279` Carried from the an earlier head/an earlier head verdicts, patch unchanged since (range-diff `=`): the direct read `ctx.services.inspector.local_checksum` is safe because `InspectorValidationService.__init__` assigns it unconditionally and `_observe_local_digest` runs on the same path that awaits `validate_rented_executor` on the same object; the only test reaching it with a digest fact gives the double `local_checksum=DIGEST`. · `neurons/validators/src/services/task/pipeline.py:21` Carried from the prior verdicts, unchanged by c945e2c: the runtime import `from services.local_verify_facts import LocalFacts` adds no cycle (`local_verify_facts` imports stdlib and `services.local_verify_client`, already imported by `pipeline.py`; nothing under it imports `services.task`). · `neurons/validators/tests/test_pipeline_default_scenarios.py:42` Carried from the prior verdicts, unchanged by c945e2c: the two `MockContainerCleanup.cleanup` doubles return `0, [], []`, matching `ContainerCleanup.cleanup -> tuple[int, list[str], list[str]]` and the 3-tuple unpack in `StaleContainerCleanupCheck`; the F541 reports at lines 421/605 are 2025-10-29 print lines outside every hunk.

### Verified

Gate: PASS (c945e2c) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1345)

EC2 aws_verify 20260915T160124Z-o32b at the stack top cb9bfa1 (carries c945e2c): validators 2,773 + 2 loopverify, executor 374, miners 43, all passed.

</details>


neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -189,7 +189,12 @@ async def _run(self, ctx: Context, started: list[RentalProbe]) -> CheckResult:
             )
 
         client = self._client_factory(ctx)
-        capabilities = await client.capabilities(ctx.executor)
+        # Phase 2: the early facts call (checks/local_facts) already read /version this cycle.
+        facts = ctx.state.local_facts
+        if facts is not None and getattr(facts, "capabilities", None):
```
taiberium: NIT: an executor without `local_verify/1` stores an empty capability set, so this test still costs a second `/version` call. `if facts is not None` avoids it.

neurons/validators/src/services/local_verify_facts.py
```diff
@@ -0,0 +1,177 @@
...
+_SHA256_RE = re.compile(r"[0-9a-f]{64}")
+_RFC3339_FRACTION_RE = re.compile(r"(\.\d{1,9})(?=Z|[+-]\d{2}:\d{2}$)")
+
+
+@dataclass(frozen=True)
+class HostContainer:
+    name: str
+    status: str  # one of DOCKER_STATUSES
+    created_at: int | None  # epoch seconds, None when the executor's string did not parse
+    image: str | None = None
```
taiberium: NIT: nothing reads `HostContainer.image`. Remove the field and `IMAGE_MAX_CHARS` until the cached-template check needs them.

neurons/validators/src/services/local_verify_facts.py
```diff
@@ -0,0 +1,177 @@
...
+    """What the early facts call learnt; every field is optional evidence, never a verdict."""
+
+    containers: tuple[HostContainer, ...] | None = None  # None = the docker fact was unusable
+    host_now: int | None = None  # the host clock at collection, epoch seconds
+    published_ports: frozenset[int] | None = None  # None = the ports fact was unusable
+    inspector_lib_sha256: str | None = None
+    capabilities: frozenset[str] = field(default_factory=frozenset)
+    round_trip_ms: int = 0
+    executor_elapsed_ms: int = 0
+    steps: dict[str, str] = field(default_factory=dict)  # step -> status as answered
```
taiberium: NIT: this field holds one status per step, not the steps themselves. `step_statuses` reads true to the type.

neurons/validators/src/services/task/checks/port_connectivity.py
```diff
@@ -53,6 +65,7 @@ async def run(self, ctx: Context) -> CheckResult:
                 "executor_uuid": ctx.executor.uuid,
                 "executor_ip": ctx.executor.address,
             },
+            **({"published_ports": published_ports} if published_ports is not None else {}),
```
taiberium: NIT: `published_ports` already defaults to None, so the conditional splat changes nothing. Pass `published_ports=published_ports` directly.

neurons/validators/src/services/container_cleanup.py
```diff
@@ -445,6 +478,24 @@ async def force_remove_health_checks(
             )
             return 0
 
+    @staticmethod
+    def _rental_container_ages_from_facts(
+        host_facts: "LocalFacts | None",
+    ) -> dict[str, Optional[float]] | None:
+        """{name: age_minutes | None} for the fact's rental-prefixed containers, in the SSH
+        listing's order; None when the fact cannot age containers (no listing, or no host clock)."""
+        if host_facts is None or not getattr(host_facts, "can_age_containers", lambda: False)():
```
taiberium: NIT: `LocalFacts` always has this method, so the `getattr` default only hides a typo. Call `host_facts.can_age_containers()`.

taiberium (approved):

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `c3f1dde`: Done in `c3f1dde8`: `if facts is not None` — an empty set read by the facts call is "nothing advertised", not a gap, so no second `/version`.

neurons/validators/src/services/local_verify_facts.py
surcyf123: Done in `c3f1dde`: Done in `c3f1dde8`: `HostContainer.image` and `IMAGE_MAX_CHARS` removed; the executor still sends `image`, the parser drops it until the cached-template check reads it.

neurons/validators/src/services/local_verify_facts.py
surcyf123: Done in `c3f1dde`: Done in `c3f1dde8`: `LocalFacts.step_statuses` (the event key stays `steps`).

neurons/validators/src/services/task/checks/port_connectivity.py
surcyf123: Done in `c3f1dde`: Done in `c3f1dde8`: `published_ports=published_ports`. The splat had been hiding that the `DummyConnectivityService` in `test_port_connectivity_check.py` lacked the parameter (8 tests failed without it); the double has the real signature now.

neurons/validators/src/services/container_cleanup.py
surcyf123: Done in `c3f1dde`: Done in `c3f1dde8`: `host_facts.can_age_containers()` directly, and the same for `facts.published_ports` in the port check.

neurons/validators/src/services/task/checks/stale_container_cleanup.py
```diff
@@ -50,10 +50,13 @@ def __init__(self) -> None:
         self._last_sweep_at: dict[str, float] = {}
 
     async def run(self, ctx: Context) -> CheckResult:
+        # liumd phase 2: the executor's own listing (checks/local_facts) replaces the `docker ps`
+        # and the per-candidate age pair when present; the `docker rm` stays SSH-proven.
         removed_count, removed_names = await ctx.services.container_cleanup.cleanup(
             ssh_client=ctx.ssh,
             rented_data=ctx.state.rented_data,
             executor_uuid=ctx.executor.uuid,
+            **({"host_facts": ctx.state.local_facts} if ctx.state.local_facts is not None else {}),
```
taiberium: NIT: `host_facts` already defaults to None, so the splat changes nothing. It also hides that `MockContainerCleanup.cleanup` has no such parameter.

```suggestion
            host_facts=ctx.state.local_facts,
```

neurons/validators/src/services/container_cleanup.py
```diff
@@ -445,6 +478,24 @@ async def force_remove_health_checks(
             )
             return 0
 
+    @staticmethod
+    def _rental_container_ages_from_facts(
```
taiberium: NIT: these ages are minutes, not seconds. Rename to `_rental_container_age_minutes_from_facts`, and write the return as `dict[str, float | None] | None`.

neurons/validators/src/services/local_verify_facts.py
```diff
@@ -0,0 +1,169 @@
...
+    return frozenset(ports)
+
+
+def parse_inspector_digest(data: Any) -> str | None:
+    digest = data.get("lib_sha256") if isinstance(data, dict) else None
+    return digest if isinstance(digest, str) and _SHA256_RE.fullmatch(digest) else None
+
+
+def parse_facts(
+    steps: dict[str, Any],
```
taiberium: NIT: `dict[str, Any]` hides the type the docstring already names. Import `StepEvidence` and write `dict[str, StepEvidence]`.

neurons/validators/src/services/task/checks/inspector.py
```diff
@@ -170,6 +176,30 @@ async def run(self, ctx: Context) -> CheckResult:
         )
 
 
+def _observe_local_digest(ctx: Context) -> None:
+    """liumd phase 2, observe-only: the executor's `inspector.lib_sha256` fact against the digest
+    `validate_rented_executor` is about to require over SSH (`sha256_from_executor`). Logged, never
+    consumed — the SSH pre-check and the inspector run below are unchanged; Loki's agreement rate
+    is what decides whether the fact may ever stand in for the pre-check (jam6099's call)."""
+    facts = ctx.state.local_facts
+    reported = getattr(facts, "inspector_lib_sha256", None) if facts is not None else None
```
taiberium: NIT: `LocalFacts` always has this field, so the getattr default only hides a typo. Use `facts.inspector_lib_sha256`.

neurons/validators/src/services/task/checks/local_facts.py
```diff
@@ -0,0 +1,188 @@
...
+        if not (settings.VALIDATOR_LOCAL_VERIFY_ENABLED and settings.LOCAL_VERIFY_FACTS_ENABLED):
+            return CheckResult(
+                passed=True, event=render_message(Msg.DISABLED, ctx=ctx, check_id=self.check_id)
+            )
+        try:
+            return await self._run(ctx)
+        except Exception as exc:  # noqa: BLE001 — the pipeline has no guard; a bug here must not end the node's cycle
+            return self._unavailable(ctx, "internal_error", f"{type(exc).__name__}: {exc}")
+
+    async def _run(self, ctx: Context) -> CheckResult:
```
taiberium, line 88: NIT: this method runs 79 lines and does two jobs. Move the event and metric build into its own method.

taiberium (approved):

neurons/validators/src/services/task/checks/stale_container_cleanup.py
surcyf123: Done in `3792747`: `host_facts=ctx.state.local_facts` as a plain keyword; the three test doubles grew the parameter.

neurons/validators/src/services/container_cleanup.py
surcyf123: Done in `3792747`: `_rental_container_age_minutes_from_facts`, returning `dict[str, float | None] | None`.

neurons/validators/src/services/local_verify_facts.py
surcyf123: Done in `3792747`: `parse_facts(steps: dict[str, StepEvidence], …)`, imported from the client module.

neurons/validators/src/services/task/checks/inspector.py
surcyf123: Done in `3792747`: `facts.inspector_lib_sha256` read directly.

neurons/validators/src/services/task/checks/local_facts.py
surcyf123: Done in `3792747`: `_run` ends at `parse_facts`; the event and metric build moved to `_report(ctx, facts)`.

taiberium (approved):

neurons/validators/src/services/task/pipeline.py
```diff
@@ -145,6 +145,11 @@ class ContextState:
     # not attempted or fell back entirely; the capability and VerifyX checks consume a judged
     # step when present and run over SSH otherwise.
     local_verify: LocalVerifyOutcome | None = None
+    # liumd phase 2: the executor's read-only host facts from the early facts-only `POST /verify`
+    # (services.local_verify_facts.LocalFacts), bounded and parsed. None = not asked or unusable;
+    # the stale cleanup and the port selector read them when present and run their SSH listings
+    # otherwise. Never a verdict: every fact only narrows what the SSH-proven steps go on to do.
+    local_facts: Any | None = None
```
taiberium, line 152: NIT: `Any` hides the type here. `LocalFacts | None` works, because this module already imports `local_verify_client`.

neurons/validators/src/services/task/checks/inspector.py
```diff
@@ -265,6 +267,30 @@ async def _act_on_provider_origin(
...
+def _observe_local_digest(ctx: Context) -> None:
+    """liumd phase 2, observe-only: the executor's `inspector.lib_sha256` fact against the digest
+    `validate_rented_executor` is about to require over SSH (`sha256_from_executor`). Logged, never
+    consumed — the SSH pre-check and the inspector run below are unchanged; Loki's agreement rate
+    is what decides whether the fact may ever stand in for the pre-check (jam6099's call)."""
+    facts = ctx.state.local_facts
+    reported = facts.inspector_lib_sha256 if facts is not None else None
+    if reported is None:
+        return
+    expected = getattr(ctx.services.inspector, "local_checksum", None)
```
taiberium, line 279: NIT: same point as your earlier getattr notes. `ctx.services.inspector` always has `local_checksum`, so read it directly.

```suggestion
    expected = ctx.services.inspector.local_checksum
```

neurons/validators/src/services/task/checks/local_facts.py
```diff
@@ -0,0 +1,236 @@
...
+LOCAL_VERIFY_OUTCOME_EVENT = "[local_verify] outcome"
+# The executor caps each fact collector at 20 s (FAST_STEP_TIMEOUT_SECONDS); ask for no more.
+FACTS_DEADLINE_S = 20
+
+
+class LocalFactsCheck:
+    check_id = "executor.local_facts"
+    fatal = False
+
+    def __init__(self, client_factory=None):
```
taiberium, line 48: NIT: `client_factory` has no type. Annotate it `Callable[[Context], LocalVerifyClient] | None`.

neurons/validators/tests/test_pipeline_default_scenarios.py
```diff
@@ -38,7 +38,7 @@
 
 class MockContainerCleanup:
     """Mock container cleanup service for tests."""
-    async def cleanup(self, ssh_client, rented_data, executor_uuid):
+    async def cleanup(self, ssh_client, rented_data, executor_uuid, host_facts=None):
```
taiberium, line 41: NIT: this double returns two values while `cleanup` returns three. Return `0, [], []` here and in `test_rented_machine_check.py`.

taiberium (approved):

neurons/validators/src/services/task/pipeline.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/task/checks/inspector.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/task/checks/local_facts.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/tests/test_pipeline_default_scenarios.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/task/pipeline.py
surcyf123: Done in `1d4f696`: `local_facts` is now `LocalFacts | None`; the import comes from `services.local_verify_facts`, which has no import back into the pipeline.

neurons/validators/src/services/task/checks/inspector.py
surcyf123: Done in `1d4f696`: `expected` reads `ctx.services.inspector.local_checksum` directly.

neurons/validators/src/services/task/checks/local_facts.py
surcyf123: Done in `1d4f696`: `client_factory` is annotated `Callable[[Context], LocalVerifyClient] | None`.

neurons/validators/tests/test_pipeline_default_scenarios.py
surcyf123: Done in `1d4f696`: both test doubles return `0, [], []` now, here and in `test_rented_machine_check.py`.

neurons/validators/src/services/task/checks/local_facts.py
```diff
@@ -0,0 +1,237 @@
...
+
+from ..messages import LocalFactsMessages as Msg
+from ..messages import render_message
+from ..pipeline import CheckResult, Context
+
+logger = logging.getLogger(__name__)
+
+LOCAL_VERIFY_OUTCOME_EVENT = "[local_verify] outcome"
+# The executor caps each fact collector at 20 s (FAST_STEP_TIMEOUT_SECONDS); ask for no more.
+FACTS_DEADLINE_S = 20
```
taiberium: NIT: a 20 s deadline in a 25 s budget leaves 5 s of margin, where the GPU call takes 30 s from `executor_deadline_s()`. Derive this budget the same way.

neurons/validators/src/services/container_cleanup.py
```diff
@@ -68,9 +71,18 @@ async def cleanup(
         ssh_client,
         rented_data: Optional[RentedExecutorsResponse],
         executor_uuid: str,
+        host_facts: "LocalFacts | None" = None,
```
taiberium, line 74: NIT: `cleanup` is now 99 lines and does two jobs. Move the choice between the fact and the SSH listing into its own method.

neurons/validators/src/services/local_verify_facts.py
```diff
@@ -0,0 +1,174 @@
...
+    host_now: int | None = None  # the host clock at collection, epoch seconds
+    published_ports: frozenset[int] | None = None  # None = the ports fact was unusable
+    inspector_lib_sha256: str | None = None
+    capabilities: frozenset[str] = field(default_factory=frozenset)
+    local_verify_port: int | None = None  # the loopback port /version named; None = nothing to tunnel to
+    round_trip_ms: int = 0
+    executor_elapsed_ms: int = 0
+    step_statuses: dict[str, str] = field(default_factory=dict)  # step -> status as answered
+
+    def can_age_containers(self) -> bool:
```
taiberium: NIT: a blind reader splits `can_age_containers` between "the host reports ages" and "we may reap containers".

```suggestion
    def has_container_ages(self) -> bool:
```

neurons/validators/src/services/task/checks/inspector.py
```diff
@@ -265,6 +267,30 @@ async def _act_on_provider_origin(
         return CheckResult(passed=not acts, event=event, updates=updates)
 
 
+def _observe_local_digest(ctx: Context) -> None:
```
taiberium: NIT: the name does not say whose digest or what it observes.

```suggestion
def _log_inspector_digest_agreement(ctx: Context) -> None:
```

neurons/validators/src/services/task/checks/port_connectivity.py
```diff
@@ -37,6 +37,18 @@ async def run(self, ctx: Context) -> CheckResult:
         # fillers has no rented_executor entry at all — hence the separate lookup.
         filler_ports = rented_data.get_filler_ports(ctx.executor.uuid) if rented_data else []
 
+        # liumd phase 2: ports the executor's docker already publishes (checks/local_facts) leave
```
taiberium: NIT: the same "never a verdict, SSH-proven, flag off" note now stands in seven files. Keep it in the `local_verify_facts` docstring and leave one line here.

neurons/validators/tests/test_local_verify_facts.py
```diff
@@ -0,0 +1,697 @@
...
+        services=build_services(
+            connectivity=connectivity, redis=SimpleNamespace(renting_in_progress=AsyncMock(return_value=False))
+        ),
+        state=build_state(local_facts=facts, rented_data=rented_data, sysbox_runtime=True),
+    )
+    return ctx, connectivity
+
+
+@pytest.mark.asyncio
+async def test_published_ports_are_handed_to_the_selector_only_when_the_fact_is_present():
```
taiberium: NIT: the test hands the fact to the connectivity service, not the selector, and the sibling test proves the selector never sees it. Rename it.

taiberium (approved):

taiberium:
**Necessity: question.**

The ticket proves the pain: 25 to 40 serial SSH commands per cycle, 8 of them 6.0 s over SSH against 0.95 s in one call, and an express lane at 162 s p50 against a 70-95 s target. The gain for this slice was measured on the stub, not on a node.

On a node with no unrented rental container this trades one `docker ps -a` for a new `POST /verify` round trip, so it can come out slower there.

Smaller version: keep the two cheap readers, the port shrink and the inspector digest log, about 60 lines. Hold the container-cleanup reader until Loki shows the agreement rate.

This stays worth landing because #1346 and #1347 build on the same facts call.

neurons/validators/src/services/task/checks/local_facts.py
surcyf123: Looking at it now. Fix and reply by 11:50Z.

neurons/validators/src/services/container_cleanup.py
surcyf123: Looking at it now. Fix and reply by 11:50Z.

neurons/validators/src/services/local_verify_facts.py
surcyf123: Looking at it now. Fix and reply by 11:50Z.

neurons/validators/src/services/task/checks/inspector.py
surcyf123: Looking at it now. Fix and reply by 11:50Z.

neurons/validators/src/services/task/checks/port_connectivity.py
surcyf123: Looking at it now. Fix and reply by 11:50Z.

neurons/validators/tests/test_local_verify_facts.py
surcyf123: Looking at it now. Fix and reply by 11:50Z.

surcyf123:
@taiberium Looking at it now. Fix and reply by 11:50Z.

surcyf123:
@taiberium The facts call is made once for all three readers. Dropping the container-cleanup reader keeps the `POST /verify` round trip and only gives back the `docker ps -a` it replaced. On a node with no unrented container the reader costs nothing extra: the fact already answered and no SSH listing or age pair runs. It stands in only when the fact carries ages (`has_container_ages`); otherwise the SSH path runs as today. The slice is off by default (`LOCAL_VERIFY_FACTS_ENABLED=false`), so the agreement rate comes from staging before any node runs it. I kept the reader in. If you want it behind its own flag, say so and I add one.

neurons/validators/src/services/task/checks/local_facts.py
surcyf123: Done in 4df2c40: `facts_deadline_s()` is `executor_deadline_s(LOCAL_VERIFY_FACTS_TIMEOUT_SECONDS)` capped at the executor's 20 s collector timeout. With the default 25 s budget that is 5 s, enough for the roughly 1 s the facts take; the test asserts 5. A re-approval click is needed, the push came after your approve.

neurons/validators/src/services/container_cleanup.py
surcyf123: Done in 4df2c40: The fact-or-SSH choice moved into `_fact_ages_or_none` and `_list_rental_containers`; `cleanup` keeps the removal loop.

neurons/validators/src/services/local_verify_facts.py
surcyf123: Done in 4df2c40: Renamed to `has_container_ages`, callers and tests updated.

neurons/validators/src/services/task/checks/inspector.py
surcyf123: Done in 4df2c40: Renamed to `_log_inspector_digest_agreement` with a two-line docstring.

neurons/validators/src/services/task/checks/port_connectivity.py
surcyf123: Done in 4df2c40: The full note lives in the `local_verify_facts` module docstring; the seven other places now carry one line that points there.

neurons/validators/tests/test_local_verify_facts.py
surcyf123: Done in 4df2c40: Renamed to `test_published_ports_reach_the_connectivity_service_only_when_the_fact_is_present`.

neurons/validators/src/services/container_cleanup.py
```diff
@@ -154,6 +171,31 @@ async def cleanup(
 
         return len(removed_names), removed_names, unremovable_names
 
+    def _fact_ages_or_none(self, host_facts: "LocalFacts | None", extra: dict) -> "dict[str, float] | None":
```
taiberium: NIT: the helper returns `_rental_container_age_minutes_from_facts`, whose values are `float | None`. The annotation here and on `_list_rental_containers` needs the same `float | None`.

taiberium (dismissed):

neurons/validators/src/services/container_cleanup.py
surcyf123: Done in `c945e2c`: both readers now say `dict[str, float | None] | None`, the helper's own type. The one consumer already skips a None age.

surcyf123:
Parked, not closed (Muhammad, DECISION-LIUMD-STACK-VALUE, 16 Sep 04:52Z): the stack stays as it is until lium-io#1348's staging measurement is in. #1348 is being cut out of the stack and based on main. No pushes here until then.

surcyf123:
Closed on DECISION-LIUMD-STACK-VALUE default B (Muhammad silent since 16 Sep 22:15Z; #1348 closed 22:07Z on the same reading): the stack speeds a path that is OFF in prod (VALIDATOR_LOCAL_VERIFY default False, the new flags also off; staging dropped VALIDATOR_LOCAL_RENT_ENABLED in lium-io-deployment#49), staging arm B measured 742 vs 806 ms p50 (64 ms) and prod was never measured; about 3.7k lines across the four PRs is not worth that. Phase 1 (#1339, #1340) stays merged; design/LIUMD_RUST_PLAN.md stays for the diagram review. Re-open if staging measures a gain with the flags on. Audit: https://github.com/Datura-ai/lium-loop/blob/main/reports/VALUE_AUDIT_20260917.md

PR #1347, opened 2026-09-09
DAH-2834 - [P0] GPU /verify leaves right after the facts call as a background task

surcyf123:
Implements [DAH-2834](https://app.notion.com/p/liumd-v1-Rust-CLI-on-executor-answering-validator-commands-specs-verifyx-3cfb8bfdbde98140b3aefa0c2d657720) · stacked on #1346 · reviewer: @jam6099
**Merge order:** this PR merges INTO #1346's branch, not main. Merge #1346 first; merge this one only to fold it into #1346.

**TL;DR** — `speed/LIUMD_PHASE2.md` §5 "what still stands" item (1): after 2c the port check's connect-back batch (up to 300 ports), the `hello-world` under sysbox and the `docker rm`. GPU /verify leaves right after the facts call as a background task. Touches validator/executor (`neurons/validators/.env.template`) — read that file first.

**What changed**
- `checks/local_verify.py`: `LocalVerifyStartCheck` starts the call as `PendingVerify` (`cancel_and_await`); `LocalVerifyCheck` consumes it; `_PrepareFailed` replaces #1340's `_NothingToSend` (#1340 merged 15 Sep)
- `pipeline_factory.py` / `pipeline.py`: start check after `LocalFactsCheck`; the settle step cancels an unconsumed early call through the `BackgroundProbe` Protocol (`step=gpu_early`)
- `checks/gpu_usage.py`: `gpu_usage_check_will_read_the_live_card` — no early start when that check would re-read the card
- `config.py` / `.env.template` / `messages.py`: `LOCAL_VERIFY_GPU_EARLY_START` (off), `EARLY_STARTED` / `EARLY_SKIPPED`
- `dind_probe.py` / `orchestrator.py`: a removed prestarted DinD fails at the first refusal (`DindContainerGone`); the prestarted port is one of `selected_ports`

**Risks**
- **Merge order:** this PR merges INTO #1346's branch, not main. Merge #1346 first; merge this one only to fold it into #1346.

**How to verify**
- `gh pr checks 1347 -R Datura-ai/lium-io` → all green

**Verified by the loop:** Gate: PASS 95b45c8

**Ran it:** `tools/aws_verify.sh lium-io 95b45c8 --app io --bundle` → EC2 run 20260915T154828Z-tqfh at 95b45c8 (this head, on #1346 242a740): validators 2,689 passed + 2 as loopverify, executor 310…

**Self-review:** clean at 95b45c8 (15 checks, 4 low)

**Parts:** backend ✓ · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — validator-internal · docs n/a — flag off · tests ✓ 38 items (28 functions, one 11-case parametrize) · dashboards n/a — Loki line only

**Docs:** none needed because the flag is off by default and the change is validator-internal.

### Origin
`speed/LIUMD_PHASE2.md` §5 "what still stands" item (1): after 2c the port check's connect-back batch (up to 300 ports), the `hello-world` under sysbox and the `docker rm`, then the image check and the tenant check, are ≈ 10–12 s of SSH round trips that precede the GPU call on the first pass. The executor's GPU steps (matmul ≈ 10 s ∥ VerifyX ≈ 45 s) do not need the validator during that time — the call can leave earlier and be judged later, the same shape as 2b's rental probe. Projected first-pass p50: ≈ 112 s → ≈ 100 s (§5 method).

### What this PR adds (flag off)
- **`LocalVerifyStartCheck`** (`checks/local_verify.py`, subclass of `LocalVerifyCheck`, `check_id=executor.local_verify_start`, `fatal=False`). Runs only with `VALIDATOR_LOCAL_VERIFY_ENABLED` and `LOCAL_VERIFY_GPU_EARLY_START`. Starts nothing (`EARLY_SKIPPED`, `step=gpu_early reason=…`) when: a call is already pending; there was no facts call this cycle (`no_facts` — the facts `/version` answer is what says the executor takes the call; no second GET here); `/version` names no `local_verify_port` (`no_tunnel_port` — tested before the steps, since without a port the facts POST was never made and its steps are empty too); the facts POST itself did not come back (`facts_unanswered`: busy, timeout, refused — the executor is not taking intents right now); `local_verify` is not advertised; the cycle is scored (`not ctx.config.unscored`, whatever `LOCAL_VERIFY_FIRST_PASS_ONLY` says); this cycle's `rented_data` shows a customer pod on the node (`rented` — `TenantEnforcementCheck` would halt it before the judge and its GPUs are a renter's); `GpuUsageCheck` (fatal, between the start and the judge) would read the live card this cycle (`gpu_reread_pending`: the scrape shows a wedge candidate with no process — its cure and re-sample — or ownerless VRAM above the floor — its confirmation; `reads_the_live_card` in `checks/gpu_usage.py`, on the same snapshot that check reads; with our matmul as the compute app those reads would say "cured" / "owned"); or `_gate_reason` says the later check would not call either (`no_specs`, `filler_only`, `not_first_pass`, `no_keypair`, `allcards_ssh`). A native error while preparing (`_PrepareFailed`) is this check's own `step=gpu_early` line and the later check builds and calls as today. Otherwise `_prepare` builds the same challenges and signed intent the later check builds, `client.verify` becomes an `asyncio.Task` and `ctx.state.local_verify = LocalVerifyOutcome(pending=PendingVerify(task, challenges))`.
- **`LocalVerifyCheck`** now reads `_pending_of(ctx)`: with a pending, unconsumed call it skips the capability read and `_prepare`, starts the rental probe (2b) as today, marks the call consumed right before it `await`s the task instead of `client.verify` (anything raising before that — the probe's request — leaves the task to `run`'s settle, which cancels it); the judge, the caps, the `LocalVerifyUnavailable` fallback and the outcome line are unchanged. `early_lead_ms` (how long the call had been running when the check reached it) and `early_wait_ms` (how long the check still waited) go on the outcome line and the `consumed step=gpu_early` metric — the two numbers that check §5's projection. When `_run` returns before reaching the call (a gate that reads state the start check could not) the check settles the pending call itself (`fallback step=gpu_early reason=unconsumed_cancelled|unconsumed_done`).
- **Settle.** `_settle_background_work` (pipeline.py) handles `pending` and `health_check_probe` (#1346's name for the 2b rental probe) with one loop: cancel if running, retrieve, `close()` the matmul challenge, one `[local_verify] outcome` line with `step=gpu_early` / `health_check_probe`. A fatal check between the start and the judge (port check, image check, tenant check) therefore cancels the task; the executor stops at its own `deadline_s`.
- **One call per cycle either way.** The executor runs one intent at a time (`BusyError`); the facts call has returned before the start check runs, and no other `/verify` is made this cycle while a call is pending (`already_pending`).
- **Facts deadline.** `facts_deadline_s()` in `checks/local_facts.py` is #1346's (timeout − 30, clamped to [5, 20]); this PR adds no line to that file and its 3 facts-deadline tests moved there with it.
- **Prestart lows from 2c.** (a) A rental landing between the facts call and the probe force-removes the container (`wait_for_port_check_containers`); the verifier's connect loop used to run the whole 30 s ready timeout on refused connects. Now, after a *failed* attempt only, it asks `docker inspect <name> --format {{.State.Running}}` (`DockerCommand.inspect_running`, `2>/dev/null` so the exit status is the answer); a non-running/absent container raises `DindContainerGone` (the read itself is bounded by `DIND_LIVENESS_READ_TIMEOUT_SECONDS` = 5 s) → logged as a warning, `docker rm -fv`, `success=False, log_text="dind: prestart gone"`. The rerun of today's `docker run` probe on the freed port is #1346's (`orchestrator.py`, the `not dind_result.success` branch); this PR's orchestrator hunks are the batch size and the `selected` list in (b). An error in the inspect itself keeps polling (an unreadable host is not "gone"). Never on the path of a container that answers. (b) `selected_ports` used to be the batch without the prestarted port while `successful_ports` / `dind_port` included it; now the batch asks for `BATCH_PORT_VERIFICATION_SIZE − 1` and `selected = batch + prestarted port`, so a host is probed on the same number of ports as without a prestart and every successful/failed port is a selected one. A prestarted port that did not answer stays in `selected_ports` and lands in `successful_ports` or `failed_ports` by the rerun's verdict on that same port (`test_a_failed_prestart_is_reprobed_on_its_freed_port_which_stays_in_the_count`).

### Not changed
The SSH-proven rows of `LIUMD_PHASE2.md` (sysbox proof, connect-back, filler liveness, stale removal, inspector) are untouched. The executor is untouched. A scored cycle, a rented node, a wedge/held-VRAM snapshot, a node without the facts flag: exactly today's path. A rental landing *during* the call is today's exposure (the later call has the same window), moved ≈ 10 s earlier; the port check's `rented_data` refresh and the tenant check halt it as they do today, and the halt cancels the task.

### Self-review

Verdict `findings` at `95b45c8` · 15 checks · by subagent-drain-r227-sr1347 · 2026-09-15 15:54Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | STALE-BODY | `neurons/validators/tests/test_local_verify_early_start.py:300` | The Parts line says 'tests ✓ 37 items (28 functions, one 10-case parametrize)' but 95b45c8 added the `no_tunnel_port` row, so the parametrize has 11 cases and the file collects 38 items (`rg -c '^(async )?def test_'` gives 28; the list at :294-:308 has 11 tuples). | Change the Parts line to '38 items (28 functions, one 11-case parametrize)' when recording this verdict. |
| low | STALE-BODY | `neurons/validators/src/services/task/checks/local_verify.py:698` | The body's list of reasons the start check does not start (`no_facts`, `facts_unanswered`, not advertised, scored, `rented`, `gpu_reread_pending`, the `_gate_reason` labels) omits `no_tunnel_port`, which 95b45c8 made reachable (tested before `step_statuses`) and covered with a test row. | Add '`/version` named no loopback port (`no_tunnel_port`)' to the reason list in the body's `LocalVerifyStartCheck` bullet. |
| low | CODE-QUALITY | `neurons/validators/tests/test_local_verify_early_start.py:7` | The module docstring still says 'Also here: the derived facts deadline', but no test in the file exercises `facts_deadline_s` (`rg facts_deadline` on the file returns nothing; those tests moved with #1346) — the same nit the 158bfe4 review recorded, unchanged in 95b45c8. | Delete the 'the derived facts deadline,' clause from the docstring. |
| low | CODE-QUALITY | `neurons/validators/tests/test_local_verify_early_start.py:523` | The docstring of `test_the_settle_step_handles_both_kinds_of_in_flight_work` says 'one outcome line each (`gpu_early`, `rental_probe`)' while the assertion expects the step label `health_check_probe` (#1346's rename) — the same nit the 158bfe4 review recorded, unchanged in 95b45c8. | Change `rental_probe` to `health_check_probe` in that docstring. |

Mechanical notes dismissed: `neurons/validators/src/services/executor_connectivity/dind_probe.py:268` `# noqa: BLE001 — an unreadable host is not "gone"; keep polling` carries its reason after the suppression and the broad catch is the documented property (a failed liveness read must not count as gone); it predates 95b45c8 (in 158bfe4). · `neurons/validators/src/services/task/checks/local_verify.py:686` `# noqa: BLE001 — the pipeline has no guard; a bug here must not end the node's cycle` carries its reason and mirrors the identical guard already in `LocalVerifyCheck.run`; it predates 95b45c8 (in 158bfe4). · `neurons/validators/src/services/task/pipeline.py:406` `# noqa: BLE001 — cleanup must not replace the pipeline's own result` carries its reason and is the pre-existing `_cancel_unconsumed_probe` line moved into the settle loop (it was on the base as well); it predates 95b45c8. · `neurons/validators/src/services/task/pipeline.py:431` `# noqa: BLE001 — same: the pipeline's own result stands` carries its reason and is the pre-existing force-remove guard moved into the `health_check_probe` branch of the settle loop; it predates 95b45c8. · `neurons/validators/src/services/task/checks/local_verify.py:128` taiberium NIT: `MatmulChallenge | None` / `VerifyXChallenge | None` on the three dataclasses is safe at runtime — the module has `from __future__ import annotations`, so the annotations are strings; `dataclasses` only pattern-matches string annotations for `ClassVar`/`InitVar` and never evaluates them, and `rg` shows no `isinstance` or other runtime reference to either name outside annotations (:128-:168, :526-:527). · `neurons/validators/tests/test_local_verify_early_start.py:324` jam_bot ordering defect: the `no_tunnel_port` row builds `LocalFacts(capabilities={CAPABILITY}, step_statuses={})` with `local_verify_port=None` and bypasses `with_tunnel_port`, so on the old order (`if not facts.step_statuses` first) the check returns `facts_unanswered` and the `what_we_saw["reason"] == "no_tunnel_port"` assertion fails; on the new order it passes, and the `facts_unanswered` row still gets a port from `with_tunnel_port`, so it keeps hitting its own branch — the row fails without the fix and passes with it. · `neurons/validators/src/core/docker_utils.py:49` jam_bot quoting defect: after 95b45c8 the only call site that quoted at the call site (`rental_probe._remove_over_shell`) passes the name bare; the other callers (`dind_probe.py` x4, `container_runner.cleanup`, `container_cleanup._remove_container`) never quoted, so every name is quoted exactly once; names are `container_<hotkey>_<port>`, `pod_<uuid4>` or `docker ps` names (`[a-zA-Z0-9][a-zA-Z0-9_.-]+`), all inside `shlex.quote`'s safe set, so the exact-string assertions (`test_local_verify_dind.py:458,473,502,516`, `test_local_verify_early_start.py:736`, `test_rental_probe_check.py:782,810,1074,1225`) are unchanged; `shlex` stays imported and used in `rental_probe.py:548`.

### Verified

Gate: PASS (95b45c8) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1347)

Rebase onto #1346 (now 242a740; the first rebase was onto its earlier head) reconciled by intent: main's `_matmul_ssh_reason` (`allcards_ssh` / `scored_ssh`, #1340) is read by phase 3's `_gate_reason`, by `_prepare` (no matmul challenge when the reason is set, for the start check and the late call alike) and by `_judge(..., matmul_ssh_reason=)`; main's `asyncio.wait_for(..., UNCONSUMED_PROBE_CLEANUP_TIMEOUT_S)` sits in the settle loop's `health_check_probe` branch. #1346 renamed `rental_probe` to `health_check_probe`, writes the two-key settle log (`health_checks_removed`, `cleanup_error`) and owns `facts_deadline_s()`; this PR takes all three as they are. `LOCAL_VERIFY_OUTCOME_EVENT = RENTAL_PROBE_OUTCOME_EVENT` was on both sides.

Cross-PR: #1340 (merged 15 Sep) added `_NothingToSend`; this PR replaces it with `_PrepareFailed`.

### Ran it

`tools/aws_verify.sh lium-io 95b45c8 --app io --bundle` → EC2 20260915T154828Z-tqfh at 95b45c8 (this head, on #1346 242a740): validators 2,689 passed, 15 skipped (+ the 2 root-sensitive tests as loopverify: 2 passed), executor 310 passed, miners 43 passed; c6i.2xlarge warm box, 3 min 54 s wall. The review-round commit's `no_tunnel_port` row of `test_nothing_starts_early_when_the_check_itself_would_not_call_or_the_gpu_is_not_ours` fails on the previous head's ordering with `facts_unanswered` (the steps test came first). `test_an_early_call_without_the_matmul_is_judged_with_the_reason_it_stayed_on_ssh` is in that count; it fails with `{'matmul': 'not_asked'}` when the judge is given no reason (checked in the worktree with the kwarg forced to None). No real node.

</details>


neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -139,7 +225,19 @@ async def run(self, ctx: Context) -> CheckResult:
                 elif not probe.task.cancelled():
                     # nothing to cancel; read the exception so asyncio does not report it unretrieved
                     probe.task.exception()
+            if pending is not None and not pending.consumed:
+                pending.consumed = True
+                pending.task.cancel()
```
taiberium, line 230: NIT: a task that already failed keeps its exception unread here, unlike the probe loop above. Add the same `elif not pending.task.cancelled(): pending.task.exception()` branch.

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -88,13 +88,98 @@ def _over_time(name: str, answer: LocalVerifyAnswer) -> bool:
...
+    reason: str
+    detail: str
+
+
+def _pending_of(ctx: Context) -> PendingVerify | None:
+    pending = getattr(ctx.state.local_verify, "pending", None)
+    return pending if pending is not None and not pending.consumed else None
+
+
+def _gate_reason(ctx: Context) -> tuple[str, str, str] | None:
```
taiberium, line 146: NIT: callers read `gate[1]` and `gate[2]` from this tuple. Return a frozen dataclass with `kind`, `reason` and `detail`, like `_PrepareFailed`.

neurons/validators/src/services/task/checks/gpu_usage.py
```diff
@@ -536,6 +536,23 @@ def _held_vram_without_owner(
     ]
 
 
+def reads_the_live_card(
```
taiberium, line 539: NIT: this name does not say what it decides. Rename it to `gpu_usage_check_will_read_the_live_card`, and the caller to match.

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -340,11 +444,7 @@ async def _call_and_judge(
         )
 
     def _judge(
-        self,
-        ctx: Context,
-        answer: LocalVerifyAnswer,
-        matmul_challenge: MatmulChallenge | None,
-        verifyx_challenge: VerifyXChallenge | None,
+        self, ctx: Context, answer: LocalVerifyAnswer, matmul_challenge, verifyx_challenge
```
taiberium, line 447: NIT: `_judge` lost the challenge types when the imports went. Import `MatmulChallenge` and `VerifyXChallenge` under `TYPE_CHECKING` and put the annotations back.

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -305,21 +327,89 @@ async def _call_and_judge(
...
             deadline_s=executor_deadline_s(settings.LOCAL_VERIFY_TIMEOUT_SECONDS),
         )
+        return _Prepared(matmul_challenge, verifyx_challenge, intent)
 
-        # Started right before the call so it overlaps the executor's GPU steps; every return
-        # from here on carries it in `ctx.state.local_verify` for RentalVerificationCheck.
-        probe = self._start_rental_probe(ctx)
-        if probe is not None:
-            started.append(probe)
+    async def _run(self, ctx: Context, started: list[RentalProbe]) -> CheckResult:
```
taiberium, line 332: NIT: `_run` is 112 lines with two phases. Move the code that finds the call into its own method.

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -305,21 +327,89 @@ async def _call_and_judge(
...
+            verifyx_challenge = prepared.verifyx_challenge
+        else:
+            # Phase 3: the call is already in flight since LocalVerifyStartCheck.
+            matmul_challenge = pending.matmul_challenge
+            verifyx_challenge = pending.verifyx_challenge
 
+        early_lead_ms = (
+            int((time.perf_counter() - pending.started_at) * 1000) if pending is not None else None
+        )
+        waited = time.perf_counter()
```
taiberium, line 380: NIT: `waited` holds a start time, not a duration. Rename it to `wait_started_at`.

taiberium (approved):

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/task/checks/gpu_usage.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `3a70853`: the settle branch now cancels a running task and reads the exception of a finished one, the same shape as the probe loop.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `3a70853`: `_gate_reason` returns a frozen `_Gate` with `kind`, `reason` and `detail`; both callers read the fields by name.

neurons/validators/src/services/task/checks/gpu_usage.py
surcyf123: Done in `3a70853`: renamed to `gpu_usage_check_will_read_the_live_card`; the wrapper in `local_verify.py` carries the same name with an underscore.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `3a70853`: `MatmulChallenge` and `VerifyXChallenge` are imported under `TYPE_CHECKING` and back on the `_judge` signature.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `3a70853`: renamed to `wait_started_at`.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `3a70853`: the code that finds the call lives in `_find_call`. It returns a frozen `_Call`, or the `CheckResult` that ends the check. `_run` keeps the wait, the judge and the metrics.

surcyf123:
@jam6099 Rebased onto #1346 at 54faee6 (carries the merged main, #1340 and #1361). Reconciled by hand: main's `_matmul_ssh_reason` (`allcards_ssh` / `scored_ssh`) is now read by phase 3's `_gate_reason`, `_prepare` and `_judge`, and main's 15 s `wait_for` on the unconsumed probe cleanup sits in the settle loop's `rental_probe` branch; one test added for the early call without a matmul. Validators suite green on EC2 at the rebased head. No re-request needed.

taiberium:
**Necessity:** I do not think we should build this one.

The PR states the gain itself: first-pass validation goes from about 112 s to about 100 s. Validation is not on the renter's clock, and a first pass is rare. Twelve seconds costs 4,400 lines and a new background task inside the validation pipeline.

Please close it. #1348 sits on this branch, but the two endpoints are independent, so it only needs a rebase onto main.

taiberium (changes requested):
Requested changes.

surcyf123:
@taiberium Answered on the issue thread with the measured numbers. Add node to listed is p50 20 min and p90 44 min for about 11 new executors a day (backend_prod, 14 days). So the 12 s is 1 percent of that path. The PR's own patch is +1,418 / -175 in 15 files, of which +841 is one test file; the 4,400 was a stale base. Rebased onto #1346 as `b7ce10b` so the diff reads. No fix is pushed and this is not a re-request. The decision is with Fish and Muhammad under DECISION-LIUMD-STACK-VALUE and your requested-changes verdict stands until then.

surcyf123:
@taiberium Facts only, no verdict from our side. The decision on the stack is with Fish and Muhammad under DECISION-LIUMD-STACK-VALUE with the default on 17 Sep. We push no code on the stack until it is taken, only rebases. This PR is not re-requested.

What first-pass time buys on the platform, measured. Source: backend_prod, 14 days to 15 Sep 12:20Z. The path is provider adds the node, miner_executor.created_at, to listed, executor.created_at. That is the backend row the validator writes on the first pass, and verified_date equals it. 215 registrations, 15 a day. 151 got a backend row after the registration, about 11 a day; the other 58 are re-registrations whose backend row predates them. Add node to listed: p50 1217 s, 20 min, and p90 2645 s, 44 min. First passed executor_validation_report to listed is 0 s, the same event. So the 12 s this PR projects on the first pass is 1 percent of the p50 path. That path is set by the validation cycle cadence, and it serves about 11 new executors a day.

The 4,400 lines. GitHub showed +4,397 / -230 because the base moved: #1346 was rewritten to 7a2b597 with its review rounds and the compare included the stack's shift. The PR's own patch is +1,418 / -175 in 15 files. One new test file, +841. Production Python +572 / -174 across 12 files: local_verify.py +392 / -130, dind_probe.py +52 / -1, pipeline.py +43 / -38, the other nine 18 lines or fewer each. .env.template +4. No Rust, no generated code. The background task is one asyncio task per first pass, settled by the same settle step that already settles the 2b probe and the 2c DinD container.

Rebased onto #1346's head 7a2b597 as one commit, `b7ce10b`, so the diff reads. GitHub now shows +1,371 / -176 in 14 files. Reconciled by hand with #1346's later rounds: the probe is health_check_probe there and the settle step keeps #1346's two-key cleanup log. The facts deadline is #1346's facts_deadline_s, so the three facts-deadline tests that rode here with the earlier 2c fix went with it. EC2 20260915T123230Z-1z7x at `b7ce10b`: validators 2,688 passed plus 2 as loopverify, executor 310, miners 43. The self-review is re-run at this head because the patch changed.

jam6099 (changes requested):

- Medium: the `gpu_early` outcome metric reads `consumed/ok` whenever a pending call returns, even when the judge consumed neither GPU step and the run fell back to SSH. It is the line Loki counts for the hit rate this feature exists to measure, so the reported overlap is inflated and contradicts the per-step metrics.
- Medium: no test pins cancellation of the phase-3 pending. The check cancels an early call only while it is unconsumed, and `_run` marks it consumed just before awaiting, so a `wait_for` timeout depends entirely on asyncio propagating the outer cancel. A later shield/wrap of that task would orphan the tunnel and the executor's GPU work with the suite still green.
- Low: the `no_tunnel_port` skip reason is unreachable. The facts state that carries no port also carries no step statuses, so `facts_unanswered` fires first and a capability/port mismatch is reported as a facts POST that never came back.
- Low: on the new gone path the DinD liveness inspect runs outside the ready-timeout budget the docstring promises, so a probe can finish up to 5 s past `DIND_SSH_READY_TIMEOUT_SECONDS`.
- Low: the new gone-path `docker rm -fv` leaves the container name unquoted, unlike the inspect added beside it. Not reachable today, but it is a new sink call site on a root SSH command.

_- jam_bot_

surcyf123:
@jam6099 Checked all five rows against b7ce10b and against the patch as it stood before the rebase (05e8c78 on the old #1346 head). None of them is a reconciliation slip: the rebase carried only the health_check_probe rename, the two-key settle log and the facts deadline that now lives in #1346. Each of the five points was already in the patch before the rebase, so each is a design point of this PR. Not changing them now because the stack is on hold. Waiting on DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build; numbers posted 12:4xZ). No design change until it is taken. Per row, what I verified on the head:

1. Medium, the `gpu_early` outcome line. Holds: the line says consumed ok whenever the early call returns, even when the judge sent both GPU steps back to SSH. Design; waits on the decision.
2. Medium, no test for a cancel that lands while the consumed pending call is awaited. Holds: the four settle tests cover the unconsumed paths and an error before the await only. Design; waits on the decision.
3. Low, `no_tunnel_port` unreachable in the start check. Holds: a facts state without a port has no step statuses, so `facts_unanswered` fires first. Design; waits on the decision.
4. Low, the liveness inspect runs after the ready deadline check. Holds: one inspect of up to 5 s can land after the 30 s deadline. Design; waits on the decision.
5. Low, the gone path removes by the unquoted name. That is main's `remove_with_volumes` helper, used as it is by the three other call sites in the same file. Quoting it is a change to the shared helper, so it waits on the same decision.

Nothing pushed; the head stays b7ce10b. If the decision is to build, all five are fixed in one round.

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -356,6 +497,20 @@ async def _call_and_judge(
             "deadline_hit": answer.deadline_hit,
             "health_check_probe_started": probe is not None,
         }
+        if pending is not None:
+            # How much of the call ran under the checks before this one (the overlap won) and how
+            # long this check still had to wait — the two numbers that check §5's projection.
+            what["early_lead_ms"] = early_lead_ms
+            what["early_wait_ms"] = int((time.perf_counter() - wait_started_at) * 1000)
+            self._metric(
```
jam6099, line 506: metric mislabel + fix (branch on `what["consumed"]`, else `fallback` with a reason from `outcome.fallbacks`).

_- jam_bot_

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -162,7 +276,23 @@ async def run(self, ctx: Context) -> CheckResult:
                 elif not probe.task.cancelled():
                     # nothing to cancel; read the exception so asyncio does not report it unretrieved
                     probe.task.exception()
+            if pending is not None and not pending.consumed:
```
jam6099, line 280: cancellation test gap + fix (mirror `test_cancelling_the_check_mid_call_cancels_the_probe_it_started`).

_- jam_bot_

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -497,3 +652,110 @@ def _metric(
...
+        if capabilities is None:
+            return self._not_started(ctx, "no_facts", "no facts call this cycle; the check decides later")
+        if CAPABILITY not in capabilities:
+            return self._not_started(ctx, "not_advertised", "executor does not advertise local verification")
+        if not facts.step_statuses:
+            # `/version` answered but the facts POST did not (busy, timeout, refused): the executor
+            # is not taking intents right now; the later check asks again where it does today.
+            return self._not_started(ctx, "facts_unanswered", "the facts call did not come back; the check decides later")
+        if facts.local_verify_port is None:
+            return self._not_started(ctx, "no_tunnel_port", "/version names no local_verify_port; nothing to tunnel to")
```
jam6099: unreachable `no_tunnel_port`.

_- jam_bot_

neurons/validators/src/services/executor_connectivity/dind_probe.py
```diff
@@ -185,6 +224,10 @@ async def _connect_retrying_until_sshd_answers(
                 )
             except Exception as error:
                 last_error = error
+                if still_running is not None and await self._is_gone(still_running, log_ctx):
```
jam6099, line 227: budget the liveness read against `deadline - loop.time()` via `asyncio.wait_for(..., min(5, remaining))`, or re-check the deadline after it.

_- jam_bot_

neurons/validators/src/services/executor_connectivity/dind_probe.py
```diff
@@ -131,6 +153,17 @@ async def verify(
                 port=port,
             )
 
+        except DindContainerGone as e:
+            # Not the host's failure: the container the validator was coming for was removed
+            # under it (a rental landing in the window). The orchestrator runs today's probe next.
+            logger.warning(_m("DinD prestart gone", extra=get_extra_info({**log_ctx, "error": str(e)})))
+            await ssh_client.run(DockerCommand.remove_with_volumes(name))
```
jam6099, line 160: `shlex.quote(name)` inside `remove_with_volumes`, as `rental_probe.py:1048` already does at its call site.

_- jam_bot_

neurons/validators/src/services/task/checks/local_verify.py
```diff
@@ -110,13 +115,122 @@ def _matmul_ssh_reason(ctx: Context) -> str | None:
...
+@dataclass
+class PendingVerify:
+    """Phase 3: the GPU `/verify` in flight, started by `LocalVerifyStartCheck` right after the facts
+    call so the executor's GPU steps overlap the SSH-driven checks in between. The challenges are
+    the ones the intent was built from — `LocalVerifyCheck` judges the answer against them exactly
+    as it judges its own call's. `consumed` tells `Pipeline.run`'s settle step that nothing is left
+    to cancel; `close()` releases the matmul challenge's native handle on every path."""
+
+    task: asyncio.Task | None  # None only while `LocalVerifyStartCheck._start` is creating it
+    matmul_challenge: Any | None
```
taiberium: NIT: `MatmulChallenge` and `VerifyXChallenge` are already imported under TYPE_CHECKING above. Use them here instead of `Any`.

taiberium (approved):

taiberium (dismissed):
Requested changes.

taiberium:
**Critical:** this branch conflicts with `liumd-phase2-c`. Rebase it, then I re-read.

surcyf123:
@taiberium Head is `95b45c8`, on #1346 `242a740`, mergeable. Your NIT and two of jam_bot's rows are fixed in that commit; the other three rows stay on the DECISION-LIUMD-STACK-VALUE line (Rustam 15 Sep: do not build).

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `95b45c8`: the port test now comes before the steps test. Without a port `LocalFactsCheck` never made the POST, so `step_statuses` was empty and the reason always read `facts_unanswered`. A `no_tunnel_port` row is added to the parametrized start-check test; it fails on the old order.

neurons/validators/src/services/executor_connectivity/dind_probe.py
surcyf123: Done in `95b45c8`: `DockerCommand.remove_with_volumes` quotes the name itself. `rental_probe._remove_over_shell` now passes the name bare, so it is quoted once. The other five callers never quoted at the call site.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Not changing this now. It is a design point of a PR on hold under DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build). The only code moves in `95b45c8` are the two plain defects from this review (`no_tunnel_port` order, `shlex.quote`).

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Not changing this now. It is a design point of a PR on hold under DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build). The only code moves in `95b45c8` are the two plain defects from this review (`no_tunnel_port` order, `shlex.quote`).

neurons/validators/src/services/executor_connectivity/dind_probe.py
surcyf123: Not changing this now. It is a design point of a PR on hold under DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build). The only code moves in `95b45c8` are the two plain defects from this review (`no_tunnel_port` order, `shlex.quote`).

surcyf123:
@taiberium Rebased. The branch has sat on #1346's head `242a740` since 14:3xZ and GitHub shows it mergeable. This round adds `95b45c8` on top with your NIT and jam_bot's two plain defects (the unreachable no_tunnel_port branch, quoting in remove_with_volumes). Suite green on EC2 at that head: validators 2,689 + 2, executor 310, miners 43. #1348 is restacked on it at `cb9bfa1`.

neurons/validators/src/services/task/checks/local_verify.py
surcyf123: Done in `95b45c8`: the three dataclasses (PendingVerify, _Prepared, _Call) now carry `MatmulChallenge | None` and `VerifyXChallenge | None`. The module has `from __future__ import annotations`, so the TYPE_CHECKING imports are enough at runtime.

surcyf123:
Closed on Muhammad's decision (DECISION-LIUMD-STACK-VALUE, 16 Sep 04:52Z, #lium-agents): 12 s saved on a path no customer waits on; Rustam asked for the same on 15 Sep. The branch stays; nothing here merges. Whatever comes back from the liumd work goes on main, one PR at a time.

PR #1343, opened 2026-09-09
DAH-2834 - [P2] liumd 2d: one-call /verify on the first pass only

surcyf123:
Implements [DAH-2834](https://app.notion.com/p/liumd-v1-Rust-CLI-on-executor-answering-validator-commands-specs-verifyx-3cfb8bfdbde98140b3aefa0c2d657720) · reviewer: @jam6099

**Problem** — `speed/LIUMD_PHASE2.md` item (d). Liumd 2d: one-call /verify on the first pass only.

**Value** — not measured — the body has no Worth it line (SO §69) and the ticket gives no prod number — a human decides.

**Goal** — [DAH-2834](https://app.notion.com/p/liumd-v1-Rust-CLI-on-executor-answering-validator-commands-specs-verifyx-3cfb8bfdbde98140b3aefa0c2d657720) (P2).

**Fits** — Sits in `lium-io` — `neurons/validators`; touches backend, tests.

**Flag:** no flag

**Risks**
- Changes runtime configuration (`config.py`) on the next deploy, no pod restart — noticed only if a value is wrong — rollback: revert the file and redeploy.

**How to verify**
- `cd neurons/validators && python -m pytest tests/test_local_verify.py tests/test_first_pass_fast_path.py -q` → the gate, the express-lane pass, the 4-case context test

**Verified by the loop:** Gate: PASS 8bc5f21

**Ran it:** `tools/aws_verify.sh lium-io 8bc5f21 --app io` → EC2 20260915T140808Z-oyn0 at this head: validators 2,572 passed, 15 skipped, plus 2 as loopverify; not on a real node

**Self-review:** clean at 8bc5f21 (15 checks, 1 low)

**Parts:** backend ✓ · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — validator-internal · docs n/a — flag semantics in .env.template · tests ✓ 3 new + the 4-case context test re-pinned · dashboards n/a — existing `[local_verify] outcome` line

**Docs:** none needed because `LOCAL_VERIFY_FIRST_PASS_ONLY` is operator-only validator config (`neurons/validators/.env.template`), not a provider-facing surface; a scored cycle keeps today's SSH path.

<details><summary>Details</summary>

### What changed

- `LocalVerifyCheck._run` (`checks/local_verify.py`): `if settings.LOCAL_VERIFY_FIRST_PASS_ONLY and not ctx.config.unscored` → the existing fallback, reason `not_first_pass`
[...]

PR #1337, opened 2026-09-09
DAH-3265 - [P2] warm container pool: a created-never-started slot per pre-pulled image

surcyf123:
Implements [DAH-3265](https://app.notion.com/p/3d6b8bfdbde9810e8aedce32d403b490) · reviewer: @taiberium

**Problem** — Warm container pool, flag off. Behind `WARM_POOL_ENABLED` a filler start leaves one created-never-started `warm_<uuid>` container per pre-pulled image; a whole-host rental of that image adopts it — volume read from the plugin, live `docker inspect` checked against the spec the rental would create now, then rename → update → start instead of volume creation + `docker run`.

**Value** — measured on prod (tsdb `container_profiler_events`, 14 d to 15 Sep 11:25Z, 5079 rents): volume stage p50 2623 ms / p90 6750 ms, docker run p50 1763 ms; volume + run p50 4.6 s = 21 % of the mean rent (27.4 s). The volume span holds sizing and create together and a hit still pays the sizing..

**Goal** — [DAH-3265](https://app.notion.com/p/3d6b8bfdbde9810e8aedce32d403b490) (P2) · serves SO §69.

**Fits** — Sits in `lium-io` — `neurons/validators`; touches validator, docs (`.env.template`), tests (199).

**Flag:** `WARM_POOL_ENABLED`, off by default, turned on by @taiberium on 2026-09-30 (proposed by the loop; change it in one reply and the body follows)

**Risks**
- Changes runtime configuration (`config.py`) on the next deploy, no pod restart — noticed only if a value is wrong — rollback: revert the file and redeploy.

**How to verify**
- `cd neurons/validators && pytest tests/test_warm_pool.py tests/test_volume_fast_path.py tests/test_container_cleanup.py -q` → 199 passed
- flag off: no slot lookup, rename or maintenance; the slot-volume listing still runs; a `warm_*` past `WARM_POOL_MAX_AGE_HOURS` is swept with its volume
- flag on, staging: filler start → `warm_pool slot=create`; whole-host rent → `adopt=hit`

**Verified by the loop:** sandbox EC2 · Gate: PASS e99a1ee

**Ran it:** real dockerd 27.3.1… · at e99a1ee (main `219c4f1` merged in after #1391; EC2 `20260917T154227Z-gdki`): validators 2798 passed, 15 skipped, + 2 root-sensitive tests as `loopverify` · … +2 under Details

[...]

PR #1348, opened 2026-09-09
DAH-2834 - [P0] executor POST /rent: rental container from one signed run spec

surcyf123:
Implements [DAH-2834](https://app.notion.com/p/3cfb8bfdbde98140b3aefa0c2d657720) · reviewer: @jam6099

**Problem** — Standalone on `main` per Muhammad's decision (16 Sep): out of the liumd stack, both flags off, measured on staging before the merge call; Rustam's bar: 5 s pays, 1–2 s does not.

**Value** — The segment `/rent` replaces costs 2.0 s p50 / 4.4 s p90 per rent on prod today (below); the PR is +1,973 / −240 production lines (≈ 1,810 test lines beside them) behind two flags that ship off. Rustam's bar (15 Sep): 5 s or more pays, 1–2 s does not. Staging AFTER (16 Sep, n=30 vs 30, same L4 node): 806 → 742 ms p50, 879 → 830 ms p90, 30/30 rents through `/rent`; staging shows the mechanism (~60 ms on a 1 ms RTT node), the prod-sized saving stays the projected ≈ 1.2–1.7 s p50.

**Goal** — [DAH-2834](https://app.notion.com/p/3cfb8bfdbde98140b3aefa0c2d657720) (P0) · serves SO §52.

**Fits** — Sits in `lium-io` — `neurons/executor`, `neurons/validators`, `datura/datura`; touches backend, docs, tests.

**Flag:** `VALIDATOR_LOCAL_RENT_ENABLED`, off by default, turned on by @jam6099 on 2026-09-18 (the day after the staging AFTER number lands, DECISION-LIUMD-STACK-VALUE; proposed by the loop; change it in one reply and the body follows)

**Risks**
- Changes runtime configuration (`config.py` in both neurons) on the next deploy, no pod restart; both flags stay off — noticed only if a value is wrong — rollback: revert and redeploy.

**How to verify**
- `cd neurons/executor && python -m pytest tests/test_local_rent.py -q` — `test_flag_off_is_404`, refusals, a network peer's 403, deadline cut + rollback
- `cd neurons/validators && python -m pytest tests/test_local_rent.py -q` — `test_the_flag_ships_off_and_nothing_is_posted`: a live fake executor gets zero requests

**Verified by the loop:** EC2 20260916T084415Z-ll6n at de253ef — green (numbers in Ran it); the round's first cut (EC2 20260916T082428Z-zlao) failed 10 tests the fresh review had already flagged…

**Ran it:** `tools/aws_verify.sh lium-io de253ef --app io-all --bundle head3.bundle` → EC2 20260916T084415Z-ll6n at de253ef: validators 2,660 + 2 loopverify, executor 374, miners 43 — all passed

**Self-review:** clean at de253ef (9 checks)

**Parts:** backend ✓ · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — internal · docs ✓ `neurons/executor/README.md` · tests ✓ · dashboards n/a — Loki line

**Docs:** none needed because both flags are off by default; operator rows in `neurons/executor/README.md`, design in `speed/DEPLOY_LOCAL_RENT.md` (lium-ads).

### Metrics

**Worth it?** The segment `/rent` replaces costs 2.0 s p50 / 4.4 s p90 per rent on prod today (below); the PR is +1,973 / −240 production lines (≈ 1,810 test lines beside them) behind two flags that ship off. Rustam's bar (15 Sep): 5 s or more pays, 1–2 s does not. The staging AFTER number is in (arm B row below: 806 → 742 ms p50 on a same-region L4, n=30 vs 30); the honest expectation from the design (`speed/DEPLOY_LOCAL_RENT.md` §2–3) is a delta of ≈ 1.2–1.7 s p50 fleet-wide (≈ 2 s on 4090/5090 nodes, ≈ 4 s at p90), so the decision rests on whether the p90 tail counts.

| | docker_run_ms | container_running_check_ms | sum (the segment) | source |
|---|---|---|---|---|
| **BEFORE, prod** | p50 1,765 / p90 3,871 | p50 259 / p90 490 | **p50 2,047 / p90 4,354 / mean 2,548 ms = 9.3 % of the mean rent (27,388 ms)** | tsdb `container_profiler_events`, 14 d to 15 Sep 11:55Z, 5,076 rents (all `is_cached_template`); one SQL over the per-rent row, read-only (`tools/db_query.sh`) |
| BEFORE by GPU, docker run p50 | RTX 4090 2,217 · RTX 5090 2,013 · RTX 3090 2,502 · B300 2,097 · H100 1,126 · H200 1,079 · RTX PRO 6000 825 · A100 769 ms | | | same query grouped by GPU type |
| **staging arm A (flag off, SDK path), 16 Sep 10:43–11:07Z** | p50 744 / p90 822 | p50 62 / p90 67 | **p50 806 / p90 879 / mean 1,388 ms** (median 95 % CI 801–811; two outliers: the first cold rent 5.4 s, one 13.3 s) | 30 scripted rents of the default PyTorch template on one loop-owned AWS g6.xlarge (L4) staging executor, read from the staging validator connector's per-rent profiler steps (`Sending back created container info` log line); staging `container_profiler_events` received no row during the run |
| **AFTER, staging arm B (both flags on, `/rent`), 16 Sep 19:45–20:08Z** | p50 742 / p90 830 (the whole signed `/rent` call, running-state wait included) | 0 (the executor already waited for `State.Running`) | **p50 742 / p90 830 / mean 876 ms** (median 95 % CI 738–762; one outlier 4,266 ms, rent 5, where the executor's own work was 714 ms) | 30 scripted rents, same node type (a fresh AWS g6.xlarge L4), same executor image digest `05efe344…`, same template, `EXECUTOR_LOCAL_RENT_ENABLED=true` on the box, staging validator + connector built from this branch (`lium-staging-validator:de253ef`, Staging Branch Deploy 18:43Z) with `VALIDATOR_LOCAL_RENT_ENABLED=true`; the connector log carries **30 `[local_rent] created` lines for the 30 rents** (`round_trip_ms` p50 732 / p90 819, `executor_ms` p50 720 / p90 774), no `fell_back`; read from the same profiler steps as arm A |

**AFTER delta on staging: 806 → 742 ms p50 (−64 ms), 879 → 830 ms p90 (−49 ms), n=30 vs n=30, same L4 node type and image; the two medians' 95 % intervals (801–811 vs 738–762) do not overlap.** What staging shows is the mechanism, not the prod number: on a same-region node (~1 ms RTT to the validator) with the template cached, the segment is the host's own `docker run` (`executor_ms` p50 720 ms; arm A's `docker_run_ms` p50 744 ms) and `/rent` can only remove the SDK round trips and the `docker ps` poll on top of it — about 60 ms here. In prod the same segment costs 2,047 ms p50 / 4,354 ms p90 (BEFORE row): the difference to staging is the far node's RTT and slower disks, which is the part `/rent` compresses to one round trip; the honest projection stays the design's ≈ 1.2–1.7 s p50 fleet-wide, under Rustam's 5 s bar. Method and per-rent rows: `RECOVERY/workers/liumd-1348-arm-b/` (`arm_B_profile.tsv`, `connector_armB.log`, `armB_stats.py`) beside arm A's `RECOVERY/workers/liumd-1348-staging/arm_A_profile.tsv` in lium-ads. One staging-only correction found on the way: the executor's `MINER_HOTKEY_SS58_ADDRESS` must be the provider hotkey the node is registered under in the portal (`/rent` binds the intent to it and refuses `intent is for another miner's executor` otherwise); the central-miner hotkey goes in `DEFAULT_MINER_HOTKEY`. Three rents made with the wrong hotkey fell back to the SDK path and are excluded (`arm_B_refused_sdk_fallback.tsv`).

### Design notes
Design, measurement and the per-stage budget: `speed/DEPLOY_LOCAL_RENT.md` (lium-ads). NEVER-replace list (LIUMD_PHASE2 §4) applies: keys exec, sshd bootstrap proof, gocryptfs mount, `rm`, the verification proofs stay SSH/SDK.

- **Gate = the flag, the spec, the host key, then `/version`.** `_create_with_local_rent` runs when `VALIDATOR_LOCAL_RENT_ENABLED`, the spec carries nothing private (`carries_only_public_fields`), every volume target is a plain absolute path the executor's wire parse accepts (`CONTAINER_PATH_PATTERN`; else `not_taken reason=volume_target_charset`, no round trip wasted) and the executor's SSH host key is known. It then reads `/version` once over the API port (`LocalVerifyClient.advertised`, connect-timeout bounded): an executor with the route on names `local_rent_port`; an old image or `EXECUTOR_LOCAL_RENT_ENABLED=false` names none → `[local_rent] not_taken reason=no_tunnel_port`, nothing posted, the SDK path. With a port, the signed intent goes through `ssh_client` (`post_signed`: `forward_local_port` to `127.0.0.1:<port>` on the executor, the listener closed after the call) to `POST /rent`, which answers loopback peers only — the same trust root as the SSH path (#1339, merged). Any non-200 is one `[local_rent] unavailable` line, then the SDK path.
- **Executor refusals** (`local_rent_service.py`, before any docker call): `refuse_spec` — no command/entrypoint, environment == the validator's own public set, volumes by docker volume name, or the one host path a rental binds — the quote broker socket `/var/run/lium-dstack/dstack.sock`, read-only only (`ALLOWED_HOST_BINDS_READ_ONLY`), runtime ∈ {daemon default, `sysbox-runc`} (what `_build_rental_container_run_spec` emits), caps ⊆ {NET_ADMIN, IPC_LOCK}, sysctls ⊆ {`net.ipv4.conf.all.src_valid_mark=1`}, ulimit names ⊆ {memlock}, devices by name (`ALLOWED_DEVICE_PATTERN`: `/dev/nvidiaN`, `nvidiactl`, `nvidia-modeset`, `nvidia-uvm[-tools]`, `nvidia-nvswitch*`, `nvidia-nvlinkN`, `nvidia-caps/nvidia-capN`, `nvidia-caps-imex-channels/channelN`, `/dev/infiniband/uverbsN` + `rdma_cm`, `/dev/net/tun`, `/dev/fuse`, `/dev/dri/cardN` + `renderDN` — a sibling such as `/dev/nvidia-foo` is refused), network == `lium-rentals` (the icc-off rental bridge — absent means docker0, which the SSH path never builds; `host`, `none`, `container:<id>` never) (the image, `device_requests`, `shm_size`, `cpu_count`, `memory_gb`, `storage_limit_gb` are not pinned to values — a renter picks the image and the sizes vary per rental — but `spec_from_wire` bounds their shape: `shm_size` is `\d+[kmg]?`, volume targets and device paths in the container are absolute paths of plain characters, no `:` or whitespace); `refuse_ports` — every published host port is on the INTERNAL side of `RENTING_PORT_MAPPINGS` / the port range (the side docker binds on the host; the external side is the renter's NAT address) and never the executor's own sshd port; `refuse_intent` — the intent's `ssh_host_key_sha256` must be this host's key, and `check_intent_target` (shared with `/verify`) its `miner_hotkey` this executor's miner; the `nonce` is lower-case hex on the wire (it is the docker label value the rollback filters by); `refuse_ready` — `ready.ssh_host_port` must be a port the spec publishes, so a signed intent cannot turn the executor into a probe of the host's other ports; the container name must start with `pod_` / `filler_` (`RENTAL_CONTAINER_NAME_PREFIXES`, shared with the validator's `get_container_name`).
- **Rollback and `rolled_back`.** Removal by container id, or by the `lium.local_rent.nonce` label when the create was cut — never by name, so the validator's fallback `docker run` of the same name is never taken. `rolled_back=true` when no create was ever issued (image absent, spec/port/ready refused, deadline before the create), when the daemon itself refused the create (an HTTP error — nothing exists), or when the removal by id or label was seen; a create that failed on the socket (no daemon answer) is looked up by label first; a create the deadline cut answers `false` and a second label pass runs 5 s later off the request.
- **The validator frees the name after any non-answer the executor may have acted on** (`may_have_acted`, `local_rent_client.py`): timeout, a connection that broke after the send, 5xx, an unreadable answer, or an answer with `rolled_back=false` → `_remove_failed_rental_container_for_retry(container_name=<pod name>)` before the SDK `run`. Also after a tunnel that closed without a status line (`refused`, detail `tunnel to …`: sshd could not open the channel OR the session broke after the send — aiohttp sees the same disconnect). Not after 404 / 401 / 403 / 409 / 422, a listener that was never bound (`tunnel: …` / `tunnel listener not bound`) or a failed connect to that listener (`ClientConnector*`): the intent never left.
- **Budget.** `LOCAL_RENT_TIMEOUT_SECONDS=45`; the executor's `deadline_s` = timeout − `EXECUTOR_ROLLBACK_MARGIN_S` (15), floor 5, counted from the top of the executor's run (the docker client open included): the answer and a by-id rollback (10 s) fit inside the client's wait; the rollback runs on its own worker thread (`_rollback_executor`), so a create the deadline cut that still holds a `_rent_executor` worker cannot queue it; `deadline_hit` is the step task's state, not the timer's, so a run whose steps had all finished as the timeout fired is kept; a by-label rollback (list + remove, up to 20 s) that runs past it is a client timeout, which frees the name on the validator (`may_have_acted`).
- **Trust.** The executor's answer is evidence, not a verdict: the keys exec, the sshd bootstrap and `_wait_for_container_exec_ready` follow over SSH exactly as today and are what prove the container reachable. The executor enforces the rental-only shape itself, so the validator hotkey alone cannot run anything but a rental on a provider's host.

### Cross-PR
What this PR no longer sits on: the liumd verification stack (lium-io#1343–#1347: facts-only `/verify`, DinD-in-intent, GPU early start, the local first-pass verifier). #1347 is closed and #1343–#1346 are parked (Muhammad, 16 Sep); nothing in `/rent` imports them — the base is `main` at f6f24bf (#1340 merged, which is where `LocalVerifyClient`, `post_signed`'s tunnel and the `/verify` signing blob live). The executor's `_capabilities()` lists `local_verify/1` and `local_rent/1` only.

main's #1357 (DAH-3394, merged) renamed the executor's `dependencies.auth.VALIDATOR_HOTKEY_SS58` to the `VALIDATOR_HOTKEYS_SS58` map; this PR's `tests/test_local_rent.py` fixture patches both names (`raising=False`) so it runs on either. lium-io#1337 (warm pool) and this PR both touch `neurons/validators/src/services/docker_service.py::create_container` and `rental_docker_sdk.py`; #1337's `warm_pool._port_bindings(host)` reads a docker inspect document and is a different function from the `rental_docker_sdk._port_bindings(ports)` this PR moves into `datura.rental_spec.build_host_config_kwargs` — a name clash only, no call crosses the two. Either may land first; the second rebases the `create_container` hunk (this PR's local-rent branch sits right before the SDK `run`, #1337's adopt branch right before the volume stage). Already on `main` and folded in: #1332 (the volume stage), #1321 / DAH-3199 (the icc-off `lium-rentals` network — `ContainerRunSpec.network`, `RENTAL_NETWORK_*` and the ensure/refuse check are datura's now and the executor's `/rent` runs the same check before its create), #1323 (the executor middleware's `SELF_AUTHENTICATED_POST_PATHS` allowlist — `/rent` joins `/verify` there).

lium-io#1370 (DAH-3475, P2, held draft) extends `rental_docker_sdk._restart_policy` — which this PR deletes into `datura.rental_spec.build_host_config_kwargs` — to map `on-failure:N` onto `MaximumRetryCount`, and sends `on-failure:5` for fillers, a value `rental_spec.RESTART_POLICIES` here refuses. This PR lands first; #1370 rebases and carries the `on-failure:N` parse and the whitelist widening into `rental_spec.py` (details in #1370's Cross-PR paragraph). Nothing in this PR changes for it.

### Known follow-ups
Muhammad's 12 review threads of 15 Sep and Rustam's codec NIT were parked on DECISION-LIUMD-STACK-VALUE; the decision is in (16 Sep 04:52Z: standalone, measure on staging) and this round's commit fixes 11 of the 12 in code (host-bind pin + read-only, `nonce` charset, `miner_hotkey` target on `/rent`, device nodes by name, deadline clock from the top of the run, the rollback's own worker, `refuse_spec`'s docstring narrowed to what it pins, `deadline_hit` from the task's state, one busy guard that never burns a nonce, wire charset on `volume.target` / `device.path_in_container` / `shm_size`, one `host_key_sha256` in `datura`) and pins the 12th (`proxy_headers=False` in `executor.py`) with an AST regression test instead of a loopback-only listener — a second uvicorn listener is a deploy change (compose ports, `/version`'s port), not this PR's. Still open: Rustam's NIT (the wire codec as a pydantic model: a size refactor with no behaviour change, ≈150 lines; a follow-up commit on this branch if he wants it before the merge) and the self-review low `ssh_timeout_s` not clamped to `deadline_s` (the deadline cuts the ready step either way). None of these changes the measured segment.

### Ran it

`tools/aws_verify.sh lium-io de253ef --app io-all --bundle head3.bundle` → EC2 20260916T084415Z-ll6n at de253ef (the review round's one commit on fa9f7cf): validators 2,660 passed, 15 skipped (+ the 2 root-sensitive tests as loopverify: 2 passed), executor 374 passed, miners 43 passed; c6i.2xlarge warm box, 3 min 46 s wall. The round's 13 new regressions (executor 10, validator 3) are in those counts; its first cut (EC2 20260916T082428Z-zlao) failed 9 validator tests (call sites without the new `miner_hotkey` kwarg) and 1 executor test (a racy `rolled_back` assertion) — both were the fresh review's two high findings, fixed in the same commit before anything was pushed. Previous head: `tools/aws_verify.sh lium-io 66623ca --app io-all --bundle head3.bundle` → EC2 20260916T054958Z-kozz at 66623ca (= fa9f7cf, an empty commit that only re-triggers CI; same tree) (the branch rebuilt on `main` f6f24bf: the same 9 commits cherry-picked, 3 conflict hunks resolved by dropping the stack's pieces — `executor/routes/apis.py` `_capabilities` no longer lists the DinD capability, `validators/.env.template` and `validators/core/config.py` carry only the three `*_LOCAL_RENT_*` keys, `datura/requests/validator_requests.py` carries the three Rent* models and not `DindStep`): validators 2,653 passed, 15 skipped (+ the 2 root-sensitive tests as loopverify: 2 passed), executor 360 passed, miners 43 passed; c6i.2xlarge warm box, 3 min 45 s wall. Regression proof for the flag-off test (SO §52): the same suite file at a throwaway commit with the validator's `if not settings.VALIDATOR_LOCAL_RENT_ENABLED: return None` removed → EC2 20260916T053748Z-8y3k: `test_the_flag_ships_off_and_nothing_is_posted` FAILED (the fake executor answered a created container), 81 passed; that commit was never pushed. Earlier, at the pre-tunnel head: RTX 4090 pod create+start 0.35 s over `/rent` (the create/start calls are unchanged since; the start goes by the id).

### Self-review

Verdict `clean` at `de253ef` · 9 checks · by subagent-r238-fresh2 · 2026-09-16 08:49Z (tools/pr_self_review.py)

Mechanical notes dismissed: `datura/datura/rental_spec.py:428` `startswith("/dev/")` is the device-path shape check in `spec_from_wire`; nothing is mounted there. · `neurons/executor/tests/test_local_rent.py:206` `VolumeMount("pod_abc_vol", "/root")` — the source is a docker volume NAME, `/root` is the target inside the container, as the SSH path mounts today. · `neurons/executor/tests/test_local_rent.py:274` Asserts docker-py's `volumes=["/root"]` (the container-side target list) for the same named volume. · `neurons/executor/tests/test_local_rent.py:392` The `curl evil | sh` command is a REFUSED parametrisation: the test asserts the executor answers `failed` with 'command' and touches no docker call. · `neurons/executor/tests/test_local_rent.py:397` The docker.sock bind is a REFUSED parametrisation ('host path'); asserted refused before any docker call. · `neurons/executor/tests/test_local_rent.py:412` `SYS_ADMIN` is a REFUSED parametrisation ('capabilities'); asserted refused. · `neurons/executor/tests/test_local_rent.py:439` Named volumes `pod_abc_vol` and `ext-vol_1` with container targets `/root` and `/mnt`, plus the one read-only broker socket — the shape the validator's SSH path emits today. · `neurons/executor/tests/test_local_rent.py:442` `NET_ADMIN` and `IPC_LOCK` are a rental's own capabilities (WireGuard / RDMA), the `ALLOWED_CAPABILITIES` set; the test asserts they pass, `SYS_ADMIN` is refused at :412.

### Verified

Gate: PASS (de253ef) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1348)

### What changed

- `datura/rental_spec.py`: `ContainerRunSpec` (+ DAH-3199 `network`), `create_and_start`, `ensure_rental_network` — one definition both ends import
- executor `POST /rent`: `/verify`'s signature, nonce, window and `miner_hotkey`; loopback peers only; bound to this host's `ssh_host_key_sha256`
- refusals before any docker call (`refuse_spec`, `refuse_ports`, `refuse_ready`); rollback by id or nonce label, never by name
- step evidence as named `datura` models (`RentImageData` / `RentContainerData` / `RentReadyData`)
- validator `_create_with_local_rent`: nothing posted while `VALIDATOR_LOCAL_RENT_ENABLED=false` (default); on, `/version` names `local_rent_port`, the intent rides `ssh_client`

</details>


neurons/executor/src/payloads/rent.py
```diff
@@ -0,0 +1,89 @@
...
+    steps: RentSteps = Field(default_factory=RentSteps)
+
+
+
+class RentIntent(RentIntentBody):
+    signature: str = Field(min_length=1, max_length=1024)
+
+
+class RentStepResult(WireModel):
+    status: str  # ok | failed | timeout | skipped
```
taiberium: NIT: the comment is the type. `/verify` already uses a Literal here.

```suggestion
    status: Literal["ok", "failed", "timeout", "skipped"]
```

neurons/executor/src/payloads/rent.py
```diff
@@ -0,0 +1,89 @@
...
+    ms: int = 0
+    data: dict[str, Any] | None = None
```
taiberium: NIT: `/verify`'s sibling model types this as a union of named step models. Here the client reads it back with `isinstance` guards.

datura/datura/rental_spec.py
```diff
@@ -0,0 +1,530 @@
...
+        volumes=container_volumes(spec.volumes) or None,
+        name=spec.name,
+        entrypoint=spec.entrypoint or None,
+        host_config=host_config,
+        **({"labels": labels} if labels else {}),
+    )
+    container_id = created.get("Id") if isinstance(created, dict) else None
+    if on_created is not None and container_id:
+        on_created(container_id)
+    api_client.start(spec.name)
```
taiberium: NIT: the rollback never goes by name, but the start does. Start `container_id` when the daemon gave one.

taiberium (approved):

taiberium (changes requested):
Merge conflict.

taiberium (changes requested):
Requesting changes: the branch cannot merge into its base.

It forked before phases 2a/2b/2c/3 landed on `liumd-phase3`. A merge of the current base now gives 39 conflict hunks in 15 files, with add/add conflicts on `neurons/validators/src/services/local_verify_client.py` (9 hunks) and `neurons/validators/tests/test_local_verify_facts.py` (6 hunks). Those files belong to the lower PRs of the stack, so a hand resolution here can revert their work.

Please restack onto the current `liumd-phase3` and run the suites again.

The code itself is fine. I compared the moved host-config build field by field: `datura/rental_spec.py::build_host_config_kwargs` matches the old `_build_host_config_kwargs`, the dataclasses and defaults are the same, and `create_and_start` issues the same docker-py calls. With both flags off the live rental path does not change.

neurons/executor/src/payloads/rent.py
surcyf123: Done in `1c720e9`: `status` is `Literal["ok", "failed", "timeout", "skipped"]`; a status outside that set now fails to build, with a test for it.

datura/datura/rental_spec.py
surcyf123: Done in `1c720e9`: `create_and_start` starts by the id the daemon answered and falls back to the name only when it gave none; tests cover both paths.

neurons/executor/src/payloads/rent.py
surcyf123: Done in `1c720e9`: each step's data is now a named datura model (image, container, ready) joined as `RentStepData`, the way `/verify` types its steps. The validator client parses them and the isinstance guards are gone. Round-trip and malformed-evidence tests cover it.

neurons/validators/src/services/local_rent_client.py
```diff
@@ -0,0 +1,273 @@
...
+    INTENT_TTL_SECONDS,
+    LocalVerifyClient,
+    LocalVerifyUnavailable,
+    StepEvidence,
+    _wire_int,
+)
+
+SCHEMA = LOCAL_RENT_SCHEMA
+CAPABILITY = LOCAL_RENT_CAPABILITY
+STEP_NAMES = ("image", "container", "ready")
```
taiberium, line 54: NIT: nothing in `neurons/validators` reads `STEP_NAMES`. Delete the constant.

neurons/validators/src/services/local_rent_client.py
```diff
@@ -0,0 +1,273 @@
...
+    @property
+    def created(self) -> bool:
+        """The container is up on the executor's word: made (the container step `ok` with its
+        evidence), running, sshd answering when asked. The validator's own execs are what confirm it."""
+        return (
+            not self.deadline_hit
+            and not self.rolled_back
+            and self.step("container").status == "ok"
+            and self.container is not None
+            and self.step("ready").status == "ok"
```
taiberium, line 192: NIT: `created` checks `self.container` but not `self.ready`, so an `ok` ready step with no data still reads as created. Add `and self.ready is not None`.

neurons/executor/src/routes/apis.py
```diff
@@ -516,6 +524,88 @@ async def local_verify(request: Request):
...
+    except ValueError:
+        raise HTTPException(status_code=422, detail="Body is not JSON")
+    if not isinstance(raw, dict):
+        raise HTTPException(status_code=422, detail="Body is not an object")
+    try:
+        intent = RentIntent.model_validate(raw)
+    except ValidationError as exc:
+        raise HTTPException(status_code=422, detail=exc.errors())
+    await verify_signature(SignaturePayload(signature=intent.signature), canonical_intent_message(raw))
+    body = RentIntentBody.model_validate({k: v for k, v in raw.items() if k != "signature"})
```
taiberium, line 580: NIT: `RentIntent` already is a `RentIntentBody`, so this second parse does nothing.

```suggestion
    body: RentIntentBody = intent
```

neurons/validators/src/services/rental_docker_sdk.py
```diff
@@ -904,102 +812,13 @@ def _create_paramiko_client(self, base_url):
...
-
-    return [
-        DeviceRequest(
-            count=device_request.count,
-            device_ids=list(device_request.device_ids) or None,
-            capabilities=[list(capability) for capability in device_request.capabilities],
-        )
-        for device_request in device_requests
-    ]
+    return build_host_config_kwargs(ContainerRunSpec(image="-", name="-", volumes=volumes))["binds"]
```
taiberium, line 821: NIT: only `test_cvm_quote_broker` calls `_binds` now. Delete it and let that test read `build_host_config_kwargs(spec)["binds"]`.

neurons/executor/src/services/local_rent_service.py
```diff
@@ -0,0 +1,664 @@
...
+        """Every published host port must be one of this executor's rental ports and never its own
+        sshd port — the intent chooses among what the executor already offers."""
+        offered = self.offered_ports()
+        for binding in spec.ports:
+            if binding.host_port == self.ssh_port or binding.host_port not in offered:
+                return f"host port {binding.host_port} is not one of this executor's rental ports"
+        return None
+
+
+class _Run:
```
taiberium, line 359: NIT: `_Run` does not say what the class holds. Rename it, for example `_RentAttempt`.

taiberium (approved):

neurons/validators/src/services/local_rent_client.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/local_rent_client.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/executor/src/routes/apis.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/rental_docker_sdk.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/executor/src/services/local_rent_service.py
surcyf123: Agreed. Fixing it now, one commit on this branch within the hour.

neurons/validators/src/services/local_rent_client.py
surcyf123: Done in `7f3ed33`: `STEP_NAMES` is gone from the validator client.

neurons/validators/src/services/local_rent_client.py
surcyf123: Done in `7f3ed33`: `created` now also needs `self.ready is not None`; a new test row covers an `ok` ready step with no data.

neurons/executor/src/routes/apis.py
surcyf123: Done in `7f3ed33`: `body` is the validated intent itself; the second parse is gone.

neurons/validators/src/services/rental_docker_sdk.py
surcyf123: Done in `7f3ed33`: `_binds` is deleted and the test reads `build_host_config_kwargs(run_spec)["binds"]`.

neurons/executor/src/services/local_rent_service.py
surcyf123: Done in `7f3ed33`: renamed to `_RentAttempt`.

surcyf123:
@jam6099 Rebased onto the new #1347 head 05e8c78 (which itself was rebased onto the merged main). Rebase only: the patch is unchanged (same patch-id), nothing was reconciled here. Same review, no re-request needed.

taiberium:
**Necessity:** I cannot weigh this one yet.

The PR says what it replaces (3 to 5 SDK calls over SSH plus a `docker ps` poll) but not what that costs today. A rent is 22.9 s p50; #1295 measured the SSH connect at 2.1 s of it. Please give the same kind of number for the calls this endpoint removes.

The price is high: a new HTTP entry point on the provider's host that creates the rental container. At 5 s or more it pays. At 1 to 2 s it does not.

surcyf123:
@taiberium Measured: the segment this endpoint replaces is 2.0 s p50 and 4.4 s p90 of a rent today.

Source: tsdb `container_profiler_events`, 14 days to 15 Sep 11:55Z, 5076 rents, all cached templates. The two profiler steps that cover the SDK create and start over the tunnel and the `docker ps` poll are `docker_run_ms` and `container_running_check_ms`. Docker run p50 1765 ms and p90 3871 ms. Running check p50 259 ms and p90 490 ms. Both together p50 2047 ms, p90 4354 ms, mean 2548 ms, which is 9.3 percent of the mean rent of 27.4 s (rent p50 23.8 s, p90 54.8 s).

By fleet, docker run alone: RTX 4090 p50 2217 ms, RTX 5090 2013 ms, RTX 3090 2502 ms, B300 2097 ms. H100 1126 ms, H200 1079 ms, RTX PRO 6000 825 ms, A100 769 ms. The table has no region column, so the GPU type stands in for far and near.

What stays after the change, from the PR's own bench: about 0.35 s of host work (create 47 ms plus start 250 ms). Plus one validator to executor round trip, 0.2 to 0.6 s to the far nodes. So the net the call can remove is about 1.2 to 1.5 s p50 on the fleet average and about 2 s p50 on the 4090 and 5090 nodes. The `docker image inspect` extra step is already skipped for cached images and is not counted.

Method: one SQL over the per-rent profiler row, percentiles of the two step columns and of their sum, share of the mean total. The decision on the stack is with Fish and Muhammad under DECISION-LIUMD-STACK-VALUE. We push no code on this branch until it is taken, only rebases.

taiberium (approved):

taiberium (changes requested):
Merge conflict.

surcyf123:
@taiberium The conflict comes from the #1347 rebase at 13:15Z: this branch still sits on #1347's old head 05e8c78 while #1347 now is b7ce10b on #1346. It is a base move, not a code change here. The restack onto b7ce10b is queued for the next lane because the loop account hit the GitHub GraphQL limit at 13:31Z, which blocks pr_restack for about an hour. Rebase only, patch unchanged, no re-request; the necessity number you asked for is in the thread above.

surcyf123:
@taiberium Restacked onto #1347 b7ce10b at 3dbcb31. The rebase was clean and the patch is unchanged (same patch-id as 5416654), so the PR merges again. The lium-io suites are green at 3dbcb31 on EC2 (validators 2,772 + 2, executor 374, miners 43). No re-request from us: your necessity question is still open and no fix was pushed.

neurons/executor/src/services/local_rent_service.py
```diff
@@ -0,0 +1,664 @@
...
+    """None when the spec is a rental container's; otherwise why it is refused (the validator's
+    log label). Enforced HERE, not only by the sender: the validator hotkey alone must not be able
+    to run anything but a rental's image with a rental's mounts on this host."""
+    if not carries_only_public_fields(spec):
+        return "spec carries a command, an entrypoint or a private environment"
+    if not spec.name.startswith(RENTAL_CONTAINER_NAME_PREFIXES):
+        return f"container name {spec.name!r} is not a rental's (pod_/filler_)"
+    for volume in spec.volumes:
+        if volume.source.startswith("/"):
+            if not volume.source.startswith(ALLOWED_HOST_BIND_PREFIXES) or ".." in volume.source:
```
jam6099: `refuse_spec` accepts any volume whose `source` starts with `ALLOWED_HOST_BIND_PREFIXES` (`/var/run/lium-dstack/`, `:101`) and has no `..`, with no condition on the mount mode or exact path. The one host path the honest path ever binds is the quote-broker socket, read-only (`cvm_quote_broker.py:106-108`). A validator key can therefore bind the whole dstack IPC directory (or the socket) read-write into an image of its choosing. Low (the signer already has host SSH; not renter-reachable), but the executor-side policy should enforce exactly what it documents. Require `read_only` for any host-path source and pin the source to the exact socket path, keeping the `..` check.

_- jam_bot_

neurons/executor/src/payloads/rent.py
```diff
@@ -0,0 +1,98 @@
...
+    image: bool = False
+    # The wire form of `ContainerRunSpec` (`datura.rental_spec.spec_to_wire`); parsed and bounded
+    # by `spec_from_wire`, never by hand here.
+    container: dict[str, Any] | None = None
+    ready: ReadyStep | None = None
+
+
+class RentIntentBody(WireModel):
+    schema_id: Literal["lium.local_rent/1"] = Field(alias="schema", default=SCHEMA)
+    nonce: str = Field(min_length=16, max_length=128)
```
jam6099: `nonce` is bounded in length but not in character set, while the neighbouring `ssh_host_key_sha256` gets a strict pattern. It is the only caller-influenced input in the destructive rollback path: a docker label value (`local_rent_service.py:451`) and a label-filter value that drives `remove_container(id, force=True)` (`:576`). Not exploitable today (docker-py JSON-encodes the filter and `NonceCache.claim` refuses reuse), but a destructive path should not rest on daemon filter-parsing. Constrain it, e.g. `^[0-9a-f]{16,128}$` to match the validator's `secrets.token_hex(16)`.

_- jam_bot_

neurons/executor/src/routes/apis.py
```diff
@@ -516,6 +524,88 @@ async def local_verify(request: Request):
...
+    if not isinstance(raw, dict):
+        raise HTTPException(status_code=422, detail="Body is not an object")
+    try:
+        intent = RentIntent.model_validate(raw)
+    except ValidationError as exc:
+        raise HTTPException(status_code=422, detail=exc.errors())
+    await verify_signature(SignaturePayload(signature=intent.signature), canonical_intent_message(raw))
+    body: RentIntentBody = intent
+
+    refused = check_intent_window(body, time.time(), settings.LOCAL_VERIFY_INTENT_WINDOW_SECONDS)
```
jam6099: Unlike `/verify` (`:478-483`, which adds `check_intent_target(intent, settings.MINER_HOTKEY_SS58_ADDRESS)`), `/rent` binds an intent only with the host-key digest, and its nonce cache is per executor process. If two executors ever share one SSH host key file (`SSH_HOST_KEY_PATH`, default a per-host `/etc/ssh/ssh_host_ed25519_key.pub`), the same signed intent and nonce are accepted by both and each creates a container. Add `miner_hotkey` to `RentIntentBody` and enforce `check_intent_target` in `/rent`, or document/assert that host keys are unique per executor.

_- jam_bot_

neurons/executor/src/services/local_rent_service.py
```diff
@@ -0,0 +1,664 @@
...
+    if spec.runtime not in ALLOWED_RUNTIMES:
+        return f"runtime {spec.runtime!r} is not a rental's"
+    if not set(spec.cap_add) <= ALLOWED_CAPABILITIES:
+        return f"capabilities {sorted(set(spec.cap_add) - ALLOWED_CAPABILITIES)} are not a rental's"
+    if any(ALLOWED_SYSCTLS.get(k) != v for k, v in spec.sysctls.items()):
+        return f"sysctls {sorted(k for k, v in spec.sysctls.items() if ALLOWED_SYSCTLS.get(k) != v)} are not a rental's"
+    if not {u.name for u in spec.ulimits} <= ALLOWED_ULIMIT_NAMES:
+        return f"ulimits {sorted({u.name for u in spec.ulimits} - ALLOWED_ULIMIT_NAMES)} are not a rental's"
+    for device in spec.devices:
+        if not device.path_on_host.startswith(ALLOWED_DEVICE_PREFIXES) or ".." in device.path_on_host:
```
jam6099: `ALLOWED_DEVICE_PREFIXES` (`:100`) has a non-slash-terminated `/dev/nvidia`, so this `startswith` check also accepts a bogus sibling such as `/dev/nvidia-foo`. `spec_from_wire` bounds paths to `/dev/` and rejects `..` (`rental_spec.py:410-416`) and Docker rejects most bad paths, so this is hardening only. Tighten the prefix to an explicit digits/`ctl`/`uvm`/`-cap` suffix rule.

_- jam_bot_

neurons/executor/src/services/local_rent_service.py
```diff
@@ -0,0 +1,664 @@
...
+    import docker  # local import: the module is mocked in unit tests
+
+    return docker.APIClient(base_url="unix:///var/run/docker.sock", timeout=CREATE_TIMEOUT_SECONDS)
+
+
+def host_gateway_ip(route_table: str | None = None) -> str | None:
+    """The default gateway of this container's network namespace (Linux `/proc/net/route`), which
+    on a bridge is the host itself; None when not readable."""
+    try:
+        text = route_table if route_table is not None else open(_HOST_GATEWAY_ROUTE).read()
```
jam6099: `contrib/STYLE.md:53` requires context managers for resource management. This line reads `/proc/net/route` with a bare `open(...).read()`; `_wait_ssh_banner` also calls `writer.close()` without `wait_closed` (`:657`). Short-lived fds, so severity is a judgement call, but both are new and both breach the documented rule. Use `with open(...) as fh` and `await writer.wait_closed()`.

_- jam_bot_

jam6099 (changes requested):

- Low: the executor host-bind allowlist accepts any path under the dstack IPC dir and does not require a read-only mount, so a signed intent can bind the host socket read-write into any image.
- Info: the /rent nonce is length-bounded but has no character-set bound and drives a destructive label-filter rollback.
- Low: /rent has no miner_hotkey target check, so two executors that share one SSH host key can each accept the same captured intent.
- Info: the device allowlist prefix also matches a crafted sibling device name.
- Low: a new service path reads the route table with a bare open().read() and closes a socket writer without wait_closed, against the documented context-manager rule.
- Low/Medium: 8 commits, none imperative and 83-561 chars, bundling unrelated work; the style guide wants atomic, imperative, <=50-char subjects.

_- jam_bot_

neurons/executor/src/services/local_rent_service.py
```diff
@@ -0,0 +1,664 @@
...
+    # --- the run -------------------------------------------------------------------------------
+
+    async def _run(self, body: RentIntentBody) -> RentResult:
+        started_wall = int(time.time())
+        started = time.perf_counter()
+        deadline = min(body.deadline_s, self.max_deadline_s)
+        results: dict[str, RentStepResult] = {}
+        try:
+            # docker-py's client constructor asks the daemon for its API version: off the loop too.
+            api = await self._open_api()
```
jam6099, line 282: The create clock starts after `_open_api`, not at the top of `run`: the client constructor (`INSPECT_TIMEOUT_SECONDS = 10`, `:291`) runs before `asyncio.wait(..., timeout=deadline)` (`:262`), while the validator sends `deadline_s = LOCAL_RENT_TIMEOUT_SECONDS - 15` (`docker_service.py:1320`). With the default 45 s the executor's worst case is 10 + 30 + a 10 s by-id rollback = 50 s, past the client's wait. On that gap the validator takes `may_have_acted` and force-removes the pod name while the create/rollback may still run, so the SDK fallback can then hit a name conflict. Start the deadline clock before `_open_api`, or add its bound to `EXECUTOR_ROLLBACK_MARGIN_S`.

_- jam_bot_

neurons/executor/src/services/local_rent_service.py
```diff
@@ -0,0 +1,664 @@
...
+    RentResult,
+    RentStepResult,
+)
+from services.local_verify_service import parse_port_range
+
+logger = logging.getLogger(__name__)
+
+# docker-py is synchronous; its calls run here, never on the event loop. Two workers: a create in
+# flight and its rollback, nothing wider — the route runs one intent at a time.
+_rent_executor = ThreadPoolExecutor(max_workers=2, thread_name_prefix="local-rent")
```
jam6099, line 77: `asyncio.wait_for` cancels the await, not the worker thread, and this pool has only two workers. A hung `create_container` (docker-py client timeout 60 s, `:75`) keeps its thread; `run.rollback()`, the late label pass and `_open_api` all submit here via `run_in_executor` (`:405-407`), and the rollback is awaited through `asyncio.shield` (`:267`, `:285`) with no bound on the queue wait. Two wedged calls therefore strand the rollback and every later `/rent` (409 busy) until a thread times out. Give the rollback its own executor and/or shorten the create client timeout.

_- jam_bot_

neurons/executor/src/services/local_rent_service.py
```diff
@@ -0,0 +1,664 @@
...
+
+def host_key_sha256(host_key_line: str | None) -> str | None:
+    """The digest both ends compute of the executor's SSH host public key line (the one the miner
+    reports and the validator pins for its SSH connections)."""
+    if not host_key_line or not host_key_line.strip():
+        return None
+    return hashlib.sha256(host_key_line.strip().encode("utf-8")).hexdigest()
+
+
+def refuse_spec(spec: ContainerRunSpec) -> str | None:
```
jam6099, line 167: The docstring says the hotkey alone must not run anything but a rental's image, but `refuse_spec` never checks `spec.image` (only the shape check in `IMAGE_PATTERN`, `rental_spec.py:293,370`) and leaves `device_requests`, `shm_size`, `cpu_count`, `memory_gb`, `storage_limit_gb` unpinned — parsed with wide bounds (`rental_spec.py:417-429`) and forwarded verbatim (`:188-191`). A signed intent can still ask for arbitrary GPU/resource values; docker rejects absurd ones with a 500 (a wasted create), but the policy is weaker than it claims. Constrain the image and those fields, or narrow the docstring.

_- jam_bot_

neurons/executor/src/routes/apis.py
```diff
@@ -516,6 +524,88 @@ async def local_verify(request: Request):
...
+    spec nor rewrite what was made. `/version` names the port as `local_rent_port`.
+
+    Auth, nonce and window exactly as `/verify` (the intent is signed as sent, replay refused).
+    Flag off → 404; the validator treats every non-200 as "use SSH". Nonces are shared with
+    `/verify`: one cache, one rule. Busy → 409 before the nonce is claimed, so the same signed
+    intent may be re-sent.
+    """
+    if not settings.EXECUTOR_LOCAL_RENT_ENABLED:
+        raise HTTPException(status_code=404, detail="Not Found")
+    if not _is_loopback_client(request):
```
jam6099, line 555: This peer check is the only thing separating a network caller from the unsigned answer, while the executor port is published (`docker-compose.app.yml:6-7`) and the same app serves `/rent`. It holds only because `executor.py` starts uvicorn with `proxy_headers=False` and nothing fronts it; enabling proxy headers, adding a proxy that sets `X-Forwarded-For`, or a host-network forwarder that connects from loopback removes the control (the signature still stops the create, but the answer and `/verify` outputs leak). Bind a loopback-only listener for `/rent` and `/verify`, or pin `proxy_headers=False` with a regression test.

_- jam_bot_

neurons/executor/src/services/local_rent_service.py
```diff
@@ -0,0 +1,664 @@
...
+            api = await self._open_api()
+        except Exception as exc:  # noqa: BLE001 — no client, nothing created: the validator falls back
+            logger.error("local rent: docker client unavailable: %r", exc)
+            results["container"] = RentStepResult(status="failed", error=f"docker client: {type(exc).__name__}: {exc}")
+            return self._result(body, started_wall, started, deadline, results, deadline_hit=False, rolled_back=True)
+        run = _RentAttempt(self, api, body, results)
+
+        task = asyncio.ensure_future(run.steps())
+        try:
+            done, pending = await asyncio.wait({task}, timeout=deadline)
```
jam6099: `deadline_hit = bool(pending)` is derived from the timeout expiry, not from `task.done()`. If `run.steps()` finishes at the instant the wait expires, `asyncio.wait` can still report the task pending; `task.cancel()` is then a no-op and `_succeeded` returns False (`:330-339`), so the rollback removes a container that was just created and observed running. The validator then re-creates it over SSH: wasteful, not wrong. Recompute `deadline_hit = bool(pending) and not task.done()`, or base `_succeeded` on the step statuses when the container and ready steps are both ok.

_- jam_bot_

neurons/executor/src/routes/apis.py
```diff
@@ -516,6 +524,88 @@ async def local_verify(request: Request):
...
+    body: RentIntentBody = intent
+
+    refused = check_intent_window(body, time.time(), settings.LOCAL_VERIFY_INTENT_WINDOW_SECONDS)
+    if refused:
+        raise HTTPException(status_code=401, detail=f"Intent refused: {refused}")
+    service = _get_local_rent_service()
+    refused = service.refuse_intent(body)
+    if refused:
+        raise HTTPException(status_code=401, detail=f"Intent refused: {refused}")
+    if service.busy:
```
jam6099: The route checks `service.busy`, claims the nonce (`:591`), then calls `service.run`, which re-checks `self._busy` and raises `RentBusyError` (`local_rent_service.py:238-242`). If that second guard ever fires, the 409 returns with the nonce already burnt, so a re-send of the same signed intent gets "nonce already used" instead of the documented "busy -> re-sendable" contract (`:552-556`). Unreachable today (no `await` between the check and the claim), but the duplicated guard makes the invariant implicit and easy to break. Drop the route-level check and let `run` raise before the claim.

_- jam_bot_

datura/datura/rental_spec.py
```diff
@@ -0,0 +1,531 @@
...
+
+
+# --- docker-py arguments (the one HostConfig) ---------------------------------------------------
+
+
+def build_host_config_kwargs(spec: ContainerRunSpec) -> dict:
+    """`APIClient.create_host_config(**kwargs)` for this spec. Both ends call this."""
+    kwargs = {
+        "port_bindings": {_port_key(port): port.host_port for port in spec.ports},
+        "binds": [
```
jam6099, line 178: `volume.target` and `device.path_in_container` are only required to be absolute (`:386-399`, `:408-422`), and `shm_size` is free-form, then interpolated into the `binds`/device strings docker-py passes on. The other checks stop a host escape, but a `:` or a control character in a mount target can confuse docker's bind/device parsing. Validate these against a strict charset at wire-parse time.

_- jam_bot_

neurons/validators/src/services/local_rent_client.py
```diff
@@ -0,0 +1,273 @@
...
+# runs past the budget is a client timeout, which frees the name here (`may_have_acted`).
+EXECUTOR_ROLLBACK_MARGIN_S = 15
+EXECUTOR_DEADLINE_MIN_S = 5
+
+
+def executor_deadline_s(timeout_s: int) -> int:
+    return max(EXECUTOR_DEADLINE_MIN_S, timeout_s - EXECUTOR_ROLLBACK_MARGIN_S)
+
+
+def host_key_sha256(host_key_line: str | None) -> str | None:
```
jam6099: Byte-identical to `host_key_sha256` in `neurons/executor/src/services/local_rent_service.py:137` apart from the docstring. The PR moved the run spec into `datura` for one definition shared by both ends (`rental_spec.py`), but this helper was left copied. Put it in `datura` once.

_- jam_bot_

jam6099 (changes requested):
_Follow-up review by jam_bot._

All five previous inlines are unchanged at this head, so they stay open. This pass adds the defects and spec axes that aborted the first time:

- Medium: the executor's answer budget excludes its own docker-client open, so a slow daemon can push the answer past the validator's wait and race the fallback name removal.
- Medium: a wedged docker call holds one of two pool workers; two strand rollback and every later rent until restart.
- Low: the executor-side policy does not constrain the image or the forwarded GPU/resource fields, though its docstring claims a rental-only shape.
- Low: rent/verify authz rests only on the peer being loopback while the port is published; enabling proxy headers or adding a fronting proxy removes it.
- Low: a deadline that fires as the steps finish can roll back a container that was created and running.
- Low: the nonce is claimed before the service's own busy check, so a busy answer burns the nonce against the documented re-send contract.
- Low: the local path moves only docker run; cleanup, volume, keys, sshd-ready and the sshd TCP-check stay on SSH, so the spec's local-rent scope and external check are not delivered.
- Low: the same host-key digest helper is duplicated on both ends, and feature plus refactor are mixed in one PR.
- Low: the bridge-gateway banner path and the cancellation rollback are untested.
- Info: mount targets and shm reach docker argument strings with no charset bound.

_- jam_bot_

neurons/executor/src/services/local_rent_service.py
jam6099: Still open on 3dbcb317.

neurons/executor/src/payloads/rent.py
jam6099: Still open on 3dbcb317.

neurons/executor/src/routes/apis.py
jam6099: Still open on 3dbcb317.

neurons/executor/src/services/local_rent_service.py
jam6099: Still open on 3dbcb317.

neurons/executor/src/services/local_rent_service.py
jam6099: Still open on 3dbcb317.

datura/datura/rental_spec.py
```diff
@@ -0,0 +1,531 @@
...
+        "memory_gb": spec.memory_gb or None,
+        "storage_limit_gb": spec.storage_limit_gb or None,
+        "shm_size": spec.shm_size,
+        "entrypoint": spec.entrypoint,
+        # the icc-off rental bridge (DAH-3199); None = the daemon's default bridge
+        "network": spec.network,
+    }
+
+
+def spec_from_wire(raw: Any) -> ContainerRunSpec:
```
taiberium, line 377: NIT: this codec and the nine `_str`/`_int`/`_objects` helpers rebuild pydantic. `datura/requests/base.py` already imports it, so a model drops about 150 lines.

taiberium (dismissed):

neurons/executor/src/services/local_rent_service.py
surcyf123: Holds: `refuse_spec` checks the prefix and `..` only, not the mount mode or the exact socket path. Pinning both is a design change. Waiting on DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build; numbers posted 12:4xZ). No design change until it is taken.

neurons/executor/src/payloads/rent.py
surcyf123: Holds: `nonce` is bounded in length only. A character-set pattern is a wire change. Waiting on DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build; numbers posted 12:4xZ). No design change until it is taken.

neurons/executor/src/routes/apis.py
surcyf123: Holds: `/rent` binds the intent to the host-key digest only; the `check_intent_target` call at `:480` is `/verify`'s. A `miner_hotkey` field on the intent is a wire change. Waiting on DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build; numbers posted 12:4xZ). No design change until it is taken.

neurons/executor/src/services/local_rent_service.py
surcyf123: Holds: `/dev/nvidia` has no slash, so `/dev/nvidia-foo` passes the prefix check. A suffix rule is a design change. Waiting on DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build; numbers posted 12:4xZ). No design change until it is taken.

neurons/executor/src/services/local_rent_service.py
surcyf123: Done in `44feec5`: `host_gateway_ip` reads the route table in a `with` block and `_wait_ssh_banner` awaits `wait_closed` after `close`.

surcyf123:
@jam6099 Checked all six rows on the head. The context-manager row is fixed in `44feec5`. The four hardening rows hold as read and are design points; the commit subjects get squashed at the fix round. Waiting on DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build; numbers posted 12:4xZ). No design change until it is taken.

neurons/executor/src/services/local_rent_service.py
surcyf123: Not changing this now. The budget's start point is a design point of a PR on hold under DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build). It is on the list for the fix round after the decision.

neurons/executor/src/services/local_rent_service.py
surcyf123: Not changing this now. The pool shape is a design point of a PR on hold under DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build). It is on the list for the fix round after the decision.

neurons/executor/src/services/local_rent_service.py
surcyf123: Not changing this now. Which fields the executor-side policy pins is a design point of a PR on hold under DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build). It is on the list for the fix round after the decision.

neurons/executor/src/services/local_rent_service.py
surcyf123: Not changing this now. The window is the instant between `asyncio.wait` returning and `task.cancel()`; the cost is one wasted create, as you say. It is on the list for the fix round after DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build).

neurons/executor/src/routes/apis.py
surcyf123: Not changing this now. Unreachable today, as you say; where the busy guard lives is a design point of a PR on hold under DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build).

datura/datura/rental_spec.py
surcyf123: Not changing this now. A charset rule on the wire is a wire change, a design point of a PR on hold under DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build).

neurons/validators/src/services/local_rent_client.py
surcyf123: Not changing this now. Moving `host_key_sha256` into `datura` is a refactor of a PR on hold under DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build). It is on the list for the fix round after the decision.

surcyf123:
@jam6099 Unchanged by design until DECISION-LIUMD-STACK-VALUE is taken (Rustam 15 Sep: do not build). The five rows you re-flagged and the new ones below are design points of this PR. The only code moves on the stack are the two plain defects fixed in #1347 at `95b45c8`: the no_tunnel_port order and the quoting in remove_with_volumes. This PR is restacked on that head at `cb9bfa1`, patch unchanged. The deadline_hit race and the duplicated host_key_sha256 helper are noted for the fix round after the decision.

neurons/executor/src/routes/apis.py
surcyf123: Not changing this now. The trust root for `/rent` and `/verify` is #1339's: the pinned SSH session, loopback peers only. A change to it is a design point on hold under DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build).

datura/datura/rental_spec.py
surcyf123: Agreed that a pydantic model would be shorter. It lands with DECISION-LIUMD-STACK-VALUE (Rustam 15 Sep: do not build). If the stack goes ahead, the codec and the nine helpers become a `datura/requests/base` model in the fix round. If not, the PR closes.

surcyf123:
@jam6099 Rebuilt as you asked on 16 Sep: this PR is now standalone on `main` (base retargeted from `liumd-phase3`; head `66623ca`), with nothing from #1343–#1347 in it. Both flags ship off (`EXECUTOR_LOCAL_RENT_ENABLED`, `VALIDATOR_LOCAL_RENT_ENABLED`); the tests for that now run against a live fake executor and fail when the gate is removed (checked on EC2 with the gate deleted).

The body has a **Metrics** section: the prod BEFORE for the segment `/rent` replaces is `docker_run_ms + container_running_check_ms` p50 2,047 ms / p90 4,354 ms over 5,076 rents in 14 days (to 15 Sep). The staging AFTER is pending; the plan (one loop-owned g6.xlarge executor on staging, ≥ 30 rents with the flags off and ≥ 30 on, same node) needs three things from you, which the loop posts in #lium-agents.

Your 12 review threads from 15 Sep are still open. They were parked on the stack decision; that decision is in, so they are listed under "Known follow-ups" in the body and are the next round on this branch. None of them changes the measured segment, so the staging number can be taken at this head.

jam6099:
@surcyf123  When are you going to address the un-addressed comments?

surcyf123:
@jam6099 the 12 threads from 15 Sep are being worked now (they were parked during the stack decision, then the standalone cut took the last round). One push with the fixes and a decision line per remaining thread within about an hour; the staging deploy pins the head it started from.

neurons/executor/src/services/local_rent_service.py
surcyf123: Done in `de253ef`: a host-path volume must be exactly `/var/run/lium-dstack/dstack.sock` and read-only. The directory, another socket, or the socket read-write is refused. Three new `refuse_spec` rows test it.

neurons/executor/src/payloads/rent.py
surcyf123: Done in `de253ef`: `nonce` is `^[0-9a-f]{16,128}$` on the wire, the alphabet of the validator's `token_hex`. Test: `test_a_nonce_outside_lower_case_hex_is_422_before_any_signature_check`.

neurons/executor/src/routes/apis.py
surcyf123: Done in `de253ef`: the intent carries a signed `miner_hotkey` and `/rent` runs the same `check_intent_target` as `/verify`; the validator sends the payload's miner hotkey. With the host-key digest, that binds the intent to one provider's one host. What stays: two executors of one miner that share a key file are told apart only by the tunnel, because the executor cannot check its own uuid. The field comment says so. Test: `test_an_intent_for_another_miners_executor_is_refused`.

neurons/executor/src/services/local_rent_service.py
surcyf123: Done in `de253ef`: device nodes are matched by name, not by prefix (`ALLOWED_DEVICE_PATTERN`). The names are the ones `shared_device_nodes_command` lists: nvidiaN, nvidiactl, modeset, uvm and uvm-tools, nvswitch and nvlink, the caps and imex channel nodes, infiniband uverbs and rdma_cm, tun, fuse, dri. Siblings such as `/dev/nvidia-foo`, `/dev/nvidia0extra` and `/dev/fuse2` are refused; every listed node passes.

neurons/executor/src/services/local_rent_service.py
surcyf123: Done in `de253ef`: the deadline counts from the top of `_run`. The wait is the deadline minus the time already spent, so the client open is inside the budget. Test: `test_the_deadline_counts_the_docker_client_open_too`. A 0.7 s open plus a 0.6 s create do not fit a 1 s deadline; under the old clock they did.

neurons/executor/src/services/local_rent_service.py
surcyf123: Done in `de253ef`: the rollback runs on its own `_rollback_executor` with two workers: by id, by label and the late pass. Two creates wedged in docker-py can no longer queue it. `CREATE_TIMEOUT_SECONDS` stays 60 because the executor accepts deadlines up to 120 s and the deadline, not the client timeout, cuts a slow create. Test: `test_the_rollback_runs_on_its_own_worker_while_both_rent_workers_are_wedged`.

neurons/executor/src/services/local_rent_service.py
surcyf123: Done in `de253ef`: the docstring now says what `refuse_spec` pins and what it leaves to the validator and the daemon. Not pinned in value: the image, since a renter picks any image, and `device_requests`, `shm_size`, the cpu, memory and storage sizes. Those are bounded in shape by `spec_from_wire`; `shm_size` is now a byte count with a k, m or g unit.

neurons/executor/src/services/local_rent_service.py
surcyf123: Done in `de253ef`: `deadline_hit` is `bool(pending) and not task.done()`, and the comment says what holds. CPython's `asyncio.wait` sorts by `f.done()` when the waiter wakes, so a task it reports pending was not done at that moment. The second clause is belt and braces; the state of the task decides.

neurons/executor/src/routes/apis.py
surcyf123: Done in `de253ef`: one guard. `LocalRentService.run` takes the claim as a callback, decides busy and claims the nonce under the same lock. The route's own busy check is gone and `NonceUsedError` is the 409. A busy 409 never calls the claim. Test: `test_a_busy_refusal_never_claims_the_nonce_and_a_used_nonce_creates_nothing`.

datura/datura/rental_spec.py
surcyf123: Done in `de253ef`: volume targets and device paths in the container must match `CONTAINER_PATH_PATTERN` at wire parse: absolute, plain characters, no colon or whitespace. `shm_size` must match the validator's own sanitiser form, a byte count with a k, m or g unit. The validator's `eligible` checks the targets first and takes the SDK path on a miss, so a template path the pattern refuses costs no round trip. Nine new wire rows test it.

neurons/validators/src/services/local_rent_client.py
surcyf123: Done in `de253ef`: `host_key_sha256` lives once, in `datura.rental_spec`; both ends import it.

neurons/executor/src/routes/apis.py
surcyf123: Two options: A. pin `proxy_headers=False` with a test. B. a second, loopback-only uvicorn listener for the rent and verify routes. Done A in `de253ef`: `test_executor_starts_uvicorn_without_proxy_headers` reads the uvicorn.run call in executor.py by AST and fails when the keyword is not False. Not doing B in this PR because it is a deploy change: a second port in the compose file, the port `/version` answers, the tunnel target. The verify route on main shares that trust root, so B would be one PR for both routes. Your call; the thread stays open.

datura/datura/rental_spec.py
surcyf123: Not changing this in this round because it is a size refactor with no behaviour change, about 150 lines, and this round is the twelve correctness threads. If you want the codec as a pydantic model before the merge, say so and it is one more commit on this branch.

surcyf123:
@jam6099 Now. 12 threads: 11 fixed in `de253ef` and resolved, 1 answered with two options and left open for your call. That one is the loopback peer check: `proxy_headers=False` is pinned by a test, and a loopback-only listener is a deploy change shared with the verify route. 0 obsolete: everything you commented on is still in the standalone diff. Rustam's pydantic NIT stays open as a follow-up. One commit on `fa9f7cf`; the lium-io suites are green on EC2 at this head. The push moved the head the staging deploy would build; the staging lane knows.

surcyf123:
Closed on Muhammad's call (jam_bot 20:28Z, no 'keep' by the 21:45Z default): staging arm B measured 742 ms vs 806 ms p50 (n=30), the most the box can show because both paths wait on the same docker run; prod was never measured and the 1–2 s there is a projection; 3.7k lines is not worth that. The measurement and method stay on DECISION-LIUMD-STACK-VALUE and the ticket; the staging flag flip-back is lium-io-deployment#49. Children of this stack are handled next.

PR #1304, opened 2026-09-07
DAH-2768 - [P1] sysbox installer adds NVIDIA's apt repo and shows apt errors

surcyf123:
Implements [DAH-2768](https://app.notion.com/p/The-Sysbox-installer-apt-installs-nvidia-container-toolkit-without-adding-the-NVIDIA-repository-and--3c7b8bfdbde981f19029ff45420366e0) · reviewer: @jam6099

**TL;DR** — `neurons/executor/nvidia_docker_sysbox_setup.sh` — the first thing a new provider runs (`sysbox.md:18`, `quickstart.md:33` pipe it into `sudo bash`). Sysbox installer adds NVIDIA's apt repo and shows apt errors. Not proven by the loop: a real chain and real GPU hosts (the #1312 stack runs the executor without a GPU).

**What changed**
- `ensure_nvidia_container_toolkit_repo()`: NVIDIA's documented sequence — key to `/usr/share/keyrings/nvidia-container-toolkit-keyring.gpg`, list to `/etc/apt/sources.list.d/nvidia-container-toolkit.list` with `signed-by`.
- `apt_install()`: every `apt-get` (update, toolkit+jq, sysbox `.deb`) keeps its output and prints the last 20 lines on failure.
- `APT_ROOT` (default empty) prefixes the four paths so the functions can be exercised against a temp tree.

**How to verify**
- `gh pr checks 1304 -R Datura-ai/lium-io` → all green
- re-run the loop's suites (Details → Verification) → executor 132 passed

**Verified by the loop:** 7 Sep 2026 · sandbox EC2 · executor 132 passed · Result: PASS a5272766 · Gate: not run

**Docs:** lium-docs#215 (companion)

<details><summary>Details</summary>

Origin: Discord ticket-0286 (provider HornsWithAHalo, 22 Aug 2026: "The script failed the first time … It is missing adding the nvidia container toolkit repo"; node off market 2026-08-21 23:29 → 08-25 13:33 UTC); ticket [DAH-2768](https://app.notion.com/p/The-Sysbox-installer-apt-installs-nvidia-container-toolkit-without-adding-the-NVIDIA-repository-and--3c7b8bfdbde981f19029ff45420366e0) (Oxygen 68).

## Problem
`neurons/executor/nvidia_docker_sysbox_setup.sh` — the first thing a new provider runs (`sysbox.md:18`, `quickstart.md:33` pipe it into `sudo bash`) — did
```
apt-get update -qq 2>/dev/null
[...]

PR #1349, opened 2026-09-10
DAH-3319 - [P0] sysbox installer preflight: PASS/FIX lines with the fix, --check mode

surcyf123:
Implements [DAH-3319](https://app.notion.com/p/Sysbox-installer-preflight-PASS-FIX-lines-with-the-exact-fix-before-anything-is-installed-check-3d7b8bfdbde9819faef4ed3ae8ff05a7) · reviewer: @taiberium

**TL;DR** — ticket-0309: a provider reinstalled Sysbox three times; the cause (iptables modules the validators' probe needs) surfaced only after staff logged in. The installer now checks the host first, one PASS / FIX line per requirement with its fix; `--check` adds what the script installs. Risk: a kernel below 5.19 now stops the install (main let sysbox-mgr decide); a driver, iptables, disk or port FIX is printed and the install goes on (round 1).

**What changed**
- Preflight in `nvidia_docker_sysbox_setup.sh`: root, x86_64, kernel ≥ 5.19, Docker up — a FIX stops the install
- Driver ≥ 580.65.06, iptables modules, disk ≥ 1.5× VRAM, ports — a FIX is printed and the install goes on (`preflight_advisory`)
- `--check`: those plus NVIDIA toolkit, Docker 29 `features.cdi` / `time-namespaces`, sysbox-runc probe; exit 1 on a FIX
- Ports from `EXECUTOR_PORT` / `SSH_PORT`, else the `.env` next to the script, else 8080 / 2200
- `cleanup()` EXIT trap no longer turns clean exits into status 1
- 68 tests, stub commands on PATH + fixture tree (`tests/test_sysbox_setup_preflight.py`)

**How to verify**
- `cd neurons/executor && pdm run pytest tests/test_sysbox_setup_preflight.py -q` → 68 passed
- `sudo bash nvidia_docker_sysbox_setup.sh --check` → PASS/FIX per line, `Preflight: N PASS, M FIX, K SKIP.`, exit 1 if M > 0
- `bash nvidia_docker_sysbox_setup.sh </dev/null`: Docker stopped → `Nothing was installed.`, exit 1; no GPU → driver FIX, `do not stop the install`, step 2 onwards
- `shellcheck -S style nvidia_docker_sysbox_setup.sh` → no output

**Verified by the loop:** Gate: pending at 379c9c8

**Ran it:** Ubuntu 22.04, no GPU (EC2 20260910T094640Z-mqjz): Docker stopped → `Nothing was installed.` rc=1; no GPU → driver + iptables FIX, sysbox 0.6.6 installed; ufw `ALLOW OUT` → FIX · executor 200 passed -kmrp

**Self-review:** clean at 379c9c8 (15 checks, 4 low)

**Parts:** backend n/a — host script · migration n/a — no schema · frontend n/a — host script · CLI/SDK n/a — PR-1 · docs ✓ lium-docs#268 · tests ✓ 68 · dashboards n/a — nothing emitted

**Docs:** [lium-docs#268](https://github.com/Datura-ai/lium-docs/pull/268) — `docs/providers/nodes/sysbox.md` "Check the host first" (merge after this PR is on main).

### Evidence (SO §25 origin)

- **ticket-0309** (Discord support, 3–4 Sep 2026): the provider had installed Sysbox, the documented verify command printed `nvidia-smi` output, the ports were open — and the validator scored the node 0 with `sysbox_runtime: false` for six hours across three reinstalls ("Honestly, I don't know what to do anymore"). Root cause after staff logged in: the host runs nftables and the legacy `iptable_nat` / `iptable_filter` modules were not loaded; the validators' Docker-in-Docker probe image runs legacy iptables, so its inner `dockerd` failed with `can't initialize iptables table 'nat'`, sshd never started, and the probe reported sysbox missing. The fix was one `modprobe`. Nothing on the host said so.
- **DAH-2768** (#1304, merged 8 Sep): the installer failed silently without NVIDIA's apt repository. **DAH-2919**: stale preflight image in `lium mine`. **VerifyX first-pass failures 11.6 %** (`speed/SWEEP_provider_verify.md`: 1,314 of 11,330 idle runs, 747 `VERIFYX_FAILED` + 563 `NETWORK_SPEED_TOO_SLOW`).
- Program: design/PROVIDER_BARRIER_PROGRAM.md §B.1 / §C PR-0 (owner P105.4, 10 Sep 2026: "faster GPU add, fully tested"). Metric: reinstall tickets → 0.
- Search first (L-261): #1304 already introduced the pattern of testing this script's functions under bash with stubs on PATH (`tests/test_sysbox_setup_apt_repo.py`); this PR follows it. No open lium-io PR touches `nvidia_docker_sysbox_setup.sh` (#1266 / #1274 touch the CVM scripts under `dstacktee/`).

### Design

- **Three groups of checks.** `preflight_blocking` is what sysbox itself needs on this host (x86_64 — the .deb is amd64 only; a kernel with ID-mapped mounts, or the override; Docker installed and running — the runtime goes into its daemon.json and it is restarted): a FIX there, or no root, ends install mode with `Nothing was installed.`, exit 1. `preflight_advisory` is what the validators need once the node runs (NVIDIA driver, iptables modules, disk, ports): sysbox installs without any of it and the node cannot be listed at all without sysbox, so a FIX is printed with its command, install mode says `The FIX lines above do not stop the install` and goes on; after `SUCCESS` / `Nothing to do.` a reminder names the count and the `--check` re-run (round 1, taiberium). `preflight_stack` is what the script installs (NVIDIA container toolkit, Docker 29 features, sysbox-runc); `--check` runs both and its FIX lines point at running the installer (or the manual daemon.json edit). Install mode does not run `preflight_stack` — it is about to do those things.
- **Exit codes.** `--check`: 0 when no FIX, 1 otherwise — advisory or blocking, it reports. Install mode: 1 on a blocking FIX before anything changes (the previous script also exited 1 on its one-line checks for root, Docker and the kernel); an advisory FIX does not change the exit code, which is the install's. `--help` 0, unknown option 2. The `cleanup` EXIT trap used `[ -n "$DOWNLOADED_DEB" ] && rm`, which under `set -e` made **every** exit report status 1 when nothing had been downloaded — `sudo bash nvidia_docker_sysbox_setup.sh; echo $?` printed 1 after SUCCESS and after "Sysbox is already working. Nothing to do." on `main`. It is an `if` now; `test_help_lists_check_and_exits_zero` covers it.
- **Kernel.** `kernel_supports_idmapped` (5.19+) as before; the FIX text depends on `/etc/os-release` (`22.04` → HWE kernel, plus the `SYSBOX_SKIP_KERNEL_CHECK=1` override line for a kernel known to carry the backport; `20.04` → upgrade the release; else generic). `SYSBOX_SKIP_KERNEL_CHECK=1` still passes an older kernel. Because the preflight settles the kernel first, the later `fail_old_kernel` branch after the sysbox-mgr report could no longer be reached and is removed; a `no` from sysbox-mgr on a 5.19+ kernel keeps `fail_no_idmapped` (filesystem / sysbox-mgr config).
- **Docker 29.** `check_docker_features` reads `/etc/docker/daemon.json` with `jq` when present and a grep fallback before jq is installed (jq is one of the packages this script installs). ≥ 29.2 needs `features.cdi = false`; ≥ 29.5 also `features["time-namespaces"] = false` — the same two settings the install step writes and lium-docs `docs/providers/nodes/sysbox.md` documents; the by-hand block in the FIX names only the keys that daemon needs. 29.0–29.1 passes with "untested" in the line (the docs say untested).
- **NVIDIA.** `nvidia-smi` present, a dotted version reported (nvidia-smi prints its errors on stdout — "Failed to initialize NVML: Driver/library version mismatch" lands in the "not loaded" line with the reboot fix, `test_nvidia_smi_error_text_is_not_read_as_a_version`), `/proc/driver/nvidia` present; then the version against `MIN_NVIDIA_DRIVER = 580.65.06` — `settings.MIN_NVIDIA_DRIVER_VERSION` in `neurons/validators/src/core/config.py`; `incentive/default.py::get_min_driver_multiplier` returns 0.0 for an idle node below it since `MIN_DRIVER_CUTOFF` (30 Jun 2026). A provider adding a node today with a 575 driver would be listed and paid nothing while idle; the FIX names the Ubuntu package (`nvidia-driver-580-server`, present in the 22.04 and 24.04 archives — Ran it, section K).
- **iptables modules (ticket-0309).** `ip_tables`, `iptable_nat`, `iptable_filter` loaded (`/proc/modules`) or built in (`/sys/module/<m>`). On both EC2 boxes Docker from get.docker.com (iptables-nft) had loaded `ip_tables` only. The FIX is `modprobe -a` — without `-a` modprobe reads the second and third names as parameters of the first and loads only `ip_tables` (the first proof run showed exactly that; section E) — plus a `modules-load.d` file so it survives a reboot.
- **Disk ≥ 1.5× VRAM.** The rule is `MIN_DISK_TO_VRAM_RATE = 1.5` in `neurons/validators/src/incentive/rental_price.py` (DAH-2520): `_insufficient_disk` compares `hard_disk.total` of the scrape — the **total** size of the filesystem the executor container sees as `/` (`miner_jobs/machine_scrape.py`: `shutil.disk_usage("/")`), which lives under Docker's data-root — against the sum of `gpu.details[].capacity` (MiB / 1024); free space is not the rule ("free is distorted by preallocated volumes"). The check does the same arithmetic on `df -Pk $(docker info --format '{{.DockerRootDir}}')` and the sum of `nvidia-smi --query-gpu=memory.total`, in the same order (VRAM and disk rounded to 0.1 GB, then VRAM × 1.5 unrounded against the disk — `test_disk_rule_compares_like_the_validator_at_the_boundary`). The gate is behind `ENABLE_UNRENTED_VRAM_OVER_DISK_LIMIT` (default False in `config.py`); production runs it on — `lium-io-deployment/infrastructure/Pulumi.prod.yaml:52` sets `"true"` — so the line is a FIX, not a note. Without a GPU the line is SKIP (not FIX) — the rule cannot be sized.
- **Ports.** Executor and SSH port: not held by a foreign process (`ss -Hltnp`); a `docker-proxy` listener is resolved to its container with `docker ps --filter publish=<port>` and passes only when that container is `executor-*` / `executor_*` (the names step 2 of this script already treats as the executor) — any other container is a FIX naming it; and not blocked by an active ufw (an inbound `ALLOW` in any column: plain, `ALLOW IN`, `on eth0`, `(v6)` rows; single-port and `lo:hi` range rules, tcp or any; an `ALLOW OUT` or `ALLOW FWD` row never counts — `test_ufw_allow_out_row_does_not_open_an_inbound_port`). The FIX line says the port has no inbound allow rule in the active ufw (Docker-published ports may bypass ufw's INPUT chain, so it does not claim the port is unreachable). Without `ss` the line is SKIP. Without root the preflight stops after the root FIX — the other checks read the process table and Docker's socket. Cloud firewalls and routers are outside the host, so the check prints the documented probe to run from another machine (`nc -vz <public IP> 8080 2200`) rather than a PASS it cannot prove. A probe from the host to its own public IP was tried on EC2 (section L: blocked by the security group, so it does detect a closed cloud firewall there) but a bare-metal host with the public IP on its NIC answers itself regardless of the upstream firewall, and a NAT without hairpin fails even when the forward is right — a PASS/FIX line from it would be wrong in both common provider shapes. The values come from `EXECUTOR_PORT` / `SSH_PORT`, else `EXTERNAL_PORT` / `SSH_PORT` in the `.env` next to the script, else the `lium mine` defaults 8080 / 2200 (lium-docs `docs/providers/nodes/quickstart.md`).
- **sysbox probe (`--check` only).** `sysbox-runc` on PATH, registered in `docker info` runtimes, and `docker run --rm --runtime=sysbox-runc alpine echo ok` — the probe lium-docs `docs/providers/nodes/sysbox.md` uses for the time-namespaces case. The GPU probe (`--gpus all … nvidia-smi`) stays where it was: the install path's step 6.
- **Idempotent.** `--check` writes nothing on the host (the sysbox probe pulls `alpine` once when the image is missing, as the docs' probe does); the second run prints the same lines (Ran it, section C2). Install mode's later steps are unchanged.
- **Re-run line.** `self_cmd` prints `sudo bash <path> [--check]` when `$0` is a file and the documented `curl … | sudo bash [-s -- --check]` one-liner when the script is piped from curl (Ran it, section I). The kernel override is printed after `sudo` (`sudo SYSBOX_SKIP_KERNEL_CHECK=1 bash …`) because sudo's `env_reset` drops a variable set in front of it. Piped from curl there is no file next to `$0`, so no `.env` is read (a `.env` in the provider's working directory must not steer the port check).
- **Driver compare** is on all three components (`version3_ge`): 580.65.05 is below 580.65.06. The kernel and Docker compares stay major/minor (`version_ge`) as before.
- **Not in this PR.** `lium mine` calling `--check` and the rewrite of Step 1 into one command (PR-1, design §C); a from-outside port probe (needs a Lium-side service); Ansible / fleet mode (DAH-2257). Docs for `--check`: lium-docs#268 (reviewer arhangel66 — the docs default `pr_ready.py` applies).

### Tests

`neurons/executor/tests/test_sysbox_setup_preflight.py` — 68, same harness as `test_sysbox_setup_apt_repo.py` (#1304): the script is sourced with `SYSBOX_SETUP_LIB=1` (or run from a copy in `tmp_path`, cwd there, so a developer's `neurons/executor/.env` cannot steer a test), commands are stubs on a PATH that also carries the real coreutils but not the host's `jq` unless the test asks, files under `SYSBOX_SETUP_HOST_ROOT` (test-only, named in the script header). `run_check` returns a `CheckResult` NamedTuple. Negative controls per check:

- kernel: 6.8 PASS · 5.15 on 22.04 FIX with the HWE command and the override after `sudo` · 5.15 on 20.04 FIX "upgrade the release" (and no HWE line) · 5.15 with `SYSBOX_SKIP_KERNEL_CHECK=1` PASS
- docker: 28.5.2 PASS · missing FIX `get.docker.com` · daemon down FIX `systemctl` · 29.1 PASS "untested" · 29.7 without settings FIX naming both keys and the two-key block (jq path and grep path) · 29.7 with both PASS (both paths) · 29.3 needs only cdi · 29.3 with `"cdi": true` FIX with the cdi-only block (jq path and grep path) · 28 needs nothing
- nvidia: 580.65.06 with 8 GPUs PASS · no nvidia-smi FIX · installed but `/proc/driver/nvidia` missing FIX · NVML error text FIX "not loaded" with the reboot fix · 575.57.08 / 580.65.05 / 580.64.99 FIX below minimum · 580.65.06 / 580.65.10 / 581.0.0 PASS · toolkit present PASS / missing FIX
- iptables: all loaded PASS · none loaded FIX with `modprobe -a` and the modules-load.d file · built-in (`/sys/module`) PASS
- disk: 8× 81559 MiB = 637.2 GB VRAM vs 1000.0 GB PASS · 500.0 GB FIX "below 955.8 GB" · the boundary the validator draws (79.7 GB VRAM: 119.5 GB FIX, 119.6 GB PASS) · no GPU SKIP
- ports: free PASS with the nc hint · 8080 held by nginx FIX · 2200 published by `executor-1` PASS · 8080 published by `nginx-proxy` FIX `docker stop nginx-proxy` · no `ss` SKIP · ufw verbose / `on eth0` / `(v6)` rows PASS · ufw active without a rule FIX `ufw allow 2200/tcp` · ufw `ALLOW OUT` and `ALLOW FWD` rows on 8080 FIX, `ALLOW IN` on 2200 PASS · ufw `2000:9000/tcp` range covers PASS · env override · `.env` next to the script · a `.env` in the working directory is ignored when piped from curl
- sysbox: installed + registered + runs PASS · missing / not registered / container fails FIX
- the script itself: `--check` good host → 12 PASS exit 0 · no GPU → 9 PASS 2 FIX 1 SKIP exit 1 · `--check` with a 575 driver → 11 PASS 1 FIX exit 1 · install mode on 5.15 / Docker down / aarch64 stops with `Nothing was installed.`, exit 1, step 2 never printed · its control: the same stubs on a good host → `Nothing to do.`, exit 0 · install mode with a 575 driver / no nvidia-smi / 500 GB disk / port held / ufw without rules → the FIX line, `do not stop the install`, `Nothing to do.`, the reminder, exit 0 · no iptables modules → the modprobe FIX and exit 0 · non-root in `--check` and in install mode: one FIX naming `sudo bash …` (`--check` kept), nothing else runs, install mode adds `Nothing was installed.` · piped from curl the FIX names the one-liner, never `sudo bash bash` · `--bogus` exit 2 · `--help` exit 0

### Ran it

EC2 sandbox, plain Canonical Ubuntu AMIs, t3.medium, no GPU — `20260910T022251Z-r0et` (Ubuntu 22.04.5, kernel 6.8.0-1063-aws) and `20260910T022251Z-lv2r` (Ubuntu 24.04.4, kernel 7.0.0-1012-aws), the script at `40c778c`; both logs read the same apart from the kernel line. Job: one script that runs each mode, then runs each FIX command and re-runs `--check`.

```
$ sudo -u ubuntu bash nvidia_docker_sysbox_setup.sh --check        # A. as a user, before Docker
  FIX  Not running as root — the checks read Docker's socket and the kernel modules.
         sudo bash /job/in/nvidia_docker_sysbox_setup.sh --check
  Preflight: 0 PASS, 1 FIX, 0 SKIP.                                 rc=1
$ curl -fsSL https://get.docker.com | sh                            # B. the Docker FIX line's command → server 29.8.0
$ bash nvidia_docker_sysbox_setup.sh --check                        # C. root, Docker up, no GPU
  PASS Running as root.
  PASS Architecture x86_64.
  PASS Kernel 6.8.0-1063-aws (>= 5.19, ID-mapped mounts available).
  PASS Docker 29.8.0.
  FIX  nvidia-smi not found — no NVIDIA driver installed.
         sudo apt-get install -y nvidia-driver-580-server && sudo reboot   # Ubuntu; or your vendor's driver package
  FIX  Kernel modules iptable_nat iptable_filter are not loaded — the validators' Docker-in-Docker probe needs legacy iptables …
         sudo modprobe -a ip_tables iptable_nat iptable_filter   # -a: without it modprobe reads the 2nd and 3rd name as parameters of the 1st
         printf 'ip_tables\niptable_nat\niptable_filter\n' | sudo tee /etc/modules-load.d/lium-iptables.conf >/dev/null   # survives reboots
  SKIP Disk >= 1.5x VRAM — no NVIDIA driver, VRAM unknown.
  PASS TCP 8080 (executor port) is free and not blocked by ufw.
  PASS TCP 2200 (SSH port) is free and not blocked by ufw.
         Cloud firewalls and routers are outside this host: after 'docker compose up', from ANY other machine run
           nc -vz <this host's public IP> 8080 2200    # both must say 'succeeded'
  FIX  NVIDIA container toolkit is not installed — Docker cannot pass GPUs into containers.
         sudo bash /job/in/nvidia_docker_sysbox_setup.sh   # adds NVIDIA's apt repository and installs nvidia-container-toolkit
  FIX  Docker 29.8.0 without features.cdi, time-namespaces = false in /etc/docker/daemon.json — sysbox rejects its containers.
         sudo bash /job/in/nvidia_docker_sysbox_setup.sh   # writes the features block and restarts Docker; stop rentals first
         or by hand: add {"features":{"cdi":false,"time-namespaces":false}} to /etc/docker/daemon.json && sudo systemctl restart docker
  FIX  sysbox-runc is not installed — validators reject a node without it.
         sudo bash /job/in/nvidia_docker_sysbox_setup.sh
  Preflight: 6 PASS, 5 FIX, 1 SKIP.                                 rc=1
$ bash … --check > c2; bash … --check > c3; diff c2 c3             # C2. idempotent → identical output on the second run
$ bash nvidia_docker_sysbox_setup.sh </dev/null                     # D. install mode — SUPERSEDED at the earlier head: since round 1 the same host
  …                                                                 #    goes on past the preflight; see "Round 1" below
$ sudo modprobe -a ip_tables iptable_nat iptable_filter && … tee /etc/modules-load.d/lium-iptables.conf   # E. the iptables FIX
  PASS Legacy iptables modules loaded (ip_tables iptable_nat iptable_filter) for the validators' Docker-in-Docker probe.
$ ufw --force enable; bash … --check | grep TCP                     # F. ufw without rules (wording at this head: "has no inbound allow rule in the active ufw")
  FIX  TCP 8080 (executor port) is blocked by ufw — validators cannot reach the node.
  FIX  TCP 2200 (SSH port) is blocked by ufw — validators cannot reach the node.
$ ufw allow 8080/tcp; ufw allow 2200/tcp; bash … --check | grep TCP   # the FIX lines' commands
  PASS TCP 8080 (executor port) is free and not blocked by ufw.
  PASS TCP 2200 (SSH port) is free and not blocked by ufw.
$ python3 -m http.server 8080 & bash … --check | grep 'TCP 8080'    # G. port held
  FIX  TCP 8080 (executor port) is already in use by users:(("python3",pid=4669,fd=3)) — the executor cannot bind it.
$ EXECUTOR_PORT=8001 SSH_PORT=2201 bash … --check | grep TCP        # H. env → TCP 8001 / TCP 2201; a .env next to a copy → TCP 8123 / TCP 2299
$ cat nvidia_docker_sysbox_setup.sh | bash -s -- --check | grep curl   # I. piped: the FIX re-run line is the one-liner
         curl -fsSL https://raw.githubusercontent.com/Datura-ai/lium-io/main/neurons/executor/nvidia_docker_sysbox_setup.sh | sudo bash   # adds NVIDIA's apt repository …
$ apt-cache show nvidia-driver-580-server | grep -E '^(Package|Version)'   # K. the driver FIX's package exists
Package: nvidia-driver-580-server
Version: 580.178.04-0ubuntu0.22.04.1                                # 24.04: 580.178.04-0ubuntu0.24.04.1
$ shellcheck -S style nvidia_docker_sysbox_setup.sh                 # M. → shellcheck clean (version: 0.8.0)
$ bash nvidia_docker_sysbox_setup.sh --help                         # N. → usage, rc=0
```

Also on the box (section L, not pasted — it prints the box's public IP): a listener on 8080 and `curl http://<own public IP>:8080/` from the same host timed out with the security group closed, so a self-probe through the public IP does see a closed cloud firewall on EC2; it is not a PASS/FIX line for the reasons in Design → Ports.

First proof run at the earlier head (`20260910T014638Z-agau`) is what found two wrong FIX commands: `modprobe ip_tables iptable_nat iptable_filter` loaded only `ip_tables` (fixed with `-a`), and `ubuntu-drivers` is not on the minimal Ubuntu images (the FIX names the package now).

**Suite.** `tools/aws_verify.sh lium-io 40c778c` (warm box, `20260910T022251Z-iyfa`): `executor pytest tests/: rc=0 — 189 passed` (132 on `main` + 57 here); the validators and miners trees are untouched. The tests also pass without the executor `conftest.py` (`--noconftest`) and on macOS bash 3.2.

### Round 1 (taiberium, 10 Sep 09:29Z) — Ran it

EC2 sandbox, plain Canonical Ubuntu 22.04.5 (kernel 6.8.0-1063-aws), t3.medium, no GPU — `20260910T094640Z-mqjz`, the script at the round-1 head.

```
$ systemctl stop docker; bash nvidia_docker_sysbox_setup.sh </dev/null   # B. a blocking FIX still stops the install
  PASS Running as root.  PASS Architecture x86_64.  PASS Kernel 6.8.0-1063-aws (>= 5.19, ID-mapped mounts available).
  FIX  Docker daemon is not running (docker ps failed).
         sudo systemctl enable --now docker
  Preflight: 3 PASS, 1 FIX, 0 SKIP.
  Fix the lines above and re-run. Nothing was installed.            rc=1; sysbox-runc not installed
$ ufw allow out 8080/tcp; ufw allow 2200/tcp; bash … --check | grep TCP   # C. an ALLOW OUT row does not open the port
  FIX  TCP 8080 (executor port) is blocked by ufw — validators cannot reach the node.      # wording at this head: "has no inbound allow rule in the active ufw"
  PASS TCP 2200 (SSH port) is free and not blocked by ufw.
$ ufw allow 8080/tcp; bash … --check | grep 'TCP 8080'                → PASS TCP 8080 (executor port) is free and not blocked by ufw.
$ bash nvidia_docker_sysbox_setup.sh --check                        # D. --check still exits 1 on any FIX
  Preflight: 6 PASS, 5 FIX, 1 SKIP.                                 rc=1
$ bash nvidia_docker_sysbox_setup.sh </dev/null                     # E. advisory FIXes: the install goes on
  FIX  nvidia-smi not found — no NVIDIA driver installed.
         sudo apt-get install -y nvidia-driver-580-server && sudo reboot   # Ubuntu; or your vendor's driver package
  FIX  Kernel modules iptable_nat iptable_filter are not loaded — …
  Preflight: 6 PASS, 2 FIX, 1 SKIP.
  The FIX lines above do not stop the install: the node is not listed at all without sysbox. Run those commands afterwards.
[2/7] Checking running containers
[3/7] Installing packages
  ✓ Sysbox v0.6.6 installed.
[4/7] Configuring Docker … [5/7] Restarting Docker … ✓ Docker is running.
[6/7] Verifying sysbox + GPU
  ✗ Verification FAILED. Diagnostics: … NVIDIA driver: FAILED             rc=1 — expected on a box without a GPU; sysbox-runc is installed and
                                                                            registered in daemon.json (printed after the run)
$ shellcheck -S style nvidia_docker_sysbox_setup.sh                 # F. → shellcheck clean (version: 0.8.0)
```

Section C ran at the commit before the fresh review's wording fix; the awk logic is the same. **Suite, round 1.** `tools/aws_verify.sh lium-io 483e1a9 --app io-all` (warm box, `20260910T095859Z-kmrp`): `executor pytest tests/: rc=0 — 200 passed` (132 on `main` + 68 here), `miners: 27 passed`; `validators: 2 failed, 1999 passed` — the same two (`test_an_unreadable_sysfs_tree_is_not_reported_as_no_devices`, `test_adopt_path_hardens_config_for_key_only_auth`) fail on every lium-io run on the warm box since 9 Sep (it runs as root with a live sshd); this PR does not touch the validators tree. **Round 2 (two rename NITs, 379c9c8):** `tools/aws_verify.sh lium-io 379c9c8` (warm box, `20260910T105551Z-k7ez`): `executor pytest tests/` → 200 passed, 0 skipped; `git diff 483e1a9 379c9c8 -w` is a pure rename, so the Ubuntu proof runs above stand.

### Self-review

Verdict `findings` at `379c9c8` · 15 checks · by subagent-fresh-r59m · 2026-09-10 11:02Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | STALE-BODY | `neurons/executor/nvidia_docker_sysbox_setup.sh:489` | GitHub already shows head 379c9c8 while the live body's `Verified by the loop: Gate: PASS 483e1a9`, `Self-review: clean at 483e1a9` and the `Suite, round 1 … 483e1a9 … -kmrp` line all name the previous head; the 379c9c8 run (EC2 20260910T105551Z-k7ez, executor 200 passed) is not in the body yet. | Record this verdict and the k7ez suite line at 379c9c8 with tools/pr_self_review.py --record --apply and let the gate re-stamp `Verified` at this head; no code change. |
| low | CORRECTNESS | `neurons/executor/nvidia_docker_sysbox_setup.sh:430` | Known low from the 483e1a9 review, unchanged by the rename: in `--check` with Docker installed but down, check_sysbox reports `installed but not registered in Docker's runtimes` although `docker info` merely failed, while check_docker_features handles the same state with a SKIP. | Before the runtimes grep, `docker ps &>/dev/null // { pf_skip "sysbox-runc — Docker is not running."; return 0; }`. |
| low | TEST-GAP | `neurons/executor/tests/test_sysbox_setup_preflight.py:111` | Known low, unchanged: the `with_jq=True` variants skip silently where jq is absent; the k7ez run shows `200 passed` with no skips, so jq was present there, but a `-q` summary on a box without jq would hide the jq path. | Have tools/aws_verify.sh (and CI) install or assert jq before the executor suite and quote the skipped count in the Suite line. |
| low | STALE-BODY | `RECOVERY/workers/sysbox-preflight-1349-fix/body.md:107` | Known low, unchanged: the first Ran-it block's section F still quotes the pre-change FIX text `is blocked by ufw — validators cannot reach the node` without the `wording at this head` note that section C of the round-1 block carries. | Add the same one-line `# wording at this head: …` note after F's two FIX lines, or replace them with the head's text. |

Mechanical notes dismissed: `neurons/executor/nvidia_docker_sysbox_setup.sh:6` A `#` header comment stating the script's documented distribution one-liner; nothing on this line executes, and it is unchanged by 379c9c8. · `neurons/executor/nvidia_docker_sysbox_setup.sh:153` A comment on self_cmd; the one-liner at line 160 is only `echo`ed as the re-run text of a FIX line, never run by the script. · `neurons/executor/nvidia_docker_sysbox_setup.sh:224` `curl -fsSL https://get.docker.com | sudo sh` is a string argument to pf_fix, printed as the fix text for a missing Docker (Docker's documented convenience install); the script does not execute it. · `neurons/executor/tests/test_sysbox_setup_preflight.py:246` An `assert … in out` on the printed FIX text of check_docker with the docker stub removed; no command is run. · `neurons/executor/tests/test_sysbox_setup_preflight.py:680` A comment; the test reads the local checkout's script and feeds it to `bash -s -- --check` over stdin with PATH set to the stub dir, to reproduce `$0 == bash`; no network access. · `neurons/executor/tests/test_sysbox_setup_preflight.py:695` An assertion that the FIX line names the curl one-liner when the script is piped; nothing is fetched or run. · `neurons/executor/tests/test_sysbox_setup_preflight.py:111` An environment `pytest.skip('jq not installed')` for the with_jq=True variants only; the grep variant of every case still runs, jq was present on the k7ez box (200 passed, 0 skipped), and it is recorded above as the known low TEST-GAP rather than a medium.

### Verified

Gate: pending at 379c9c8

</details>

### DIFF
--- a/neurons/executor/nvidia_docker_sysbox_setup.sh
+++ b/neurons/executor/nvidia_docker_sysbox_setup.sh
@@ -9,6 +9,7 @@ set -e
 # Env:
 #   SYSBOX_SKIP_KERNEL_CHECK=1  install even when the ID-mapped mounts check rejects the host
 #   EXECUTOR_PORT / SSH_PORT    the ports the preflight checks (else neurons/executor/.env next to this script, else 8080 / 2200)
+#   SYSBOX_SETUP_HOST_ROOT      test-only: prefix for the host files the preflight reads (/proc/modules, /etc/os-release, ...)
 
 SYSBOX_VERSION="0.6.6"
 SYSBOX_DEB_URL="https://github.com/nestybox/sysbox/releases/download/v${SYSBOX_VERSION}/sysbox-ce_${SYSBOX_VERSION}-0.linux_amd64.deb"
@@ -124,9 +125,11 @@ fail_no_idmapped() {
 
 # ── Preflight ───────────────────────────────────────────
 # One PASS / FIX line per requirement, each FIX with the command that fixes it. The host checks
-# run before anything is installed (a FIX there ends the run with nothing changed); `--check`
-# runs them plus the checks on what this script installs. Paths under HOST_ROOT so tests can
-# point them at a fixture tree; commands are stubbed on PATH.
+# run before anything is installed: a FIX in what sysbox itself needs (root, x86_64, kernel,
+# Docker) ends the run with nothing changed; the rest (driver, iptables modules, disk, ports)
+# print the fix and the install goes on, because the node cannot be listed at all without
+# sysbox. `--check` runs them all plus the checks on what this script installs. Paths under
+# SYSBOX_SETUP_HOST_ROOT so tests can point them at a fixture tree; commands are stubbed on PATH.
 
 MIN_NVIDIA_DRIVER="580.65.06"   # validators' MIN_NVIDIA_DRIVER_VERSION: an idle node below it earns nothing after the cutoff
 MIN_DISK_TO_VRAM_RATE="1.5"     # validators' MIN_DISK_TO_VRAM_RATE (rental_price.py): idle pay needs total disk >= 1.5x total VRAM
@@ -144,7 +147,7 @@ pf_fix() {
     return 1
 }
 
-host_path() { echo "${HOST_ROOT:-}$1"; }
+host_path() { echo "${SYSBOX_SETUP_HOST_ROOT:-}$1"; }
 
 self_cmd() {
     # how to run this script again: the file when there is one, else the one-liner (curl | bash). A leading VAR=VALUE
@@ -315,7 +318,7 @@ check_iptables_modules() {
         "printf 'ip_tables\\niptable_nat\\niptable_filter\\n' | sudo tee /etc/modules-load.d/lium-iptables.conf >/dev/null   # survives reboots"
 }
 
-check_disk_for_vram() {
+check_total_disk_for_vram() {
     # The validators compare the TOTAL size of the filesystem the executor sees as / (its
     # rootfs lives under Docker's data-root) with total GPU VRAM; free space is not the rule.
     local vram_mib vram_gb data_root total_kb total_gb needed_gb
@@ -339,7 +342,7 @@ check_disk_for_vram() {
         "Add disk, or move Docker's data-root to a filesystem of at least ${needed_gb} GB (\"data-root\" in /etc/docker/daemon.json, then sudo systemctl restart docker)."
 }
 
-preflight_ports() {
+resolve_preflight_ports() {
     # EXECUTOR_PORT / SSH_PORT from the environment, else the executor .env next to this script, else the defaults
     local env_file=""
     [ -f "$0" ] && env_file="$(dirname "$0")/.env"   # piped from curl there is no file next to the script
@@ -363,13 +366,14 @@ port_container() {
 }
 
 ufw_blocks_port() {
-    # true when ufw is active and no ALLOW rule covers TCP $1 (single port or lo:hi range)
+    # true when ufw is active and no inbound ALLOW rule covers TCP $1 (single port or lo:hi range)
     local status
     status=$(ufw status 2>/dev/null) || return 1
     echo "$status" | grep -q '^Status: active' || return 1
     ! echo "$status" | awk -v p="$1" '
         {
-            allow = 0; for (i = 2; i <= NF; i++) if ($i == "ALLOW") allow = 1   # "ALLOW", "ALLOW IN", "… on eth0 ALLOW"
+            # "ALLOW", "ALLOW IN", "… on eth0 ALLOW" open the port; "ALLOW OUT" / "ALLOW FWD" (ufw route) do not
+            allow = 0; for (i = 2; i <= NF; i++) if ($i == "ALLOW") { if ($(i+1) == "OUT" || $(i+1) == "FWD") next; allow = 1 }
             if (!allow) next
             split($1, spec, "/"); if (spec[2] != "" && spec[2] != "tcp") next
             n = split(spec[1], range, ":")
@@ -379,8 +383,8 @@ ufw_blocks_port() {
 }
 
 check_ports() {
-    local ports p listener container label fixed=0
-    read -r -a ports <<< "$(preflight_ports)"
+    local ports p listener container label has_fix=0
+    read -r -a ports <<< "$(resolve_preflight_ports)"
     command -v ss &>/dev/null || { pf_skip "Ports — 'ss' (iproute2) is missing, cannot tell what listens."; return 0; }
     for p in "${ports[0]}:executor" "${ports[1]}:SSH"; do
         label=${p#*:} p=${p%%:*}
@@ -392,18 +396,18 @@ check_ports() {
                 executor-*|executor_*) ;;   # the executor itself, from an earlier install
                 *) pf_fix "TCP $p ($label port) is published by container ${container:-unknown} — the executor cannot bind it." \
                        "docker stop ${container:-<name>}, or choose another port (EXTERNAL_PORT / SSH_PORT in neurons/executor/.env) and open it instead."
-                   fixed=1
+                   has_fix=1
                    continue ;;
             esac
         elif [ -n "$listener" ]; then
             pf_fix "TCP $p ($label port) is already in use by $listener — the executor cannot bind it." \
                 "Stop that process, or choose another port (EXTERNAL_PORT / SSH_PORT in neurons/executor/.env) and open it instead."
-            fixed=1
+            has_fix=1
             continue
         fi
         if ufw_blocks_port "$p"; then
-            pf_fix "TCP $p ($label port) is blocked by ufw — validators cannot reach the node." "sudo ufw allow $p/tcp"
-            fixed=1
+            pf_fix "TCP $p ($label port) has no inbound allow rule in the active ufw — open it for the validators." "sudo ufw allow $p/tcp"
+            has_fix=1
             continue
         fi
         if [ -n "$container" ]; then
@@ -414,7 +418,7 @@ check_ports() {
     done
     echo "         Cloud firewalls and routers are outside this host: after 'docker compose up', from ANY other machine run"
     echo "           nc -vz <this host's public IP> ${ports[0]} ${ports[1]}    # both must say 'succeeded'"
-    return $fixed
+    return $has_fix
 }
 
 check_sysbox() {
@@ -433,7 +437,8 @@ check_sysbox() {
             "$(self_cmd)   # re-applies the Docker 29 settings and re-verifies; then: journalctl -u sysbox-mgr --no-pager -n 20"
         return 1
     fi
-    pf_pass "sysbox-runc $(sysbox-runc --version 2>/dev/null | head -1 | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' || echo installed) runs a container."
+    # `sysbox-runc --version` prints its name on the first line and "version: 0.6.6" on the second
+    pf_pass "sysbox-runc $(sysbox-runc --version 2>/dev/null | grep -oE '[0-9]+\.[0-9]+\.[0-9]+' | head -1 | grep . || echo installed) runs a container."
 }
 
 preflight_summary() {
@@ -442,17 +447,30 @@ preflight_summary() {
     [ "$PREFLIGHT_FIX" -eq 0 ]
 }
 
-preflight_host() {
-    # what this script cannot install for you; a FIX here stops the run before anything changes.
-    # Returns 1 without root: every other check reads Docker's socket or the process table.
-    check_root || return 1
+preflight_blocking() {
+    # what sysbox itself needs before it can go on this host: x86_64 (the .deb is amd64 only), a kernel with
+    # ID-mapped mounts (or the override), Docker installed and running (the runtime is registered in its
+    # daemon.json and it is restarted). Returns 1 when any of them is a FIX; install mode stops there.
+    local fixes_before=$PREFLIGHT_FIX
     check_arch || true
     check_kernel || true
     check_docker || true
+    [ "$PREFLIGHT_FIX" -eq "$fixes_before" ]
+}
+
+preflight_advisory() {
+    # what the validators need once the node runs; sysbox installs without any of it and the node cannot be
+    # listed at all without sysbox, so a FIX here prints the fix and install mode goes on
     check_nvidia_driver || true
     check_iptables_modules || true
-    check_disk_for_vram || true
+    check_total_disk_for_vram || true
     check_ports || true
+}
+
+preflight_reminder() {
+    # after a successful install: the advisory FIX lines are far up the screen by now
+    [ "$PREFLIGHT_FIX" -gt 0 ] || return 0
+    warn "The preflight printed $PREFLIGHT_FIX FIX line(s) at the top. Run those commands, then: $(self_cmd --check)"
 }
 
 preflight_stack() {
@@ -474,7 +492,8 @@ case "${1:-}" in
     --check) PREFLIGHT_MODE_FLAG="--check" ;;
     -h|--help)
         echo "Usage: sudo bash nvidia_docker_sysbox_setup.sh [--check]"
-        echo "  (no option)  preflight the host, then install sysbox + NVIDIA container toolkit, configure Docker, verify"
+        echo "  (no option)  preflight the host, then install sysbox + NVIDIA container toolkit, configure Docker, verify;"
+        echo "               a FIX on root, x86_64, kernel or Docker stops the install, any other FIX is printed and the install goes on"
         echo "  --check      preflight only: host requirements and what this script installs, PASS/FIX per line, exit 1 on any FIX"
         echo "Env: SYSBOX_SKIP_KERNEL_CHECK=1, EXECUTOR_PORT, SSH_PORT (see the header of this script)"
         exit 0
@@ -486,7 +505,12 @@ esac
 
 if [ -n "$PREFLIGHT_MODE_FLAG" ]; then
     echo -e "\n${B}Preflight${N} — host requirements, then what this script installs"
-    if preflight_host; then preflight_stack; fi
+    # without root nothing else is checked: every other check reads Docker's socket or the process table
+    if check_root; then
+        preflight_blocking || true
+        preflight_advisory
+        preflight_stack
+    fi
     preflight_summary && exit 0
     echo "  Fix the lines above, then run: $(self_cmd)"
     exit 1
@@ -494,8 +518,13 @@ fi
 
 step 1 7 "Pre-flight checks"
 
-preflight_host || true
-preflight_summary || { echo "  Fix the lines above and re-run. Nothing was installed."; exit 1; }
+if ! check_root || ! preflight_blocking; then
+    preflight_summary || true
+    echo "  Fix the lines above and re-run. Nothing was installed."
+    exit 1
+fi
+preflight_advisory
+preflight_summary || echo "  The FIX lines above do not stop the install: the node is not listed at all without sysbox. Run those commands afterwards."
 
 echo -e "\n  This will install sysbox, configure Docker, restart Docker, and verify."
 if [ -t 0 ]; then
@@ -513,6 +542,7 @@ if command -v sysbox-runc &>/dev/null && docker info 2>/dev/null | grep -q sysbo
     fi
     if docker run --rm --runtime=sysbox-runc --gpus all "$VERIFY_IMAGE" nvidia-smi &>/dev/null; then
         ok "Sysbox is already working. Nothing to do."
+        preflight_reminder
         exit 0
     fi
 fi
@@ -695,6 +725,7 @@ if docker run --rm --runtime=sysbox-runc --gpus all "$VERIFY_IMAGE" nvidia-smi &
     echo -e "  ${G}╚══════════════════════════════════════════╝${N}"
     echo ""
     ok "Your executor now supports Docker-in-Docker."
+    preflight_reminder
 
     # ── 8. Restart executor ──────────────────────────────
     if [ -n "$EXECUTOR_COMPOSE_DIR" ]; then
--- a/neurons/executor/nvidia_docker_sysbox_setup.sh
+++ b/neurons/executor/nvidia_docker_sysbox_setup.sh
@@ -5,8 +5,10 @@ set -e
 # Usage:
 #   curl -fsSL https://raw.githubusercontent.com/Datura-ai/compute-subnet/main/neurons/executor/nvidia_docker_sysbox_setup.sh | sudo bash
 #   or: cd compute-subnet/neurons/executor && sudo bash nvidia_docker_sysbox_setup.sh
+#   sudo bash nvidia_docker_sysbox_setup.sh --check   only the preflight, one PASS/FIX line per requirement; exit 1 on any FIX
 # Env:
 #   SYSBOX_SKIP_KERNEL_CHECK=1  install even when the ID-mapped mounts check rejects the host
+#   EXECUTOR_PORT / SSH_PORT    the ports the preflight checks (else neurons/executor/.env next to this script, else 8080 / 2200)
 
 SYSBOX_VERSION="0.6.6"
 SYSBOX_DEB_URL="https://github.com/nestybox/sysbox/releases/download/v${SYSBOX_VERSION}/sysbox-ce_${SYSBOX_VERSION}-0.linux_amd64.deb"
@@ -20,7 +22,9 @@ warn() { echo -e "  ${Y}!${N} $1"; }
 fail() { echo -e "  ${R}✗${N} $1"; }
 step() { echo -e "\n${B}[$1/$2]${N} $3"; }
 
-cleanup() { [ -n "$DOWNLOADED_DEB" ] && rm -f "$DOWNLOADED_DEB"; }
+# an `if`, not `[ … ] && rm`: under `set -e` the failing test made the EXIT trap end every run
+# with status 1, including "Nothing to do." and SUCCESS
+cleanup() { if [ -n "$DOWNLOADED_DEB" ]; then rm -f "$DOWNLOADED_DEB"; fi; }
 trap cleanup EXIT
 
 version_ge() {
@@ -99,16 +103,6 @@ nvidia_hook_symptom() {
     fail "  nvidia-container-cli: mount error: .../merged/proc/driver/nvidia: no such file or directory"
 }
 
-fail_old_kernel() {
-    fail "Kernel $(uname -r) is too old for sysbox with GPUs — overlayfs gained ID-mapped mounts in 5.19 (6.x recommended)."
-    nvidia_hook_symptom
-    fail "Fix on Ubuntu 22.04: sudo apt-get install -y linux-generic-hwe-22.04 && sudo reboot"
-    fail "  Ubuntu 20.04 tops out at 5.15 even with HWE — upgrade the distro instead."
-    fail "Before rebooting: stop any rentals, and confirm the NVIDIA driver is DKMS-managed ('dkms status'),"
-    fail "otherwise a .run-installed driver will not load on the new kernel and the node comes back without GPUs."
-    fail "If this kernel is known to carry the backport, override with: sudo SYSBOX_SKIP_KERNEL_CHECK=1 bash $0"
-}
-
 fail_no_idmapped() {
     fail "Sysbox reports it cannot use ID-mapped mounts, although kernel $(uname -r) supports them."
     nvidia_hook_symptom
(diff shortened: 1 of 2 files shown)


neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -118,18 +122,380 @@ fail_no_idmapped() {
...
+}
+
+preflight_host() {
+    # what this script cannot install for you; a FIX here stops the run before anything changes.
+    # Returns 1 without root: every other check reads Docker's socket or the process table.
+    check_root || return 1
+    check_arch || true
+    check_kernel || true
+    check_docker || true
+    check_nvidia_driver || true
```
taiberium, line 465: **Critical:** A driver, disk or port FIX stops the install, but the node cannot be listed at all without sysbox. Make these three checks advisory.

neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -118,18 +122,380 @@ fail_no_idmapped() {
...
+}
+
+ufw_blocks_port() {
+    # true when ufw is active and no ALLOW rule covers TCP $1 (single port or lo:hi range)
+    local status
+    status=$(ufw status 2>/dev/null) || return 1
+    echo "$status" | grep -q '^Status: active' || return 1
+    ! echo "$status" | awk -v p="$1" '
+        {
+            allow = 0; for (i = 2; i <= NF; i++) if ($i == "ALLOW") allow = 1   # "ALLOW", "ALLOW IN", "… on eth0 ALLOW"
```
taiberium: NIT: The awk accepts any `ALLOW` token, so an `ALLOW OUT` row hides a blocked inbound port.

```suggestion
            allow = 0; for (i = 2; i <= NF; i++) if ($i == "ALLOW") { if ($(i+1) == "OUT") next; allow = 1 }   # "ALLOW", "ALLOW IN", "… on eth0 ALLOW"
```

neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -118,18 +122,380 @@ fail_no_idmapped() {
...
+    # the validator rounds VRAM and disk to 0.1 GB, then compares VRAM x 1.5 unrounded with the disk
+    if awk -v t="$total_gb" -v v="$vram_gb" -v r="$MIN_DISK_TO_VRAM_RATE" 'BEGIN {exit !(t >= v * r)}'; then
+        pf_pass "Disk ${total_gb} GB on $data_root >= ${needed_gb} GB (${MIN_DISK_TO_VRAM_RATE}x of ${vram_gb} GB VRAM) — idle pay eligible."
+        return 0
+    fi
+    pf_fix "Disk ${total_gb} GB on $data_root is below ${needed_gb} GB (${MIN_DISK_TO_VRAM_RATE}x of ${vram_gb} GB VRAM) — the node is listed but earns nothing while idle." \
+        "Add disk, or move Docker's data-root to a filesystem of at least ${needed_gb} GB (\"data-root\" in /etc/docker/daemon.json, then sudo systemctl restart docker)."
+}
+
+preflight_ports() {
```
taiberium: NIT: `preflight_ports` only resolves port numbers. Rename it `resolve_preflight_ports`.

neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -118,18 +122,380 @@ fail_no_idmapped() {
...
+    if [ -z "$missing" ]; then
+        pf_pass "Legacy iptables modules loaded (ip_tables iptable_nat iptable_filter) for the validators' Docker-in-Docker probe."
+        return 0
+    fi
+    pf_fix "Kernel modules $missing are not loaded — the validators' Docker-in-Docker probe needs legacy iptables and reports sysbox missing without them." \
+        "sudo modprobe -a ip_tables iptable_nat iptable_filter   # -a: without it modprobe reads the 2nd and 3rd name as parameters of the 1st" \
+        "printf 'ip_tables\\niptable_nat\\niptable_filter\\n' | sudo tee /etc/modules-load.d/lium-iptables.conf >/dev/null   # survives reboots"
+}
+
+check_disk_for_vram() {
```
taiberium: NIT: `check_disk_for_vram` compares total disk, not free. Rename it `check_total_disk_for_vram`.

neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -118,18 +122,380 @@ fail_no_idmapped() {
...
+            if (!allow) next
+            split($1, spec, "/"); if (spec[2] != "" && spec[2] != "tcp") next
+            n = split(spec[1], range, ":")
+            if ((n == 1 && range[1] == p) || (n == 2 && range[1] + 0 <= p + 0 && p + 0 <= range[2] + 0)) found = 1
+        }
+        END { exit !found }'
+}
+
+check_ports() {
+    local ports p listener container label fixed=0
```
taiberium: NIT: `fixed=0` is a flag, not a count. Rename it `has_fix`.

neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -118,18 +122,380 @@ fail_no_idmapped() {
...
+    # $1 what is wrong; every further argument is one line of the fix
+    PREFLIGHT_FIX=$((PREFLIGHT_FIX + 1))
+    echo -e "  ${R}FIX${N}  $1"
+    shift
+    local line
+    for line in "$@"; do echo "         $line"; done
+    return 1
+}
+
+host_path() { echo "${HOST_ROOT:-}$1"; }
```
taiberium: NIT: `HOST_ROOT` is a test-only knob with a generic name. Rename it `SYSBOX_SETUP_HOST_ROOT` and mark it test-only in the header.

neurons/executor/tests/test_sysbox_setup_preflight.py
```diff
@@ -0,0 +1,614 @@
...
+    proc = subprocess.run(
+        ["bash", "-c", program],
+        capture_output=True,
+        text=True,
+        cwd=tmp_path,
+        env={"PATH": str(stubs), "HOST_ROOT": str(root), **(env or {})},
+    )
+    assert proc.returncode == 0, proc.stdout + proc.stderr
+    out = ANSI.sub("", proc.stdout)
+    counts = dict(kv.split("=") for kv in out.strip().splitlines()[-1].split())
```
taiberium: NIT: Rename `program` to `bash_snippet` and `counts` to `summary_counters`.

neurons/executor/tests/test_sysbox_setup_preflight.py
```diff
@@ -0,0 +1,614 @@
...
+def _script_copy(tmp_path):
+    """The script in its own directory, so the `.env` it may read next to itself is the test's, never the checkout's."""
+    copy_dir = tmp_path / "executor"
+    copy_dir.mkdir(exist_ok=True)
+    copy = copy_dir / "nvidia_docker_sysbox_setup.sh"
+    shutil.copy(SCRIPT, copy)
+    return copy
+
+
+def run_check(tmp_path, function, *, env=None, files=None, with_jq=False, without=()):
```
taiberium: NIT: These helpers carry no types and `run_check` returns a bare 3-tuple. Add annotations and a NamedTuple.

taiberium (changes requested):
Requested changes.

neurons/executor/nvidia_docker_sysbox_setup.sh
surcyf123: Done in `483e1a9`: your awk, plus the same skip for ALLOW FWD rows from ufw route. Test: test_ufw_allow_out_row_does_not_open_an_inbound_port, and on a real ufw an ALLOW OUT row on 8080 now gives the FIX line.

neurons/executor/nvidia_docker_sysbox_setup.sh
surcyf123: Done in `483e1a9`: renamed to resolve_preflight_ports.

neurons/executor/nvidia_docker_sysbox_setup.sh
surcyf123: Done in `483e1a9`: renamed to check_total_disk_for_vram, in the script and the five tests.

neurons/executor/nvidia_docker_sysbox_setup.sh
surcyf123: Done in `483e1a9`: renamed to has_fix.

neurons/executor/nvidia_docker_sysbox_setup.sh
surcyf123: Done in `483e1a9`: renamed to SYSBOX_SETUP_HOST_ROOT; the header's Env list marks it test-only.

neurons/executor/tests/test_sysbox_setup_preflight.py
surcyf123: Done in `483e1a9`: renamed to bash_snippet and summary_counters.

neurons/executor/tests/test_sysbox_setup_preflight.py
surcyf123: Done in `483e1a9`: run_check returns a CheckResult NamedTuple with rc, output and fix_count; _bin_dir, _host_root, _script_copy, run_check and run_script carry annotations.

neurons/executor/nvidia_docker_sysbox_setup.sh
surcyf123: Done in `483e1a9`: driver, iptables modules, disk and ports are advisory now. The FIX and its command are printed, the install goes on, and a reminder follows the SUCCESS banner. Still blocking: root, x86_64, kernel 5.19 or the override, and Docker installed and running. Those are what sysbox itself needs. The .deb is amd64 only. The runtime goes into Docker's daemon.json and Docker is restarted. Without ID-mapped mounts the GPU probe at step 6 fails anyway. The iptables modules are advisory too, by the same rule: sysbox installs without them. Proof on a plain Ubuntu box without a GPU: the install went past the preflight and installed sysbox 0.6.6. With Docker stopped it still ends with Nothing was installed. Tests: test_install_mode_goes_on_after_an_advisory_fix, test_install_mode_goes_on_without_the_iptables_modules, test_install_mode_stops_before_installing_on_a_blocking_fix. Compare: https://github.com/Datura-ai/lium-io/compare/40c778c...483e1a9.

neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -118,18 +123,408 @@ fail_no_idmapped() {
...
+    check_sysbox || true
+}
+
+
+# sourced by the tests: functions only, nothing below runs
+if [ -n "${SYSBOX_SETUP_LIB:-}" ]; then
+    return 0
+fi
+
+PREFLIGHT_MODE_FLAG=""
```
taiberium: NIT: `PREFLIGHT_MODE_FLAG` holds an option string (`--check`), not a flag.

```suggestion
CHECK_MODE_OPTION=""
```

neurons/executor/nvidia_docker_sysbox_setup.sh
```diff
@@ -118,18 +123,408 @@ fail_no_idmapped() {
...
+    fi
+    daemon_feature_off cdi || missing="cdi"
+    if docker_version_ge 29 5 && ! daemon_feature_off time-namespaces; then
+        missing="${missing:+$missing, }time-namespaces"
+    fi
+    if [ -z "$missing" ]; then
+        pf_pass "Docker $version has the sysbox settings in /etc/docker/daemon.json (features.cdi$(docker_version_ge 29 5 && echo ' and features.time-namespaces') = false)."
+        return 0
+    fi
+    local block='{"features":{"cdi":false}}'
```
taiberium: NIT: `block` does not say what it holds. Rename it `daemon_features_block`.

taiberium (approved):

neurons/executor/nvidia_docker_sysbox_setup.sh
surcyf123: Done in `379c9c8`: `CHECK_MODE_OPTION` at the four sites; executor tests 200 green on EC2.

neurons/executor/nvidia_docker_sysbox_setup.sh
surcyf123: Done in `379c9c8`: `daemon_features_block` now; the printed hint still says features block.

PR #1355, opened 2026-09-10
DAH-3386 - [P1] the reset that flips a rented node names its check and carries diagnostics

surcyf123:
Implements [DAH-3386](https://app.notion.com/p/3d7b8bfdbde981068710f95288af1852) · reviewer: @taiberium

**TL;DR** — The reset that flips a rented node inactive reaches the backend as reason 0/1, so the penalty it raises cannot say which check fired or why (165 in 90 d, 11 reverted by hand). Every clearing check now names itself and `POD_NOT_RUNNING` carries the container's death diagnostics; the new wire fields are optional both ways. No check outcome or score changes.

**What changed**
- `updates_with_clear_verified_job_evidence` fills `clear_verified_job_evidence` (`reason_code`, `check_id`) for any check that sets `clear_verified_job_info` (`services/task/pipeline.py`)
- `TenantEnforcementCheck` adds `pod_id`, `container_name` and the container's death diagnostics on `POD_NOT_RUNNING` — strings cut to 256 chars, exit code only as int, flags only as bool (`services/task/checks/rented_machine.py`)
- `clear_verified_job_info(evidence=)` publishes `reason_code`, `check_id`, `evidence` next to `reason` (`services/redis_service.py`, `services/task/result_handler.py`)
- `ResetVerifiedJobRequest` gains optional `reason_code`, `check_id`, `evidence`; the compute client forwards them (`protocol/vc_protocol/validator_requests.py`, `clients/compute_client.py`)

**How to verify**
- `pytest neurons/validators/tests/test_reset_verified_job_evidence.py -q` → 6 passed (diagnostics on the reset; hostile inspect state bounded; an unnamed clearing check is named; transport-unreachable adds nothing; wire shape with and without the new keys; the published payload reaches the backend queue)
- `pytest neurons/validators/tests/test_rented_machine_check.py neurons/validators/tests/test_pipeline_default_scenarios.py -q` → green (existing behaviour)

**Ran it:** `aws_verify.sh lium-io 4eff09c --bundle` full suite → 2223 passed, 2 pre-existing env failures in untouched files (Details) · EC2 20260911T081217Z-t1mc

**Verified by the loop:** Gate: PASS 4eff09c

**Self-review:** clean at 4eff09c (19 checks)

**Parts:** validator ✓ · backend ✓ lium-platform#325 (reads the new fields) · migration n/a — wire fields only · frontend n/a — nothing renter-facing changes here · CLI/SDK n/a — validator-internal · docs ✓ lium-docs#284 · tests ✓ 6 · dashboards n/a — no metric

**Docs:** docs: https://github.com/Datura-ai/lium-docs/pull/284 — `penalties.mdx` documents `evidence.validator`, the field this PR fills

### Why (lium-ads reports/PENALTY_ACCURACY_20260910.md)

Backend `reset_verified_job` marks the executor inactive and raises `EXECUTOR_INACTIVE_MID_RENTAL` on every `ResetVerifiedJobRequest`; the request carries `reason` ∈ {DEFAULT, POD_NOT_RUNNING}. Spec change, GPU fingerprint, duplicate executor, banned GPU, NVML spoof and rental verification all arrive as DEFAULT. The 11 hand reverts on this path in 90 d were one `POD_NOT_RUNNING` cycle between two `RENTED` cycles (ticket-0292) and renters' own containers exiting (image without `MINER_KEY`, entrypoint exit 2, sglang wedging the kernel) — each needed an SSH to the host to establish. With the diagnostics on the row the backend (DAH-3385) can show them, and the owner can decide a rule on them (DECISION-PENALTY-RENTER-EXIT in lium-ads).

### Compatibility

Publish payload gains three keys; `compute_client` reads them with `.get`, so a validator on this build talking to an older backend sends fields the older `ResetVerifiedJobRequest` model ignores (pydantic default `extra=ignore` on `BaseRequest`), and an older validator's payload parses on the new backend (all three default to None — test 5). `evidence` is a `ResetVerifiedJobEvidence` model (`pod_id`, `container_name`, `container` with the eight `_PENALTY_EVIDENCE_FIELDS`), `extra="allow"` so a key a newer check adds never drops the reset; unanswered fields stay absent on the wire (a `None`-dropping serializer), so the JSON is the redis payload unchanged. Logs and host context stay in Loki (`_PENALTY_EVIDENCE_FIELDS` is the allow-list); every forwarded value is typed and bounded (`_penalty_evidence_from_diagnostics`: strings ≤ `_PENALTY_EVIDENCE_STR_MAX`, exit code int, flags bool) because it is the provider host's `docker inspect` answer and the backend writes it into JSONB and a log line. The backend bounds the whole payload again on receipt (lium-platform#325 `bounded_reset_evidence`).

### Ran it

```
aws_verify: run 20260911T081217Z-t1mc (4eff09c, rebased onto main 7b53a06) validators pytest tests/: 2 failed, 2223 passed, 15 skipped — failures: tests/test_scrape_infiniband.py::test_an_unreadable_sysfs_tree_is_not_reported_as_no_devices, tests/test_sshd_bootstrap_script.py::test_adopt_path_hardens_config_for_key_only_auth (files not in this diff)
aws_verify: run 20260910T225326Z-d94n --only tests/test_reset_verified_job_evidence.py tests/test_rented_machine_check.py tests/test_pipeline_default_scenarios.py: 43 passed (the first head, before the rebase)
aws_verify: run 20260910T230648Z-gum3 same targets at the head taiberium approved (before the rebase): 44 passed
```

### Self-review

Verdict `clean` at `4eff09c` · 19 checks · by subagent-r91-1355 · 2026-09-11 08:29Z (tools/pr_self_review.py)

Mechanical notes dismissed: `neurons/validators/src/protocol/vc_protocol/validator_requests.py:143` Medium 1 from 555cf73 is fixed: `_AbsentWhenNone` (extra=allow + `@model_serializer(mode="wrap")` dropping None) is inherited by both evidence models; re-checked on pydantic 2.13 with the same shapes — `model_dump_json()` of the queued request equals the redis payload's evidence key for key, extras (`future_key`, `other_new`) are kept, `0`/`False` are kept, an all-None container serialises as `{}` exactly as `_penalty_evidence_from_diagnostics` publishes it, `evidence: None` at the request level stays `null` (the request model has no serializer), python-mode dump matches, and `model_validate_json` round-trips; the docstring now says what the wire does, including the deliberate-None drop. · `neurons/validators/tests/test_reset_verified_job_evidence.py:243` The wire assertion now uses `request.model_dump_json()` with nothing excluded — the exact call `send_model` makes — and pins it to `payload["evidence"]` and the literal dict, so a regression to per-field nulls or to a dropped key fails the test; the bare case still asserts `plain.evidence is None`. · `neurons/validators/src/services/task/pipeline.py:211` Medium 2 is fixed: merge-base is origin/main 7b53a06f, `git merge-tree --write-tree origin/main 4eff09c8` writes a tree with no conflict, `STEP_SUMMARY_MIN_MS` is gone from the file (main's #1356 removal kept), `updates_with_clear_verified_job_evidence` is defined once (line 211) and used at the single call site (line 290); `git diff origin/main 4eff09c8 --stat` is the PR's own 7 files (+386 −3). · `neurons/validators/src/services/task/pipeline.py:217` Body low carried from 555cf73 (the bullet still names `clear_evidence_for`; How-to-verify `5 passed`, Parts `tests ✓ 5`, Compatibility's `test 4` pointer) — the author fixes these in the body before the push per the brief; not re-flagged. · `neurons/validators/src/clients/compute_client.py:394` Unchanged since 555cf73: pydantic validates the `data.get("evidence")` dict into ResetVerifiedJobEvidence and leaves None as None; nothing else in neurons/ constructs ResetVerifiedJobRequest or reads `.evidence`; result_handler.py:309 passes the Context dict to the publisher. · `neurons/validators/src/services/task/pipeline.py:163` `clear_verified_job_evidence: dict[str, Any] | None` on the Context is a deliberate superset of the wire model (it carries reason_code/check_id that redis_service.py pops into top-level request fields); typing it as ResetVerifiedJobEvidence would be wrong. · `neurons/validators/tests/test_reset_verified_job_evidence.py:191` The CancelledError exit is reliable (BaseException past `except Exception`, `finally` still awaits `pubsub.aclose()` and the test asserts it, `await asyncio.sleep(1)` never reached; same pattern as test_incentive_reasons_payload.py:50); deleting the three `data.get` kwargs fails at `request.reason_code == "POD_NOT_RUNNING"` and at `request.evidence.pod_id`, so the SO §52 gap is guarded. · `neurons/validators/src/protocol/vc_protocol/validator_requests.py:147` `extra="allow"` remains the right trade-off: known keys are named and typed as the reviewer asked, a key a newer check adds flows to the backend's `evidence: dict | None` instead of a ValidationError in the redis loop that would drop the reset and tear the pubsub down for a second; the only publisher is the validator's own bounded `_penalty_evidence_from_diagnostics`, so lax coercion of mistyped known keys is unreachable on this path and the backend re-bounds on receipt (RESET_EVIDENCE_MAX_BYTES=4096).

### Verified

Gate: PASS (4eff09c) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1355)

Gate: not yet run for this head

</details>

### DIFF
--- a/neurons/validators/src/protocol/vc_protocol/validator_requests.py
+++ b/neurons/validators/src/protocol/vc_protocol/validator_requests.py
@@ -146,6 +146,12 @@ class ResetVerifiedJobRequest(BaseValidatorRequest):
     miner_hotkey: str
     executor_uuid: str
     reason: ResetVerifiedJobReason = ResetVerifiedJobReason.DEFAULT
+    # DAH-3386: the check that cleared the verified job (its event reason_code and check_id) and what it saw —
+    # for POD_NOT_RUNNING the container's status, exit code, OOM flag and finish time. The backend writes them
+    # into the penalty it raises (lium-platform DAH-3385 details.evidence.validator). Optional both ways.
+    reason_code: str | None = None
+    check_id: str | None = None
+    evidence: dict | None = None
 
 
 class DuplicateExecutorsRequest(BaseValidatorRequest):
--- a/neurons/validators/src/protocol/vc_protocol/validator_requests.py
+++ b/neurons/validators/src/protocol/vc_protocol/validator_requests.py
@@ -140,6 +140,42 @@ class ResetVerifiedJobReason(int, enum.Enum):
     POD_NOT_RUNNING = 1         # container for pod is not running
 
 
+class _AbsentWhenNone(pydantic.BaseModel):
+    """A field the host did not answer stays absent on the wire (as the redis payload had it), instead of a
+    `null` the backend's JSONB row would keep; `None` set on purpose (`container_finished_at: None`) is dropped
+    too — the row reads the same either way."""
+
+    model_config = pydantic.ConfigDict(extra="allow")
+
+    @pydantic.model_serializer(mode="wrap")
+    def _drop_none(self, handler: pydantic.SerializerFunctionWrapHandler) -> dict[str, Any]:
+        return {key: value for key, value in handler(self).items() if value is not None}
+
+
+class ResetVerifiedJobContainerEvidence(_AbsentWhenNone):
+    """What the rented-machine check saw of the pod's container (`docker inspect`, typed and bounded on the
+    validator: rented_machine.py `_penalty_evidence_from_diagnostics`). Every field is optional."""
+
+    container_status: str | None = None
+    container_exit_code: int | None = None
+    container_oom_killed: bool | None = None
+    container_error: str | None = None
+    container_started_at: str | None = None
+    container_finished_at: str | None = None
+    container_missing: bool | None = None
+    diagnostics_capture_error: str | None = None
+
+
+class ResetVerifiedJobEvidence(_AbsentWhenNone):
+    """DAH-3386: what the check that cleared the verified job saw. The backend copies it into the penalty row
+    (lium-platform DAH-3385 `details.evidence.validator`), so the keys are named here and on that side alike;
+    `extra="allow"` keeps a key a newer check adds on the wire instead of dropping the reset."""
+
+    pod_id: str | None = None
+    container_name: str | None = None
+    container: ResetVerifiedJobContainerEvidence | None = None
+
+
 class ResetVerifiedJobRequest(BaseValidatorRequest):
     message_type: RequestType = RequestType.ResetVerifiedJobRequest
     validator_hotkey: str
@@ -151,7 +187,7 @@ class ResetVerifiedJobRequest(BaseValidatorRequest):
     # into the penalty it raises (lium-platform DAH-3385 details.evidence.validator). Optional both ways.
     reason_code: str | None = None
     check_id: str | None = None
-    evidence: dict | None = None
+    evidence: ResetVerifiedJobEvidence | None = None
 
 
 class DuplicateExecutorsRequest(BaseValidatorRequest):
--- a/neurons/validators/src/services/task/pipeline.py
+++ b/neurons/validators/src/services/task/pipeline.py
@@ -208,7 +208,7 @@ class LoggerSink:
         getattr(self.logger, level)(_m(event.event, extra=event.model_dump(mode="json")))
 
 
-def clear_evidence_for(res: "CheckResult", check_id: str) -> dict[str, Any]:
+def clear_evidence_for(res: CheckResult, check_id: str) -> dict[str, Any]:
     """The check's updates, with ``clear_verified_job_evidence`` filled when the check clears the verified job.
 
     DAH-3386: every check that sets ``clear_verified_job_info`` names itself to the backend — its event's
--- a/neurons/validators/src/services/task/pipeline.py
+++ b/neurons/validators/src/services/task/pipeline.py
@@ -208,7 +208,7 @@ class LoggerSink:
         getattr(self.logger, level)(_m(event.event, extra=event.model_dump(mode="json")))
 
 
-def clear_evidence_for(res: CheckResult, check_id: str) -> dict[str, Any]:
+def updates_with_clear_verified_job_evidence(res: CheckResult, check_id: str) -> dict[str, Any]:
     """The check's updates, with ``clear_verified_job_evidence`` filled when the check clears the verified job.
 
     DAH-3386: every check that sets ``clear_verified_job_info`` names itself to the backend — its event's
@@ -287,7 +287,7 @@ class Pipeline:
             events.append(res.event)
 
             if res.updates:
-                current_ctx = current_ctx.model_copy(update=clear_evidence_for(res, chk.check_id))
+                current_ctx = current_ctx.model_copy(update=updates_with_clear_verified_job_evidence(res, chk.check_id))
 
             if failed:
                 return False, events, current_ctx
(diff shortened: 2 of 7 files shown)


neurons/validators/src/services/task/pipeline.py
```diff
@@ -209,6 +214,24 @@ async def emit(self, event: ValidationEvent) -> None:
 STEP_SUMMARY_MIN_MS = 1000
 
 
+def clear_evidence_for(res: CheckResult, check_id: str) -> dict[str, Any]:
```
taiberium: **NIT:** `clear_evidence_for` reads as a verb that erases evidence. Rename it to `updates_with_clear_verified_job_evidence`, after the field it fills.

neurons/validators/src/protocol/vc_protocol/validator_requests.py
```diff
@@ -146,6 +146,12 @@ class ResetVerifiedJobRequest(BaseValidatorRequest):
     miner_hotkey: str
     executor_uuid: str
     reason: ResetVerifiedJobReason = ResetVerifiedJobReason.DEFAULT
+    # DAH-3386: the check that cleared the verified job (its event reason_code and check_id) and what it saw —
+    # for POD_NOT_RUNNING the container's status, exit code, OOM flag and finish time. The backend writes them
+    # into the penalty it raises (lium-platform DAH-3385 details.evidence.validator). Optional both ways.
+    reason_code: str | None = None
+    check_id: str | None = None
+    evidence: dict | None = None
```
taiberium: **NIT:** `evidence: dict` hides the shape of a payload that crosses two repos. Give it a small model with named fields.

taiberium (approved):

taiberium (commented):
Second pass. No new Redis call is added — the change only widens the payload of a publish that already ran on this path — so Redis-down behaviour is unchanged, and `clear_evidence_for` never sets `clear_verified_job_info` itself, so no healthy node can lose its verified job. All seven checks that clear the job are fatal, so a later check cannot overwrite the evidence. Nothing blocks.

NIT: `src/clients/compute_client.py:392` — nothing tests this branch. Delete the three new `data.get(...)` kwargs and all five new tests plus the existing 43 stay green while every reset reaches the backend unnamed again. One test that feeds a published payload and asserts the queued request closes it.

NIT: `src/services/redis_service.py:445` — the two `evidence.pop(...)` calls mutate the dict the next line reads as `evidence or None`, so correctness rests on dict-literal evaluation order. Pop into two named variables before the publish.

NIT: body, "Ran it" — the `2141 passed` run is from the commit before the head. Re-run at head or say which commit it belongs to.

neurons/validators/src/services/task/pipeline.py
surcyf123: Done in `4eff09c`: renamed to `updates_with_clear_verified_job_evidence`, definition, call site and tests.

neurons/validators/src/protocol/vc_protocol/validator_requests.py
surcyf123: Done in `4eff09c`: `evidence` is a `ResetVerifiedJobEvidence` model: pod_id, container_name and a container model with the eight bounded fields. Extras are allowed so a newer check never drops the reset. Unanswered fields stay absent on the wire, so the JSON is the redis payload unchanged.

PR #1356, opened 2026-09-11
DAH-3019 - [P1] the run summary lists every step, not only the slow ones

taiberium:
The provider's node page (DAH-3019) lists the per-step summary the validator writes on the last event. That summary dropped every check under a second, so a finished run reads as "three checks ran" when two dozen did.

Now every step of the run is listed. A step under a second carries three decimals, so a 4 ms check does not print as `0.0`; the portal shows those as "under 1 s".

Cost: the final event grows from ~300 bytes to ~900. At the current prod rate (~14.6k verdict rows per 6 hours) that is about 9 MB a day more in Loki and `prod_executors`.

Tests: `tests/test_pipeline_steps_summary.py` updated, validator suite green (1257 passed).

### DIFF
--- a/neurons/validators/src/services/task/pipeline.py
+++ b/neurons/validators/src/services/task/pipeline.py
@@ -203,12 +203,6 @@ class LoggerSink:
         getattr(self.logger, level)(_m(event.event, extra=event.model_dump(mode="json")))
 
 
-# DAH-3012: steps shorter than this are left out of the final event's summary — they are the
-# ~25 sub-second bookkeeping checks; the 6–8 that carry the run's time are what a provider (and a
-# Loki query) needs to see, in ~300 bytes.
-STEP_SUMMARY_MIN_MS = 1000
-
-
 def summarize_steps(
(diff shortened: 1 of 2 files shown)


surcyf123 (approved):
Approved (loop review, requested by Rustam 06:18Z). The change does what the body says: every executed step lands in the final event, sub-second steps keep three decimals, the slow ones one. Read against `summarize_steps` callers — nothing downstream keys on `STEP_SUMMARY_MIN_MS`, and the portal's "under 1 s" rendering is the right place for the rounding.

One non-blocking NIT: the body's motivating case — "a 4 ms check does not print as 0.0" — is not in the tests; the fixtures only carry 0 ms and ≥ 200 ms sub-second steps, so a regression to `round(ms / 1000, 1)` would still pass `test_passing_run_lists_every_step_and_the_total_on_the_last_event`. One `_TimedCheck("x", 0.004, clock)` asserting `0.004` would pin it. Fine to do in a follow-up.

Cost line (+9 MB/day in Loki and `prod_executors`) acknowledged; well under the retention budget noted in I-42.

PR #1358, opened 2026-09-11
DAH-3405 - [P0] validator: no verdict for nodes unreachable in a known image rollout

surcyf123:
Implements [DAH-3405](https://app.notion.com/p/3d8b8bfdbde981ac8c86cf027caa65d8) · reviewer: @taiberium

**TL;DR** — 11 Sep 07:50Z executor-v1.126 went live mid-cycle; watchtower restarts cut the validator's SSH; one cycle read 168 unreachable, 51 scrape-failed, 50 insufficient-ports, 70 OUTDATED rows (tsdb 506/514 at 0). The validator now records the cycle the digest changed in and, for that cycle and the next, withholds those verdicts for executors not yet on the new image: neither scored nor published.

**What changed**
- `ExecutorRolloutTracker`: digest, last change, cycles since (Redis key `executor_image_rollout`; `services/executor_rollout.py`)
- `withhold_rollout_verdicts`: score-0 results ending unreachable / upload or scrape failed / insufficient ports (digest ≠ new) or OUTDATED leave the cycle; one `outcome=ROLLOUT_GRACE` log line each
- digest read at cycle start and end; withheld results leave `all_job_results` before incentive and publish (`core/validator.py`)
- `JobResult.failure_reason_code`: the event a run without a score ended on (`services/task/models.py`, `service.py`)
- `EXECUTOR_ROLLOUT_GRACE_CYCLES` default 2, capped at 2 (backend 1-h sweep), 0 = off; OUTDATED counts only with `EXECUTOR_IMAGE_CHECK_ENFORCE` on (`core/config.py`)
- 25 tests (28 cases), each naming its regression (`tests/test_executor_rollout_grace.py`)

**Risks**

**How to verify**
- `cd neurons/validators && pdm run pytest tests/test_executor_rollout_grace.py -q` → 28 passed
- log after a push: `[rollout-grace]` detected → `verdict withheld` per executor → `window ended`

**Verified by the loop:** Gate: PASS a0d7821

**Ran it:** `pdm run pytest tests -q` → 2319 passed, 2 env failures also on main · warm EC2 20260912T142315Z-pjav, this head · no live registry/Redis

**Self-review:** clean at a0d7821 (15 checks, 3 low)

**Parts:** backend ✓ · migration n/a — no schema change · frontend n/a — validator only · CLI/SDK n/a — no command or SDK surface · docs ✓ `.env.template` · tests ✓ 25 · dashboards n/a — Loki `outcome=ROLLOUT_GRACE`

**Docs:** none needed because the only new surface is validator operator config (`EXECUTOR_ROLLOUT_GRACE_CYCLES`, `neurons/validators/.env.template`); a withheld result publishes nothing to providers

### Numbers and sources

- Rustam's validator cycle log, 11 Sep 2026 07:50–08:00Z (Notion comment 09:39Z): 168 transport-unreachable, 51 scrape-failed, 50 insufficient-ports rows, and 70 `EXECUTOR_IMAGE_OUTDATED` rows against a normal background of 31 (units: log rows in one cycle; they are not summed here because a row is not a distinct executor). The orchestrator's count of zeroed executors in the same log: 239 of 476 (distinct executors, one cycle).
- tsdb `prod_executors` (ticket body): 506 of 514 distinct executors at 0 in that cycle vs 161 of 517 the cycle before (executors per cycle).
- Cause (Rustam, 09:39Z): watchtower recreated the executor container mid-cycle when the new image went live, killing the validator's SSH sessions; OUTDATED rows were the fleet not yet pulled (or the validator's snapshot predating the push). No incentive was lost (09:21Z weights unchanged, sum 5.0000) — the damage was the published cycle score (Grafana dip, 7 inactive marks, 4 `EXECUTOR_INACTIVE_MID_RENTAL` penalties = $113.61, reverts with Mikhail). The money path is lium-platform#325 (DAH-3385), out of scope here.
- Ceiling for the window: lium-io-backend `apps/server/src/worker.py` `check_and_update_executors` marks an executor inactive when its row was not updated for 1 h and applies `EXECUTOR_INACTIVE_MID_RENTAL`. A withheld executor publishes nothing, so at most two consecutive cycles may be withheld: the gap between its publishes is then three cycle periods (job blocks 900 s apart) plus the duration of the next scored cycle minus the duration of the last one — about 2700 + 850 s (`JOB_TIME_OUT − 50`) = 3550 s when cycles start 900 s apart — a lower bound, not a ceiling: a cycle that runs past 900 s delays the next start (`sync` waits for 75 blocks since the previous start), and the row lands after incentive, estimates and the publish loop. A third withheld cycle is past the hour whenever the next scored cycle is not shorter than the last; two are under it only when cycles keep their 15-min cadence. `MAX_ROLLOUT_GRACE_CYCLES = 2` in `executor_rollout.py` caps the setting and logs the cap. Measuring the real publish cadence from `prod_executors.updated_at` and keying the backend sweep to validator cycles (or 75 min) is the durable fix — follow-up below.

### Design

- Source of truth for "a rollout is happening": the registry digest of `EXECUTOR_IMAGE_REF` (`fetch_executor_image_digest`), the same tag watchtower pulls (`neurons/executor/docker-compose.yml`: `nickfedor/watchtower` with `--interval 60 --cleanup --label-enable`). The validator already fetched it once per cycle for `ExecutorImageCheck`; it is now fetched a second time at cycle end, because the 07:50Z push landed mid-cycle and a start-of-cycle read alone would have opened the window one cycle late.
- Decision point is the cycle (`Validator.sync`), not the pipeline: the pipeline cannot know a digest that changed after it started. `withhold_rollout_verdicts` runs after every job returned and before `IncentiveFactory.create`, so withheld results are in neither the incentive nor `publish_machine_specs` nor `silence_availability_errors_on_our_own_outage`. Their miners stay in `active_hotkeys` (DAH-2622: the u16=1 floor that keeps a UID); their ids are added to the express lane's validated set so it does not start a first pass on them.
- The window is counted in cycles as the validator observes them (`cycles_seen`, incremented when `observe` is called with a job block it has not seen; the cycle-start and cycle-end observations of one cycle share the job block), never by the clock or by job-block arithmetic: `sync` starts a cycle on the first 12-s tick at ≥ 75 blocks since the last, so a late-starting long cycle makes the next one land two job blocks later — it is still the next cycle. `RolloutWindow.covers(job_block)` is true while `cycles_seen ≤ 2`. A validator restart inside the window resumes it from Redis.
- Classifier (`rollout_grace_reason`): score > 0 ⇒ stands. Reason ∈ {`EXECUTOR_SSH_UNREACHABLE`, `UPLOAD_FAILED`, `EXECUTOR_TRANSPORT_UNREACHABLE`, `FILLER_TRANSPORT_UNREACHABLE`, `SCRAPE_FAILED`, `INSUFFICIENT_PORTS`, `EXECUTOR_IMAGE_OUTDATED`}, or the run ended without failing (the rented halt `RENTED`, or `VALIDATION_COMPLETED`) with score 0 and the image report says OUTDATED (a rented executor's image check passes, so `TenantEnforcementCheck` halts its run as RENTED with score 0 instead of failing) ⇒ candidate; a rented executor that failed a later check of its own (pod not running, filler killed) stands, OUTDATED or not. For every reason but OUTDATED, an observed digest (`executor_image.observed_executor_digest` over the scraped specs, or the report) equal to the new digest ⇒ a real failure on the rolled-out image ⇒ stands. OUTDATED is withheld regardless of the observed digest: old = not pulled yet, new = the validator's snapshot predated the push.
- `failure_reason_code` is set by `TaskService.create_task`: the last event's reason when the run did not succeed or ended with score 0 (the rented halt included); `EXECUTOR_SSH_UNREACHABLE` when the connect failed; `EXECUTOR_TRANSPORT_UNREACHABLE` when the shell had opened and an `asyncssh.Error`/`OSError` escaped a check (the recreate landing under a `ctx.ssh.run` that a check does not catch) — the name the rented-machine check gives the same death. Any other exception leaves it None and the result stands.
- Redis state (`executor_image_rollout`, JSON: `digest`, `previous_digest`, `started_at`, `opened_job_block`, `last_job_block`, `cycles_seen`, `withheld`, `closed`) survives the validator redeploy that usually accompanies a release. A second digest change while the window is open, or before one uncovered cycle has published (`cycles_seen < grace_cycles + 2`), makes the new digest current but does not reopen the window — withheld cycles would otherwise chain to three or four in a row; that second push gets no grace, and the log says so. A Redis error at cycle start leaves the cycle as it is today (every failure a verdict) and logs `[rollout-grace] rollout state unreadable; every verdict stands this cycle`; a Redis error at cycle end keeps the window seen at cycle start; a value that is not JSON is replaced. A registry error (`fetch_registry_digest` returns None on any HTTP error) changes nothing: a window already open stays open, no window is opened. The first digest a fresh Redis sees opens no window. `EXECUTOR_ROLLOUT_GRACE_CYCLES=0`: the digest is still tracked, no cycle is covered, no window lines are logged.
- Incentive effect: a withheld executor is absent from the cycle, so its miner's mining score for that cycle is what a 0 would have given (nothing) — no scoring formula changes. Accepting the previous digest as CURRENT for the window (ticket option a) would score those nodes normally; that is a scoring change and needs a design note first (PR_PROCESS §1b).

### Known follow-ups

- Watchtower delaying the recreate until the validator's job on the node has finished (brief part 5): the standard stack runs upstream `nickfedor/watchtower`, so the mechanism is a lifecycle pre-update hook in the executor image (label `com.centurylinklabs.watchtower.lifecycle.pre-update`, exit 75 skips this round) that returns non-zero while a validator SSH session / uploaded key is present — an executor image + compose change on every provider, not small; separate PR on this ticket. The in-repo `watchtower/` (CVM `executor-runner`) would take the same check before `container.stop`.
- Pushing `:latest` right after a cycle ends is a release-process rule, not code.
- A provider-facing note on release behaviour ("during an executor image release your node keeps its last cycle's score for up to two cycles") on lium-docs `providers/troubleshooting.md` — on this ticket, if Rustam wants providers told.
- A withheld result's `availability_errors` (DAH-2748) leave the cycle with it: a node that really died during the window is not hidden from the market until its next scored cycle (one or two cycles later). Publishing only the availability part of a withheld result would need the backend to accept a row without a score — follow-up on this ticket if Rustam wants it.
- `RedisService.set_verified_job_info` still counts a withheld run as `failed` (written inside `TaskService.create_task`, before the cycle-level decision). The counter is read by nothing that scores — the branch that acted on it is commented out.
- The second registry read at cycle end is a token GET plus a manifest HEAD against Docker Hub, each with a 30-s timeout (`fetch_registry_digest`), on the path between the last job and the weight computation; an unresponsive Docker Hub costs the cycle up to 60 s there.
- The backend's 1-h sweep vs a 2-cycle window leaves under a minute of margin (numbers above); keying `check_and_update_executors` to validator cycles, or 75 min, is a lium-io-backend follow-up on this ticket.
- The express lane (`core/express_lane.py`) is unchanged: a never-validated executor whose first pass fails during a rollout is still published with score 0 (there is no previous row to keep); the next scored cycle overwrites it.
- No test drives `Validator.sync` end to end (it needs a subtensor, backend, Redis and miner service); the wiring is the `withhold_rollout_verdicts(` call in `core/validator.py` — the rebinding of `all_job_results` before `IncentiveFactory.create` — and the fail-open helpers `observe_executor_rollout` / `record_withheld_verdicts` are tested directly.
- Self-review round 8 finding, fixed in this head: a rented executor's run halts in `TenantEnforcementCheck` with success=True and score 0 (reason `RENTED`), so the reason is now recorded for every score-0 run and the classifier accepts `RENTED` as well as `VALIDATION_COMPLETED`; the rented case is driven through `TaskService.create_task` with the real halt event.
- Self-review round 7 findings, fixed in this head: the OUTDATED fallback applies only to a run that completed (`VALIDATION_COMPLETED`), so a rented executor's own later failure stands; the cycle-end block is `Validator.withhold_verdicts_for_rollout` and tested with the registry stubbed OLD → NEW inside one job block; the tracker normalises the registry digest the way the image check normalises the node's; any `OSError` after the connect is named `EXECUTOR_TRANSPORT_UNREACHABLE` on purpose (stated in `service.py`).
- Self-review round 6 findings, fixed in this head: the grace-0 log guard; the `MAX_ROLLOUT_GRACE_CYCLES` comment now calls 3550 s the lower bound it is; a Redis error at the cycle-end observation falls back to the window seen at cycle start; the cycle-start observation moved below the `rented_executors is None` abort so an aborted iteration is not a cycle of the window; a test for the headline mid-cycle push (OLD at cycle start, NEW at cycle end of the same job block). Open lium-io#1351 adds `JobResult` fields in the adjacent hunk (merge order: either first, trivial conflict).
- lium-platform#325 (DAH-3385, confirm an automatic penalty before charging it): the money path for the first cycle after a push.

### Ran it

EC2 sandbox, warm lium-io box i-0551cd50560a42a0d (c6i.2xlarge, Linux), this head:

```
$ pdm run pytest tests -q                                       # run 20260911T115100Z-jzd0, this head
2 failed, 2240 passed, 15 skipped, 165 warnings in 121.33s (0:02:01)
```

The two failures are the box's (sysfs readable as root; an sshd already running) and appear identically in main runs 20260911T084014Z-pldc, 20260911T070308Z-h5fd, 20260910T172331Z-qqh0. Not run: a live validator against Docker Hub and Redis.

### Self-review

Verdict `clean` at `a0d7821` · 15 checks · by subagent-drain-r125b · 2026-09-12 14:34Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | PROSE-VS-CODE | `RECOVERY/workers/rollout-grace/body.md:43` | Property: the body's Classifier bullet must state the same rule as rollout_grace_reason. The bullet lists the OUTDATED-without-failing arm ('the run ended without failing ... with score 0 and the image report says OUTDATED => candidate') with no mention of EXECUTOR_IMAGE_CHECK_ENFORCE, while executor_rollout.py:386 fires that arm only when the flag is on; the only body sentence that states the condition is the 'What changed' bullet, and it points at core/config.py, where nothing about the rule lives (the flag's comment at config.py:336-340 says nothing about rollout grace). | In the Classifier bullet add 'the OUTDATED-without-failing arm counts only while EXECUTOR_IMAGE_CHECK_ENFORCE is on (off by default since DAH-3439/#1360); off, an OUTDATED node scores normally and a rented 0 under an OUTDATED report stands', and in the 'What changed' bullet point that clause at services/executor_rollout.py rather than core/config.py. |
| low | CODE-QUALITY | `neurons/validators/src/services/task/service.py:222` | Property: a comment that names a condition must name the whole condition. '(success=True, score 0 when the image is OUTDATED)' describes the rented halt, but score_calculator.py:44-50 zeroes the score for OUTDATED only while EXECUTOR_IMAGE_CHECK_ENFORCE is on; with the flag off (main's default) the rented halt carries the run's normal score and this line does not record a reason for it. | Say '(success=True, score 0 when the image is OUTDATED and EXECUTOR_IMAGE_CHECK_ENFORCE is on, or when another gate zeroed the score)'. |
| low | STALE-BODY | `RECOVERY/workers/rollout-grace/body.md:69` | Property: every run id and count in the body must come from one run at the head it describes. The top-level 'Ran it' line renders as run 20260912T142315Z-pjav / 2319 passed (matches aws_run/20260912T142315Z-pjav/job.log, checked out at a0d7821), while the Details 'Ran it' block (body.md:67-75) still says run 20260911T115100Z-jzd0 / 2240 passed 'this head' and lists the two failures twice; the Details 'Self-review' block cites e4335e9 and 1b0831e, neither reachable from a0d7821 (git merge-base --is-ancestor fails for both), and 'Verified' says 'not run at 80a81e4'. | Replace the Details 'Ran it' block with the pjav run (2 failed, 2319 passed, 15 skipped, 128.11 s; same two env failures as on main) and let tools/pr_self_review.py --record --apply replace the Self-review/Verified blocks so only reachable shas remain. |

Mechanical notes dismissed: `neurons/validators/tests/test_executor_rollout_grace.py:17` Mechanical note: datura.requests.miner_requests resolves. neurons/validators/pyproject.toml:19 installs the sibling package `datura @ file:///${PROJECT_ROOT}/../../datura`, ../../datura/datura/requests/miner_requests.py exists in the worktree, tests/conftest.py:8 imports the same name today, and the EC2 run at this head collected and passed the module (2319 passed). · `neurons/validators/src/services/executor_rollout.py:386` The r125 medium is fixed: `settings` is imported at executor_rollout.py:28 (from core.config import settings, already used for EXECUTOR_ROLLOUT_GRACE_CYCLES), the fallback reads settings.EXECUTOR_IMAGE_CHECK_ENFORCE at classification time, and monkeypatching core.config.settings in the tests reaches the same object. The new test fails without the fix: on 80a81e4 the same input (reason RENTED, report status OUTDATED, score 0) sets ended_without_failing=True and outdated=True, relabels the reason EXECUTOR_IMAGE_OUTDATED and returns it, which is not None. · `neurons/validators/src/services/executor_rollout.py:46` ROLLOUT_FAILURE_REASONS keeps EXECUTOR_IMAGE_OUTDATED without a flag check, but with EXECUTOR_IMAGE_CHECK_ENFORCE off that reason can never be a run's failure_reason_code: checks/executor_image.py:84-98 passes the check with OUTDATED_WARNING, Pipeline.run (pipeline.py:295-301) continues past a passing check, and FinalizeCheck / TenantEnforcementCheck's halt supply the last event, so service.py:224 never records it; the except-block at service.py:227-250 records None or EXECUTOR_TRANSPORT_UNREACHABLE only. Nothing in core/validator.py mentions OUTDATED, and incentive/default.py:110-116 already gates its own OUTDATED exclusion on the flag. · `neurons/validators/tests/test_executor_rollout_grace.py:374` TEST-GAP: the two tests that set the flag True encode the enforced 11 Sep behaviour and their docstrings say so; the unrented arm (reason EXECUTOR_IMAGE_OUTDATED in the set) does not depend on the flag, the rented arm and the create_task-driven rented halt (tests:571) do, and each assertion still fails against a classifier that drops the fallback. 25 test functions plus the 4-way parametrize at tests:359 give the 28 cases the body claims; the local sibling venv is empty, so the run rests on the EC2 log at this head. · `neurons/validators/src/core/validator.py:456` total_gpu_model_count_map is built before withhold_verdicts_for_rollout (validator.py:529), but a withheld unrented result has score 0 and job_score 0 and never enters the map, and a withheld rented result is counted exactly as today's code counts it, so the mining denominator does not move. · `neurons/validators/src/core/validator.py:826` PERF: the cycle-end Docker Hub read (token GET + manifest HEAD, 30-s timeouts) runs even with EXECUTOR_ROLLOUT_GRACE_CYCLES=0 and sits between the last job and the weight computation, not on a request path; it is fail-open (None leaves the window as it was) and the body lists the up-to-60-s cost as a follow-up. · `neurons/validators/src/services/executor_rollout.py:108` SECURITY/CORRECTNESS on the Redis value: from_json is wrapped in except (ValueError, TypeError) at _load, which covers json.loads, int(), datetime.fromisoformat and a non-dict payload, and the corrupt-value test drives both a non-JSON string and a list; the key is written only by this validator. · `neurons/validators/src/services/task/models.py:86` CROSS-PR: open lium-io#1351 (base DAH-2991-reap-unkillable-orphan-container, not main) touches models.py, pipeline.py and result_handler.py but not executor_rollout.py or service.py's except-block; the body states the merge order and the models.py hunk is a field-list merge. #1358 is based on main (merge-base a120dda == GitHub main tip) and GitHub reports MERGEABLE.

### Verified

Gate: PASS (a0d7821) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1358)

**Why Redis, not a stateless tag_last_pushed window (taiberium NIT, 13:29Z):** the window is counted in validator cycles, not seconds, so it must outlive a validator restart in the middle of a rollout: with a stateless comparison a restart at cycle 1 would re-read the registry, see the digest as already current, and score the still-unreachable nodes 0 in cycle 2. The two Redis keys (observed digest, cycles remaining) are the only state; they expire on their own and a flush just means one more grace cycle, never a lost score.

</details>

### DIFF
--- a/neurons/validators/src/services/executor_rollout.py
+++ b/neurons/validators/src/services/executor_rollout.py
@@ -22,7 +22,7 @@ from __future__ import annotations
 
 import json
 import logging
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from datetime import UTC, datetime
 
 from core.config import settings
@@ -85,6 +85,68 @@ _RUN_ENDED_WITHOUT_FAILING = frozenset(
 
 
 @dataclass(frozen=True)
+class RolloutState:
+    """What Redis remembers between cycles, under `ROLLOUT_STATE_KEY`, as one JSON object.
+
+    `opened_job_block` stays None until a digest change opens a window. `cycles_seen` counts the
+    cycles observed since that change, the opening one included, and `last_job_block` is the cycle
+    the last count was made for. `closed` records that the window-end line was logged, so it is
+    logged once.
+    """
+
+    digest: str | None = None
+    previous_digest: str | None = None
+    started_at: datetime | None = None
+    opened_job_block: int | None = None
+    last_job_block: int | None = None
+    cycles_seen: int = 0
+    withheld: int = 0
+    closed: bool = False
+
+    @classmethod
+    def from_json(cls, raw: str | bytes) -> RolloutState:
+        """Raises ValueError or TypeError on anything this class did not write."""
+        fields = json.loads(raw)
+        if not isinstance(fields, dict):
+            raise ValueError("rollout state is not a JSON object")
+        started_at = fields.get("started_at")
+        opened_job_block = fields.get("opened_job_block")
+        last_job_block = fields.get("last_job_block")
+        return cls(
+            digest=fields.get("digest"),
+            previous_digest=fields.get("previous_digest"),
+            started_at=datetime.fromisoformat(started_at) if started_at else None,
+            opened_job_block=int(opened_job_block) if opened_job_block is not None else None,
+            last_job_block=int(last_job_block) if last_job_block is not None else None,
+            cycles_seen=int(fields.get("cycles_seen", 0)),
+            withheld=int(fields.get("withheld", 0)),
+            closed=bool(fields.get("closed", False)),
+        )
+
+    def to_json(self) -> str:
+        return json.dumps(
+            {
+                "digest": self.digest,
+                "previous_digest": self.previous_digest,
+                "started_at": self.started_at.isoformat() if self.started_at else None,
+                "opened_job_block": self.opened_job_block,
+                "last_job_block": self.last_job_block,
+                "cycles_seen": self.cycles_seen,
+                "withheld": self.withheld,
+                "closed": self.closed,
+            }
+        )
+
+
+@dataclass(frozen=True)
+class WithheldVerdict:
+    """One result the rollout took out of the cycle, with the miner that reported it."""
+
+    miner_hotkey: str
+    result: JobResult
+
+
+@dataclass(frozen=True)
 class RolloutWindow:
     """What the validator knows about the current executor image and its last change.
 
@@ -121,25 +183,24 @@ class RolloutWindow:
 class ExecutorRolloutTracker:
     """Remembers the authorized executor digest, and when and in which cycle it last changed.
 
-    State lives in Redis under `ROLLOUT_STATE_KEY` as one JSON object: `digest`, `previous_digest`,
-    `started_at` (ISO-8601), `opened_job_block`, `last_job_block` and `cycles_seen` (the cycles
-    observed since the change, the opening one included), `withheld` (results withheld so far in
-    this window) and `closed` (the window-end line was logged). A Redis error propagates; the caller decides what
-    a cycle without this knowledge does. An unreadable value is treated as no state and re-seeded.
+    State lives in Redis as a `RolloutState`. A Redis error propagates; the caller decides what a
+    cycle without this knowledge does. An unreadable value is treated as no state and re-seeded.
     With `grace_cycles` 0 the digest is still tracked but no window ever covers a cycle and neither
     window line is logged.
     """
 
     def __init__(self, redis_service: RedisService, grace_cycles: int | None = None):
         self.redis_service = redis_service
-        wanted = settings.EXECUTOR_ROLLOUT_GRACE_CYCLES if grace_cycles is None else grace_cycles
-        self.grace_cycles = min(wanted, MAX_ROLLOUT_GRACE_CYCLES)
-        if wanted > MAX_ROLLOUT_GRACE_CYCLES:
+        configured_grace_cycles = (
+            settings.EXECUTOR_ROLLOUT_GRACE_CYCLES if grace_cycles is None else grace_cycles
+        )
+        self.grace_cycles = min(configured_grace_cycles, MAX_ROLLOUT_GRACE_CYCLES)
+        if configured_grace_cycles > MAX_ROLLOUT_GRACE_CYCLES:
             logger.warning(
                 _m(
                     "[rollout-grace] EXECUTOR_ROLLOUT_GRACE_CYCLES capped: more cycles without a "
                     "published row would trip the backend's 1-hour inactive sweep",
-                    extra={"wanted": wanted, "used": self.grace_cycles},
+                    extra={"wanted": configured_grace_cycles, "used": self.grace_cycles},
                 )
             )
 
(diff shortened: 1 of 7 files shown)


neurons/validators/src/services/executor_rollout.py
```diff
@@ -0,0 +1,434 @@
...
+    PortCountMessages,
+    RentalVerificationMessages,
+    TenantEnforcementMessages,
+    UploadFilesMessages,
+)
+from services.task.models import JobResult
+
+logger = logging.getLogger(__name__)
+
+ROLLOUT_STATE_KEY = "executor_image_rollout"
```
taiberium, line 47: **NIT:** The ticket picked a stateless window (`tag_last_pushed`, nothing to survive a restart). Say in the body why this keeps state in Redis instead.

taiberium (approved):

neurons/validators/src/services/executor_rollout.py
surcyf123: Done in the body, paragraph *Why Redis, not a stateless window*. The window counts validator cycles, so it must survive a validator restart mid-rollout. With a stateless comparison, a restart at cycle 1 would see the digest as current and score the still-unreachable nodes 0 in cycle 2. The two keys are the only state and expire on their own.

PR #1351, opened 2026-09-10
DAH-3338 - [P1] pod container states to the backend; unreachable is not an unknown id

surcyf123:
Implements [DAH-3338](https://app.notion.com/p/3d7b8bfdbde981d3be15f8f4fffe2b81) · reviewer: @taiberium

**Problem** — The backend cannot tell a dead host from an unknown executor id, so it neither closes nor credits (48 rent + 160 delete failures in 7 d).

**Value** — The backend can close and credit the 48 rent + 160 delete failures per 7 d it now files as 'Invalid executor id'; prod 16 Sep: 122 rentals of 42 renters on 22 nodes closed `executor_offline` in 7 d (1,497 rental-hours).

**Goal** — [DAH-3338](https://app.notion.com/p/3d7b8bfdbde981d3be15f8f4fffe2b81) (P1).

**Fits** — Sits in `lium-io` — `neurons/validators`, `lium_protocol/lium_protocol`, `neurons/miners`…; touches docs (lium_protocol/README.md), tests (47 new).

**Flag:** `POD_STATES_REPORT_ENABLED`, off by default, turned on by @taiberium on 2026-09-23 (proposed by the loop; change it in one reply and the body follows)

**Risks**
- No migration, no restart, no data change. Rebased onto #1361 (keep-both in `pipeline.py`) and onto #1331: main's compat test now requires the new wire types in `lium_protocol`.
- Merge order: deploy lium-platform#312 first. Its `FailedContainerErrorCodes` is strict, so a backend without `ExecutorUnreachable` drops that failure message and the rent or delete waits out its timeout.
- `POD_STATES_REPORT_ENABLED` defaults off; turn it on after lium-platform#312 deploys (an older backend logs and drops each report).

**How to verify**
- `cd neurons/validators && pytest -q tests/test_pod_states_payload.py tests/test_stale_container_cleanup_check.py tests/test_protocol_compat.py` → green
- `cd neurons/miners && pytest -q tests/test_executor_service.py` → 4 passed
- `cd lium_protocol && pip install . pytest 'pydantic==2.13.*' && python -m lium_protocol.schema --check && pytest -q tests` → `schema ok … (protocol 1.1.0)`, 83 passed

**Verified by the loop:** EC2 run 20260916T093936Z-23g7 at `18b49f8` (rebased onto #1383's main): validators 2676 passed + 2 root-sensitive, executor 296, miners 45; all rc=0

[...]

PR #1361, opened 2026-09-12
DAH-3436 - [P1] validator: synthetic rental probe delists a node a renter cannot use

surcyf123:
Implements [DAH-3436](https://app.notion.com/p/3d8b8bfdbde9812d9b8af4e3334a15e8) · reviewer: @taiberium

**TL;DR** — Node 939b3e38 broke 6 renter pods on 11 Sep while the validator scored it 1.0 each cycle (DAH-3432). New check `executor.validate.rental_probe` rents the idle node through `create_container`, waits for sshd's banner, logs in, runs `nvidia-smi -L` and tears down; a failed step is score 0 plus a cleared verified job. Off by default; one open question for Rustam in Details.

**What changed**
- `RentalProbeCheck` (`rental_probe.py`), fatal, last before `ScoreCheck`; steps `container_start`, `sshd_listen`, `ssh_login`, `gpu_count`, `teardown`
- review NITs: the may-we-probe gate is `RentalProbeCheck._may_probe`, the verdict stays in `run`; `_probe` is orchestration over `_step_container_start`, `_step_sshd_listen`, `_step_ssh_login`, `_step_gpu_count`, `_step_teardown` (no behaviour change)
- `sshd_listen` needs the `SSH-2.0` banner, not a TCP accept; `ssh_login` retries until the deadline
- failure: score 0, verified job cleared; the failed step stands in Redis, no lifetime, until a probe passes (`_stamp_failure`, review)
- per-executor create lock: renters hold it around `create_container`, the probe skips when held (review)
- skips (`_skip_reason`, `_busy_now`): rented, filler, image not pulled, within the interval, a renter's create in progress

**Risks**
- Changes runtime configuration (`config.py`) on the next deploy, no pod restart — rollback: revert the file and redeploy.
- Every renter create takes a per-executor Redis lock (`miner_service.py`): behind a probe's create it waits up to 360 s, never fails on the lock — rollback: revert `miner_service.py`.

**How to verify**
- `cd neurons/validators && pdm run pytest tests/test_rental_probe_check.py tests/test_miner_service.py -q` → 54 passed
- staging, `RENTAL_PROBE_ENABLED=true`: closed SSH port → `RENTAL_PROBE_FAILED` at `sshd_listen`; the next skipped cycle carries `standing_failure`

**Verified by the loop:** Gate: PASS 1aaecd1

**Ran it:** `pdm run pytest tests -q` → 2355 passed, 2 known box env failures (infiniband, sshd bootstrap) · warm EC2 20260914T130325Z-u232 at 1aaecd1

**Self-review:** clean at 1aaecd1 (14 checks, 3 low)

**Parts:** backend ✓ · migration n/a — no schema change · frontend n/a — portal reads reason_code · CLI/SDK n/a — no surface · docs ✓ `.env.template` + lium-platform#366 · tests ✓ 54 (e2e: follow-up on this ticket) · dashboards n/a — Loki events only

**Docs:** `neurons/validators/.env.template`; provider page lium-platform#366 (same ticket and reviewer)

### Why the existing check did not catch it

`executor.validate.rental_verification` asks the backend for its rental-check status. That status is `RENTAL_CHECK_PENDING` for 455 of 476 verified executors (I-77, DAH-3382), so it gates nothing. The 15-minute cycle proves GPUs, bandwidth and the filler container; it never starts a renter container with the executor's port mapping and logs in. Node 939b3e38 (2x RTX 5090): 4 renter pods from 4 users and 2 loop test pods went BROKEN within 60 s (`RECOVERY/workers/node-redteam-939b/REPORT.md`), validator rows `VALIDATION_COMPLETED score 1` at 21:00, 21:15, 21:30Z.

### Design as built

- Placement: after `RentalVerificationCheck`, before `ScoreCheck`. It needs `verified_port_pairs` (PortConnectivityCheck), the GPU UUIDs (scrape), `sysbox_runtime`, and the rented/filler state (`rented_data`).
- The probe rents the node exactly as the backend does: `ContainerCreateRequest` with the default renter image for the node's GPU and driver (`backend.get_default_docker_image`, the same call CachedTemplateVerificationCheck makes), `user_public_keys=[probe key]` (a fresh ed25519 pair from `SSHService.generate_keypair`, thrown away after the run), `gpu_uuids` from the scrape, `is_sysbox` from the port check, `ships_sshd=True` (the default image starts its own sshd), `available_ports` = the verified pairs, `active_container_names=[]`. Then `DockerService.create_container(payload, executor, keypair, encrypted key)`: the same code path as a renter's pod, including the pre-run sweeps, the GPU device config, the 90 s port-retry, key injection and the rented-pod cache. Teardown is `DockerService.delete_container`. `PodRecoverer` declares both so the check reaches them through `ctx.services.pod_recovery`.
- The encrypted key: `create_container`/`delete_container` decrypt the key themselves (`ssh_service.decrypt_payload`). `TaskService.create_task` keeps the encrypted form it received and `build_context` stores it as `Context.executor_ssh_private_key_encrypted`. No re-encryption, no change to `ssh_service.py`.
- Steps and verdicts: `container_start` fails when `create_container` returns `FailedContainerRequest` with a `failure_step` at or after `ssh_connect`, or does not return within 300 s; a failure before that (`port_mapping`, `pending_pod`, `attestation`, `docker_sdk_ssh_host_key`, ...), a result without a failure step, a created container with no mapping for port 22, or a raise is the validator's and the verdict is `RENTAL_PROBE_INCONCLUSIVE` (pass, no penalty). `sshd_listen`: connect to `executor.address:<external port mapped to 22>` every 2 s until `RENTAL_PROBE_SSH_DEADLINE_SECONDS` and read the first line; only an `SSH-2.0` banner ends the wait. A bare TCP accept is not enough: docker-proxy accepts on a published port as soon as the container exists, before the image's start.sh has run `service ssh start`. `ssh_login`: `asyncssh.connect` as root with the probe key, retried with backoff (1, 2, 4, 8 s) until the same deadline; `PermissionDenied` is final (authorized_keys is written before sshd starts, so it cannot clear itself); the step records the attempt count; a connection that never opened is this step's failure. `gpu_count`: `nvidia-smi -L` over that connection (timed as the command; a session that opened but could not run the command is this step's failure), lines starting `GPU ` must equal `ctx.state.gpu_count`. `teardown`: `delete_container` (120 s deadline); when it fails, or when the create returned no `ContainerCreated` after reaching the host (failed after `docker_run` at `add_public_keys`, `set_environment` or `finalize`; cut off by its 300 s deadline; cancelled; raised), the container and its volume are removed by name (`docker rm -fv pod_<id>`, `docker volume rm volume_<id>`) over the validation shell and the rented-pod cache entry is dropped, so no verified port stays bound into the next cycle. Both results are read (review, 14 Sep): `DockerCommand.volume_remove` masks its exit with `|| true`, so the probe runs `volume_remove_strict` and reads exit status and stderr; "no such volume" is nothing left behind, any other non-zero exit is a leftover the teardown step reports. Only a create that failed before `ssh_connect` removes nothing. The run's outer timeout (`JOB_TIME_OUT`, a `CancelledError`) mid-create takes the same by-name path: the cleanup runs as its own task in the `finally`, awaited to the end through `asyncio.shield`; a cancel that lands during it is re-raised once it is done, so the run still ends as cancelled and no verdict is built from a half-settled outcome. A teardown that needed the fallback after a clean run is INCONCLUSIVE (no penalty, no interval stamp, probed again next cycle). A leftover the fallback could not remove is reaped by `StaleContainerCleanupCheck` (orphan `pod_*` older than 15 min, not in rented data).
- Failure mechanics: `updates={"score": 0.0, "job_score": 0.0, "score_warning": ..., "clear_verified_job_info": True, "clear_verified_job_evidence": {...}}`, the same shape as the GPU runtime quarantine in `rental_verification.py`. The pipeline fills nothing extra: `reason_code` and `check_id` are set by the check (DAH-3386). The event is the run's last event (fatal check), so it is what the backend stores as the executor's `log_text` and the portal shows as `last_validation`, with `failed_step` and the per-step timings in `what_we_saw.probe_steps` (`steps` on that event is the pipeline's own per-check duration summary).
- Idle now, not at cycle start: `ctx.state.rented_data` is the snapshot the cycle started from, minutes before this last check. Right before the create the check re-reads `backend.get_all_rented_executors()` and `redis.renting_in_progress(miner, executor)` (a rent this validator is creating at that moment); a pod, a filler or a pending rent skips the probe, and a read that fails reaches no verdict.
- Per-executor create lock (review, 14 Sep): a renter's create could still start after that re-read, and the probe's create (an empty `active_container_names`) then swept the renter's `pod_*` container. `RedisService.executor_create_exclusion` (`lock:executor-create:<executor_id>`, TTL 360 s) is held by `miner_service` around both of its `create_container` calls, from before the call until it returns. The probe takes the same lock with `acquire(blocking=False)` before the idleness re-read and skips the node when a renter holds it; it holds the lock until its own create returns (released in the create's `finally`, shielded, on return, deadline, raise and cancel), not through the sshd wait or the teardown. A renter's create arriving in that window waits for the lock, at most 360 s (the probe's 30 s re-read budget plus its 300 s create deadline; in practice the tens of seconds the create takes); a renter never fails on the lock (Redis down or the wait exhausted logs a warning and the create goes ahead). A re-read that overruns its 30 s budget gives the lock back and reaches no verdict rather than sweep late. The key is distinct from `lock:executor:` (port allocation inside `create_container`); no deadlock, no inversion.
- Race with a real renter: on a failed verdict the check re-reads the backend and this validator's pending-pod marks (the probe's own mark is already cleared by then); a pod or a filler on the node (a filler's create takes the same path and sweeps the same way) or a rent being created means another create swept the probe container away and the verdict is INCONCLUSIVE; a read that fails is INCONCLUSIVE too. Not covered: a renter's create on another validator, or one the backend has not listed and this validator has not started yet.
- Interval: Redis key `rental_probe_ok:<executor_uuid>` holds the epoch of the last passed probe, written on success and deleted on failure, so a failing node that stays idle with the image pulled is re-probed every cycle until it passes (a pass stamped earlier in the interval cannot shield a failure). The backend-forced probe (`rental_probe_requested_executor_ids`) was removed in review: nothing on the backend produced it. It returns with its producer in a follow-up PR on this ticket.
- How long the penalty holds (review, 14 Sep): until a probe passes. Redis key `rental_probe_failed:<executor_uuid>` holds the failed step (and the create step), written on a failed verdict and deleted on a pass, with no lifetime of its own (review: with the OK stamp's 30 days a node skipped or inconclusive for longer was relisted without a clean probe); the OK stamp expires after 30 days, which only makes the probe run again. The failure key is one short string per executor that failed and never passed again. A later cycle that reaches no verdict of its own — skipped for a filler, a missing image or an unreadable interval stamp, inconclusive, an unreadable input, a create lock that could not be taken — fails the node again with that step (`what_we_saw.standing_failure=true`, `no_verdict_this_cycle` says what the cycle would have reported), so a node is not relisted without a clean probe while Redis is readable; an unreadable Redis is logged and the cycle decides as if no failure stood, like every other unreadable input. A rented node is left alone, whichever read says so (the cycle's snapshot, the re-read before the create, a renter's create holding the lock, a rental during the probe): the probe cannot run there and the renter's pod is the backend's to judge; the standing failure applies again once the node is idle. A filler is not a renter: a filler seen at any of those reads carries the failure. `RedisService.set` gained `ex` for the OK stamp's lifetime. A marked node that always has a filler stays off the market until the filler ends (open question 2 is what would let the probe run beside one).
- `create_container` side effects the probe clears: it marks the pod pending in Redis (a second create of the same pod id is declined while it stands); the probe removes that mark in a `finally` that also runs on cancellation (the run's outer `JOB_TIME_OUT` arrives as `CancelledError`, which passes every `except Exception` in `create_container`), shielded so a cancelled task still finishes it. `delete_container` removes the rented-pod cache entry itself. Log lines the create publishes on `STREAMING_LOG_CHANNEL` carry the probe's pod id, which the backend has no pod for.
- One validator: the probe assumes the single Lium validator runs it. Two validators with the flag on would sweep each other's probe container (a renter's create does the same to any `pod_*` it was not told about) and read a false `RENTAL_PROBE_FAILED`.

### Deviations from the brief

- No GPU-less probe beside a filler. The brief asked for a "container start + sshd" probe while the filler holds the GPUs. `create_container` force-removes every `pod_*`/`filler_*` container not in `active_container_names` and restores GPU power caps before a container without a cap starts (`restore_tracked_gpu_power_limits`, `raise_low_power_limits_to_default`), which would uncap a running PEARL filler. A probe-aware branch in `create_container` is a design decision for Rustam (open question below); this PR skips the probe while a filler runs.
- The probe skips a node whose default image is not pulled (`docker image inspect` over the validation shell, 60 s cap; an inspect that cannot be read is INCONCLUSIVE) instead of pulling it: a multi-GB pull inside the validation cycle on every such node is a cost the cached-template check already reports as `recommended_image_cached=false`. This makes removing the image a way to opt out of the probe; the cached-template check still reports it.
- Volume encryption is off for the probe container (`enable_volume_encryption=False`). A renter's pod may get gocryptfs; the probe tests container start, sshd, login and GPU visibility, not the encrypted volume.
- The backend field `rental_check_verified_status` is not set from the validator; the verdict reaches the backend on the existing paths (the validation event and the verified-job reset with `reason_code=RENTAL_PROBE_FAILED`). Consuming it is DAH-3382's side.
- No e2e scenario in this PR: the stub executor in lium-platform `e2e/` would need a real Docker daemon reachable over the rental Docker SDK and an sshd it can refuse; that is a stack change, not small.

### Security as properties

- Bounded input: every string copied from the host or from `create_container` into the event is cut to 600 chars (`_TAIL_CHARS`); step names and ports are the validator's own.
- The probe's private key exists only in the check's frame for one run; it is never logged, never written to the event, never stored. The public key is injected the way a renter's is.
- The container login uses `known_hosts=None`: the probe sends nothing secret over that connection (the key it authenticates with is discarded after the run) and reads one command's output, bounded.
- The image reference comes from the backend and is `shlex.quote`d in the one shell command that names it.
- The check never runs on a rented node or beside a filler (test `test_never_rents_beside_a_renter_or_a_filler`) and never in the dry-run pipeline (test `test_the_probe_runs_last_before_scoring_and_never_in_dry_run`).
- Nothing new is written to the backend; the settings default to off.

### Cost when enabled

One container per idle node per 6 h: image already pulled, `docker run` with the GPUs, sshd up, one `nvidia-smi -L`, `docker rm`. Expect 20 to 60 s per probe on top of the node's cycle, plus the failing case's 90 s deadline, shared by the sshd banner wait and the login retries (a banner poll in flight at the deadline adds up to 10 s, then the one login attempt that always runs adds up to 30 s); hard caps 300 s for the create and 120 s for the teardown. Worst case, every cap summed (image check 60 s, create 300 s, sshd and login 90 s + 10 s + 30 s, command 30 s, teardown 120 s, by-name removal 120 s), about 12.7 min plus the two backend reads (each tenacity-retried, up to about 150 s), the run's whole 780 s `wait_for` (`miner_service.py`) after every other check; on a slow cycle a hung create ends as the outer cancel (no verdict, the container removed by name) rather than a `container_start` failure. A renter's rent dispatched during the probe's create waits for the create lock until that create returns (see the lock note above); one dispatched later runs beside the sshd wait and the teardown and sweeps the probe container (the race note above). A validator restart does not reset the interval (Redis). Three backend reads per probe: `get_default_docker_image`, `get_all_rented_executors` before the create, and again on a failed verdict.

### How the team enables it

1. Staging validator: `RENTAL_PROBE_ENABLED=true` in its env, watch one cycle in Loki for `check_id=executor.validate.rental_probe`: `RENTAL_PROBE_OK` with `probe_steps[]`, `RENTAL_PROBE_SKIPPED` with `reason`, `RENTAL_PROBE_INCONCLUSIVE` with `reason`.
2. Prod: same flag; `RENTAL_PROBE_INTERVAL_HOURS` and `RENTAL_PROBE_SSH_DEADLINE_SECONDS` as needed. Nothing else to deploy on the validator side.

### Provider docs text (lium-platform#366, `apps/docs/docs/providers/troubleshooting.md`, same ticket and reviewer)

Rental probe: what it checks and how to fix each failed step

Every few hours the validator rents your idle node the way a customer does: it starts the default template image with the node's port mapping, waits for SSH to answer on the mapped port, logs in with its own key, runs `nvidia-smi -L`, and removes the container. If a step fails, the node's score is 0 and the node is taken off the market until a probe passes. While the node stays idle, the validator probes it again every cycle; it is listed again when a probe passes. While a customer pod is on the node the probe does not run and does not change the node's standing. The node page shows the reason code `RENTAL_PROBE_FAILED` and the step.

- `container_start`: the container did not start. Run `docker run --rm --gpus all <default image> nvidia-smi -L` on the host. Check the Docker daemon, the NVIDIA container runtime, and that the default image is present.
- `sshd_listen`: the container started but sshd did not answer on the mapped port within 90 s. Check that sshd starts inside the container (`docker logs <container>`), then the host firewall and Docker port publishing: port 22 inside the container must be reachable from the internet on the mapped port. `docker port <container>` shows the mapping; `nc -vz <public ip> <port>` from outside must print an `SSH-2.0` banner, not only connect.
- `ssh_login`: sshd answered but the login was refused, or the connection kept dropping until the 90 s deadline. Check that the container's `/root/.ssh/authorized_keys` is written and that sshd inside the container allows public-key login as root.
- `gpu_count`: SSH worked but `nvidia-smi -L` inside the container did not list every GPU the node advertises. Check `docker run --gpus all ... nvidia-smi -L` and the NVIDIA container runtime.

The probe never runs while a customer pod or a Lium default job is on the node.

### Open questions for Rustam

1. Decided in review (14 Sep): a failed probe holds the node off the market until a probe passes (Redis mark, see "How long the penalty holds"). The consequence stands as a question: a marked node that gets a PEARL or DPHN filler stays off the market until the filler ends, because the probe never runs beside a filler (question 2).
2. Should `create_container` get a probe-aware branch (no container sweep, no power-cap restore, no GPUs) so the probe can run beside a filler? Today the probe skips a node while a filler runs, so a node that always has a filler is never probed. This PR does not touch `create_container`.

### Changed since the first draft (12 Sep, drain-r124's review)

- `sshd_listen` waits for the `SSH-2.0` banner, not a TCP accept: docker-proxy accepted before sshd listened, the single login was dropped, and a healthy slow-starting container would have scored 0 at `ssh_login`. The login is retried with backoff until the same deadline.
- A create that failed after `docker_run` (`add_public_keys`, `set_environment`, `finalize`) or raised returned early from the settle, so `pod_<id>` stayed on the node whenever create_container's own best-effort `cleanup_failed_container_creation` had not removed it (its failure is logged and swallowed). The container and volume are now removed by name, as the backstop, in every path that reached the host.
- Provider text for `sshd_listen` and `ssh_login` updated to match (`_REMEDIATION_BY_STEP`); lium-platform#366 carries the docs page.

### Changed in review (14 Sep, taiberium's four comments)

- Per-executor create lock shared by `miner_service`'s creates and the probe (`RedisService.executor_create_exclusion` / `executor_create_lock`); the probe re-reads idleness under it and releases when its create returns.
- The failed step stands in Redis (`rental_probe_failed:<uuid>`) until a probe passes; a cycle with no verdict of its own fails the node again; a rented node is left alone.
- `rental_probe_requested_executor_ids` and the forced-probe path removed: no backend producer existed.
- The teardown reads the `docker volume rm` result (`DockerCommand.volume_remove_strict`).
- Wording in `core/config.py`, `.env.template`, `messages.py` and the class docstring follows.

### Known follow-ups

- `rental_probe.py` `run`: `_stamp_last_ok` runs before `_clear_failure`; if the OK stamp is written and the delete then fails, the node skips as within interval and carries the stale failure for up to `RENTAL_PROBE_INTERVAL_HOURS`. Pre-existing; clear first, stamp OK only when the clear succeeded (self-review low at `543d35e`).
- `tests/test_rental_probe_check.py` `test_the_real_redis_service_accepts_the_stamps_lifetime` docstring still says both stamps pass `ex`; only the OK stamp does now (self-review low at `543d35e`).
- Backend producer for a forced probe after a BROKEN pod (a field on `RentedExecutorsResponse` plus the validator hook, both in one follow-up on this ticket, with contract coverage), and reading `RENTAL_PROBE_FAILED` from the reset evidence into `rental_check_verified_status` (DAH-3382, lium-platform#301 is the backend circuit).
- A GPU-less probe beside a filler needs `create_container` to skip the container sweep and the power-cap restore for a probe payload (open question 2).
- A test that runs the real `RedisService.executor_create_exclusion` against a stub lock (acquire False / raises / True) instead of the MagicMock `test_miner_service.py` uses (self-review, low).
- e2e scenario: stub executor refuses sshd, node disappears from `GET /executors` within one cycle.
- `STREAMING_LOG_CHANNEL` lines for the probe's pod id reach the backend with no pod to attach to.
- Adjacent hunks in open loop PRs (textual merge either way, none changes this PR's behaviour): lium-io#1328 (`build_checks`, `checks/__init__.py`, `messages.py`), #1340 (`checks/__init__.py`, `pipeline_factory.py`, `ContextState`, `messages.py`), #1335 (`messages.py`), #1360, #1358 and #1337 (`core/config.py`, `.env.template`; #1358 also `task/service.py`, #1360 also `messages.py`), #1303 (`ContextState`, `core/config.py`), #1302 (`ContextState`) and #1351 (`pipeline.py`, stacked on #1302). lium-io#1344 (liumd phase 2b) edits `pipeline.py`, `core/config.py` and `.env.template` in adjacent hunks and calls the backend's health-check pod a "rental probe" in its prose; textual merge either way; no identifier collides (#1344 defines `RentalProbe`, `RENTAL_PROBE_OUTCOME_EVENT`, `LOCAL_VERIFY_RENTAL_PROBE_PARALLEL` for the liumd local-verify flow; this PR's are `RentalProbeCheck`, `RentalProbeMessages`, `RENTAL_PROBE_ENABLED`, `RENTAL_PROBE_INTERVAL_HOURS`, `RENTAL_PROBE_SSH_DEADLINE_SECONDS`).

### Ran it

EC2 sandbox, warm lium-io box i-0f8da1d147408016a (c6i.2xlarge, Linux), at the head as pushed (the earlier heads ran the same way on i-04620f7ff08cc2238):

```
$ pdm run pytest tests -q                                      # run 20260914T130325Z-u232, head 1aaecd1 (from the local bundle, before the push)
2 failed, 2355 passed, 15 skipped, 165 warnings in 136.23s (0:02:16)
$ pdm run pytest tests/test_rental_probe_check.py tests/test_miner_service.py -q   # run 20260914T131104Z-emog, head 1aaecd1
54 passed, 102 warnings in 15.14s
```

The two failures are the box's (sysfs readable as root; an sshd already running) and appear identically in every lium-io run on the box today (20260914T074603Z-ily1, 20260914T081604Z-ht95, 20260914T091107Z-z60v at 6c4e5a7) and in runs of main on the earlier box. Not run: a live validator cycle with RENTAL_PROBE_ENABLED=true against a real executor (the staging step under How to verify). Not run on macOS: the validator runs on Linux hosts only.

### Self-review

Verdict `findings` at `1aaecd1` · 14 checks · by subagent-r153-review · 2026-09-14 13:11Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | STALE-BODY | `RECOVERY/workers/responder-r153/selfreview/lium-io#1361.1aaecd1.brief.md:91` | The body's How-to-verify line says `pytest tests/test_rental_probe_check.py tests/test_miner_service.py -q` -> 53 passed, and the Ran-it block cites run 20260914T120736Z-xhe4 at 543d35e. This commit adds test_gpu_count_step_fails_a_listing_with_the_right_count_but_a_non_zero_exit (tests/test_rental_probe_check.py:1353), so the pair now collects 54 (44 test defs + parametrize in the probe module = 53, plus 1 in test_miner_service.py; 43 defs = 52 at 543d35e). No run at 1aaecd1 is recorded anywhere in the brief, and the new test could not be run on this macOS box (neurons/validators/.venv has no pytest; the validator's deps are Linux-only per the body), so the only evidence for it is static: py_compile passes, ruff check/format pass, the fixture returns a 3-tuple (make_probe_context line 283), _Login is a dataclass with defaults so `_Login(result=smi, command_seconds=0.5)` constructs, make_probe_context sets gpu_count=2 so _expected_gpu_count(ctx) == 2 == _count_gpus(two_gpus). | Re-run the pair on the EC2 box at 1aaecd1 and write the new count (54 passed) and run id into How-to-verify and Ran-it; the gate at the pushed head covers the rest. |
| low | PROSE-VS-CODE | `neurons/validators/tests/test_rental_probe_check.py:1409` | Carried over unchanged from the 543d35e verdict (was line 1397; the new test above it shifts it by 12 lines). The docstring of test_the_real_redis_service_accepts_the_stamps_lifetime still says both stamps pass `ex`; only _stamp_last_ok does (rental_probe.py:569-585), _stamp_failure (620-636) uses set(key, value) with no lifetime. Listed in the PR body's Known follow-ups. | Reword the docstring: the OK stamp passes `ex` (30-day lifetime); the failure stamp uses the plain set(key, value) shape (the second call here). |
| low | CORRECTNESS | `neurons/validators/src/services/task/checks/rental_probe.py:220` | Carried over unchanged from the 543d35e verdict (was line 281; the verdict logic is at the same position in `run`, now lines 219-225). On a passed probe _stamp_last_ok runs before _clear_failure; if the set succeeds and the delete raises, the node skips 'within interval' and carries the stale failure for up to RENTAL_PROBE_INTERVAL_HOURS. Pre-existing, listed in the PR body's Known follow-ups; this commit does not touch the order. | Clear the failure first and stamp the OK only when the clear succeeded. |

Mechanical notes dismissed: `neurons/validators/src/services/task/checks/rental_probe.py:953` Scanner note '_settle not awaited' (was line 845 at 543d35e, moved into _step_teardown by this commit): `_settle(...)` is the argument of `asyncio.ensure_future(...)` at 952-954 and the task is awaited to completion through the `asyncio.shield` loop at 956-960; a CancelledError interrupts only the wait and is re-raised at 961-962. The loop body is byte-for-byte the old `finally` block (old 844-854). · `neurons/validators/src/services/task/checks/rental_probe.py:62` Lines 62, 67, 73, 74 and 80 are values of `_REMEDIATION_BY_STEP`, templates that `_remediation` (now line 1262) fills with `.format(...)`; they must not be f-strings. Untouched by this commit. · `neurons/validators/tests/test_rental_probe_check.py:1129` `run_case(...)` at 1129 is the argument of `asyncio.run(` opened on 1128 and closed on 1130, inside a sync test; likewise 1150 inside `asyncio.run(` at 1149-1151 and 1155-1159. Untouched by this commit (the new test is appended at 1353, below these lines). · `neurons/validators/src/services/task/checks/rental_probe.py:806` The one statement that moved relative to the probe's outer try: `docker = ctx.services.pod_recovery` was read before `try:` (old line 720) and is now read inside _step_container_start, i.e. inside the try whose finally runs _step_teardown. The only observable difference would be a raise from that attribute read (then the teardown would now run with created=None and create_step=None: one by-name `docker rm` of a name that never existed plus the pending-pod clear). `ctx.services` is the pipeline's fixed services container; a plain attribute read on it does not raise. Not a behaviour change in any reachable path. · `neurons/validators/src/services/task/checks/rental_probe.py:325` Lock release paths, checked one by one against old 240-263: old had one `try/finally: await create_lock.release(ctx)` around the re-read AND the probe. New: _may_probe releases (a) in `except BaseException: release; raise` when _idle_under_lock raises or is cancelled (old: finally, then propagate; same exception, same __context__ chaining if release itself raised), (b) before `return refusal` for busy_now None / busy_now truthy / budget overrun (old: return inside try -> finally), and does NOT release on the gate path; `run` 200-203 releases in `finally` after _probe on return, raise or cancel (old 262-263). _step_container_start's shielded release in its `finally` (843) is the old 760-761 unchanged; _CreateLock.release is idempotent via `held` (658-661), so the double release per path is the pre-existing shape. Exactly one check-side release per path, same as before. · `neurons/validators/src/services/task/checks/rental_probe.py:260` Order of awaits/side effects in _may_probe (268-333) + _idle_under_lock (343-358) equals old run 184-261 statement for statement: flag -> _standing_failure -> _skip_reason -> _missing_inputs -> _default_renter_image -> _image_present_on_host -> _take_create_lock -> held check -> monotonic() -> _busy_now -> budget check. `standing` passed on each refusal is identical: rented skip and _RENT_IN_PROGRESS skip pass no standing (None), busy_now passes `None if busy_now in _RENTED_NOW else standing`, every other skip/inconclusive passes `standing`. The verdict half of `run` (204-258) is the old 264-319 verbatim. · `neurons/validators/src/services/task/checks/rental_probe.py:744` _probe split, checked per step against old 705-854: _step_container_start (806-859) is old 725-779 with `return outcome` -> `return None` and `created = result` -> `return result` (so `created` in _probe is set only when isinstance ContainerCreated, None on every other return, unset on raise: same as old); the TimeoutError / CancelledError (`create_step = 'cancelled'`, re-raise) / Exception / shielded-release finally / _failure_detail branches set failed_step, create_step, inconclusive_reason identically. _step_sshd_listen (862-895) sets outcome.ssh_port before the None check as old 781-782, appends the same _Step records, and returns (ssh_port, deadline_at) with deadline_at computed after the port check as before (old 792-793). _step_ssh_login (898-914) and _step_gpu_count (917-935, sync as the old inline code had no await) append the same records and set the same failed_step. Every `return outcome` in _probe (772, 775, 779, 781) is inside the try whose finally awaits _step_teardown, so the settle runs on every early return, raise and cancel as before; a cancel landing during the teardown is delivered to the same innermost `await asyncio.shield(settle)` and re-raised after the settle completes, unchanged. · `neurons/validators/tests/test_rental_probe_check.py:1353` TEST-GAP check on the new test: it names a real regression (dropping the `smi.exit_status == 0 and` conjunct at 926 makes gpu_ok True, failed_step None, and the first assert fails); no other test in the module feeds nvidia-smi a non-zero exit with the right count (the only `exit_status=[1-9]` literal in the file is this test's). It asserts the check's output (failed_step, step.ok, the `listed 2 (exit 1)` detail), not its own setup. Could not be executed on this box (see the STALE-BODY finding); statically it constructs and its expectations follow from _count_gpus (1212-1213) and gpu_count=2 in make_probe_context.

### Verified

Gate: PASS (1aaecd1) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1361)

</details>


neurons/validators/src/services/task/checks/rental_probe.py
```diff
@@ -0,0 +1,944 @@
...
+        ships_sshd=True,
+        enable_jupyter=False,
+        enable_volume_encryption=False,
+        available_ports=[
+            PayloadPortMapping(internal_port=internal, external_port=external)
+            for internal, external in ctx.state.verified_port_pairs
+        ],
+        pod_mapping=[],
+        # nothing to protect: the probe runs only when the backend lists no pod and no filler here
+        active_container_names=[],
```
taiberium, line 739: A real rental can start after `_busy_now()`, and this empty list lets the probe sweep its `pod_*` container. Please add shared per-executor exclusion around both create paths.

neurons/validators/src/services/task/checks/rental_probe.py
```diff
@@ -0,0 +1,944 @@
...
+    GPUs, the bandwidth and the filler, not the one thing a renter needs. This check starts the
+    default renter image through `DockerService.create_container`, the path a renter's pod takes,
+    with a probe-owned SSH key and the ports PortConnectivityCheck verified; waits for sshd's banner
+    on the mapped port; logs in, retrying until the same RENTAL_PROBE_SSH_DEADLINE_SECONDS deadline;
+    runs `nvidia-smi -L`; and tears the container down through `delete_container`. Each step is
+    timed and lands in `what_we_saw["probe_steps"]`.
+
+    A failed step zeroes the score and clears the verified job with reason code RENTAL_PROBE_FAILED
+    and the step's name, the same mechanics as the GPU runtime quarantine in rental_verification.py.
+    The penalty is the cycle's: the failure is not carried forward, so the next cycle verifies the
```
taiberium: A failed probe only clears verification for one cycle, so a skipped or inconclusive next cycle re-lists the node without a clean probe. Please persist failure until a probe passes.

neurons/validators/src/protocol/vc_protocol/compute_requests.py
```diff
@@ -122,6 +122,10 @@ class RentedExecutorsResponse(BaseModel):
     # FillerRunDao.CREATE_KILL_*). Such a run never reaches RUNNING, so the ISSUE-050 liveness
     # probe has nothing to check. Additive with an empty default: an older backend penalizes nobody.
     filler_create_kill_executor_ids: list[str] = []
+    # DAH-3436: executor_ids the backend wants rental-probed this cycle regardless of the probe
+    # interval (a renter pod on the node went BROKEN since the last probe). Additive with an empty
+    # default: a backend that does not set it leaves the probe on its interval alone.
+    rental_probe_requested_executor_ids: list[str] = []
```
taiberium: Nothing currently populates this field, so a BROKEN pod cannot trigger the promised immediate probe. Please add the backend producer and contract coverage before enabling the feature.

neurons/validators/src/services/task/checks/rental_probe.py
```diff
@@ -0,0 +1,944 @@
...
+    rented-pod cache entry dropped; None when nothing is left on the node."""
+    try:
+        removed = await asyncio.wait_for(
+            ctx.ssh.run(
+                DockerCommand.remove_with_volumes(shlex.quote(container_name)), check=False
+            ),
+            timeout=_SHELL_COMMAND_TIMEOUT_SECONDS,
+        )
+        await asyncio.wait_for(
+            ctx.ssh.run(DockerCommand.volume_remove(volume_name), check=False),
```
taiberium: This ignores the volume command result, and `volume_remove()` masks failures with `|| true`, so cleanup can report success while the probe volume remains. Please preserve and inspect that result.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/protocol/vc_protocol/compute_requests.py
surcyf123: Done in `6c4e5a7`: removed the field and the forced-probe path, so the PR promises only what runs. The BROKEN-pod trigger comes back with its backend producer and contract coverage in one follow-up PR on this ticket; the body's Known follow-ups names it.

neurons/validators/src/services/task/checks/rental_probe.py
surcyf123: Done in `6c4e5a7`: a per-executor create lock in Redis, shared by both create paths. `miner_service` holds it around both of its `create_container` calls, from before the call until it returns. The probe takes it without waiting before it re-reads idleness. It skips the node when a renter holds it. It releases the lock as soon as its own create returns, also on deadline, raise and cancel. The lock's TTL outlasts the probe's whole hold (30 s re-read budget plus the 300 s create deadline), so it cannot lapse before the probe's sweep. A renter waits behind it for that long at most and never fails on it. Tests cover the order acquire, re-read, create, release, teardown; the skip when a renter holds it; the release when the create is cut short.

neurons/validators/src/services/task/checks/rental_probe.py
surcyf123: Done in `6c4e5a7`: the failed step now stands in Redis until a probe passes. A later cycle with no verdict of its own fails the node again with that step: filler running, image gone, inconclusive, an unreadable input, a lock it could not take. The event says what the cycle would otherwise have reported. A rented node is left alone at every read (snapshot, re-read, a renter's create holding the lock, a rental during the probe). The probe cannot run there and the renter's pod is the backend's to judge; the failure applies again once the node is idle. A filler is not a renter and carries it. Both stamps expire after 30 days. Tests cover four no-verdict paths, the rented exemptions, and the clear on a pass.

neurons/validators/src/services/task/checks/rental_probe.py
surcyf123: Done in `6c4e5a7`: the teardown now runs `docker volume rm` without the `|| true` (a strict variant next to the masked one) and reads its exit status and stderr. "No such volume" is nothing left behind. Any other non-zero exit is a leftover the teardown step reports and logs. Tested with an in-use volume, a missing one and a clean removal.

neurons/validators/src/services/task/checks/rental_probe.py
```diff
@@ -0,0 +1,1157 @@
...
+    step, _, create_step = text.partition(":")
+    return _Failure(step, create_step or None) if step else None
+
+
+async def _stamp_failure(ctx: Context, failure: _Failure) -> None:
+    try:
+        await ctx.services.redis.set(
+            f"{_REDIS_FAILED_PREFIX}:{ctx.executor.uuid}",
+            f"{failure.step}:{failure.create_step or ''}",
+            ex=_REDIS_STAMP_TTL_SECONDS,
```
taiberium: Please keep the failure until a probe passes; the 30-day TTL can re-list a node after a long skipped or inconclusive period.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/services/task/checks/rental_probe.py
surcyf123: Will do.

surcyf123: Done in `543d35e`: `_stamp_failure` now sets `rental_probe_failed:<uuid>` with no lifetime. The failure stands however long the node is skipped or inconclusive; only `_clear_failure` on a passed probe removes it. The OK stamp keeps its 30-day TTL, since expiring it only makes the probe run again. Test: `test_a_standing_failure_outlives_every_stamp_lifetime_until_a_probe_passes` (30+ days of skipped cycles, still delisted, relisted on the first pass).

neurons/validators/src/services/task/checks/rental_probe.py
```diff
@@ -0,0 +1,1161 @@
...
+    renter holds it, and re-reads the node's idleness from the backend and this validator's pending-pod
+    marks under the lock, not from the cycle-start snapshot; the lock is released as soon as the
+    create returns. Anything the probe cannot read (backend, Redis) makes it skip or reach no verdict;
+    it never penalises on missing data alone. RENTAL_PROBE_ENABLED is off by default.
+    """
+
+    check_id = "executor.validate.rental_probe"
+    fatal = True
+
+    async def run(self, ctx: Context) -> CheckResult:
```
taiberium, line 194: NIT: 136 lines, two jobs. Move the may-we-probe gate (flag, skip, inputs, image, lock, idleness) into its own method and keep the verdict here.

neurons/validators/src/services/task/checks/rental_probe.py
```diff
@@ -0,0 +1,1161 @@
...
+            for internal, external in ctx.state.verified_port_pairs
+        ],
+        pod_mapping=[],
+        # nothing to protect: the probe runs only when the backend lists no pod and no filler here
+        active_container_names=[],
+        active_volume_names=[],
+    )
+
+
+async def _probe(ctx: Context, image_ref: str, create_lock: _CreateLock) -> _ProbeOutcome:
```
taiberium, line 744: NIT: 138 lines for five named steps. One method per step would read like the list in the docstring.

taiberium (approved):

neurons/validators/src/services/task/checks/rental_probe.py
surcyf123: Will do. Refactoring now, no behaviour change. The gate moves into its own method; the verdict stays here.

neurons/validators/src/services/task/checks/rental_probe.py
surcyf123: Will do. Refactoring now, no behaviour change. One private method per step, named after the docstring list.

neurons/validators/src/services/task/checks/rental_probe.py
surcyf123: Done in `1aaecd1`: the gate is `_may_probe` (flag, standing failure, skip, inputs, image, lock, idleness re-read); `run` keeps the verdict, 137 lines to 65. No behaviour change: 54 probe tests and 2355 validator tests pass unchanged.

neurons/validators/src/services/task/checks/rental_probe.py
surcyf123: Done in `1aaecd1`: `_probe` is now orchestration over one `_step_*` function per step (container_start, sshd_listen, ssh_login, gpu_count, teardown), 150 lines to 40. Same steps, same records, one new test on the gpu_count step.

surcyf123: Done in `1aaecd1`: `_probe` is now orchestration over one `_step_*` function per step (container_start, sshd_listen, ssh_login, gpu_count, teardown), 150 lines to 40. Same steps, same records, one new test on the gpu_count step.

PR #1365, opened 2026-09-14
DAH-3467 - [P2] validator: inspect after a timed-out docker rm before failing the delete

surcyf123:
Implements DAH-3467 · reviewer: @taiberium

**Problem** — A `docker rm` finishing after the SDK's 60 s read timeout was reported as a failed delete though the container was gone (13 prod deletes 12–14 Sep, 9–19 min late).

**Goal** — DAH-3467 (P2) · serves B-168.

**Fits** — Sits in `lium-io` — `neurons/validators`; touches backend, tests (23).

**Flag:** no flag

**Risks**
- Validator pod restart on deploy; consumer websocket gone ~40–70 s, renters see nothing — rollback: previous tag
- A container reported gone that is still there would leave the executor marked free — only an `inspect` 404 by exact name on the same host counts as gone — rollback: revert
- A timed-out `docker rm` answers up to ~75 s later (~25 s more while dockerd holds a volume) instead of failing — rollback: revert
- A volume still held after 6 retries stays on the host (a warning and the step's non-fatal error line) until the stale-volume sweep — as before; rollback: n/a

**How to verify**
- `cd neurons/validators && uv run pytest tests/test_docker_service.py tests/test_rental_docker_sdk.py -q` → 272 passed, 1 skipped
- Loki `{namespace="validator"} |= "read timeout"` after deploy → `inspect confirms it is gone` lines, no `Read timed out` failures

**Verified by the loop:** Gate: PASS 52ca475

**Ran it:** validators suite on a sandbox EC2 box (`aws_verify` run 20260918T201856Z-nizp at 52ca475): 2656 passed, 15 skipped, plus the 2 root-only tests as the CI user: 2 passed

**Self-review:** clean at 52ca475 (15 checks, 4 low)

**Parts:** backend ✓ · migration n/a — no schema · frontend n/a — validator-internal · CLI/SDK n/a — no API change · docs n/a — nothing user-visible · tests ✓ 25 · dashboards n/a — no panel change

**Docs:** none changed in this PR — internal to the validator's delete path; no page describes the `ContainerDeletionFailed` events it removes.

### Review round 3 (Rustam, 18 Sep 10:04Z)

- The exhausted volume reached the backend's penalty path: a delete whose 6 in-use retries ran out answered `DeletionInProgress` with the container already gone. The backend keeps a `DeletionInProgress` pod in DELETING and re-asks from `retry_failed_pods` every 10 min (`get_pod_ids_to_retry(DELETING, minutes_ago=9)`), and `retry_remove_rent` gives up at `POD_DELETE_MAX_ATTEMPTS` (3) into `undeploy_verdict` → `penalize_undeploy_give_up`, so a volume dockerd held through three attempts penalised a node that had removed the container. Now `_remove_named_volume` raises after the last attempt like any other volume error, the `_best_effort_delete_step` logs it (`leaving it on the host`, warning) and the delete completes with `ContainerDeleted`; the redis rental record is removed and the marker cleared. Only the container path (still `removing` after the read timeout, dockerd's 409) answers `DeletionInProgress` now. Tests: `…_then_gone_with_volume_still_in_use_stays_best_effort_once_retries_run_out`, `…_then_gone_with_external_volume_still_in_use_stays_best_effort` (the s3fs plugin removal still does not run after a failed `volume rm`), `test_delete_re_asked_after_in_progress_retries_a_volume_still_in_use_then_gives_it_up` (the marker keeps the 2 retries on the re-ask, then the delete completes and the marker is cleared).
- `pending_deletions` never evicted: a pod the backend gave up on, or a rental-probe pod (a fresh id per run, never re-asked), stayed for the life of the process. The registry now keeps `pod_id → marked_at` on an injectable monotonic clock and drops entries older than `PENDING_DELETION_TTL_SECONDS` (3600 s: the backend's whole re-ask span is 3 attempts 10 min apart, so nothing older is coming) on every `mark`, `is_pending` and `len`. Tests: `test_pending_deletions_expire_after_the_ttl_and_a_new_mark_restarts_it`, `test_pending_deletions_do_not_accumulate_probe_pods`.
- `rental_probe._teardown` capped `delete_container` at 120 s while the delete's slowest path is about 190 s (30 s stop grace + the SDK's 60 s read timeout + the 60 s inspect window with one 15 s inspect past it + five 5 s volume waits). `_TEARDOWN_DEADLINE_SECONDS` is 300: the cap wins — a delete still running at 300 s is cancelled, the probe records `delete_container did not finish in 300 s` and `_step_teardown`'s by-name `docker rm -fv` runs as the backstop, as before; a delete that finishes inside the windows is no longer cut off at 120 s mid-inspect.

### Review round 2 (Rustam, 18 Sep 08:42Z)

- (As built in round 2; since round 3 above only the container path answers `DeletionInProgress`, and a volume still held after the retries is given up, not answered in progress.) Volume cleanup across delete retries: after a `DeletionInProgress` answer the backend re-asks, and the re-asked delete finds the container already absent (`No such container`), which is not the inspect-confirmed path, so a `volume is in use` on its volume step fell back to the one best-effort attempt and the pod closed over the volume left behind. Now `pending_deletions` (an in-process per-pod marker next to `inflight_creates`) is set whenever `delete_container` answers `DeletionInProgress`, from the container path (still `removing`, dockerd's 409) or the volume path, and a later delete for a marked pod runs `_remove_named_volume(retry_in_use=True)` whether or not this request's removal was confirmed by inspect: `volume is in use` is retried and, still held, answered `DeletionInProgress` again. The marker is cleared once every named volume of the pod is gone; an absent container with no marker (DAH-2345, removed by failed-create cleanup) keeps the one best-effort attempt. The marker is in-process, like `inflight_creates`: a validator restart between the in-progress answer and the re-ask forgets it, and that re-ask runs the one best-effort attempt as before. Tests: `test_delete_re_asked_after_in_progress_keeps_a_volume_still_in_use_in_progress` (×2: round 1 answered in progress from the container path or the volume path; the re-ask with an absent container and the volume still in use → `DeletionInProgress`, no inspect, redis record untouched), `test_delete_re_asked_after_in_progress_completes_once_the_volume_is_gone` (×2: the re-ask's volume goes on the second attempt → `ContainerDeleted`, marker cleared), `test_delete_of_an_absent_container_without_a_pending_deletion_stays_best_effort`.

### Review round 1 (Rustam, 18 Sep 06:35Z)

- Volume references after the 404: dockerd can answer `inspect` 404 before it has released the container's named-volume references, so the `volume rm` that follows fails with `409 … volume is in use` and the old best-effort step swallowed it while the pod closed on our success, leaving the volume on the executor. Now `_force_remove_container` returns `_ForcedRemoval(confirmed_by_inspect=True)` on that path and `delete_container` removes each named volume through `_remove_named_volume(retry_in_use=True)`: `volume is in use` gets `REMOVE_CONFIRM_VOLUME_ATTEMPTS` (6) attempts, `REMOVE_CONFIRM_VOLUME_RETRY_SECONDS` (5 s) apart; still in use after the last one returns `DeletionInProgress` (`…; volume <name> still in use after 6 attempts`) and the backend re-asks. The re-asked delete finds the container absent (`No such container`); round 2 above keeps its volume step on the retried path. After a remove that answered in time, and for every error other than in-use, the step is the one best-effort attempt it always was. Tests: `…_then_gone_retries_a_volume_still_in_use` (two in-use answers, then success → `ContainerDeleted`, three attempts), `…_then_gone_with_volume_still_in_use_reports_in_progress` (in-use on every attempt → `DeletionInProgress`, the redis record untouched), `…_then_gone_with_external_volume_still_in_use_reports_in_progress` (the s3fs plugin removal does not run), `test_delete_container_volume_in_use_after_a_plain_remove_stays_best_effort`, `…_then_gone_other_volume_error_stays_best_effort`, `test_docker_volume_in_use_detection`.
- The wedged-GPU sweep: a filler's read-timed-out remove used to raise, and the failed-remove path swept the wedged GPUs before propagating (DAH-2427); answering `DeletionInProgress` returned without the sweep. `delete_container` now runs `_sweep_wedged_gpus_after_teardown` (best-effort, a no-op while the container's processes are alive) before returning either in-progress answer for a filler: the read-timeout `removing` one and dockerd's 409 `removal already in progress`, which had the same gap. Tests: `test_delete_filler_answering_deletion_in_progress_sweeps_wedged_gpus_first` (×2), `test_delete_customer_rental_answering_deletion_in_progress_does_not_sweep`.
- The nit: the warning for `unknown` and for a real `State.Status` reads `Container not confirmed gone after the remove timed out` (`unknown` never established that the container exists).

### Why

Prod, 12–14 Sep (Loki, `{namespace="validator"}` connector logs): 13 customer-pod deletes failed with `Docker SDK remove container failed: SSHConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)` at exactly 60 s after `docker rm -f`. In every one of the 13 the backend's retry sweep (9–19 min later) got `404 No such container` — dockerd had finished the remove after the SDK stopped waiting. The renter's pod sat in DELETING for those minutes and the backend recorded a `ContainerDeletionFailed` it did not need to.

Where the 60 s go: > 50 % of graceful stops in the same window ran the full 30 s grace and were SIGKILLed; dockerd bounds the kill itself to ~20 s and fails loudly (`could not kill container: did not receive an exit event`, DAH-2991) when it cannot. A remove that is silent past 60 s is therefore in the teardown after the kill — unmounting and unlinking the rw layer (`overlay2 … directory not empty` on the nested-Docker/IO-heavy workloads in the same logs) — and completes on its own. The old code turned "no answer yet" into "failed".

### What the fix does

`_force_remove_container` catches the wrapped SDK error; when `_is_docker_read_timeout_error` matches (cause chain holds `requests.ReadTimeout`/`urllib3.ReadTimeoutError`, or the text carries "read timed out"), `_confirm_removal_after_timeout` polls `RentalDockerSdkClient.container_status` (a `docker inspect` by name over the same SSH client) every 5 s for up to 60 s, each inspect bounded to 15 s by `asyncio.wait_for` (the window is checked between inspects, so the whole confirmation is ~75 s worst case):

| inspect says | delete answers |
|---|---|
| 404 (None) | success — the container is gone; the named volumes are then removed with the in-use retry below |
| `removing` for the whole window | `FailedContainerRequest` with `DeletionInProgress` — the backend keeps the pod DELETING and re-asks from its sweep; no failure event |
| any other `State.Status` (`running`, `exited`, `dead`, …) | the original read-timeout error is re-raised — a present, not-removing container is a real failure |
| inspect fails, hangs, or its body has no `State.Status` (`unknown`) | the original read-timeout error is re-raised |

A connect timeout (`Connection to … timed out`) never reached dockerd and is not confirmed. `_deletion_in_progress` is the existing 409 `removal already in progress` response, factored out so both paths build the same request.

Constants: `REMOVE_CONFIRM_TIMEOUT_SECONDS = 60.0`, `REMOVE_CONFIRM_POLL_SECONDS = 5.0`, `REMOVE_CONFIRM_INSPECT_TIMEOUT_SECONDS = 15.0`, `_DOCKER_REMOVING_STATUS = "removing"`. `miner_service` does not wrap `delete_container` in an outer timeout, so the extra minute is not cut off upstream.

### Related

- B-168 report (`RECOVERY/workers/bug-b168-rm-close-lag/REPORT.md`): first saw the 60 s timeouts as the second-largest share of `rm` close lag.
- DAH-2991 (`could not kill container`): the wedged-process path is untouched; that error is a 500, not a read timeout.
- DAH-3436 / DAH-2928 are different mechanisms (rental probe, disk health); the ticket body says why this is not a duplicate.

### Ran it

EC2 warm lium-io box `i-0d2847e63c80eec7b` (c6i.2xlarge, the validators venv), review round 3 (18 Sep 10:04Z), head 52ca475:

```
20260918T201856Z-nizp  validators pytest tests/ (2 root-sensitive deselected, run below as loopverify): rc=0 in 184 s — 2656 passed, 15 skipped, 2 deselected
20260918T201856Z-nizp  validators pytest 2 root-sensitive tests as loopverify: rc=0 — 2 passed
20260918T184827Z-nhcf  validators pytest tests/test_docker_service.py tests/test_rental_docker_sdk.py: rc=0 in 5 s — 272 passed, 1 skipped (pre-push tree; 52ca475 differs from it by one comment and one test bound)
```

The round-3 tests fail on 13c8d1b: the exhausted volume answered `DeletionInProgress` where `…_stays_best_effort_once_retries_run_out` and `…_then_gives_it_up` expect `ContainerDeleted`; the set-based registry has no clock and `test_pending_deletions_expire_after_the_ttl…` cannot construct it.

EC2 warm lium-io box (c6i.2xlarge, the validators venv), review round 2 (18 Sep), head 13c8d1b:

```
20260918T090624Z-nkwm  validators pytest tests/ (2 root-sensitive deselected, run below as loopverify): rc=0 in 187 s — 2656 passed, 15 skipped, 2 deselected
20260918T090624Z-nkwm  validators pytest 2 root-sensitive tests as loopverify: rc=0 — 2 passed
20260918T090547Z-m43c  validators pytest tests/test_docker_service.py: rc=0 in 5 s — 222 passed, 1 skipped
```

The round-2 tests fail on 1ff8359: the re-asked delete with an absent container answers `ContainerDeleted` after one swallowed `volume rm` where `…_keeps_a_volume_still_in_use_in_progress` expects `DeletionInProgress` and two attempts.

Round 1 (18 Sep), head 1ff8359:

```
20260918T070719Z-ounw  validators pytest tests/ (2 root-sensitive deselected, run below as loopverify): rc=0 in 187 s — 2651 passed, 15 skipped, 2 deselected
20260918T070719Z-ounw  validators pytest 2 root-sensitive tests as loopverify: rc=0 — 2 passed
20260918T071418Z-cnfj  validators pytest tests/test_docker_service.py: rc=0 in 5 s — 217 passed, 1 skipped
```

The round-1 tests fail on 5de3a4d: the in-use `volume rm` was one swallowed attempt (`…_retries_a_volume_still_in_use` sees one attempt and `ContainerDeleted` where it expects three; `…_reports_in_progress` gets `ContainerDeleted`), and the filler's `DeletionInProgress` return ran no sweep (`sweep.assert_awaited_once()` fails).

Round 0 (14 Sep), head 6434c3b32 (rebased to 5de3a4d without change): EC2 run `20260914T091145Z-kt2u` (warm lium-io box i-0f8da1d147408016a, c6i.2xlarge), round 1 of the self-review: `tests/test_docker_service.py tests/test_rental_docker_sdk.py -v`: 258 passed, 1 skipped. Full validator suite on b6d203668: EC2 run `20260914T085122Z-oo2o`, 2372 passed, 15 skipped, 2 failed — the two failures (`test_an_unreadable_sysfs_tree_is_not_reported_as_no_devices`, `test_adopt_path_hardens_config_for_key_only_auth`) failed identically on `origin/main` on that box (sysfs and file-mode assumptions of the warm image), not from the diff; they pass in today's run under the loopverify user.

Not run against a live dockerd: reproducing a > 60 s `docker rm` needs a multi-GB rw layer on a slow disk, and the volume-reference release lag is dockerd-internal; the prod evidence in Why is the measurement.

### Self-review

Verdict `clean` at `52ca475` · 15 checks · by subagent-sr1365-resume1 · 2026-09-18 20:26Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | PROSE-VS-CODE | `neurons/validators/src/services/task/checks/rental_probe.py:101` | The comment calls 30+60+60+15+25 = 190 s 'the slowest path the delete takes', but the delete has slower legal paths still under the cap: docker-py's stop() adds the 30 s grace to its own 60 s HTTP read timeout (a silent stop is ~90 s, not 30), the DAH-2991 could-not-kill escalation runs a second rm -f with its own 60 s timeout plus another 75 s inspect window, and prune_images and the custom-build cleanup run before the volumes; the 300 s cap still covers the common case, so the number is right but the superlative is not. | folded at 52ca475: the comment names the inspect-confirmed path and says a slower stop stays under the cap |
| low | PROSE-VS-CODE | `review-wave/readability/overrides/lium-io-1365.json:79` | The Risks bullet says an exhausted volume leaves 'one `Named volume still in use … leaving it on the host` warning', but the raise then lands in _best_effort_delete_step (docker_service.py:373), which also logs 'delete_container post-teardown step failed (non-fatal)' at ERROR; a reviewer reading Loki for the warning alone would miss the ERROR line, and the file's own _stop_container_gracefully comment treats an ERROR log as an alert source. | folded in the override: the Risks bullet names the best-effort step's error line |
| low | TEST-GAP | `neurons/validators/tests/test_docker_service.py:2334` | test_pending_deletions_do_not_accumulate_probe_pods asserts `len(registry) < 60` after 60 marks 360 s apart; the arithmetic gives at most 11 live entries (3600/360 + 1), so `< 60` would still pass a registry that evicted only one entry. | folded at 52ca475: `assert len(registry) == 10` |
| low | STALE-BODY | `review-wave/readability/overrides/lium-io-1365.json:104` | The 'Review round 2' block still describes pending_deletions as set 'from the container path (still removing, dockerd's 409) or the volume path' and 'still held, answered DeletionInProgress again'; since round 4 only the container path marks and a held volume completes the delete, so a reader of the round-2 block alone gets the pre-round-4 behaviour (the round-3 block above it corrects this, but nothing in the round-2 block says it was superseded). | folded in the override: the round-2 block says the volume path no longer answers in progress since round 3 |

Mechanical notes dismissed: `neurons/validators/src/services/docker_service.py:6739` False positive: the line is `msg=f"{exc}; container still '{status}' after "`, f-prefixed, and its continuation `f"{REMOVE_CONFIRM_TIMEOUT_SECONDS:.0f}s"` on 6740 is f-prefixed too; test_delete_container_remove_read_timeout_still_removing_reports_in_progress asserts the interpolated "still 'removing'" text. · `neurons/validators/tests/test_docker_service.py:2033` False positive: `f'Conflict ("remove {volume_name}: volume is in use - [...]")'` carries the f prefix; test_docker_volume_in_use_detection passes the interpolated text through _is_docker_volume_in_use_error. · `neurons/validators/tests/test_docker_service.py:2200` False positive: `f'("No such container: {container_name}")'` carries the f prefix (the brace-free first line of the pair is a plain string on purpose); the re-ask tests reach the absent-container path only because this text matches _is_missing_docker_container_error.

### Verified

Gate: PASS (52ca475) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1365)

### What changed

- A `docker rm` read timeout is followed by `inspect` polls, 60 s window, 5 s apart; 404 → done, `removing` → `DeletionInProgress`, anything else → timeout re-raised (`neurons/validators/src/services/docker_service.py`)
- After a removal confirmed by that inspect 404, a named volume dockerd still holds (`volume is in use`) gets 6 attempts 5 s apart; still held after the last one it is logged and left on the host like any other volume error, and the delete completes (`_remove_named_volume`); a per-pod `pending_deletions` marker set by the container-path `DeletionInProgress` keeps that retry on the backend's re-ask, which finds the container already absent, and expires after 1 h (`PENDING_DELETION_TTL_SECONDS`)
- A filler whose delete answers `DeletionInProgress` (read timeout still `removing`, or dockerd's 409) sweeps wedged GPUs before answering, as the failed-remove path does
- Only a read timeout (`ReadTimeout` cause or "read timed out" text) is confirmed, never a connect timeout; `container_status()` returns `State.Status` lowercased, None on 404 (`rental_docker_sdk.py`)
- Delete-path tests (25 cases), 3 SDK tests (`tests/test_docker_service.py`, `tests/test_rental_docker_sdk.py`); `rental_probe._teardown` caps `delete_container` at 300 s, above the delete's slowest path (~190 s)

</details>


neurons/validators/src/services/docker_service.py
```diff
@@ -6546,15 +6634,37 @@ async def _force_remove_container(
...
+                # DAH-3467: dockerd took the force-remove and has not answered yet. Ask it what
+                # happened instead of failing a delete that is most likely completing.
+                status = await self._confirm_removal_after_timeout(docker_client, payload, log)
+                if status is None:
+                    log.info(
+                        "Container removal outlived the read timeout; inspect confirms it is gone",
+                        container_name=payload.container_name,
+                        error=str(exc),
+                    )
+                    return None
```
taiberium: Inspect can return 404 before dockerd releases named-volume references, so volume cleanup can fail while the backend closes the pod. Please retry cleanup or return `DeletionInProgress`.

neurons/validators/src/services/docker_service.py
```diff
@@ -6546,15 +6634,37 @@ async def _force_remove_container(
...
+                        container_name=payload.container_name,
+                        error=str(exc),
+                    )
+                    return self._deletion_in_progress(
+                        payload,
+                        msg=f"{exc}; container still '{status}' after "
+                        f"{REMOVE_CONFIRM_TIMEOUT_SECONDS:.0f}s",
+                    )
+                log.warning(
+                    "Container is still present after the remove timed out",
```
taiberium: nit: `unknown` does not establish that the container exists, so “still present” is inaccurate. Could this say “Container not confirmed gone”? Non-blocking.

neurons/validators/src/services/docker_service.py
```diff
@@ -6546,15 +6634,37 @@ async def _force_remove_container(
...
+                        error=str(exc),
+                    )
+                    return None
+                if status == _DOCKER_REMOVING_STATUS:
+                    log.info(
+                        "Container deletion is still in progress after the read timeout",
+                        container_name=payload.container_name,
+                        error=str(exc),
+                    )
+                    return self._deletion_in_progress(
```
taiberium: Filler read timeouts used to run the wedged-GPU sweep, but this return skips it while retries stay in progress. Could you run the sweep before returning?

taiberium (changes requested):
See the inline comments.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `1ff8359`: The warning reads `Container not confirmed gone after the remove timed out`, for `unknown` and for a real `State.Status` alike; `container_status` in the extra keeps the two apart.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `1ff8359`: The sweep runs before the return. When the removal comes back with a DeletionInProgress failure for a filler, `delete_container` runs `_sweep_wedged_gpus_after_teardown` and then answers. The sweep is best-effort and a no-op while the container's processes are alive. This covers the read-timeout 'removing' return and dockerd's 409 'removal already in progress', which had the same gap. Tests: `test_delete_filler_answering_deletion_in_progress_sweeps_wedged_gpus_first` (x2, one per path) and `test_delete_customer_rental_answering_deletion_in_progress_does_not_sweep`.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `1ff8359`: Retry, then DeletionInProgress. The forced removal now reports whether the inspect 404 confirmed it. On that path each named volume goes through `_remove_named_volume(retry_in_use=True)`. A 'volume is in use' answer gets 6 attempts 5 s apart. Still in use after the last one answers DeletionInProgress, and the backend re-asks. The message names the volume and the attempts. Any other volume error stays the one best-effort attempt it was. So does every error after a remove that answered in time. Tests, all in test_docker_service.py. Two in-use answers, then success: `..._read_timeout_then_gone_retries_a_volume_still_in_use` sees three attempts and ContainerDeleted. In use on every attempt: `..._then_gone_with_volume_still_in_use_reports_in_progress` and its external-volume twin get DeletionInProgress. There the s3fs plugin removal and the redis record do not run. The best-effort paths stay: `..._volume_in_use_after_a_plain_remove_stays_best_effort` and `..._then_gone_other_volume_error_stays_best_effort`. The matcher: test_docker_volume_in_use_detection. Validator suite on EC2 at this head: 2651 passed.

neurons/validators/src/services/docker_service.py
```diff
@@ -6716,36 +6922,43 @@ async def delete_container(
...
-                            ),
-                            volume_name=payload.local_volume,
-                            volume_role="local",
+                        volume_in_progress = await self._remove_named_volume(
+                            docker_client,
+                            payload,
+                            payload.local_volume,
+                            "local",
+                            log,
+                            retry_in_use=removal.confirmed_by_inspect,
```
taiberium: Please keep volume cleanup in progress across delete retries; the next request sees an absent container and can swallow another `volume is in use` error.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `13c8d1b`: The in-progress state now survives the re-ask. A per-pod marker, `pending_deletions`, is set whenever the delete answers DeletionInProgress, from the container path or the volume path. A later delete for a marked pod runs the volume remove with the in-use retry even though it finds the container absent. Still in use after the attempts, it answers DeletionInProgress again instead of swallowing the error. The marker clears once every named volume is gone; an absent container with no marker keeps the one best-effort attempt as before. It is in-process like the create registry, so a validator restart forgets it. Tests: `test_delete_re_asked_after_in_progress_keeps_a_volume_still_in_use_in_progress` covers both first-round paths and expects DeletionInProgress with two attempts and no inspect. `test_delete_re_asked_after_in_progress_completes_once_the_volume_is_gone` expects ContainerDeleted and the marker cleared. `test_delete_of_an_absent_container_without_a_pending_deletion_stays_best_effort` keeps the DAH-2345 case. Validator suite on EC2 at this head: 2656 passed.

neurons/validators/src/services/docker_service.py
```diff
@@ -6567,15 +6737,73 @@ async def _force_remove_container(
...
+                if attempt >= REMOVE_CONFIRM_VOLUME_ATTEMPTS:
+                    log.info(
+                        "Named volume still in use after the container was confirmed gone; "
+                        "deletion stays in progress",
+                        volume_name=volume_name,
+                        volume_role=volume_role,
+                        attempts=attempt,
+                        error=str(exc),
+                    )
+                    return self._deletion_in_progress(
```
taiberium: Once volume retries expire, this new `DeletionInProgress` response reaches the backend penalty path although the container is gone. Please keep the exhausted volume error best effort.

neurons/validators/src/services/docker_service.py
```diff
@@ -704,6 +725,35 @@ async def wait_until_done(self, pod_id: str, timeout: float) -> bool:
...
+    DAH-3467 (review): the backend re-asks a delete that answered DeletionInProgress — the container
+    still removing, or a named volume dockerd still held after the container was confirmed gone. The
+    re-ask finds the container already absent ("No such container"), which is not the inspect-confirmed
+    path, so without this marker a "volume is in use" on the retried volume remove would fall back to
+    best-effort and the pod would close over the volume left behind. A pod marked here keeps its
+    volume cleanup on the in-progress path until a delete for it completes.
+    """
+
+    def __init__(self) -> None:
+        self._pod_ids: set[str] = set()
```
taiberium: nit: This set never evicts failed or probe deletions, so unique probe IDs can accumulate indefinitely. Please add expiry; non-blocking, fine as a follow-up.

neurons/validators/src/services/docker_service.py
```diff
@@ -160,6 +160,25 @@
 # that hold GPUs and brick the executor.
 CONTAINER_STOP_GRACE_SECONDS = 30
 
+# DAH-3467: dockerd answers a force-remove only once the rw layer is unlinked, and that can outlive
+# the Docker SDK's 60 s read timeout (prod, 12-14 Sep: 13 deletes, all of them gone by the time the
+# backend retried 9-19 min later). A read timeout therefore says "no answer yet", not "failed": the
+# container is inspected by name for up to REMOVE_CONFIRM_TIMEOUT_SECONDS; only a 404 counts as gone.
+REMOVE_CONFIRM_TIMEOUT_SECONDS = 60.0
```
taiberium, line 167: `rental_probe._teardown` still caps this call at 120 seconds, while the new timeout and inspect windows can exceed 135 seconds. Please raise that cap to 300 seconds; non-blocking.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `52ca475`: After the six in-use retries, `_remove_named_volume` raises like any other volume error. The best-effort step logs it as "leaving it on the host" and the delete completes with `ContainerDeleted`. The redis rental record is removed and the pod's marker cleared. Only the container path still answers `DeletionInProgress`: the container still removing after the read timeout, or dockerd's 409. So nothing from the volume step can reach the backend's three delete attempts and the penalty path. Three tests cover it: the local volume, the external volume, and the re-ask after an in-progress answer. The s3fs plugin removal still does not run after a failed volume rm.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `52ca475`: `_PendingDeletionRegistry` keeps pod id to marked-at time on an injectable monotonic clock. Entries older than `PENDING_DELETION_TTL_SECONDS` (3600 s) are dropped on every `mark`, `is_pending` and `len`. The backend's whole re-ask span is three attempts about 10 min apart, so nothing older than an hour is coming. A probe pod's delete is never re-asked at all. Two tests: expiry with a restart on a new mark, and 60 probe ids at 6 min steps, which leave 10 and then none.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `52ca475`: `_TEARDOWN_DEADLINE_SECONDS = 300`. The slowest delete path on the inspect-confirmed route is about 190 s. It adds up from the 30 s stop grace, the SDK's 60 s read timeout and the 60 s inspect window. One 15 s inspect past the window and five 5 s volume waits complete the sum. The cap wins over the delete. A delete still running at 300 s is cancelled and the probe records that it did not finish. The by-name `docker rm -fv` in `_step_teardown` then runs as the backstop, as before. A delete that finishes inside its windows is no longer cut off at 120 s mid-inspect. The comment above the constant carries the arithmetic.

surcyf123:
@taiberium Done in `52ca475`: All three inline comments are answered on their threads. Head 52ca475: the exhausted volume is best effort again, `pending_deletions` expires after 1 h, and the probe's teardown cap is 300 s. Validators suite on EC2: 2656 passed.

neurons/validators/src/services/docker_service.py
```diff
@@ -6471,6 +6562,81 @@ def _failed_delete(self, payload: ContainerDeleteRequest, msg: str) -> FailedCon
...
+            miner_hotkey=payload.miner_hotkey,
+            executor_id=payload.executor_id,
+            pod_id=payload.pod_id,
+            workload_kind=payload.workload_kind,
+            msg=msg,
+            error_type=FailedContainerErrorTypes.ContainerDeletionFailed,
+            error_code=FailedContainerErrorCodes.DeletionInProgress,
+        )
+
+    async def _confirm_removal_after_timeout(
```
taiberium: NIT: this returns a container status, not a yes or no answer. Rename it to `_container_status_after_removal_timeout`.

neurons/validators/src/services/docker_service.py
```diff
@@ -6546,16 +6712,44 @@ async def _force_remove_container(
...
-                    pod_id=payload.pod_id,
-                    workload_kind=payload.workload_kind,
-                    msg=error_msg,
-                    error_type=FailedContainerErrorTypes.ContainerDeletionFailed,
-                    error_code=FailedContainerErrorCodes.DeletionInProgress,
+                return self._ForcedRemoval(
+                    failure=self._deletion_in_progress(payload, msg=error_msg)
                 )
 
+            if _is_docker_read_timeout_error(exc):
```
taiberium, line 6863: NIT: `_force_remove_container` grew to 73 lines. Move this read-timeout branch into a helper that returns `_ForcedRemoval | None`, and re-raise in the caller when it returns None.

taiberium (approved):

neurons/validators/src/services/docker_service.py
surcyf123: Renamed in edd4a786: `docker_service.py:6592` is now `_container_status_after_removal_timeout`, the one call site follows. Nothing else named it.

neurons/validators/src/services/docker_service.py
surcyf123: Done in 9955c70f: the read-timeout branch is `_removal_after_read_timeout` (`docker_service.py:6686`), returning `_ForcedRemoval | None`, and `_force_remove_container` (now 44 lines) re-raises the original timeout when it gets None. Same three outcomes and log lines as before; the 222 docker-service tests and the full validators suite (2658 passed) are unchanged.

surcyf123:
Final head is `9955c70`. The 8 `Done in` replies above name `13c8d1b`, `1ff8359`, `52ca475`, an earlier push of this round. Every fix named there is on `9955c70`.

surcyf123:
eng-drain-1200-io: restacking this branch on main (hold until 2026-09-23T14:03:55Z).

surcyf123:
Restacking on main 5754d83: conflicts in `neurons/validators/tests/test_rental_docker_sdk.py`; re-review from 52ca475.

This is a rebase (all commits are surcyf123's); the old head was 418a187. The only conflict was two test blocks appended at the same spot: #1377's SSH-keepalive adapter tests and this PR's `container_status` tests. Both are kept, and no source line changed in the resolution. Validators: 2900 passed, 15 skipped, and 9 failures in `test_port_mapping_dao.py` that also fail on plain main (venv sqlmodel tz). Ran on cloud VM, not EC2.

surcyf123:
**lium-io#1365 read-clean at 062dc01: all taiberium change-request threads (5 rounds) are answered in code with loop replies, the delete stays idempotent (inspect-confirmed removal, pending-deletion marker with a 1 h TTL, exhausted volume retries best-effort, not DeletionInProgress), validators suite green on this head; 0 findings.**

Fresh read at `062dc01d78ffbc4599e466d9ab3372b6227795ec` by `subagent-eng-verdict-drain-b` (it wrote none of this PR). 0 high, 0 medium, 0 low.

```json
{
 "sha": "062dc01d78ffbc4599e466d9ab3372b6227795ec",
 "verdict": "clean",
 "checks": 10,
 "change_requests": [
  {
   "reviewer": "taiberium",
   "threads": "4044383594 4044383600 4044383607 4045176560 4045753481 4045753485 4045753491 4059797003 4059797007",
   "answered": "each Done in <sha> with a loop reply; approved 52ca475; the two round-5 NITs (rename, helper extract) are 8b5074f7 and 062dc01d"
  }
 ],
 "findings": [],
 "dismissed": [
  {
   "line": 7401,
   "why": "pending_deletions.mark runs only on the DeletionInProgress return; exception paths leave no marker, and a stale one expires after PENDING_DELETION_TTL_SECONDS"
  },
  {
   "path": "neurons/validators/src/services/rental_docker_sdk.py",
   "line": 357,
   "why": "container_status returns '' for an inspect without State; the caller reads it as unknown, never as gone"
  }
 ],
 "evidence": {
  "riskiest": {
   "line": 7060,
   "why": "_container_status_after_removal_timeout decides whether a timed-out rm is reported as ContainerDeleted to the backend (billing closes the rental on it)",
   "checked_by": "read: only a 404 (None) is success; removing/unknown/other status stays DeletionInProgress or failure; tests in test_docker_service.py cover 404, removing at deadline, inspect timeout"
  },
  "ran": [
   {
    "cmd": "cd neurons/validators && pdm install && pytest tests/ -q (CI env vars)",
    "result": "2909 passed, 15 skipped in 179 s"
   }
  ],
  "edge_cases": [
   "connect timeout (not read timeout) → stays a failure, no inspect",
   "volume in use after 6 retries → logged, delete completes ContainerDeleted",
   "re-ask after DeletionInProgress finds container absent → pending marker keeps the in-use retry"
  ],
  "callers": [
   "neurons/validators/src/services/docker_service.py:7392 (_force_remove_or_kill)",
   "neurons/validators/src/services/task/checks/rental_probe.py:107 (_TEARDOWN_DEADLINE_SECONDS=300 around delete_container)"
  ],
  "callers_none_why": "",
  "reviewer_page": "checked taiberium threads: all answered"
 },
 "high": 0,
 "medium": 0,
 "low": 0,
}
```

PR #1318, opened 2026-09-07
DAH-3188 - [P1] security updates for 10 Dependabot alerts

surcyf123:
Fixes [DAH-3188](https://app.notion.com/p/security-dependency-updates-for-10-Dependabot-alerts-lium-io-3d4b8bfdbde981d79439c46b59fcb7bd) · reviewer: @taiberium

**Problem** — 10 open Dependabot alerts on the SGX key-provider's Cargo.lock (4 high); `cargo update --precise` against the pinned upstream closes 8. hickory-proto 0.25.2 stays: upstream's reqwest needs ^0.25 and one of its two alerts has no fix.

**Value** — Closes 8 of the 10 open Dependabot alerts on the key-provider `Cargo.lock` (4 high, 2 medium, 4 low open at 16 Sep 15:25Z) with a lockfile-only change.

**Goal** — [DAH-3188](https://app.notion.com/p/security-dependency-updates-for-10-Dependabot-alerts-lium-io-3d4b8bfdbde981d79439c46b59fcb7bd) (P1) · serves SO §33, SO §37.

**Fits** — Sits in `lium-io` — `neurons/executor`; touches backend.

**Flag:** no flag

**Risks**
- Merging ships nothing; the next key-provider rebuild on a TDX host changes MRENCLAVE and the sealing key, and every CVM data disk on that host (stopped CVMs included) is unreadable at its next boot — the provider notices as the `Failed to open encrypted data disk` loop — no rollback once rebuilt, so the base stays #1366 (the guard that refuses the rebuild while any disk exists) and the PR stays a draft until it merges.

**How to verify**
- `git clone https://github.com/MoeMahhouk/gramine-sealing-key-provider && git checkout 327e8b5 && cp <this>/Cargo.lock . && cargo build --release --locked` → builds with rust 1.85
- `docker compose build` in `key-provider/` (needs an SGX host for the run, not for the build)
- release: per TDX host, stop the CVMs, remove their disks by hand, then the guarded `docker compose up --build -d` in `key-provider/`, recreate each CVM; no whitelist change (Details, Release)

**Verified by the loop:** Gate: PASS 6163da4

**Ran it:** `cargo update -p quinn-proto@0.11.13 --precise 0.11.15` (+ rustls-webpki, rand ×2)… · sandbox EC2 aws_run 20260907T165436Z-1l3s; not run in an SGX enclave · head moved to 6163da4 — re-run

[...]

PR #1366, opened 2026-09-14
DAH-3471 - [P1] CVM upgrade guard: refuse the rebuild while any CVM disk exists

surcyf123:
Implements [DAH-3471](https://app.notion.com/p/3dbb8bfdbde9818f970ace4f2f88094e) · reviewer: @taiberium

**Problem** — Rebuilding the SGX key provider changes MRENCLAVE and every CVM data disk on the host is lost at its next boot; two scripts could trigger it unchecked.

**Value** — Guards the CVM data disks on the 6 attested TDX hosts (148 of 527 active nodes are TDX-capable, prod 16 Sep) from a rebuild that would wipe every one at its next boot; it is also the merge condition Mikhail set for lium-io#1318 (4 high alerts).

**Goal** — [DAH-3471](https://app.notion.com/p/3dbb8bfdbde9818f970ace4f2f88094e) (P1).

**Fits** — Sits in `lium-io` — `neurons/executor`; touches CLI/SDK, docs, tests (41).

**Flag:** no flag

**Risks**
- First `lium-cvm.sh run` after pulling adopts the running images, writes pin + registry, recreates both containers once on the same ids — rollback: previous tag
- A forgotten `hda.img` blocks a legitimate `upgrade` — the operator sees the path, and the `rm -rf` line when the manifest or `runtime.json` proves the CVM stack made it; the guard never deletes
- `new`/`run` wait ≤ 60 s for the lock, exit 5 after; a checkout outside the sweep roots is unseen until it runs once
- A `build:` or an override file put back into `key-provider/` stops the guard (exit 1, the file is named)

**How to verify**
- `cd neurons/executor && pdm run pytest tests/test_cvm_upgrade_guard.py -q` → 39 passed, 2 skipped as root on a box without docker (40 non-root; the real `docker compose config` test needs docker; Linux `flock`)
- `cd neurons/executor/dstacktee/key-provider && docker compose build` → `No services to build`, image set unchanged (EC2 aws_run 20260915T114924Z-gkc3)
- `sudo ./cvm_upgrade_guard.sh upgrade` on a host with a CVM disk → exit 3, `docker images -q | sort -u | wc -l` unchanged

**Verified by the loop:** Gate: PASS 2c2919c

**Ran it:** EC2 aws_verify 20260915T120926Z-zkp9 @ 2c2919c: executor 270 passed, 2 skipped as root · not run on macOS (`flock`, SGX); TDX run pending · … +1 under Details

**Self-review:** clean at 2c2919c (15 checks)

**Parts:** backend n/a — host tooling · migration n/a — no database · frontend n/a — no UI · CLI/SDK ✓ `cvm_upgrade_guard.sh`, `lium-cvm.sh` · docs ✓ `dstacktee/docs/host-setup.md` · tests ✓ 41 · dashboards n/a — no metric

**Docs:** `neurons/executor/dstacktee/docs/host-setup.md` §4, §6.1, troubleshooting · public: none needed because the rebuild was never documented on docs.lium.io and is an operator-only host action

### What changed, in full
- `cvm_upgrade_guard.sh inventory`: every `hda.img` in this checkout, the registered VM roots and a sweep of `/home /root /opt /srv /mnt /data`; exit 3 with the manual removal line per running or stopped disk (an orphan is named with a `#` line) (`neurons/executor/dstacktee/cvm_upgrade_guard.sh`)
- `cvm_upgrade_guard.sh upgrade`: refused while any disk exists (3) or a registered root is missing or unreadable (4); keeps the old image as `lium-key-provider:pre-upgrade-<stamp>`; `--dry-run` reports and writes nothing (`cvm_upgrade_guard.sh`)
- `cvm_upgrade_guard.sh start`: `docker compose up -d --no-build` on the image id pinned in `/var/lib/lium-cvm/key-provider.image`; adopts the running containers' images on an older host; exit 6 with recovery steps when the pinned image is missing; exit 1 when docker or `flock` is unavailable (`cvm_upgrade_guard.sh`)
- `lium-cvm.sh new` and `run` take `/var/lock/lium-cvm.lock` and register their VM root; `run` also starts the provider through the guard and creates `hda.img` before releasing the lock; `new`'s `curl ifconfig.me` gets `-m 10` so a slow lookup does not hold the lock (`neurons/executor/dstacktee/lium-cvm.sh`)
- `key-provider/run.sh` calls the guard (`run.sh` → start, `run.sh upgrade`); `docker-compose.yaml` names the images `lium-key-provider:local` and `lium-aesmd:local` and carries no `build:`; the `build:` sections live in `docker-compose.build.yaml`, and only `cvm_guard_compose_build` passes it (`docker compose -f docker-compose.yaml -f docker-compose.build.yaml build`), so `docker compose build` or `docker compose up` by hand in `key-provider/` builds nothing; the preflight exits 1 when `docker-compose.yaml` declares a `build:` again or an auto-loaded compose file (`docker-compose.yml`, `docker-compose.override.yaml`, `compose.yaml`, their variants) sits in `key-provider/`; every guard compose call names `docker-compose.yaml` with `-f`, so an override file or `COMPOSE_FILE` never changes what the guard runs (`key-provider/run.sh`, `key-provider/docker-compose.yaml`, `key-provider/docker-compose.build.yaml`, `cvm_upgrade_guard.sh`)
- `scripts/cvm_guard_validation.sh`: the SGX/TDX validation, four phases, PASS/FAIL per case, removes no disk (`neurons/executor/dstacktee/scripts/cvm_guard_validation.sh`); docs: `docs/host-setup.md` §4, §6.1, troubleshooting rows, `README.md`

### Risks, in full
- A provider who pulls this and runs `lium-cvm.sh run` gets the guard's `start`: on a host with existing `dstack-key-provider` and `dstack-aesmd` containers it adopts both images (no build), writes the pin file, the `vm-dirs` registry and three tags (`lium-key-provider:local`, `:pinned-<id>`, `lium-aesmd:local`), and `up -d --no-build` recreates both containers once on the same image ids under the new names — the provider notices a few `[cvm-guard]` lines and that recreate — rollback: `git checkout` the previous tag; the pin file and tags stay and are harmless
- The guard refuses a legitimate upgrade while a forgotten `hda.img` exists somewhere on the host — the operator running `upgrade` sees the path and, for a disk with `vm-manifest.json` or `runtime.json` beside it, the exact `rm -rf` line; an orphan is named with a comment line and nothing to run; the override is that manual removal, there is no deletion path in the guard
- `lium-cvm.sh new`/`run` wait up to 60 s for the host lock and exit 5 after that — an operator starting two CVMs at once sees the second wait
- A checkout outside the sweep roots (`/usr/local`, `/var/lib`, a top-level dir) that never ran this version of `lium-cvm.sh` is not in the registry and not swept, so an `upgrade` from another checkout does not see its disks — `LIUM_CVM_SWEEP_ROOTS` widens the sweep; the first `lium-cvm.sh run` in that checkout registers it — a real `upgrade` is a planned host operation after `stop` per CVM (#1318 Release), the operator knows the host's checkouts

### How to verify, in full
- `cd neurons/executor && pdm run pytest tests/test_cvm_upgrade_guard.py -q` → 39 passed, 2 skipped under root on a box without docker (observed on EC2: the unreadable-directory test and the real `docker compose config` test); as a non-root user (CI) the unreadable-directory test runs too; the host-layout tests skip on macOS (`flock`)
- `cd neurons/executor/dstacktee/key-provider && docker compose build` → nothing to build, image set unchanged; `docker compose up -d` → fails to pull `lium-key-provider:local`, builds nothing (EC2 with docker, run-id in Ran it)
- `sudo ./cvm_upgrade_guard.sh inventory` on a host with a stopped CVM → exit 3, its `hda.img` listed with `sudo rm -rf /<checkout>/neurons/executor/dstacktee/run/vms/<name>`
- `sudo ./cvm_upgrade_guard.sh upgrade` on that host → exit 3, `docker images -q | sort -u | wc -l` unchanged before and after
- `sudo scripts/cvm_guard_validation.sh phase1 <cvm>` on a TDX host → PASS per case (Details, Validation plan)

### Design

**What breaks and why the guard exists.** The key provider (`key-provider/Dockerfile.key-provider`) is a Gramine SGX enclave. Gramine gives it a sealing key bound to MRENCLAVE; the provider hashes it with the CVM's MRTD and returns that as the CVM's key ([gramine-sealing-key-provider at the Dockerfile's pinned commit](https://github.com/MoeMahhouk/gramine-sealing-key-provider/commit/327e8b52515ab671dcadb75014b7ca7e272d86bc), `src/gramine/interface.rs:7`, `src/crypto/keys.rs:18-19`). Inside the guest, dstack 0.5.11 derives the encrypted data-disk key from it (`dstack-util/src/main.rs:383`). So a rebuilt enclave means a new MRENCLAVE, a new sealing key, and every CVM on the host, running or stopped, fails `Failed to open encrypted data disk` at its next boot. Before this PR three paths could rebuild: `key-provider/run.sh` (`docker compose up --build -d`), `lium-cvm.sh run` (`docker-compose up -d`, which builds when the image is missing) and the documented `cd key-provider && docker compose up --build -d` in `docs/host-setup.md` §4.

**Inventory** (`cvm_guard_inventory`). A CVM disk is a directory with `hda.img`; the registry walk and the sweep share that one predicate (`cvm_guard_classify_dir`). `vm-manifest.json` and `runtime.json` (the layout `scripts/dstack.py` writes; `hda.img` is created at first `run`, `runtime.json` exists while QEMU runs) only refine the state: `running`, `stopped`, or `orphan` when the CVM is not running and its manifest is gone. An orphan blocks like any other disk (`test_swept_disk_without_manifest_is_an_orphan_and_refuses`, `test_registered_disk_without_manifest_is_an_orphan`). Sources, in order: this checkout's `run/vms`; every line of `/var/lib/lium-cvm/vm-dirs`, which `lium-cvm.sh new` and `run` append their `run/vms` to, so a host with several checkouts sees all of them whichever checkout runs the guard; a `find -name hda.img` sweep of `/home /root /opt /srv /mnt /data` (`LIUM_CVM_SWEEP_ROOTS`) for a VM directory nothing registered. A sweep root that is a symlink (`/data -> /mnt/nvme0`) is entered (`find "$root/"`) and its disks are reported through the link. The sweep stays on one filesystem (`-xdev`) and then descends into every real filesystem mounted below those roots (`cvm_guard_submounts` reads `/proc/self/mounts` and skips overlay, proc, sysfs, cgroup and the other pseudo filesystems; tmpfs is walked, it can hold files), so a data disk on a separate mount is seen and a pseudo filesystem is not walked. A registered root that is missing (an unmounted disk looks like a removed checkout), a root or a sweep directory that is not readable, or an unreadable registry makes the inventory incomplete: exit 4, and `upgrade` refuses until the operator mounts it or removes the stale registry line. A directory with a manifest and no `hda.img` (created, never run) is listed as "no data disk yet" and does not block: no key has been derived for it.

**Refusal** (`cvm_guard_upgrade`). Exit 3 prints one line per disk (`running`/`stopped`/`orphan`, path) and then, per disk, `sudo <checkout>/lium-cvm.sh stop <name>` when it runs and `sudo rm -rf <dir>` when the manifest or `runtime.json` proves the CVM stack made the directory; an `orphan` gets a `#` line naming the path instead (review round 2, Rustam). The guard runs neither; there is no flag that deletes. `upgrade --dry-run` runs the inventory and stops before the pin or adopt step; it writes nothing.

**Pin** (`cvm_guard_start`, `cvm_guard_write_pin`). `docker-compose.yaml` now names the images (`lium-key-provider:local`, `lium-aesmd:local`), so the id is addressable. Before anything else `cvm_guard_preflight` requires `flock` and a reachable Docker daemon (exit 1 with the reason), so a stopped daemon is not reported as a missing image. The pin file holds the image id; `write_pin` also tags the id `lium-key-provider:pinned-<id>` so a later build that moves `:local` leaves the old layers tagged, not dangling. `start` runs `docker compose up -d --no-build`, then checks the container's image is the pinned id (exit 6 otherwise). A hand build is no longer possible through compose: `docker-compose.yaml` has no `build:` (review round 3, Rustam: the file kept its `build:` sections, so `docker compose build` walked past the guard), the `build:` sections sit in `docker-compose.build.yaml`, and `cvm_guard_compose_build` is the one function that passes both files. The preflight greps `docker-compose.yaml` for a `build:` and exits 1 with the file name when it finds one, and exits 1 as well when a file compose loads on its own (`docker-compose.yml`, which compose prefers over `.yaml`, `docker-compose.override.yaml`, `compose.yaml`, the `.yml` and `compose.override` variants) sits in `key-provider/`, so the invariant cannot rot silently (`test_guard_refuses_a_compose_file_that_declares_build`, `test_guard_refuses_an_auto_loaded_compose_file_beside_the_runtime_file`). The guard's own `up`, `build` and the stub-checked runtime calls pass `-f docker-compose.yaml`, so neither such a file nor a `COMPOSE_FILE` variable reaches them. `COMPOSE_FILE` set by hand to the build file is the same deliberate act as typing `-f` and is not refused. A `docker build -t lium-key-provider:local` typed by hand is still possible; if `:local` was moved that way, `start` retags the pinned id back before `up`. If the pinned id is not on the host at all, exit 6 with the recovery text (restore the image, or remove the disks by hand and `upgrade`); nothing is built. A host without a pin file adopts, in order: the image the `dstack-key-provider` container runs, `lium-key-provider:local`, the name compose gave the build before this PR (`key-provider-gramine-sealing-key-provider` or the v1 form). Only a host with no pin, nothing to adopt and a complete, empty inventory builds. `aesmd` (the SGX AESM sidecar, no key material) is adopted from the running `dstack-aesmd` container or the old compose name, and built only when neither exists; the key provider is never built by `start` while a disk exists.

**Lock** (`cvm_guard_lock`). `flock -w $LIUM_CVM_LOCK_WAIT` (default 60 s) on `/var/lock/lium-cvm.lock`, fd 9, exit 5 with the path when it stays busy. `lium-cvm.sh` sources the guard as a library. `new` takes the lock before writing the VM directory. `run` takes it before `cvm_guard_start`, then creates `hda.img` itself (`qemu-img create -f qcow2 … <disk_size>G`, the size from `vm-manifest.json`; `dstack.py` keeps an existing disk) and releases the lock before `dstack.py run`. That ordering closes the race arhangel66 named: once `run` has the lock, an `upgrade` waits; once `run` releases it, `hda.img` exists and the upgrade's inventory refuses. The guest fetching its key after that is safe because no image switch can happen any more.

**Compatibility.** A provider on the current tag has a container from `docker compose up --build` and no pin file; the first `lium-cvm.sh run` after pulling this adopts that container's image, pins it, and `up -d --no-build` recreates the container under the new image name with the same image id. `lium-cvm.sh check`'s hint now names `cvm_upgrade_guard.sh start`. `lium-cvm.sh inventory` is the guard's inventory.

### Validation plan (SGX/TDX host, Muhammad or Pixel)

`scripts/cvm_guard_validation.sh` prints PASS/FAIL per case and never removes a disk. Run as root from the checkout's `dstacktee` directory with one CVM created and running; paste each phase's output here.

| phase | when | cases |
|---|---|---|
| `phase1 <cvm>` | CVM running | inventory lists it (3) · `upgrade` refused (3), pin and container unchanged, `hda.img` intact · a hand-run `docker compose build` in `key-provider/` exits 0 with the image set unchanged · SSH host key, API, `mr_enclave` recorded · `stop`, then inventory shows `stopped` and `upgrade` still refused · `start` with a bogus pin in a scratch state dir: exit 6 with recovery text, image count unchanged · `lium-cvm.sh run` boots the CVM again on the pinned image, SSH key unchanged |
| `phase2 <cvm>` | after a host reboot | key provider up on the pinned image without a build · CVM boots, API answers, no `Failed to open encrypted data disk` on the console, SSH host key identical to phase1, `mr_enclave` unchanged |
| `phase3 <cvm>` | after the operator removed the listed disks by hand | inventory 0 · `upgrade` 0, new pin, container on it, `pre-upgrade-*` tag present · new `mr_enclave` differs · `lium-cvm.sh new` + `run`: boots, API answers, new SSH key recorded |
| `phase4 <cvm>` | after `lium-cvm.sh stop` + phase4's `run` | new CVM boots, no disk error, SSH host key identical to phase3, provider still on the pin |

Data preservation is read from outside the guest: no disk error on the serial console, the executor API answers, and the SSH host key on `SSH_PUBLIC_PORT` (it lives on the encrypted data disk) is the one recorded before. `GUEST_EXEC="ssh -p <port> root@127.0.0.1"` adds a marker file written in phase1/phase3 and read back in phase2/phase4. The validator's `Attestation verified` line and the verifier's `key_provider_info` id after phase3 are recorded by hand on this PR.

### Release order for #1318

This PR lands first. #1318 rebases onto it; its Release section then reads: merge, `git pull` per host, drain, `lium-cvm.sh stop` per CVM, remove each `run/vms/<name>` by hand once its data is not needed, `sudo ./cvm_upgrade_guard.sh upgrade` (refused while any disk remains), read `mr_enclave`, `lium-cvm.sh new` + `run` per CVM.

**Cross-PRs.** [#1264](https://github.com/Datura-ai/lium-io/pull/1264) (DAH-2780) edits `docs/host-setup.md` too and renumbers §6→§7: whichever lands second rebases; if it is this one, §6.1 becomes §7.1 and the two pointers (`lium-cvm.sh check`'s hint, `README.md`) move with it. [#1274](https://github.com/Datura-ai/lium-io/pull/1274) (DAH-2859) touches `lium-cvm.sh` (`check`/`stop`, no overlap with `new`/`run`) and adds a §4 line about a by-hand `gramine:1.9` rebuild; after this PR that rebuild is `cvm_upgrade_guard.sh upgrade`, and #1274's line should say so when it rebases. [#1275](https://github.com/Datura-ai/lium-io/pull/1275) (DAH-2860) re-pins `Dockerfile.key-provider`: it is a §6.1 release like #1318, same order (this PR first, then disks removed by hand, then `upgrade`).

### Known follow-ups
- `docs.lium.io` provider page: a pointer to §6.1 when #1318 releases (docs are in lium-docs until the `apps/docs` import).

### Self-review

Verdict `clean` at `2c2919c` · 15 checks · by subagent-r218-1366-r2 · 2026-09-15 12:14Z (tools/pr_self_review.py)

Mechanical notes dismissed: `neurons/executor/dstacktee/cvm_upgrade_guard.sh:94` Fold of the round-2 medium verified: CVM_GUARD_AUTOLOAD_FILES now lists all seven names compose discovers without -f (docker-compose.yml first, plus the four override names and compose.yaml/.yml), matching compose-go DefaultFileNames minus docker-compose.yaml itself and DefaultOverrideFileNames in full; reproduced against the sourced guard with a copy of key-provider/ plus a docker-compose.yml declaring build: -> cvm_guard_check_compose_file prints '<file> exists' and returns 1. · `neurons/executor/tests/test_cvm_upgrade_guard.py:42` Fold verified: AUTOLOAD_NAMES is read from the guard by sourcing it at import (bash -c, check=True), asserted to contain docker-compose.yml/docker-compose.override.yaml/compose.yaml, and drives the 7-way parametrize at line 353 and the per-name no-extra-file assertion plus the exact {docker-compose.yaml, docker-compose.build.yaml} `*.y*ml` set at lines 295-301, which matches the directory listing; sourcing the guard has no side effects (assignments and function definitions only). · `neurons/executor/dstacktee/scripts/cvm_guard_validation.sh:168` Fold of the round-2 low verified: case 5 now compares `docker images -q | sort -u` before and after, the same kind of value case 2b stores in images_before; bash -n and shellcheck -S warning clean. · `neurons/executor/dstacktee/docs/host-setup.md:96` Fold verified: §4 'Why the guard' and the exit-1 troubleshooting row now name docker-compose.yml among the auto-loaded files. · `review-wave/readability/overrides/lium-io-1366.json:1` Body counts re-checked against the file at this head: 35 `def test_` plus a 7-case parametrize = 41 items ('tests ✓ 41'); root without docker skips the unreadable-directory test and the real-compose test -> 39 passed, 2 skipped; non-root without docker -> 40 passed ('40 non-root'); the 'new guard tests (11 cases)' = 1 parsed-files + 1 real-compose + 2 build:-in-runtime-file + 7 parametrized; 270 passed in the executor suite is 265 + the 5 added items. The file-name lists in What changed in full, Design, host-setup.md and the Ran-it paragraph match CVM_GUARD_AUTOLOAD_FILES. The How-to-verify `docker compose build` line still cites EC2 run 20260915T114924Z-gkc3 at the previous head, which is valid: neither compose file changed between 8fc5931 and 2c2919c. · `neurons/executor/tests/test_cvm_upgrade_guard.py:239` pytest.skip in the `host` fixture when util-linux flock is absent is a platform guard (macOS); stated in the module docstring and the body; the tests run on the Linux gate. · `neurons/executor/tests/test_cvm_upgrade_guard.py:259` pytest.skip in `_service_config` when the docker compose plugin is absent: the same no-build property is asserted on every box by test_compose_file_names_images_and_declares_no_build; the body states the skip. · `neurons/executor/tests/test_cvm_upgrade_guard.py:552` pytest.skip under euid 0 is correct: root reads a chmod-0 directory, so the unreadable-root case cannot be produced as root; the body states '2 skipped as root'.

### Verified

Gate: PASS (2c2919c) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1366)

### What changed

- `cvm_upgrade_guard.sh` `inventory`/`upgrade`: exit 3 with a `rm -rf` line per running or stopped disk, 4 inventory incomplete; `start`: 6 pin missing
- An `orphan` (`hda.img` with no `vm-manifest.json` or `runtime.json` beside it) still blocks; it is named, with no `rm` line
- `lium-cvm.sh new`/`run`: host lock, VM-root registry, `hda.img` created under the lock
- `docker-compose.yaml` has no `build:`; only the guard passes `docker-compose.build.yaml`, so a hand-run `docker compose build` builds nothing (round 3)
- `scripts/cvm_guard_validation.sh` (TDX validation, PASS/FAIL); docs

### Ran it

Round 3 (head 2c2919c), EC2 aws_run `20260915T114924Z-gkc3` (Amazon Linux 2023, real docker 28 + compose plugin, the `key-provider/` directory of this head uploaded, no image on the box):

```
$ cd key-provider && docker compose build
level=warning msg="No services to build"
[exit 0]      docker images -q | wc -l: 0 before, 0 after
$ docker compose up -d
 Image lium-key-provider:local Error pull access denied for lium-key-provider, repository does not exist …
Error response from daemon: No such image: lium-aesmd:local
[exit 1]      nothing built, nothing pulled
$ docker compose config --format json | jq -c '.services|map_values(has("build"))'
{"aesmd":false,"gramine-sealing-key-provider":false}
$ docker compose -f docker-compose.yaml -f docker-compose.build.yaml config --format json | jq -c '.services|map_values(has("build"))'
{"aesmd":true,"gramine-sealing-key-provider":true}
```

Tests at 2c2919c: `pdm run pytest tests/` → 270 passed, 2 skipped on the warm lium-io box, aws_verify run `20260915T120926Z-zkp9` (`RECOVERY/aws_run/`); the two skips are the unreadable-directory test (root) and `test_real_compose_sees_nothing_to_build_without_the_build_file` (no docker on that box; the same check is the `config` output above). The new guard tests (11 cases): the two compose files parsed (no `build:` in the runtime file, `build:` only in the build file, dockerfiles present, no override file), real `docker compose config` sees `build` only with the build file, a checkout whose `docker-compose.yaml` gains a `build:` is refused by `start` and `upgrade` before any docker call (exit 1, no build, pin unchanged; also on an empty host), and each name compose loads on its own (`docker-compose.yml`, `docker-compose.override.yaml`, `compose.yaml`, their variants; the list is read from the guard) beside the runtime file is refused the same way; the empty-host upgrade test now asserts exactly one build, through both files.

Earlier rounds, EC2 aws_run `20260914T115109Z-tx60` (`RECOVERY/aws_run/20260914T115109Z-tx60/out/ranit.txt`, docker stubbed, head 7d9febb): two checkouts under the sweep root, `exec-a` running in `opt/lium-io`, `exec-old` stopped in `home/operator/lium-io-old`, pin `sha256:aaaa…`.

```
$ cvm_upgrade_guard.sh upgrade
[cvm-guard] 2 CVM disk(s) exist on this host:
  running  /job/work/host/sweep/opt/lium-io/neurons/executor/dstacktee/run/vms/exec-a/hda.img
  stopped  /job/work/host/sweep/home/operator/lium-io-old/neurons/executor/dstacktee/run/vms/exec-old/hda.img
[cvm-guard] The upgrade is refused while these disks exist. Each data disk is lost when it is removed.
  sudo /job/work/host/sweep/opt/lium-io/neurons/executor/dstacktee/lium-cvm.sh stop exec-a
  sudo rm -rf /job/work/host/sweep/opt/lium-io/neurons/executor/dstacktee/run/vms/exec-a
  sudo rm -rf /job/work/host/sweep/home/operator/lium-io-old/neurons/executor/dstacktee/run/vms/exec-old
[exit 3]
$ cvm_upgrade_guard.sh start
[cvm-guard] starting the key provider from sha256:aaaa… (no build)
[exit 0]
$ cvm_upgrade_guard.sh upgrade --dry-run        # while another process holds the lock
[cvm-guard] ERROR: another lium-cvm.sh or cvm_upgrade_guard.sh holds /job/work/host/lium-cvm.lock; waited 2s. Let it finish, then retry.
[exit 5]
$ rm -rf …/exec-a …/exec-old                     # by hand, then:
$ cvm_upgrade_guard.sh upgrade
[cvm-guard] previous image sha256:aaaa… kept as lium-key-provider:pre-upgrade-20260914T115229Z
[cvm-guard] pinned key-provider image sha256:<new id> in /job/work/host/state/key-provider.image
[exit 0]
```

`cvm_guard_submounts` is unit-tested against a fake mount table (filter, `\040` decoding, root prefix); whether `find -xdev` then crosses into those roots needs a real mount boundary, which phase1 of the validation script exercises on a real host. Not run on an SGX/TDX host: the loop has none.

</details>


neurons/executor/dstacktee/cvm_upgrade_guard.sh
```diff
@@ -0,0 +1,524 @@
...
+    for root in "${sweep[@]}"; do
+        # The trailing slash makes find enter a root that is a symlink
+        # (/data -> /mnt/nvme0); in -P mode a bare symlink start point is skipped.
+        found="$(find "$root/" -xdev -type f -name hda.img -print 2>>"$err_file")" || true
+        while IFS= read -r line; do
+            [ -n "$line" ] || continue
+            dir="$(dirname "$line")"
+            [ -n "${seen[$dir]:-}" ] && continue
+            seen[$dir]=1
+            [ -f "$dir/vm-manifest.json" ] || continue
```
taiberium: NIT: the registry walk counts any `hda.img`, the sweep only one beside a manifest. A disk whose manifest is gone is then swept past and the upgrade proceeds.

taiberium (approved):

neurons/executor/dstacktee/cvm_upgrade_guard.sh
surcyf123: Done in `b11fdeb`: the sweep and the registry walk now share one predicate: any `hda.img` is a CVM disk. A disk with no manifest is listed as `orphan` and still exits 3 with its `rm -rf` line (3 new tests, 260 passed on EC2).

neurons/executor/tests/test_cvm_upgrade_guard.py
```diff
@@ -0,0 +1,702 @@
...
+
+    def seq(self) -> list[str]:
+        log = self.docker_state / "seq.log"
+        return log.read_text().splitlines() if log.exists() else []
+
+    def calls(self) -> list[str]:
+        log = self.docker_state / "calls.log"
+        return log.read_text().splitlines() if log.exists() else []
+
+    def env(self, lock_wait: int = 5) -> dict:
```
taiberium, line 163: NIT: `env` returns a bare `dict`, and `guard` and `lium_cvm` have no return type. Annotate them `dict[str, str]` and `subprocess.CompletedProcess[str]`.

taiberium (approved):

neurons/executor/tests/test_cvm_upgrade_guard.py
surcyf123: On it. Adding the return types in the test. Fix lands within the hour.

surcyf123: On it. Adding the return types in the test. Fix lands within the hour.

surcyf123: Done in `6324771`: `env` returns `dict[str, str]`; `guard` and `lium_cvm` return `subprocess.CompletedProcess[str]`.

neurons/executor/dstacktee/cvm_upgrade_guard.sh
```diff
@@ -0,0 +1,537 @@
...
+        state="${entry%%	*}"
+        dir="${entry#*	}"
+        name="$(basename "$dir")"
+        # <checkout>/run/vms/<name> is the layout lium-cvm.sh writes; a swept
+        # directory elsewhere gets only the rm line.
+        checkout="$(cd "$dir/../../.." 2>/dev/null && pwd)" || checkout=""
+        if [ "$state" = running ] && [ -x "$checkout/lium-cvm.sh" ]; then
+            printf '  sudo %q stop %q\n' "$checkout/lium-cvm.sh" "$name"
+        fi
+        printf '  sudo rm -rf %q\n' "$dir"
```
taiberium: NIT: this `rm -rf` line covers any directory that holds a file named `hda.img`, even one the CVM stack never made. Gate it on `vm-manifest.json`.

neurons/executor/dstacktee/cvm_upgrade_guard.sh
```diff
@@ -0,0 +1,537 @@
...
+    printf '%s %s\n' "$id" "$(date -u +%Y-%m-%dT%H:%M:%SZ)" >"$CVM_GUARD_PIN_FILE"
+    # A second tag keeps the image when a later build moves :local.
+    docker tag "$id" "lium-key-provider:pinned-${id#sha256:}" >/dev/null 2>&1 || true
+    cvm_guard_say "pinned key-provider image $id in $CVM_GUARD_PIN_FILE"
+}
+
+# On a host that predates the pin file, the image in use is the one the
+# dstack-key-provider container runs, else the image compose named before this
+# guard. Pins it. Returns 1 when nothing is there to adopt.
+cvm_guard_adopt_existing() {
```
taiberium: NIT: the name reads as adopting a CVM, but it adopts the key-provider image. Rename it `cvm_guard_adopt_existing_image`.

neurons/executor/dstacktee/cvm_upgrade_guard.sh
```diff
@@ -0,0 +1,537 @@
...
+        case " $CVM_GUARD_PSEUDO_FS " in *" $fstype "*) continue ;; esac
+        target="$(printf '%b' "$target")"
+        case "$target" in "$root"/*) echo "$target" ;; esac
+    done <"$CVM_GUARD_MOUNTS"
+}
+
+# Fill CVM_GUARD_DISKS / CVM_GUARD_UNREADABLE from: this checkout's run/vms,
+# every root in the registry, and a sweep of CVM_GUARD_SWEEP_ROOTS for any
+# hda.img the registry does not know about.
+cvm_guard_inventory() {
```
taiberium, line 245: NIT: 69 lines and two jobs here, the registry walk and the filesystem sweep. Move the sweep into its own function.

neurons/executor/tests/test_cvm_upgrade_guard.py
```diff
@@ -0,0 +1,704 @@
...
+
+    def pinned(self) -> str:
+        return (self.state / "key-provider.image").read_text().split()[0]
+
+    def register(self, vms_dir: Path) -> None:
+        self.state.mkdir(exist_ok=True)
+        with open(self.state / "vm-dirs", "a") as f:
+            f.write(str(vms_dir) + "\n")
+
+    def seq(self) -> list[str]:
```
taiberium: NIT: `seq` does not say what it returns, which is the order of every stub call. Rename it `stub_call_order`.

taiberium (approved):

taiberium:
**Necessity: question.**

The ticket proves the data-loss mechanism from the code, but gives no production number: no count of TDX hosts, no count of CVMs, no past incident. The change is +1631/-26 for a guard.

The two live rebuild paths it closes are `docker compose up --build` in `run.sh` and the implicit build in `lium-cvm.sh run`. Both close with the `image:` keys in `docker-compose.yaml`, the dropped `--build`, the host lock, and a refusal that reads this checkout's `run/vms`, about 80 lines.

The registry, the sweep of `/home /root /opt /srv /mnt /data`, the submount walk, the orphan state and the 281-line validation script cover a host with several checkouts, which the ticket does not show exists. A hand-run `docker compose build` in `key-provider/` still rebuilds the enclave, so the guard stays advisory either way.

taiberium (changes requested):
Requested changes.

taiberium:
**Correction: the necessity verdict is drop, not question.**

About 80 lines close both live rebuild paths. This PR is +1631/-26, so the version we need is a twentieth of it. A hand-run `docker compose build` in `key-provider/` still rebuilds the enclave, so the other 1550 lines do not close the hole they are written for.

No production number backs the rest: no count of TDX hosts, no count of CVMs, no past incident.

My APPROVE above is withdrawn. Rustam decides whether it ships.

neurons/executor/dstacktee/cvm_upgrade_guard.sh
surcyf123: Done in `e4ef2d0`: renamed to `cvm_guard_adopt_existing_image`, both call sites updated.

neurons/executor/dstacktee/cvm_upgrade_guard.sh
surcyf123: Done in `e4ef2d0`: the sweep is `cvm_guard_sweep_dirs`, which prints one directory per line; `cvm_guard_inventory` keeps the registry walk and the dedupe.

neurons/executor/tests/test_cvm_upgrade_guard.py
surcyf123: Done in `e4ef2d0`: renamed to `stub_call_order`.

neurons/executor/dstacktee/cvm_upgrade_guard.sh
surcyf123: Done in `e4ef2d0`: the rm line is printed only when `vm-manifest.json` or `runtime.json` sits beside `hda.img`. An orphan gets a comment line that names the path. The two orphan tests assert there is no rm line.

surcyf123:
@taiberium Fair point on the numbers. The ticket has no count of TDX hosts or CVMs and no past incident. The core you describe is about 80 lines: the `image:` keys, the dropped `--build`, the host lock, and a refusal that reads this checkout's `run/vms`. The rest guards a host with several checkouts, which we cannot show exists. Two options: keep this PR as approved and merge it, or I cut it to that core here and drop the registry, the sweep and the validation script. Which do you want?

taiberium:
**Correction: I over-stated the size argument. The block stands on the defect, not on the scope.**

Two things I checked after writing the drop.

The scope is not the bot's invention. DAH-3471 records that arhangel66 asked for this on lium-io#1318 and accepted this exact scope, so "it should be a twentieth of the size" argues against a decision a reviewer already made. I withdraw that.

The fleet it protects is small but real: prod has 11 TDX-capable hosts, 6 with attestation passed. The loss is irreversible, so caution is fair even with no past incident.

What still needs a change is the defect. `docker-compose.yaml` keeps its `build:` sections, so a hand-run `docker compose build` in `key-provider/` rebuilds the enclave and walks past the guard. A guard that the documented tool can bypass does not close the path it is written for. Close that, and the necessity question goes away.

surcyf123:
@taiberium Understood, the verdict is drop. The next round cuts this PR to the core you named: the `image:` keys, the dropped `--build`, the host lock, and a refusal that reads this checkout's `run/vms`. The registry, the sweep, the orphan state and the validation script go. This lane is at its time cap, so the cut is queued for the next lane on this same PR, not a new one. Agreed that a hand-run `docker compose build` still rebuilds the enclave; the guard stays advisory.

surcyf123:
@taiberium The four NITs are in `e4ef2d0`. The drop verdict on the size is answered in the thread below: the cut to the core is the next round on this PR.

surcyf123:
@taiberium Agreed on both points. The scope stays as arhangel66 accepted it on #1318, so the cut I offered above is off. The defect is real: the compose file keeps its `build:` sections, so a hand-run compose build in key-provider rebuilds the enclave past the guard. The fix is to drop `build:` from the compose file and build only from the guard's `upgrade` path, with a test that a compose build there is a no-op. This lane is at its time cap; the fix is queued as the next round on this PR.

surcyf123:
@taiberium Done in `2c2919c`: the compose file now only names the images. Its `build:` sections moved to `docker-compose.build.yaml`, and only the guard passes that file to compose. So `docker compose build` in key-provider has nothing to build. I ran it on a box with real docker: it prints "No services to build" and the image list is unchanged. A plain `docker compose up` cannot rebuild either; it tries to pull and fails. The guard also refuses to run if the file gets a build section back, or if an override file appears beside it. Tests cover each case, and the executor suite is 270 passed on EC2.

surcyf123:
Round 3 is in 2c2919c: the compose file lost its `build:` sections, so a hand-run compose build has nothing to build. Details are in my reply to the 11:33Z correction, one comment up. The four NITs were done in e4ef2d0.

taiberium (approved):

PR #1368, opened 2026-09-14
DAH-3480 - [P1] validator: GPU probe blocked by a mid-cycle pod or filler is waived

surcyf123:
Implements [DAH-3480](https://app.notion.com/p/P1-validator-skip-the-GPU-memory-verify-when-a-rental-or-filler-took-the-cards-mid-cycle-B-175-3dbb8bfdbde9815ab34ef41dc81dbc04) · reviewer: @taiberium

**TL;DR** — The GPU probe zeroes a cycle when a filler or pod took the cards after the cycle's rental snapshot (60 of 267 `GPU_VERIFY_FAILED` rows in 48 h). After a no-UUID or timed-out probe the check re-reads the backend once; a listed filler or pod gives informational `GPU_VERIFY_SKIPPED_RENTED` instead. Look at: what counts as "rented"; the fail-closed cases.

**What changed**
- After a timed-out or no-UUID probe, `_lium_workload_live_now` re-reads `GET /internal/executors/rented` once, 30 s timeout (`checks/capability.py`)
- A filler or pod the backend lists for the node waives the verdict: `GPU_VERIFY_SKIPPED_RENTED`, probe output in `what_we_saw.probe` (`checks/capability.py`)
- The waiver returns the fresh snapshot as `state.rented_data`, so the fault probe, rental verification and the result handler read it too (`checks/capability.py`)
- No backend answer, a raised read, a wrong returned UUID or a probe exception keep the failure (`checks/capability.py`)
- `BackendClient.get_rented_executors_now`: the un-retried read; the cycle's `get_all_rented_executors` keeps its retry (`clients/backend_client.py`)
- New template `CapabilityMessages.RENTED_SKIPPED`, severity info (`messages.py`); the test backend mock answers `None` to the new read (`tests/helpers.py`)

**Risks**
- A node that held a Lium filler or pod at probe time keeps that cycle's idle score instead of 0; a miner's own default job in the fresh snapshot excludes the cycle from unrented incentive, as on every cycle that knows it — provider (about 30 cycles a day fleet-wide) — revert the PR
- One extra backend read per timed-out or no-UUID probe (≤ 267 in 48 h fleet-wide; a backend outage costs one 30 s wait), none on the healthy path — nobody — revert the PR

**How to verify**
- `cd neurons/validators && pytest tests/test_capability_check_mid_cycle_workload.py -q` → 8 passed
- Loki, after release: `GPU_VERIFY_SKIPPED_RENTED` rows carry `what_we_saw.workload` in {filler, pod}

**Verified by the loop:** 14 Sep 2026 · sandbox EC2 · targeted 17/17 · Gate: PASS ddb4cd1

**Ran it:** `pytest tests/test_capability_check*.py -q` → 17 passed · EC2 aws_verify 20260915T110328Z-7bu4 @ ddb4cd1: validators 2468 passed, 15 skipped

**Self-review:** clean at ddb4cd1 (19 checks)

**Parts:** backend ✓ · migration n/a — no schema · frontend n/a — codes printed verbatim · CLI/SDK n/a — validator-only · docs ✓ [lium-platform#423](https://github.com/Datura-ai/lium-platform/pull/423) · tests ✓ 5 (8 cases) · dashboards n/a — same log line

**Docs:** docs: [lium-platform#423](https://github.com/Datura-ai/lium-platform/pull/423) (`providers/troubleshooting.md` row, same ticket, same reviewer)

### Origin
LIUM_ISSUES B-175 (P1), from Discord #ticket-0324-provider (14 Sep 2026 13:11Z, "no incentives for an epoch"): the provider's 8x RTX PRO 6000 node scored 0 in the 12:24Z cycle with `GPU_VERIFY_FAILED`, `stderr` "Failed to allocate d_A: out of memory", returned UUID 'None'. His rental closed 12:28:23Z; the backend started a filler on the node; the probe ran 12:33:35Z.

### What the data says (prod tsdb `prod_executors` + backend `rental_history` and `filler_run`, 48 h to 14 Sep 15:30Z, 267 rows, 47 executors, 25 providers)
- 0 rows fell inside a rental as `rental_history` records it. The pipeline already halts for a node the cycle snapshot lists as rented (`TenantEnforcementCheck`, `halt=True`) and already skips the probe for a filler in the snapshot (`GPU_VERIFY_SKIPPED_ACTIVE_FILLER`). So "skip while rented" was already true for what the snapshot knew.
- 60 rows had a filler run (PEARL 31, DPHN 29, all Lium default jobs) created after the cycle's `rented_data` snapshot and live at the moment the probe ran (the backend starts a filler right after a rental closes; the probe came 5 to 15 minutes later). 9 of those also carried a pod that had just closed (the `STALE_POD_NOT_RUNNING` path continues the pipeline with `rented=False`). This is the DAH-2757 race, for the capability check. No row in the 48 h was a pod alone; the pod branch covers the same race for a pod created mid-cycle.
- 205 rows sit on 2 executors that fail every cycle with nothing of the backend's on the card. Those stay failures.
- Script and raw exports: `RECOVERY/workers/bug-b175-gpu-verify-rented/correlate.py` in the loop repo.

### Design
- Ask only after a failure. A fresh read before every probe would be one fleet-wide fetch per executor per cycle (about 520 per cycle); after a failure it is 267 calls in 48 h. `PortConnectivityCheck` re-reads the same list on its failure path.
- The read is `get_rented_executors_now`, added for this: the cycle's `get_all_rented_executors` is wrapped in tenacity (`retry_if_result(None)`, 3 attempts, 30 s apart) and raises `RetryError` after about 60 s of outage, which would abort the pipeline for the executor. The new method is the same GET without the wrapper: 30 s timeout, `None` on any client error. The check also catches a raised read and treats it as `None`.
- "Rented" is the backend's current view, never the node's: `get_filler_containers(uuid)` non-empty, or `executors[uuid].pods` non-empty. The backend does not list BROKEN and DELETING pods (`daos/pod.py:get_rented_pods_with_ports`). A filler is a default-job container the backend runs on the node, Lium's or the miner's own (`MINER_DEFAULT_JOB`); both are already exempt from the probe when the snapshot knows them (`_get_filler_only_container`), so this adds no new exemption, only the cycle the snapshot missed.
- Self-rental: the rented-executors response carries no self-rent marker (`RentedExecutor.owner_flag` exists in the protocol but the backend never sets it), so this check cannot tell a customer pod from a provider renting their own node. Today's snapshot path cannot either: `TenantEnforcementCheck` halts with the rented score for any listed pod, and the incentive layer handles self-rent through `rental_history.tier_source` (`self_rent` pays no incentive). This PR adds at most one cycle to that existing window. Known follow-up: have the backend mark self-rent pods in the response, then exclude them here.
- Waived only when the probe gave no answer: `timed_out`, or the service's "UUID mismatch" against '', 'None' or 'null'. A returned UUID that does not match the per-call nonce is the anti-spoof case and stays `GPU_VERIFY_FAILED` whatever stderr says; a sealed result that failed authentication has a different `error_message` and stays failed too; an exception on the validator's side stays failed and asks the backend nothing.
- `passed=True`, and `updates` carries the fresh snapshot as `state.rented_data` (review round 2). The checks after this one read the same list the waiver saw: `GpuFaultProbeCheck` skips for the filler, `RentalVerificationCheck` takes its filler-liveness path or `rental_in_progress` for a pod instead of asking for a health-check container on the busy card, and `ResultHandler` reads the default-job owner and pause flags from it. `ctx.rented` and `ctx.verified` are untouched; the cycle scores as it would have with `GPU_VERIFY_OK`, without new `gpu_metrics`.
- Not a scoring formula change: no constant, multiplier or pool moves. It changes when one check zeroes a cycle, on evidence the backend supplies.

### Relation to open PRs
- lium-io#1335 (`GPU_VERIFY_VRAM_UNAVAILABLE`, taiberium's latest review APPROVED) and lium-io#1352 (stacked on it) classify the same failure into finer reason codes. This PR runs before that classification: a listed workload waives; otherwise their `_failure_template` picks the code. The two PRs edit adjacent lines of `capability.py` (the import block and the `template =` line this PR inserts before), so whichever lands second takes a small rebase conflict; the behaviour of both is unchanged by the other. `_NO_UUID` and the "UUID mismatch" prefix here duplicate #1335's constants until then; the follow-up shares them.

### Known follow-ups
- Backend: mark self-rent pods in `GET /internal/executors/rented` so this check and `TenantEnforcementCheck` can exclude them (see Design, self-rental).
- After #1335 merges: import `UUID_MISMATCH_ERROR_PREFIX` and `_NO_UUID` from there.

### Verification
**Verified by the loop on 14 Sep 2026** · EC2 aws_verify 20260914T161809Z-2vpn @ f4e920a · validators 2468 passed, 15 skipped (+ main's own 2 failures: `tests/test_scrape_infiniband.py::test_an_unreadable_sysfs_tree_is_not_reported_as_no_devices`, `tests/test_sshd_bootstrap_script.py::test_adopt_path_hardens_config_for_key_only_auth`, the same two #1335 reports on origin/main without this branch) · targeted 17/17 (20260914T162155Z-iya8) · Result: PASS f4e920a · Gate: PASS (f4e920a) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1368)

### Ran it

```
$ cd neurons/validators && pytest tests/test_capability_check_mid_cycle_workload.py tests/test_capability_check.py -q
17 passed, 101 warnings in 0.11s
(EC2 aws_verify 20260914T162155Z-iya8, warm lium-io box, HEAD f4e920ade3a3a00cdc533d60960d4f7b4542a2b3)
```

### Self-review

Verdict `clean` at `ddb4cd1` · 19 checks · by subagent-r205-1368 · 2026-09-15 11:25Z (tools/pr_self_review.py)

Mechanical notes dismissed: `neurons/validators/src/services/task/checks/capability.py:120` Resolved in the body (round 2, no code change): the Design bullet that said "passed=True with no updates" now states that updates carries the fresh snapshot as state.rented_data and names the three readers (GpuFaultProbeCheck filler skip, RentalVerificationCheck filler-liveness / rental_in_progress path instead of a health-check container, ResultHandler default-job owner and pause flags) with ctx.rented and ctx.verified untouched, and What-changed gained the matching bullet; each sentence matches capability.py:120, pipeline.py:300, gpu_fault_probe.py:65, rental_verification.py:172-193 and 254, result_handler.py:133-145. · `neurons/validators/src/services/task/checks/capability.py:120` Resolved in the body (round 2, no code change): the Risks line now distinguishes a Lium filler or pod (keeps the idle score) from a miner's own default job (excluded from unrented incentive that cycle, as on every cycle that knows it), which is what rental_price.py:495 does with JobResult.default_job_owner == "miner" and is_rented False. · `neurons/validators/src/services/task/checks/capability.py:195` fresh.executors.get(executor_uuid) is an in-memory dict lookup on the backend's own response keyed by ctx.executor.uuid, the executor this pipeline is already verifying; there is no caller-supplied id and no owner to scope to. · `neurons/validators/src/services/task/checks/capability.py:184` The noqa: BLE001 marks the intended fail-closed path (warn, return None, the probe failure stands), stated in the docstring and covered by test_a_backend_read_that_raises_keeps_the_failure; narrowing it would let an unexpected client exception abort the pipeline. · `neurons/validators/tests/test_capability_check_mid_cycle_workload.py:43` The braces are the literal RESULT_JSON payload from the prod transcript stdout, not a placeholder; an f prefix would break it. · `neurons/validators/tests/test_capability_check_mid_cycle_workload.py:54` Line 54 is already an f-string; the scanner mis-located the literal JSON on line 43. · `neurons/validators/src/services/task/checks/capability.py:13` TYPE_CHECKING-only import of RentedExecutorsResponse on a dataclass field is safe: from __future__ import annotations keeps the annotation a string and dataclasses only string-matches ClassVar/InitVar; py_compile passes and no caller resolves the hints at runtime. · `neurons/validators/src/services/task/checks/capability.py:120` dataclasses.replace on the frozen ContextState dataclass and a "state" key in CheckResult.updates is the pattern the success path (gpu_metrics) and PortConnectivityCheck already use; Pipeline.run applies it with Context.model_copy(update=...) on the frozen pydantic Context.

### Verified

Gate: PASS (ddb4cd1) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1368)

</details>


neurons/validators/src/services/task/checks/capability.py
```diff
@@ -101,6 +119,65 @@ async def run(self, ctx: Context) -> CheckResult:
...
+    if result is None:
+        return False
+    if result.timed_out:
+        return True
+    if not (result.error_message or "").startswith(_UUID_MISMATCH_ERROR_PREFIX):
+        return False
+    return (result.returned_uuid or "").strip().lower() in _NO_UUID
+
+
+async def _lium_workload_live_now(ctx: Context) -> dict[str, Any] | None:
```
taiberium, line 144: **NIT:** New code must not return `dict[str, Any]`. Use a small frozen dataclass with `kind` and `container_names`, then build the event dict at the caller.

taiberium (approved):

neurons/validators/src/services/task/checks/capability.py
surcyf123: On it. Replacing the dict with a frozen dataclass and building the event at the caller. Fix lands within the hour.

surcyf123: On it. Replacing the dict with a frozen dataclass and building the event at the caller. Fix lands within the hour.

surcyf123: Done in `a53cf6d`: `_lium_workload_live_now` returns a frozen `_LiumWorkload(kind, container_names)` or None; the caller builds the event dict. `Any` is gone from the file.

neurons/validators/src/services/task/checks/capability.py
```diff
@@ -91,6 +98,21 @@ async def run(self, ctx: Context) -> CheckResult:
...
+                    Msg.RENTED_SKIPPED,
+                    ctx=ctx,
+                    check_id=self.check_id,
+                    what={
+                        "workload": lium_workload.kind,
+                        "containers": list(lium_workload.container_names),
+                        "probe": failure_details,
+                    },
+                )
+                return CheckResult(passed=True, event=event)
```
taiberium: **NIT:** The waiver throws away the fresh read, so `rental_verification.py:136` and `gpu_fault_probe.py:65` still get the stale snapshot. Return it in `updates`, the way `PortConnectivityCheck` does.

neurons/validators/src/services/task/checks/capability.py
```diff
@@ -101,6 +123,74 @@ async def run(self, ctx: Context) -> CheckResult:
...
+        return False
+    return (result.returned_uuid or "").strip().lower() in _NO_UUID
+
+
+@dataclass(frozen=True)
+class _LiumWorkload:
+    """A workload the backend says holds this node's cards right now: which kind (`filler` or
+    `pod`) and the container names it listed, sorted."""
+
+    kind: str
```
taiberium: **NIT:** `kind` only ever holds `filler` or `pod`. Type it `Literal["filler", "pod"]` and add the import.

taiberium (approved):

neurons/validators/src/services/task/checks/capability.py
surcyf123: Done in `ddb4cd1`: the waiver returns `state` with the fresh snapshot as `rented_data`, the way `PortConnectivityCheck` does; a test asserts it.

neurons/validators/src/services/task/checks/capability.py
surcyf123: Done in `ddb4cd1`: `kind` is `Literal["filler", "pod"]`.

taiberium (approved):

PR #1359, opened 2026-09-11
DAH-3419 - [P2] executor updater pulls the runner by signed digest

surcyf123:
Implements [DAH-3419](https://app.notion.com/p/Executor-image-rollout-stuck-on-139-493-nodes-watchtower-never-updates-itself-label-missing-on-the-3d8b8bfdbde981e7a097fead850a5558) · reviewer: @taiberium

**Problem** — 99 of 493 nodes run an old executor: a Docker Hub mirror on their site serves an old `runner:latest` and `nickfedor/watchtower` pulls by tag through it.

**Value** — not measured — the body has no Worth it line (SO §69) and the ticket gives no prod number — a human decides.

**Goal** — [DAH-3419](https://app.notion.com/p/Executor-image-rollout-stuck-on-139-493-nodes-watchtower-never-updates-itself-label-missing-on-the-3d8b8bfdbde981e7a097fead850a5558) (P2).

**Fits** — Sits in `lium-io` — `watchtower/tests`, `watchtower/src`, `neurons/executor`…; touches docs, tests (66).

**Flag:** no flag

**Risks**
- None beyond the diff: no migration, no pod restart, no config change, no data rewrite.

**How to verify**
- `cd watchtower && pdm run pytest tests -q` → 45 passed; `cd neurons/executor && pdm run pytest tests/test_stack_updater_by_digest.py tests/test_update_status.py -q` → 21 passed
- before merge: push the git tag `watchtower-v1.1.0` at this head, then `docker manifest inspect daturaai/lium-watchtower:1.1.0 >/dev/null && echo published`

**Verified by the loop:** Gate: PASS e8ba2de

**Ran it:** executor suite at e8ba2de77 → 285 passed, EC2 20260915T145245Z-2hqt · … +2 under Details

**Self-review:** clean at e8ba2de (15 checks, 2 low)

**Parts:** backend n/a — no lium-platform change · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — `lium mine` runs the compose file unchanged · docs ✓ `EXECUTOR_UPDATE.md`, `watchtower/README.md` · tests ✓ 66 · dashboards n/a — query in Details

**Docs:** `neurons/executor/EXECUTOR_UPDATE.md`, `neurons/executor/README.md`, `watchtower/README.md` updated.

<details><summary>Details</summary>

### What changed

- compose files: `watchtower` is `daturaai/lium-watchtower:1.1.0` (`-staging` dev), not `nickfedor/watchtower`
[...]

PR #1369, opened 2026-09-14
DAH-3481 - [P2] updater holds while the NVIDIA runtime probe fails

surcyf123:
Implements [DAH-3481](https://app.notion.com/p/3dbb8bfdbde9812eacf6de75f964378a) · reviewer: @taiberium

**Problem** — A rented node went offline after its executor image updated on a host whose NVIDIA driver and libraries no longer matched: the new executor could not start (ticket-0325, B-176).

**Value** — not measured — no Worth it line (SO §69), the ticket gives no prod number — a human decides.

**Goal** — [DAH-3481](https://app.notion.com/p/3dbb8bfdbde9812eacf6de75f964378a) (P2) · serves B-176, SO §69.

**Fits** — Sits in `lium-io` — `watchtower/src`, `watchtower/tests`, `neurons/executor` compose + docs; touches docs, tests (17).

**Flag:** no flag

**Risks**
- A host whose runtime cannot start a GPU container gets no executor update until fixed — provider — `docker logs executor-watchtower-1` says `Update held` and why; the doc says wait for the rental to end, then reboot. Rollback: pin the compose back to `1.1.0`.
- No migration, no pod restart, no config change.
- The compose pins `1.1.1`; the tag exists on Docker Hub once a human pushes `watchtower-v1.1.1` at the approved head, before the merge (README order), so no host pulls a missing tag.

**How to verify**
- `cd watchtower && pdm run pytest tests -q` → 62 passed
- `cd neurons/executor && pdm run pytest tests/test_stack_updater_by_digest.py -q` → the pins and `pyproject.toml` agree on `1.1.1`
- broken host: `docker logs executor-watchtower-1 | grep "Update held"` → one line per cycle with `"reason_code": "NVIDIA_RUNTIME_MISMATCH"`

**Verified by the loop:** Gate: PASS 169d507

**Ran it:** EC2 throwaway box `20260918T090814Z-29um` at `169d5072` (c6i.2xlarge, Python 3.11.11 as CI): `cd watchtower && pdm install && pytest tests -q` →… · not run on macOS · … +3 under Details

**Self-review:** clean at 169d507 (15 checks, 2 low)

**Parts:** backend n/a — validator half cut 15 Sep (Details) · migration n/a — no schema change · frontend n/a — lium-platform#425 has no producer after the cut (Details) · CLI/SDK n/a — not a CLI feature · docs ✓ `EXECUTOR_UPDATE.md`, `watchtower/README.md` · tests ✓ 17 · dashboards n/a — log line only

**Docs:** `neurons/executor/EXECUTOR_UPDATE.md`, `watchtower/README.md` (the hold step, the `Update held` / stale-probe lines, the remedy timing, the by-hand probe) · public: lium-platform#557 (`providers/nodes/gpu-power-cap.md`, `providers/troubleshooting.md`, same ticket, same reviewer).

**Cut 15 Sep (value audit, SO §69) from `neurons/validators/**`:** the validator half (`DindProbeResult.reason_code` / `error`, `specs.runtime_probe`, 7 tests) and the separate `1.2.0` tag are dropped. One incident (ticket-0325) does not pay for two halves; the hold in the updater is the part that keeps the node online. The version files came back on 18 Sep as a `1.1.1` patch bump (Version, below), because `1.1.0` had been published without the hold. Kept: the probe (`probe_nvidia_runtime` since the 18 Sep review round), the `Update held` hold in `check_and_update`, the tests, `EXECUTOR_UPDATE.md` and `watchtower/README.md`.

### The ERROR-vs-score question (B-176 item c)

"DinD probe ERROR with `PORT_VERIFY_OK` and score 1.0 (`sysbox_runtime` false) — intended?" Code path on this branch:

1. `dind_probe.py` `DindVerifier.verify`: `docker run -d --gpus all …` (`--runtime=sysbox-runc` only when the scrape said Sysbox) exits non-zero → `logger.error("DinD creation failed")` → `DindProbeResult(success=False, sysbox_runtime=…)`.
2. `orchestrator.py` `ConnectivityOrchestrator.verify`: `dind_result.success` False → the DinD port joins `failed`, `sysbox_runtime = False`; `status = "ok" if successful` — the 99 plain TCP ports passed, so the status is `ok`.
3. `service.py` `ExecutorConnectivityService.verify_ports`: for a node with rented ports whose executor scrape reported sysbox (`ctx.state.sysbox_runtime` True), the probe's False is replaced by the scraped True (DAH-2527 fallback). The ticket-0325 node did not take this path: its scrape already said `sysbox_runtime: false` (the Loki line), so False stayed.
4. `port_connectivity.py`: status `ok` → `PORT_VERIFY_OK`, `passed=True`, `state.sysbox_runtime = False`.
5. `sysbox_required.py` `SysboxRequiredCheck`: `not sysbox_runtime and not is_rented` is False for a rented node → `SYSBOX_REQUIRED_OK` (DAH-2313: a live rental is never delisted for a missing sysbox).
6. `incentive/default.py` `calculate_executor_score`: `score` stays 1.0 (no check failed); `sysbox_multiplier = 1` when `sysbox_runtime` is True, else `1 - PORTION_FOR_SYSBOX_RENTED` = `1 - 1` = **0** for a rental created after `SYSBOX_RENTED_CUTOFF` (2026-04-03); `mining_score *= sysbox_multiplier`.

So it is intended for the listing (steps 4 and 5). Whether it is scoring depends on step 3: a rented node whose scrape reports sysbox keeps multiplier 1 and loses nothing; a rented node whose scrape also reports no sysbox (ticket-0325: Loki 11:08:10Z reads `"sysbox_runtime": false, "sysbox_multiplier": 0.0, "mining_score": 0.0` with `"score": 1`) is paid 0 while it stays listed. ticket-0325's REPORT said "non-scoring" because it read the `score` column. The part that is a bug is visibility: `PORT_VERIFY_OK` + `SYSBOX_REQUIRED_OK` name nothing, and the ERROR line stays on the validator. The validator-side visibility fix (`specs.runtime_probe`) was cut from this PR on 15 Sep (below); what this PR ships is the updater hold and its `Update held` log line, which is what stops the node going offline. The multiplier is scoring math and stays as it is (PR_PROCESS §1b)

### Design

- The hold is on the host, in the updater, not on the validator: the validator's probe result reaches the backend on the next cycle, but the updater has no per-node channel to the backend and needs none — the same failure is visible locally by starting one container. The probe uses the image the runner already runs (`container.attrs["Image"]`), so nothing is pulled while the host is broken; entrypoint `true`, `DeviceRequest(count=-1, capabilities=[["gpu"]])` — the same request the executor's compose `deploy.resources.reservations.devices` makes. The daemon runs the NVIDIA prestart hook on `start`; the hook's failure text is the daemon's 500 message. `wait(timeout=60)`; a hung probe reads as not ready. A runner with no image id holds with `RUNTIME_PROBE_FAILED` (no crash into the outer handler).
- Any probe failure holds, not only the NVML mismatch: a `true` container with a GPU request that cannot start means the executor container cannot start either. `NVIDIA_RUNTIME_MISMATCH` is logged only when the daemon's message says `driver/library version mismatch`, the phrase NVML prints for a driver package updated without a reboot; every other failure, `nvml error` and `initialization error` included (a stopped persistence daemon, a missing `libnvidia-ml.so.1`), is `RUNTIME_PROBE_FAILED` with the daemon's text, so the doc's reboot remedy is named only when it applies (review round, 18 Sep).
- No state between cycles: the probe runs every cycle an update is pending, so the update resumes the cycle after the reboot. A first boot (no runner container) skips the probe: there is no image to probe with and nothing running to break.
- The probe container carries `io.lium.watchtower.probe=nvidia-runtime` and is removed in `finally`, force, whatever happened; a failed remove is a WARNING with the id. Before each probe, every container with that label (any state) is removed first, so a held host holds one probe container at most; when a removal or the listing fails, the probe fails with `stale runtime probe container not removed` and creates nothing (review round, 18 Sep).
- Version: `1.1.1`. The 15 Sep plan was to land this on #1359's branch before `watchtower-v1.1.0` was tagged so the `1.1.0` image carried the hold. That did not happen: `watchtower-v1.1.0` was tagged at #1359's last head (before its squash merge) and `daturaai/lium-watchtower:1.1.0` was published on 18 Sep 06:34Z without the hold. So this PR bumps `watchtower/pyproject.toml` to `1.1.1`, pins both compose files and the workflow default to it (`test_stack_updater_by_digest.py` fails when they drift), and the image is published by the git tag `watchtower-v1.1.1`, which a human pushes at the approved head before the merge, as `watchtower/README.md` says (Cross-PR).

### Cross-PR

- Was stacked on lium-io#1359 (DAH-3419, same reviewer): the updater code this hooks into. #1359 merged to `main` on 18 Sep 06:41Z (`c911658`); this PR was rebased onto `main` the same morning (0 conflicts, patch unchanged). `watchtower_image.yml` refuses to move a published version tag, so the hold reaches hosts only through a new tag: this PR carries the `1.1.1` bump and pins; the git tag `watchtower-v1.1.1` is pushed by a human (@taiberium, per the review) at this PR's approved head, before the merge, and the workflow publishes `daturaai/lium-watchtower:1.1.1` and `1.1.1-staging`. The loop pushes no tags (SO §31).
- lium-platform#557 on DAH-3481 (docs half, PR_PROCESS §1.1a): the `Update held` paragraph in `providers/nodes/gpu-power-cap.md` and the Common-issues row in `providers/troubleshooting.md`; merge with or after this PR.
- lium-platform#425 on DAH-3481 (closed): `MachineSpecs.runtime_probe`, the portal node page line, the two docs rows. After this cut nothing writes `specs.runtime_probe`, so #425's node-page line has no producer; the validator half needs its own ticket (or #425 drops the line) — for the orchestrator.
- lium-io#1360 (merged) made `EXECUTOR_IMAGE_OUTDATED` a warning; unchanged.

### Security properties (PR_PROCESS §5)

- The probe container runs `true` with no binds, no network request, no env from the host; the image is the one already running.

### Ran it

Review round 18 Sep (taiberium's six comments: mismatch phrase only, labelled stale probes, rename, remedy timing, by-hand probe, `1.1.1` pins). One EC2 throwaway box at the head `169d5072`, `20260918T090814Z-29um` (c6i.2xlarge, Python 3.11.11 as CI, the head fetched from a git bundle of the worktree before the push):

`cd watchtower && pdm install && .venv/bin/python -m pytest tests -q`:

```
HEAD 169d50723c596a24ef27b02bf52ea30cb0d070f5
62 passed, 5 warnings in 1.62s
watchtower pytest rc=0
```

The B-176 tests by name (`-k "probe or classify_runtime or held"`): 15 passed, 47 deselected.

`cd neurons/executor && pdm install && .venv/bin/python -m pytest tests/ -q`:

```
403 passed, 2 skipped, 218 warnings in 15.56s
executor pytest rc=0
```

`tests/test_stack_updater_by_digest.py -v`: 9 passed — both compose pins and the workflow default equal `pyproject.toml`'s `1.1.1`.

This PR changes no validator file, so the validator suite is main's. The hold path itself was not run on a host with a broken driver (none available to the loop); the tests carry the daemon's text from the ticket-0325 host. `docker-compose.dev.yml` not run on macOS. Earlier rounds: `20260918T071235Z-sau9` / `20260918T071446Z-6ixn` at the rebased head (54 / 403 passed).

### Self-review

Verdict `findings` at `169d507` · 15 checks · by subagent-fresh-r319 · 2026-09-18 09:26Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | STALE-BODY | `PR body (Verified by the loop / Ran it):82` | The Verified and Ran-it sections report 54 passed at the previous head while How-to-verify predicts 62 at this head (60 test defs + 3 parametrized cases = 62, so the prediction is right); the sections must be re-rendered from the EC2 run at this commit once it exists, and the old pre-rebase head cited under Self-review is not reachable from the branch. | After the EC2 run at this head, replace the 54-passed block and the Verified head with the new run's output and drop the unreachable pre-rebase sha from the prose. |
| low | CODE-QUALITY | `PR body (Details):143` | The Details fold has two consecutive `### Ran it` headings. | Delete the duplicate heading. |

Mechanical notes dismissed: `neurons/executor/pdm.lock:317` Remaining `1.1.0` strings on the head (pdm.lock, Cargo.lock, `msgpack==1.1.0` in neurons/executor/pyproject.toml) are unrelated dependency versions, not the updater tag; every updater pin (pyproject, both compose files, workflow default, docker_build.sh comment, README) reads 1.1.1 and test_stack_updater_by_digest.py ties them together. · `watchtower/tests/test_watchtower.py:352` The `containers.list` side_effect returns unittest.mock.DEFAULT for every non-probe filter, which makes the Mock fall back to `return_value`; find_runner_container's two calls use only `all=` and `filters=` kwargs, and the runner-lookup and check_and_update tests that set `return_value` or a replacement `side_effect` per case keep their prior semantics. · `watchtower/src/watchtower.py:543` `entrypoint=["true"]` on the runner image is safe: Dockerfile.runner is `docker:26-cli` (Alpine/busybox), which ships `true`; the executor compose files request GPUs unconditionally (`driver: nvidia, count: all, capabilities: [gpu]`), so a host that runs the executor also has the runtime the probe exercises. · `PR body (What changed):137` The `1.1.0` publish sha is reachable via the origin tag `watchtower-v1.1.0`, which points at it; not an ancestor of HEAD but a real tagged commit. · `tools/pr_review_service.py scanners:1` No mechanical findings to confirm; the diff was read class by class instead.

### Verified

Gate: PASS (169d507) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1369)

### What changed

- `watchtower/src/watchtower.py`: `probe_nvidia_runtime` starts a labelled throwaway GPU container from the runner's image, after removing every labelled leftover; `check_and_update` holds the pull while it fails, logs `Update held` + `NVIDIA_RUNTIME_MISMATCH` (only for `driver/library version mismatch`) / `RUNTIME_PROBE_FAILED`
- `watchtower/pyproject.toml` `1.1.1`, `neurons/executor/docker-compose.yml` + `docker-compose.dev.yml` pins, `watchtower_image.yml` default, `docker_build.sh` and README: `1.1.0` was published from #1359's tagged head without the hold
- `neurons/executor/EXECUTOR_UPDATE.md`, `watchtower/README.md`: the hold step, the log lines, wait for the rental to end before the reboot, the by-hand probe from the runner image
- tests: `watchtower/tests/test_watchtower.py` +17 cases (hold, resume, first boot, classifier + 3 negative controls, stale probes removed / stuck / unlistable, two held cycles leave one container, unrelated NVML error)

</details>


surcyf123:
**Cut on 15 Sep (value audit, SO §69) — cuts-3.** Dropped the validator half — `specs.runtime_probe` in the executor spec, the validator-side reading of it and its tests — and the separate `1.2.0` image tag: `.github/workflows/watchtower_image.yml`, both executor compose files, `watchtower/pyproject.toml` and `neurons/validators/` are back to the #1359 base, so the hold now ships inside #1359's `lium-watchtower:1.1.0` (no `1.1.0` or `1.2.0` tag is on Docker Hub yet). Kept: the updater hold itself (`nvidia_runtime_ready` + `Update held` log with the reason) and its 9 tests. The portal half already lives in `lium-platform#425`; that PR needs a note that the validator no longer emits the probe field.

Merge order: this PR into `DAH-3419-watchtower-self-update` first, then `watchtower-v1.1.0` is tagged there, then #1359 to `main`.

`4a2241b` → `af2b222`: −286 / +17 lines (4 files vs the #1359 base, was 15). EC2 at `af2b222`: `watchtower` pytest 54 passed (`20260915T180456Z-mbie`); fresh self-review clean (15 checks); gate PASS at `af2b222`.

surcyf123:
Rebased onto `main` after #1359 merged (`c911658`). New head `16d9e9e`, 0 conflicts, patch unchanged (the diff against `main` is byte-identical to the old diff against #1359's branch). The three DAH-3419 commits this branch carried are gone; the five DAH-3481 commits stay.

Tests at `16d9e9e` on EC2: watchtower 54 passed, executor 403 passed / 2 skipped. Gate PASS, CI green.

One thing changed under this PR: `watchtower-v1.1.0` was tagged and `daturaai/lium-watchtower:1.1.0` published (06:34Z) before this landed, so the 15 Sep plan ("the hold ships in 1.1.0") no longer holds. Merging this PR alone puts the hold in no published image. The body's Cross-PR section now says so. @taiberium: should the version bump (`1.1.1` in `watchtower/pyproject.toml` + the compose pin) go into this PR, or into a follow-up on DAH-3481? The loop adds it either way once you say which.

watchtower/src/watchtower.py
```diff
@@ -23,6 +23,24 @@
...
+# reboot) that hook fails and the executor cannot start. An update on such a host replaces a
+# running executor with one that never comes back (ticket-0325: rented node offline for hours).
+# The probe below starts a throwaway container from the runner's OWN image (already on the
+# host, nothing pulled) with the same GPU request and the command `true`; the hook runs, the
+# container exits 0, or the daemon reports why it could not start.
+RUNTIME_PROBE_NVIDIA_MISMATCH = "NVIDIA_RUNTIME_MISMATCH"
+RUNTIME_PROBE_FAILED = "RUNTIME_PROBE_FAILED"
+_NVIDIA_MISMATCH_MARKERS = (
+    "driver/library version mismatch",
+    "nvidia-container-cli: initialization error",
```
taiberium: `initialization error` and `nvml error` cover failures besides driver/library mismatch, so this gives the wrong remedy. Match only `driver/library version mismatch`.

watchtower/src/watchtower.py
```diff
@@ -478,6 +496,56 @@ def pull_and_restart_containers(client: docker.DockerClient, image_name: str, re
...
+            return RuntimeProbe(False, RUNTIME_PROBE_FAILED, f"probe container exited {status}")
+        return RuntimeProbe(True, None, None)
+    except docker.errors.APIError as e:
+        return RuntimeProbe(False, classify_runtime_probe_error(str(e)), str(e))
+    except Exception as e:
+        return RuntimeProbe(False, RUNTIME_PROBE_FAILED, str(e))
+    finally:
+        if container is not None:
+            try:
+                container.remove(force=True)
```
taiberium, line 567: If cleanup fails, every held cycle can leave another probe container. Remove labeled stale probes before creating one, and abort if any removal fails.

neurons/executor/EXECUTOR_UPDATE.md
```diff
@@ -69,6 +70,6 @@ Read the updater's log:
...
... line `Removing leftover runner container from an interrupted update` means the leftover was removed; a line `Restoring the name of a leftover runner container from an interrupted update` means the leftover was the only runner and got its name back. A line `Leftover runner container kept: the runner is not running` means the runner under the real name did not start (the line carries the error); the updater keeps both and goes on with the digest check. Then run `docker compose up -d`, which recreates the runner, and remove the leftover with `docker rm -f <name>`. A line `Update held: the NVIDIA runtime cannot start a GPU container on this host` means the host cannot start a GPU container right now, so the updater waits. With `"reason_code": "NVIDIA_RUNTIME_MISMATCH"` the host's NVIDIA driver and its libraries do not match: reboot the host, or reload the driver, then run `docker compose up -d` here. The update runs at the next check after that. With `"reason_code": "RUNTIME_PROBE_FAILED"` the `error` field carries the daemon's message; `docker run --rm --gpus all ubuntu true` on the host shows the same. If the log does not say, send the 50 lines to Lium support with the node address.
```
taiberium: This remedy can kill the rented pod that the hold protects. Tell providers to wait until the rental ends before they reboot or reload the driver.

taiberium: This command can pull `ubuntu` and fail at the registry instead. Use the current runner image ID with `--entrypoint true` to reproduce the probe; non-blocking.

watchtower/src/watchtower.py
```diff
@@ -478,6 +496,56 @@ def pull_and_restart_containers(client: docker.DockerClient, image_name: str, re
...
+
+def classify_runtime_probe_error(error: str) -> str:
+    """`NVIDIA_RUNTIME_MISMATCH` for the driver/library disagreement, else `RUNTIME_PROBE_FAILED`."""
+    text = error.lower()
+    if any(marker in text for marker in _NVIDIA_MISMATCH_MARKERS):
+        return RUNTIME_PROBE_NVIDIA_MISMATCH
+    return RUNTIME_PROBE_FAILED
+
+
+def nvidia_runtime_ready(client: docker.DockerClient, image_id: str) -> RuntimeProbe:
```
taiberium: nit: this name reads as a yes/no question but returns a `RuntimeProbe`, so `if nvidia_runtime_ready(...)` is always true. Rename it `probe_nvidia_runtime`. non-blocking, fine as a follow-up.

taiberium (changes requested):
**watchtower/pyproject.toml:3**

`watchtower/pyproject.toml:3` still uses 1.1.0, but Docker Hub published that tag from `5f3eccd` without this guard. Please bump all updater pins and publish the new tag.

watchtower/src/watchtower.py
surcyf123: Done in `169d507`: renamed to `probe_nvidia_runtime` in the source, the tests (import and the patch target) and the body; no `nvidia_runtime_ready` is left on the branch.

watchtower/src/watchtower.py
surcyf123: Done in `169d507`: The probe container now carries the label `io.lium.watchtower.probe=nvidia-runtime`. Before each probe, the updater lists every container with that label, in any state, and force-removes each one. If the listing or any removal fails, the probe fails with `stale runtime probe container not removed` and creates nothing. Four tests cover it. Two stale probes are removed before the new one is created. A stuck removal aborts with no create. A failed listing aborts with no create. Two held cycles where the first remove fails end with one container on the host, not two. The doc names the new log line and the manual fix for a leftover the daemon refuses to drop.

neurons/executor/EXECUTOR_UPDATE.md
surcyf123: Done in `169d507`: The doc now says the pods keep running under the hold. It says a reboot or a driver reload stops every running pod. It tells the provider to wait until the rental on the node ends before either. The public pages say the same in lium-platform#557, the docs half of this ticket.

surcyf123: Done in `169d507`: The by-hand command now runs `true` from the runner image that is already on the host, with `--entrypoint true` and `--gpus all`. The image id comes from `docker inspect` on the runner container. The doc says why not to use a name like `ubuntu`: a name is pulled, and a registry error then hides the runtime error.

watchtower/src/watchtower.py
surcyf123: Done in `169d507`: The classifier now matches only the phrase `driver/library version mismatch`. Every other NVML failure stays `RUNTIME_PROBE_FAILED` with the daemon's text. So the reboot remedy is named only for the mismatch. Two new tests cover it. Three unrelated NVML messages classify as the generic code. A held cycle with an unrelated NVML error logs the generic code, not the mismatch one.

surcyf123:
@taiberium Done in `169d507`: Every updater pin is `1.1.1` now. That is the pyproject version, both compose files (`1.1.1` and `1.1.1-staging`), the workflow's default tag, the build script comment and the README. The executor test that ties the pins to the version passed on EC2 with the rest of the suite (403 passed). The watchtower suite passed too (62). The tag is the one step the loop does not do. The image is published by pushing the git tag `watchtower-v1.1.1`. The README says to push it at the head before the compose change merges. Then a fresh compose up never asks for a tag that is not on Docker Hub yet. @taiberium, would you push that tag at the approved head once you are happy with the PR, or name who should? The five inline comments are answered in their threads. The public docs half is lium-platform#557.

taiberium (approved):
I pushed the missing step to this branch: pause new rentals before the reboot. The other four findings are answered.

surcyf123:
Approved-not-merged check (loop, 19 Sep 12:1xZ): approved by @taiberium on 18 Sep 11:39Z, MERGEABLE at the current head, CI green, no thread waiting on the loop. Merge order: this one first, then its docs half lium-platform#557 (the `EXECUTOR_UPDATE.md` link there resolves once this is on main). Ready whenever you want it in.

PR #1379, opened 2026-09-15
DAH-3518 - [P2] validator: restore GPU power limits side by side, not one GPU at a time

surcyf123:
Implements [DAH-3518](https://app.notion.com/p/3dcb8bfdbde98101abb5d5085505b9ef) · reviewer: @taiberium

**Problem** — A filler's delete restored its GPUs' power limits one after another, three `nvidia-smi` round trips per GPU.

**Value** — 293 capped rents / 14 d ≈ 21 per day, each waiting ~22 s past the point the container was gone ≈ 8 min of renter waiting per day, plus the 337 rents / 14 d that waited 20 to 29 s on the same nodes. Change: 139 lines added / 23 removed in one module..

**Goal** — [DAH-3518](https://app.notion.com/p/3dcb8bfdbde98101abb5d5085505b9ef) (P2).

**Fits** — Sits in `lium-io` — `neurons/validators`; touches backend, tests (8 new).

**Flag:** no flag

**Risks**
- Up to 8 `nvidia-smi` sessions open on the host at once during a filler delete or a rent's pre-run power check, instead of one — a host whose sshd `MaxSessions` is below 8 answers the extra channel opens with `ChannelOpenError`; those GPUs are set again one at a time after the first pass (the old loop's pace for them alone), so such a host costs time, not a GPU left capped — rollback: revert the file; no pod restart beyond the validator's own deploy, no migration, no config change.

**How to verify**
- `pdm run pytest tests/test_gpu_power_limit.py -q` (validators)
- Fail-old: with `POWER_LIMIT_SET_CONCURRENCY` stubbed to 8 in the test file, three of the eight new tests fail on `origin/main` on `peak_in_flight == 1` and the two `refused` tests on `refused` 0 == 12; on 2c28e52 (side by side, no retry) the two `refused` tests read 2 == 8; on 124c796 (round 1) they read 6 == 12 and the refused `-pm 1` test reads `False is True`; the hung-GPU test pins what the old loop already did and passes on both
- Prod after deploy: Loki `{job="subnet/connector"} |= "gpu power limit restore ok"` — the eight lines of one 8-GPU delete land within ~2 s instead of ~15 s; tsdb `container_profiler_events.filler_preemption_ms ≥ 29500` share falls from 5.8 %

**Verified by the loop:** Gate: PASS 29b56c4

**Ran it:** `tools/aws_verify.sh lium-io 29b56c4a5 --only neurons/validators/tests/test_gpu_power_limit.py` (the touched module, sandbox EC2, run 20260916T141510Z-r7hs) → 54… · … +2 under Details

**Self-review:** clean at 29b56c4 (15 checks, 3 fixed in body)

**Parts:** backend ✓ validator `gpu_power_limit.py` · migration n/a — no schema change · frontend n/a — no UI · CLI/SDK n/a — no API change · docs n/a — nothing a provider sees or sets changes · tests ✓ 8 new · dashboards n/a — no panel reads the restore timing

**Docs:** none needed because nothing a provider sees or sets changes; the module docstring in `gpu_power_limit.py` is updated in this PR.

[DAH-3518](https://app.notion.com/p/3dcb8bfdbde98101abb5d5085505b9ef) · parent [DAH-3003](https://app.notion.com/p/3d3b8bfdbde981fa9ff8dd62d74e8876) (lium-platform#107).

## Problem

Origin: Rustam's bot, 15 Sep 2026 10:15Z (WORK_QUEUE row 762): 4.9 % of rents wait the backend's whole 30 s filler-preemption cap (`FILLER_STOP_WAIT_TIMEOUT_SECONDS`, lium-platform `default_job/constants.py`) and then continue without the STOPPED confirmation.

Measured on prod (read-only), 14 d to 15 Sep 2026:

- tsdb `container_profiler_events`: 5,078 rents carry `filler_preemption_ms`, 4,094 had a filler to stop, 293 (5.8 %) reached 29.5 s or more. By GPU class: RTX 4090 119 of 729 (16 %), RTX 5090 101 of 655 (15 %), H200 23 of 393 (6 %); every other class under 4 %. The top 14 nodes by capped rents are 8x 4090 and 8x 5090 nodes of two providers; the median rent on those nodes waits 29.9 to 30.2 s.
- backend `filler_run` joined to `rental_history` on executor and time: on capped rents the filler reaches STOPPED 37 s (p50) / 44 to 51 s (p90) after the rent started. The filler does stop; the callback is late. 251 of 315 capped joins are PEARL, 37 DPHN, 25 ENGY_TEE.
- Loki `subnet/connector`, 48 h, n = 1,143 filler deletes with both the "Deleting" and "Deleted" line: total p50 7.9 s, p90 25.5 s; 128 (11 %) took 25 s or more, and in 78 of those the largest phase is the GPU power restore. By strategy and GPU count: PEARL 8-GPU (n = 152) p50 27.7 s, 65 % over 25 s, `docker stop` itself 2.3 s, power-restore span p50 14.7 s; PEARL 1-GPU (n = 436) p50 7.3 s, 0.7 % over 25 s; DPHN 8-GPU (n = 129, no power cap) p50 11.1 s.

One capped rent, 8x 4090, 15 Sep 14:55Z (connector log): delete received 14:55:29.2 · graceful stop done 14:55:36.2 (2.6 s) · container removed 14:55:37.3 · `gpu power limit restore ok` x 8 from 14:55:40.9 to 14:55:54.7 (~1.8 s apart) · sweep, prune, redis, inspector stop until 14:56:00.1 · `ContainerDeleted` sent 14:56:00.1. The backend gave up at 14:55:58.5 (`filler_preemption_ms` 30,355) and the customer's create then spent 6.6 s in its own container-cleaning step.

Why ~1.8 s per GPU: `_set_and_log_power_limit` is three SSH commands (`nvidia-smi -i U -pm 1`, `nvidia-smi -i U -pl W`, the readback query), each a channel open + `nvidia-smi` on a loaded host.

Second, smaller class, not changed here: fillers that ignore SIGTERM and burn the 15 s `FILLER_CONTAINER_STOP_GRACE_SECONDS` — MINER_DEFAULT_JOB (36 of 67 deletes at the grace ceiling; 2-GPU nodes 56 % over 25 s) and ENGY_TEE on 8-GPU nodes (12 of 13 at the ceiling, 7.7 % over 25 s).

## Change

`_restore_records`: the loop body moved into `restore_one(record)`, run for every record with `asyncio.gather` under `asyncio.Semaphore(POWER_LIMIT_SET_CONCURRENCY)`. The return value is the count of records restored and cleared, as before; a record whose set failed, or whose Redis delete failed after a good set, is not counted and stays for the safety nets (`GpuPowerLimitCheck`, the create-path restore).

`raise_low_power_limits_to_default`: the below-floor GPUs are collected first (same filter as before: known state, a default limit, current under `MIN_POWER_LIMIT_RATIO` x default), then raised together under the same bound. Zero below-floor GPUs still returns 0 with only the state query on the wire.

`POWER_LIMIT_SET_CONCURRENCY = 8`: every in-flight set holds one SSH channel on the connection the delete or create already opened; OpenSSH's default `MaxSessions` is 10, so eight leaves room for the caller's own channel. A host whose `MaxSessions` is under 8 fails the extra channel opens with `asyncssh.ChannelOpenError` (Rustam's review, 16 Sep: before this round the helpers caught it as a failed set, so on such a host a create's restore and raise left every GPU past the limit capped and the customer started on them). `_set_power_limit` and `_read_back_power_state` now let that one error through (`_enable_persistence_mode`, the best-effort `-pm 1`, does not: round 2, Rustam — a refusal there used to reject a filler before `-pl` ran; it is now logged like any other failure of that step and the set goes on to `-pl`, the hard gate); `_set_side_by_side` runs the first pass under the semaphore, collects the refused GPUs, logs one WARNING naming the count, and sets them again one at a time through `_set_alone`, which turns a session refused even alone into a logged failed set (the #1115 fields). The serial cap loop in `apply_filler_gpu_power_limits` uses `_set_alone` too, so a refused session there is still a fail-closed False with the undo, never an exception into `create_container`. Tests: `test_restore_retries_the_gpus_the_host_refused_one_at_a_time` and `test_raise_retries_the_gpus_the_host_refused_one_at_a_time` (a fake sshd with MaxSessions 2: 6 of 8 GPUs refused, each on `-pm 1` and then on `-pl`, 12 refusals; all 8 restored/raised, and in the restore test every GPU's last attempt is the full verified triple; both read 2 == 8 on the head before round 1), `test_apply_fails_closed_when_the_host_refuses_the_pl_session`, `test_a_refused_pm_session_does_not_reject_the_filler_before_pl` (round 2: `-pm 1` refused, `-pl` and the readback pass, the cap is applied and `apply` returns True).

`apply_filler_gpu_power_limits` (the cap at filler start) keeps its serial loop: it is not on a customer's path, and its fail-closed undo reads more simply in order; its setter is `_set_alone` (above).

Expected after deploy: an 8-GPU PEARL delete goes from ~28 s to ~15 s (connect ~4 s + stop ~2.3 s + rm ~0.9 s + restore ~2 s + the ~6 s tail of sweep/prune/redis/inspector), under the backend's 30 s wait. The same two functions run before a customer's `docker run` (`docker_service.create_container`: `restore_tracked_gpu_power_limits` then `raise_low_power_limits_to_default`), so once DAH-3003's overlap flag is on and the create carries that restore, the create does not inherit the 15 s either.

Companion: lium-io#1249 (taiberium, DAH-2786, draft) edits the same two loop bodies to record cap reverts; whichever lands second rebases a small hunk — the per-record body is now one helper, so the merge is local.

Known follow-ups (low, from the fresh reviews): `asyncio.gather` runs without `return_exceptions=True` — the one exception the setters let through is `asyncssh.ChannelOpenError`, which `guarded` turns into a retry; anything else would propagate out of the gather as before, into the delete's `_best_effort_delete_step`, where it is logged. asyncssh raises the same `ChannelOpenError` (code `OPEN_CONNECT_FAILED`) when the whole connection dropped mid-restore, so that case also logs the MaxSessions WARNING and retries serially; every retry fails at once and the end state equals the old loop's, so the cost is one misleading WARNING line. The ENGY_TEE / MINER_DEFAULT_JOB SIGTERM class (15 s grace ceiling) is a separate ticket.

Worth it: 293 capped rents / 14 d ≈ 21 per day, each waiting ~22 s past the point the container was gone ≈ 8 min of renter waiting per day, plus the 337 rents / 14 d that waited 20 to 29 s on the same nodes. Change: 176 lines added / 22 removed in one module, 219 lines of tests (this head).

`_Setter` carries the explicit call signature (`Callable[[SSHClientConnection, Literal["cap", "restore", "raise"], str, str, int | None, int, dict[str, object] | None], Awaitable[bool]]`), so a mismatched setter is a type error (Rustam's nit, round 2).

Round 3 (Rustam's two nits after approval, 16 Sep 13:56Z): the comment in `_set_and_log_power_limit` now says a refused session from `-pl` or the readback means the set may or may not have applied (a refused readback channel comes after `-pl` ran; the retry alone re-sets the same value). The below-floor GPUs in `raise_low_power_limits_to_default` are a frozen dataclass, `_BelowFloorGpu` (`gpu_uuid`, `current_watts`, `default_watts`), instead of a tuple with a field-order comment.

### Ran it

Sandbox EC2 (Linux x86_64, c6i.2xlarge, the loop's warm lium-io verify box), run `20260916T141510Z-r7hs` at `29b56c4a5` (round 3: a comment and a dataclass in `gpu_power_limit.py`, its test module only):

```
validators pytest tests/test_gpu_power_limit.py: rc=0 in 3 s — 54 passed, 107 warnings in 0.36s
```

The full validator suite, run `20260916T130339Z-wb1y` at `42ff46de5` (round 2, the last code change beyond that module):

```
validators pytest tests/ (2 root-sensitive test(s) deselected, run below as loopverify): rc=0 in 184 s — 2577 passed, 15 skipped, 2 deselected, 265 warnings
validators pytest 2 root-sensitive test(s) as loopverify (sysfs perms, live sshd — the CI runner's privilege): rc=0 in 2 s — 2 passed
```

Fail-old for round 2: run `20260916T125810Z-jez6` at 12ff084 (this code with round 1's test expectations) fails the two `refused` tests on `12 == 6`. The earlier rounds' evidence (2c28e52: run 20260915T171111Z-lh33; 124c796: run 20260916T095439Z-p8ih, 2576 passed) stands for the side-by-side change and the retry.

### Self-review

Verdict `clean` at `29b56c4` · 15 checks · by subagent-r255 · 2026-09-16 14:22Z (tools/pr_self_review.py)

Fixed in the body: 3 body-only finding(s) answered by re-rendering the body, no code change, no new verdict (PR_PROCESS §7). `review-wave/readability/overrides/lium-io-1379.json:19` STALE-BODY (medium) — The body's 'Known follow-ups' paragraph still lists '`_set_and_log_power_limit`'s comment says a refusal made no change; the refusal can come from the readback channel after -pl applied' as an open item, but at 29b56c4… → the Known follow-ups sentence about the comment is removed; a Round 3 paragraph says the comment now reads 'may or may not have applied' and names the `_BelowFloorGpu` dataclass · `review-wave/readability/overrides/lium-io-1379.json:14` STALE-BODY (low) — Line counts are two heads old: Value says 'Change: 139 lines added / 23 removed in one module..' (double period) and Worth-it says '139 lines added / 23 removed in one module, 187 lines of tests (this head)', while `git… → Worth it now reads 176 / 22 in the module and 219 lines of tests (git diff --numstat origin/main...HEAD at 29b56c4); the override sets Fits to 'tests (8 new)' · `review-wave/readability/overrides/lium-io-1379.json:20` STALE-BODY (low) — 'Gate: PASS 42ff46d', 'Ran it: tools/aws_verify.sh lium-io 42ff46de5' and 'Self-review: clean at 42ff46d' all cite the round-2 head; round 3 changes code (the `_BelowFloorGpu` dataclass replaces the tuple in `raise_low_… → Ran it cites EC2 run 20260916T141510Z-r7hs at 29b56c4 (test_gpu_power_limit.py, 54 passed) and keeps the full-suite run wb1y at 42ff46d; Gate and Self-review are re-recorded at 29b56c4 by pr_post_push.sh / --record

Mechanical notes dismissed: `neurons/validators/src/services/gpu_power_limit.py:345` The round-3 comment is now correct and complete: ChannelOpenError escapes only from `_set_power_limit` (line 308) and `_read_back_power_state` (line 287), a readback refusal after a successful -pl leaves the GPU set with nothing logged, and `rg 'no change'` over the worktree finds no other statement of the old wording; the `_set_power_limit`, `_read_back_power_state`, `_set_alone` and `_set_side_by_side` docstrings all say 'refused session, retried by the caller' without claiming no change was made. · `neurons/validators/src/services/gpu_power_limit.py:412` `_set_alone` logs status=failed for a refusal that may have followed an applied -pl (readback channel refused); this matches the module's existing rule that an unverified set is a failed set (line 318-322 logs 'could not be read back for verification' as failed), the restore record is kept and the safety nets re-set the same idempotent value, so no GPU state is misreported beyond what the module already accepts. · `neurons/validators/src/services/gpu_power_limit.py:155` `_BelowFloorGpu` is complete: the only producer is the append at line 646 and the only consumer is `raise_one` at line 657 (attribute access, no unpacking); no test imports the class or the old tuple (`rg tuple tests/test_gpu_power_limit.py` hits only `_capped_gpus`'s return type), `test_raise_lifts_only_below_floor_gpu_to_default` and `test_raise_lifts_eight_below_floor_gpus_side_by_side` exercise it, and narrowing `default_watts` to `int` (GpuPowerState's is `int | None`) is why it is not just `GpuPowerState`. · `neurons/validators/src/services/gpu_power_limit.py:345` The comment line is 118 characters against ruff line-length 100, but E501 is ignored repo-wide in pyproject.toml and `ruff check` on both changed files passes; `ruff format --check` also rewrites pre-existing lines on main (110, 224, 315), so formatting is not enforced on this file. · `review-wave/readability/overrides/lium-io-1379.json:22` The Self-review block's mechanical note '`so no change was made` is still wrong ... already recorded under Known follow-ups' is the recorded round-2 verdict text, which tools/pr_self_review.py --record replaces with this verdict once 29b56c4 is pushed; not a separate finding. · `neurons/validators/src/services/gpu_power_limit.py:445` `asyncio.gather` without `return_exceptions=True` and the `zip(..., strict=True)` (Python >= 3.10, project pins 3.11) are unchanged from round 2, already recorded under Known follow-ups, and the only exception the setters let through is caught inside `guarded`. · `neurons/validators/src/services/gpu_power_limit.py:1` STACK-HYGIENE: branch DAH-3518-concurrent-gpu-power-restore is four commits on merge-base f6f24bfd7 of origin/main; main has moved eight commits since, none touching gpu_power_limit.py or its test, and `git merge-tree origin/main HEAD` merges clean. · `neurons/validators/src/services/gpu_power_limit.py:1` SECURITY / MIGRATION / CI-WORKFLOW / PLATFORM / DOCS-MISSING: the diff adds no route, column, workflow, image, or provider-visible behaviour; every nvidia-smi command still quotes the uuid with shlex.quote and the watts are ints from nvidia-smi's own CSV, so the §5 properties are unchanged by this PR.

### Verified

Gate: PASS (29b56c4) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1379)

### What changed

- `services/gpu_power_limit.py`: `_restore_records` and `raise_low_power_limits_to_default` set their GPUs concurrently, bounded by `POWER_LIMIT_SET_CONCURRENCY = 8`
- `_set_side_by_side`: a GPU whose SSH session the host refused (`asyncssh.ChannelOpenError`, `MaxSessions` below the bound) is retried alone after the first pass; `_set_power_limit` and `_read_back_power_state` let that one error through, `_set_alone` logs it as a failed set for the serial cap loop. `_enable_persistence_mode` (`-pm 1`, best-effort) catches it like any other failure of that step, so a refused session there no longer rejects a filler before `-pl` ran (Rustam, round 2)
- `tests/test_gpu_power_limit.py`: a suspending SSH fake counts in-flight commands; eight tests (8 GPUs, the bound, one hung GPU, the raise, a MaxSessions-2 host for the restore and the raise, a refused `-pl` session in the cap loop, a refused `-pm 1` session that does not reject the filler)

</details>


neurons/validators/src/services/gpu_power_limit.py
```diff
@@ -500,19 +518,27 @@ async def raise_low_power_limits_to_default(
...
+    # Side by side, like the restore: this runs before the customer's `docker run`.
+    limit = asyncio.Semaphore(POWER_LIMIT_SET_CONCURRENCY)
+
+    async def raise_one(gpu_uuid: str, current_watts: int, default_watts: int) -> bool:
+        async with limit:
+            return await _set_and_log_power_limit(
+                ssh, "raise", executor_id, gpu_uuid, current_watts, default_watts, log_extra
+            )
+
+    outcomes = await asyncio.gather(*(raise_one(*target) for target in below_floor))
```
taiberium: Please preserve `ChannelOpenError` and retry affected GPUs serially before returning. If `MaxSessions` is less than 8, `create_container` ignores the partial result and starts with capped GPUs.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/services/gpu_power_limit.py
surcyf123: Done in `124c796`: a refused session is set again alone. The three helpers let `ChannelOpenError` through. `_set_side_by_side` collects the refused GPUs after the first pass and sets them one at a time; the cap loop uses `_set_alone` and stays fail-closed. Two tests fake a host with MaxSessions 2: six of eight opens refused, all eight restored and raised.

neurons/validators/src/services/gpu_power_limit.py
```diff
@@ -219,6 +238,8 @@ async def _enable_persistence_mode(
         )
         if result.exit_status != 0:
             failure = f"exit={result.exit_status}, stderr={result.stderr!r}"
+    except asyncssh.ChannelOpenError:
```
taiberium: `-pm 1` is best-effort, but rethrowing its `ChannelOpenError` rejects a filler before `-pl`. Please catch it here because hard-gate failures still trigger retries.

neurons/validators/src/services/gpu_power_limit.py
```diff
@@ -55,6 +59,10 @@
 
 logger = logging.getLogger(__name__)
 
+_T = TypeVar("_T")
+# _set_and_log_power_limit or _set_alone: (ssh, action, executor_id, gpu_uuid, before, after, log_extra) -> set ok
+_Setter = Callable[..., Awaitable[bool]]
```
taiberium: nit: Please give `_Setter` an explicit call signature so type checking can reject mismatched setters. This is non-blocking and fine as a follow-up.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/services/gpu_power_limit.py
surcyf123: Done in `42ff46d`: `_Setter` now carries the explicit call signature: the SSH connection, the action literal, executor id, GPU uuid, watts before, watts after, the log extra, returning an awaitable bool. It is the one signature both setters share. mypy strict on the module reports nothing.

neurons/validators/src/services/gpu_power_limit.py
surcyf123: Done in `42ff46d`: `_enable_persistence_mode` no longer re-raises the refused session. It is logged like any other failure of that best-effort step, and the set goes on to -pl. A refusal on -pl is what the caller retries. New test: -pm 1 refused, -pl and the readback pass, the cap is applied and apply returns True. It read False on the round-1 module. The fail-closed cap test now refuses the -pl session. In the two MaxSessions-2 tests each refused GPU is refused twice on the way, on -pm 1 and then on -pl. So they count 12 refusals, and all 8 GPUs still end restored or raised. Full suite on EC2: 2577 passed.

neurons/validators/src/services/gpu_power_limit.py
```diff
@@ -293,6 +331,10 @@ async def _set_and_log_power_limit(
     log_extra: dict[str, object] | None,
 ) -> bool:
     # Reviewer contract (PR #1115): every PL change is logged with executor, GPU, before/after, status.
+    # Raises only asyncssh.ChannelOpenError, from -pl or the readback: the host refused the session,
+    # so no change was made and nothing is logged here; _set_side_by_side retries the GPU alone,
```
taiberium: nit: when the refused channel is the readback, `-pl` already applied, so "no change was made" is wrong here. Non-blocking, fine as a follow-up.
```suggestion
    # so the set may or may not have applied and nothing is logged here; _set_side_by_side retries the GPU alone,
```

neurons/validators/src/services/gpu_power_limit.py
```diff
@@ -500,19 +626,22 @@ async def raise_low_power_limits_to_default(
         _log(logging.ERROR, f"gpu power raise: state query failed: {exc}; leaving limits as-is", {}, log_extra)
         return 0
     target_uuids: list[str] = gpu_uuids if gpu_uuids else list(state_by_uuid)
-    raised = 0
+    below_floor: list[tuple[str, int, int]] = []  # gpu_uuid, current watts, default watts
```
taiberium: nit: can this be a small frozen dataclass (`gpu_uuid`, `current_watts`, `default_watts`) instead of a tuple that needs a comment for its field order? Non-blocking, fine as a follow-up.

taiberium (approved):

neurons/validators/src/services/gpu_power_limit.py
surcyf123: Done in `29b56c4`, with your text. The readback runs after -pl, so a refused readback channel comes after the set went through. The comment now says the set may or may not have applied. The retry alone sets the same value again, so the end state is the same either way.

neurons/validators/src/services/gpu_power_limit.py
surcyf123: Done in `29b56c4`: The tuple is now `_BelowFloorGpu`, a frozen dataclass with the three fields you named: gpu_uuid, current_watts and default_watts. `raise_one` reads them by name and the field-order comment is gone. The module's 54 tests pass on EC2 (run 20260916T141510Z-r7hs).

taiberium (approved):
Round 3 fixes look good.

PR #1376, opened 2026-09-15
DAH-3504 - [P2] validator: a failed custom-Dockerfile build carries its last output lines

surcyf123:
Implements DAH-3504 · reviewer: @taiberium

**Problem** — 38 of 40 custom-Dockerfile rentals in 14 d never got a container; the 22 `docker_build` failures all reached the backend as `failure_step=docker_build` and nothing else.

**Value** — not measured — the body has no Worth it line (SO §69) and the ticket gives no prod number — a human decides.

**Goal** — DAH-3504 (P2).

**Fits** — Sits in `lium-io` — `neurons/validators`; touches validator, tests (5).

**Flag:** no flag

**Risks**
- The DinD build now pipes through `tee`/`tail` (busybox `sh`): a DinD image without them fails every build at `docker_build` — `daturaai/dind:0.0.1` (Alpine) has both — rollback: revert.
- Up to 2000 chars more per failed-build CCF message on the wire — the backend ignores the new field until lium-platform mirrors it (DAH-3504 part 2).
- Not changed: CCF step names, `msg`, cleanup order, the image-pull path; no config change, no migration.

**How to verify**
- `cd neurons/validators && pytest tests/test_custom_dockerfile_build.py -q` → 21 passed (on main the 5 new fail: no `build_log_tail`, `expected 3, got 2`, `ImportError`)
- `sh -c` the rendered build command with a `docker` stub printing two lines and exiting 1 → stderr shows the two lines then `BUILD_FAILED_RC=1`, exit 1

**Verified by the loop:** Gate: PASS 7841d84

**Ran it:** `pytest tests/test_custom_dockerfile_build.py -q`… · EC2 20260915T071428Z-juve; the same file on `origin/main` → 12 failed, 9 passed · … +1 under Details · head moved to 7841d84 — re-run

**Self-review:** clean at 7841d84 (carried from 2f91e2c: rebase-only, patch unchanged; 17 checks, 4 low)

**Parts:** validator ✓ · migration n/a — no schema change · backend n/a — lium-platform mirrors `build_log_tail` in DAH-3504 part 2 · frontend n/a — no UI in lium-io · CLI/SDK n/a — `lium up` prints the backend's event · docs n/a — internal wire · tests ✓ 5 · dashboards n/a — no new metric

**Docs:** none changed in this PR — the field travels on the validator→backend websocket, not a renter surface; the renter-visible text is lium-platform's (part 2).

<details><summary>Details</summary>

### What changed

[...]

PR #1381, opened 2026-09-15
DAH-3521 - [P1] validator: a custom build reaches the node's own DNS resolver…

surcyf123:
Implements DAH-3521 · reviewer: @taiberium

**Problem** — 18 of 23 `docker_build` failures in 14 d (the top class behind 38/40 custom-Dockerfile rentals never getting a container) are our egress firewall dropping the build container's DNS on nodes whose resolver sits in a blocked range: `FROM` fails after the resolver timeout, 12 to 23 s per node.

**Value** — Removes the top cause of custom-build failure: 18 of 23 `docker_build` failures in 14 d were the DNS drop, and 38 of 40 custom-Dockerfile rentals never got a container; each affected node burned 12 to 23 s of resolver timeout before failing.

**Goal** — DAH-3521 (P1).

**Fits** — Sits in `lium-io` — `neurons/validators`; touches validator, the orphan sweep check, config, tests (21).

**Flag:** no flag

**Risks**
- A resolver inside the block now gets DNS from the build container: port 53 only, that one address. Every other private address and port stay dropped; metadata on 169.254.169.254:80 verified blocked on EC2 after the fix.
- A DinD image pull over 180 s now fails at `build_dind_start` instead of waiting; rollback: raise `CUSTOM_DOCKERFILE_SETUP_STEP_TIMEOUT_SECONDS`.
- The purge reads `iptables -S DOCKER-USER` inside the helper container and needs `xt_comment` on the host; no private resolver: only the purge line is added to today's script (A21, A22b).

**How to verify**
- `cd neurons/validators && pytest tests/test_custom_dockerfile_build.py tests/test_custom_build_orphan_sweep.py -q` → 41 + 7 passed; the same files on the round-7 head: the sweep file does not import (`dind_container_pod_id` is new) and A23b to A23e fail (the fakes serve the probe only through `create_process`); every earlier round's negative control is under Ran it
- Loki `{namespace="subnet"} |= "Failed create_container" |= "failure_step=docker_build"`, 1 to 14 Sep: 23 rows; build durations per node in Details

**Verified by the loop:** Gate: PASS 766d624

**Ran it:** validators suite on a sandbox EC2 box (`aws_verify` run 20260918T180542Z-mnk8 at the round-8 head): 2596 passed, 15 skipped, plus… · … +1 under Details · head moved to 766d624 — re-run

**Self-review:** clean at 766d624 (10 checks, 2 low, 1 fixed in body)

**Parts:** validator ✓ · config ✓ · protocol n/a — no wire change · migration n/a — no schema · backend n/a — nothing to mirror · frontend n/a — no UI in lium-io · CLI/SDK n/a — `lium up` unchanged · docs n/a — internal setting (Docs line) · tests ✓ 19 (25 cases) + 2 sweep · dashboards n/a — no new metric

**Docs:** none needed because the surface is internal and operator-only: `CUSTOM_DOCKERFILE_SETUP_STEP_TIMEOUT_SECONDS` is read by the validator, which Datura operates; providers and renters neither see nor set it, the validator's `.env.template` lists none of the `CUSTOM_DOCKERFILE_*` settings (it carries 42 of `Settings`' 111), and nothing a provider or renter types or reads changes.

### Review round 2 (taiberium, 16 Sep 11:05Z)
- Tag and purge: every DNS ACCEPT carries `-m comment --comment lium-dind-dns`. Apply and teardown read `iptables -S DOCKER-USER` and delete every rule with the DinD's source IP and the tag, as printed, before the inserts. A failed teardown plus DinD IP reuse with another resolver no longer leaves the old resolver allowed; the delete never needs the old resolver's address (A21b runs the purge line against a fake chain listing).
- `CUSTOM_DOCKERFILE_DIND_READY_TIMEOUT_SECONDS` is `gt=0`: `range(0)` ran no probe and every build failed at `build_dind_unready` (A25b).
- The nit: the outer `wait_for` capped readiness at N seconds wall time. The loop is back to N probes a second apart, each bounded by min(N, 10) s (at round 2 through `ssh_client.run(timeout=)`), so a hung `docker info` ended the wait within 10 s; round 4 changed this to one not-ready probe, and a slow one cannot stretch the stage past N × 11 s (A23b, A23c).
- Self-review round 3 (subagent-r247b) folded: the purge captures `iptables -S` and checks its exit status, since a pipe made a failed listing a silent no-op; on apply a failed listing exits 4 and the build aborts, on teardown it is tolerated so the DROP deletes still run (A21b covers both); `_teardown_dind_build` no longer takes `dns_servers`.
- Round 3 (taiberium, 13:04Z): a tagged rule the apply cannot delete aborts the apply too (`$IPT -D $spec || exit 5` inside the purge loop, `|| { echo "tagged DNS rule delete failed" >&2; exit 5; }` after `done`), so a reused DinD IP never builds with an old resolver's ACCEPT left in the chain; the purge now runs before the DROP inserts, so an abort leaves no half-applied DROP behind for a teardown that never runs; the teardown still tolerates a failed delete and goes on to the DROPs (A21c runs the whole apply and teardown scripts against a fake `iptables` whose `-D` fails).
- lium-io#1376/#1377 edit `_custom_build_image`'s later steps and will need a rebase after this lands (or this after them).
- Round 4 lows, not folded: the `CUSTOM_DOCKERFILE_DIND_READY_TIMEOUT_SECONDS` description still calls the value the bound on a single probe (the bound is min(N, 10)); an apply that fails after its DROP inserts leaves `egress_applied` False, so the teardown gets no `dind_ip` and the four DROPs for that bridge IP stay until the next build on it purges them (the remove script is idempotent; passing `dind_ip` whenever inspect succeeded is the follow-up) (both folded: the `dind_ip` half in round 5, the description in round 4's nit commit).

### Measured (prod, read-only, 15 Sep 19:5xZ; `rental_history` + `event`, Loki `{namespace="subnet"}`)

40 custom-Dockerfile rentals (`rental_history.docker_image='lium-build-pending'`, 1 to 15 Sep), 11 renters; 38 never got a container. Classes (per `event` `executor-rent.failed` `failure_step`, joined by renter + executor + 45 min window):

| class | n | note |
|---|---|---|
| `docker_build` | 23 | 8 renters, 10 executors; this PR |
| `volume_creation` | 10 | DAH-3505, lium-io#1377 |
| renter deleted during the build | 6 | incl. 2 `cancelled_by_delete` |
| `build_timeout` (1200 s) | 1 | |
| executor never reported | 1 | DAH-3351, lium-platform#449 |
| got a container | 2 | |

Build duration for the 23 (Loki `Building custom image` → `Failed create_container`, same pod_id) is a constant per node, not per Dockerfile:

| node (executor; addresses in the lane's REPORT.md, not here) | failures | build duration |
|---|---|---|
| node A | 3 | 12.6 to 14.0 s |
| node B | 2 | 12.2 s |
| nodes C and D (same provider) | 5 | 12.6 to 15.5 s |
| node E | 8 | 23.2 to 23.4 s |
| nodes F and G | 2 | 30.4 and 31.4 s |
| node H | 1 | 4.1 s |
| nodes I and J | 2 | 61 and 207 s (real build errors) |

- 12 to 15 s = Go's resolver giving up on one nameserver (2 attempts × 5 s, plus `docker exec` start); 23 s = the same with two nameservers: 18 of 23 carry that signature. 30 s (2 more) = the registry dial timeout with DNS answering. 2 are real build errors (61 s, 207 s); 1 (4 s) is too fast for either.
- One renter rented two nodes at the same minute three times on 2 Sep (17:58, 18:39, 20:12): node node A failed `docker_build` in 13 s each time; node `5FFXLu…` built for 9 to 12 min each time (then hit DAH-3505). Same renter, same Dockerfile, same minute.
- The build output that would have said `lookup registry-1.docker.io … i/o timeout` lived only in `pod_status.logs`, deleted with the pod (DAH-3504 carries it from now on).

### Why
`_custom_build_image` fences the DinD build container with four DROP rules in the host's `DOCKER-USER` chain (`-s <dind ip> -d <cidr> -j DROP` for 169.254.0.0/16, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16). Docker copies the host's upstream resolvers into the DinD's `/etc/resolv.conf` (on a systemd-resolved host, the upstream servers, not 127.0.0.53). On a node whose resolver is private (a cloud VPC resolver at the .2 of a 10/8 or 172.16/12 network, a datacenter resolver in 10/8, a router in 192.168/16) the inner dockerd's DNS queries traverse FORWARD → DOCKER-USER and are dropped. The host's own `docker pull` works, so nothing else on the node shows it.

### Design notes
- The resolver list is read from the DinD container itself (`docker exec <dind> cat /etc/resolv.conf`): that file is what its dockerd queries. Only IPv4 addresses inside a blocked range get a rule; a public resolver needs none. Unreadable → today's rules, the build runs.
- ACCEPT lines run after the DROP lines in the script, so `-I` lands them above the DROPs. Each opens udp and tcp port 53 to one address from one source IP. A malicious Dockerfile could send DNS queries to that resolver; it could not reach any other private address or any other port.
- `--dns 8.8.8.8` on the DinD was the alternative; rejected because nodes that block outbound public DNS would then fail, and the node's own resolver is what the host trusts.
- Setup bound: `docker run -d <dind image>` pulls the DinD image on a node's first custom build and had no timeout; a hang there left the pod PENDING until the backend's 1 h stale sweep. The DinD start now runs under the executor's `timeout -k 5 180` (kills the pull too; asyncssh's own `timeout=` only stops waiting) and fails at `build_dind_start`; the sysbox preflight, `inspect`, the resolver read and the Dockerfile write carry `timeout=180`; the firewall helper runs under the executor's `timeout -k 5 180` with the streamer at 195 s as the backstop (round 6) and fail at their own steps (`build_sysbox_unavailable`, `build_setup`, or through `create_container`'s handler); the readiness loop runs up to `CUSTOM_DOCKERFILE_DIND_READY_TIMEOUT_SECONDS` probes a second apart, each under the executor's `timeout -k 2 min(N, 10)` and asyncssh's bound + 5 s as the backstop, so a probe that hits its bound counts as one not-ready probe, and a dockerd that never answers fails at `build_dind_unready` after about N × 10 s; worst case for the readiness stage, every probe taking its 10 s bound and answering not ready: N × 11 s, 11 min at the default 60. Worst case before the build with every command hanging: six commands at 180 s, 15 s slack on the start, 10 s for the hung probe, then the three teardown commands (the remove helper at 195 s, `rm -f` of the two helpers and `rm -fv` of the DinD at 180 s each): about 28 min, against 1 h before. An `inspect` or resolver-read timeout returns `build_egress_setup`; coreutils' exit 124 on the start writes the same renter line as the asyncssh path.

### Review round 1 (taiberium, 16 Sep)
- `-C || -I` on the ACCEPT: `-C` matches a rule anywhere in the chain. When a teardown removed the DROPs but not the ACCEPT (a `-D` that lost the xtables lock is swallowed by `|| true`), the ACCEPT for that DinD IP and resolver is still in `DOCKER-USER`; the next build on the same DinD IP inserts its DROPs at the top, `-C` finds the stale ACCEPT below them and skips the insert, and DNS is dropped again. Now every matching ACCEPT is deleted in a `while $IPT -D …; do :; done` loop and one is inserted at the top; the teardown deletes every copy the same way. The DROPs keep `-C || -I`: a stale DROP anywhere is below an ACCEPT inserted at position 1 afterwards.
- `CUSTOM_DOCKERFILE_SETUP_STEP_TIMEOUT_SECONDS` gets `gt=0`, like `RENTAL_PROBE_SSH_DEADLINE_SECONDS`: asyncssh's `run(timeout=0)` and a negative value time out at once, so every setup command would fail on the spot; `execute_and_stream_logs(timeout=0)` disables its bound, so the firewall helper would run unbounded. Rejected at settings load (A25).

### Not this PR
- The two 30 s failures (nodes F and G) may be a registry dial timeout with DNS working: a proxy-only node (`docker info` `HTTPProxy` set on the host dockerd, not inherited by the inner dockerd) or a NAT gap. DAH-3504's `build_log_tail` will show which.
- The 4.1 s failure on node H: too fast for a timeout; a Dockerfile parse error or an immediate NXDOMAIN. Same answer: the tail.
- A DNS preflight inside the DinD (`nslookup registry-1.docker.io` with a 5 s bound, its own step name) would name this class in seconds on any node the fix misses; left out to keep the PR to the cause.

### Known follow-ups
- `$IPT` runs without `-w` (pre-existing): a concurrent xtables holder makes one `-C`/`-I`/`-D` fail; the apply then fails closed at `build_egress_setup`, the teardown logs and moves on.

Round 4 (17 Sep, Rustam): a readiness probe that hits its bound is one not-ready probe, not the end of the wait; the remaining probes run. The bound is `timeout -k 2 N` on the executor (exit 124 = not ready), as for the DinD start, so a slow probe leaves no session channel open; asyncssh's own timeout (N + 5 s) is the backstop. Worst case stays N × (10 + 1) s.

Round 5 (17 Sep, Rustam, nit): the teardown gets `dind_ip` whenever the IP is known, not only after a successful apply. An apply that timed out may have inserted some rules before the deadline; the remove script is idempotent (`-D … || true`, purge by tag), so running it after a failed apply costs one helper container and removes whatever landed. A13 now also asserts the remove script ran for the DinD IP on the egress-failed path.

Round 6 (18 Sep, Rustam): both firewall helpers (`_egress_filter_script` apply in `_custom_build_image`, remove in `_teardown_dind_build`) run under the executor's `timeout -k 5 <setup bound>` through `_bounded_helper_command`, which echoes one stderr line on exit 124 so `execute_and_stream_logs` (stderr = failure) fails the apply at `build_egress_setup`; before, the streamer's own `timeout=` only stopped reading and the remote helper went on inserting rules. Each helper is named (`<dind>-fw-apply`, `<dind>-fw-remove`) and the teardown runs `docker rm -f` on both names before `docker rm -fv <dind>` releases the DinD IP, so a helper that outlived its bound cannot keep editing rules for an IP the next build gets; the remove helper's exit 124 is logged as a failed teardown. The streamer's and asyncssh's timeouts stay as backstops at the bound + 15 s. Review references in the `docker_service.py` comments are gone (the PR keeps that history). A24 asserts the apply command; A24b runs the wrapper in `sh -c` (124 + stderr line, a quick command's own status, a failing command's own stderr); A26 (×2) asserts the remove helper's command and bound, the order remove helper → `rm -f` both helpers → `rm -fv` DinD, and the warning on exit 124.

Round 7 (18 Sep, Rustam): the apply step fails on the helper's exit status, not on a stderr line. `execute_and_stream_logs` failed a step on stderr only, so a helper that `timeout -k 5` killed with SIGKILL after the grace period (exit 137, the wrapper's 124 line never prints), an OOM-killed helper, or a `docker run` that failed before iptables ran, read as an applied firewall and the build started open. `execute_and_stream_logs(check_exit_status=True)` waits for the channel to close, reads `process.exit_status`, and fails the step on any value other than 0 (None, a signal death with no status, included), with `Process exited with status <rc>` in the error and one error line in the pod log; the default stays stderr-only, so no other caller changes. The apply call passes it; r305's `timeout -k 5` wrapper and its 124 stderr line stay (the pod log still says why). A24 asserts the flag on the apply call; A24c (137, 1) and A24d (0) run `_custom_build_image` through the real streamer with a fake process whose exit status arrives at `wait_closed()`; A24e (137, 1, None) covers the streamer itself, with and without the flag, and the raising path.

Round 8 (18 Sep, Rustam): three changes. (1) `CustomBuildOrphanSweepCheck` read `<pod_id>-fw-apply` as a pod id: `docker ps --filter name=lium-dind-build-` lists a build's two firewall helpers with its DinD, no pod has that id, and the sweep removed a helper of an ACTIVE build while it was inserting rules. `dind_container_pod_id()` strips the `-fw-apply` / `-fw-remove` suffix (the sweep's copy of `_dind_firewall_helper_names`) before the active-pod check; a helper whose pod is gone is still removed. Tests: `test_sweep_keeps_the_firewall_helpers_of_an_active_build` (a helper of an active pod survives the sweep, the orphan's helper goes) and `test_dind_container_pod_id_agrees_with_the_names_docker_service_gives` (the suffixes match `_dind_firewall_helper_names`); both fail on the round-7 head. (2) The readiness probe ran through `ssh_client.run(timeout=)`, which drops its process when the wait times out and leaves the session channel open until the remote command exits. On a host where even `timeout(1)` did not return, every probe that hit the backstop kept a channel, and sshd's MaxSessions (10) refused the eleventh. The probe is now `async with ssh_client.create_process(...)` + `process.wait(check=False, timeout=...)`, as `execute_and_stream_logs` runs its commands: the `async with` exit closes the channel (`close`, `wait_closed`) before the next probe opens one, on a timeout as on a result. Same command, same `timeout -k 2 min(N, 10)` bound, same backstop, same `exit_status` reading; the success path is unchanged. A23e: the first probe outlives the backstop, its channel is closed before the second opens, the second is ready and the build runs; the fakes now serve the probe through `create_process` with asyncssh's `wait` / `close` / `wait_closed` contract. (3) `_teardown_dind_build`'s local `from core.config import settings` is gone; the module imports it at the top.

### Ran it

EC2 warm lium-io box (c6i.2xlarge, the validators venv), review round 8 (18 Sep), the round-8 head:

```
20260918T180542Z-mnk8  validators pytest tests/ (2 root-sensitive deselected, run below as loopverify): rc=0 in 197 s — 2596 passed, 15 skipped, 2 deselected
20260918T180542Z-mnk8  validators pytest 2 root-sensitive tests as loopverify: rc=0 — 2 passed
```

Negative control, round 8 — the round-8 test files committed on the round-7 head 2c4ffe4 (negctl commit, not on the branch). The round-8 fakes serve the readiness probe only through `create_process`, as asyncssh's process contract, so the round-7 code (probe through `ssh.run(timeout=)`) never reaches them:

```
20260918T180737Z-ol2m  validators pytest tests/test_custom_dockerfile_build.py tests/test_custom_build_orphan_sweep.py: rc=2 — 1 error during collection
ERROR tests/test_custom_build_orphan_sweep.py   ImportError: cannot import name 'dind_container_pod_id' (new in round 8)
20260918T181047Z-1qrn  validators pytest tests/test_custom_dockerfile_build.py: rc=1 in 2 s — 4 failed, 37 passed
FAILED test_A23b_a_hung_readiness_probe_fails_at_build_dind_unready_within_the_ready_bound   (the round-7 probe goes through `ssh.run`, which the round-8 fake answers exit 0: "ready" at once)
FAILED test_A23c_readiness_keeps_its_probe_budget_when_probes_fail_fast                     (same: 1 probe, not 2)
FAILED test_A23d_a_probe_that_hits_its_bound_is_not_ready_and_the_next_probe_runs           (same)
FAILED test_A23e_a_probe_that_outlives_the_backstop_is_closed_before_the_next_one_opens     (no `create_process` call: `probe_processes` is empty)
```

EC2 warm lium-io box (c6i.2xlarge, the validators venv), review round 7 (18 Sep), the round-7 head 2c4ffe4:

```
20260918T083302Z-35ik  validators pytest tests/ (2 root-sensitive deselected, run below as loopverify): rc=0 in 196 s — 2593 passed, 15 skipped, 2 deselected
20260918T083302Z-35ik  validators pytest 2 root-sensitive tests as loopverify: rc=0 — 2 passed
20260918T083227Z-qk1b  validators pytest tests/test_custom_dockerfile_build.py: rc=0 in 13 s — 40 passed
```

Round 7's new cases go through the real `execute_and_stream_logs` with a fake `create_process` whose exit status is known only after `wait_closed()`, as asyncssh's is: A24c (helper exit 137 and 1, nothing on stderr) → `build_egress_setup`, no `docker build`, DinD torn down; A24d (exit 0) → the build runs; A24e (137, 1, None) → the streamer fails only with `check_exit_status=True`, and raises with `raise_exception`.

EC2 warm lium-io box (c6i.2xlarge, the validators venv), review round 6 (18 Sep), the round-6 head c3ca081:

```
20260918T070659Z-h5b4  validators pytest tests/ (2 root-sensitive deselected, run below as loopverify): rc=0 in 195 s — 2587 passed, 15 skipped, 2 deselected
20260918T070659Z-h5b4  validators pytest 2 root-sensitive tests as loopverify: rc=0 — 2 passed
20260918T071439Z-q9on  validators pytest tests/test_custom_dockerfile_build.py: rc=0 in 13 s — 34 passed
```

A24 with the round-6 assertions fails on 8530ffe (the apply command had no `timeout` prefix, no `--name`, and the streamer bound was 11, not 26); A26 fails there too (no `docker rm -f` of the helpers, the remove helper unbounded).

EC2 warm lium-io box (c6i.2xlarge, the validators venv), review round 2 (16 Sep). The head differs from the full-suite run only by one deleted assertion in A25b (a kwarg round-trip the gate called a tautology); the file was re-run at the head:

```
20260916T122238Z-78dh  validators pytest tests/test_custom_dockerfile_build.py: rc=0 in 6 s — 29 passed
20260916T121623Z-cete  validators pytest tests/ (2 root-sensitive deselected, run below as loopverify): rc=0 in 186 s — 2582 passed, 15 skipped, 2 deselected (the 29 in tests/test_custom_dockerfile_build.py included)
20260916T121623Z-cete  validators pytest 2 root-sensitive tests as loopverify: rc=0 — 2 passed
```

Review round 1 (16 Sep), the round-1 head:

```
20260916T094711Z-6s9r  validators pytest tests/ (2 root-sensitive deselected, run below as loopverify): rc=0 in 185 s — 2578 passed, 15 skipped, 2 deselected (the 25 in tests/test_custom_dockerfile_build.py included)
20260916T094711Z-6s9r  validators pytest 2 root-sensitive tests as loopverify: rc=0 — 2 passed
```

Negative control, round 1 — the round-1 test file committed on the round-0 head (negctl commit, not on the branch):

```
20260916T095034Z-fv7g  validators pytest tests/test_custom_dockerfile_build.py: rc=1 in 17 s — 3 failed, 22 passed
FAILED test_A21_egress_script_allows_port_53_to_the_resolver_above_the_drops   (the ACCEPT was still `-C || -I`)
FAILED test_A25_setup_step_timeout_rejects_zero_and_negative[0]                (no `gt=0`)
FAILED test_A25_setup_step_timeout_rejects_zero_and_negative[-1]
```

Round 0 (15 Sep), the round-0 head's tree:

```
20260915T204506Z-o54q  validators pytest tests/test_custom_dockerfile_build.py: rc=0 in 4 s — 23 passed, 108 warnings in 1.85s
20260915T203640Z-k42r  validators pytest tests/ (2 root-sensitive deselected, run as loopverify): rc=0 in 185 s — 2576 passed, 15 skipped (the head before the fold; the fold adds timeouts on the teardown, the exit-124 text and two except blocks, all covered by the file above)
```

Negative control — the same test file committed on `origin/main` f6f24bfd7 (negctl commit, not on the branch):

```
20260915T203541Z-ovd4  validators pytest tests/test_custom_dockerfile_build.py: rc=1 in 8 s — 6 failed, 17 passed
FAILED test_A20_dind_nameservers_inside_the_block_are_the_only_ones_kept
FAILED test_A21_egress_script_allows_port_53_to_the_resolver_above_the_drops
FAILED test_A22_build_on_a_host_with_a_private_resolver_lets_dns_through
FAILED test_A23_setup_commands_are_bounded_and_a_hung_dind_start_fails_the_step
FAILED test_A23b_a_hung_readiness_probe_fails_at_build_dind_unready_within_the_ready_bound
FAILED test_A24_egress_helper_runs_under_the_setup_bound
```

Reproduction on a plain EC2 box (c6i.large, AL2023, Docker 25, `daturaai/dind:0.0.1` run `--privileged` in place of sysbox-runc; the host resolver is the VPC resolver, the .2 of a 172.16/12 network), script `RECOVERY/workers/bug-custom-dockerfile-builds/repro_dind_dns.sh`, run `20260915T200456Z-bgp9`:

```
== PART 1: build under today's rules (the four DROP rules for the DinD IP)
#2 ERROR: failed to do request: Head "https://registry-1.docker.io/v2/library/alpine/manifests/3.20": dial tcp: lookup registry-1.docker.io on <resolver>:53: read udp <dind ip>:41169-><resolver>:53: i/o timeout
build exit code 1 after 20 s
== PART 2: same rules + ACCEPT :53 to the DinD's nameservers (the fix)
allowed :53 to <resolver>
#5 0.180 built-ok
build exit code 0 in under 1 s
== PART 3: metadata endpoint still blocked from the DinD
METADATA_BLOCKED
```

No executor ran this; the validator path is exercised by A22 through `_custom_build_image` with a routed `ssh.run` stub.

### Self-review

Verdict `findings` at `766d624` · 10 checks · by subagent-r324b · 2026-09-18 18:21Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | PROSE-VS-CODE | `neurons/validators/tests/test_custom_dockerfile_build.py:1373` | The body's round-6 paragraph says 'Review references in code comments are gone', but the A24c docstring still opens with 'Regression (taiberium, round 2):'; the src files and the sweep test carry none. | Drop '(taiberium, round 2)' from the docstring (the PR thread keeps the history), or soften the body sentence to 'in `docker_service.py`'. |
| low | CORRECTNESS | `neurons/validators/src/services/docker_service.py:4488` | Property: a probe whose backstop fired returns to the loop within a bound. The `async with` exit (`SSHProcess.__aexit__` → `close()` + `wait_closed()`, asyncssh 2.24) awaits the remote's CHANNEL_CLOSE reply with no timeout; sshd answers at once for a still-running command (RFC 4254 §5.3, OpenSSH `chan_rcvd_oclose`), so only a dead transport blocks it, and there the old path blocked on the next `run()`'s channel open instead — the same exposure `execute_and_stream_logs` (line 1817) already carries. | Accept as-is (the comment at 4472 already points at `execute_and_stream_logs`); if a hard bound is wanted, unroll the `async with` into `try: wait(timeout) finally: process.close(); await asyncio.wait_for(process.wait_closed(), 5)`. |

Fixed in the body: 1 body-only finding(s) answered by re-rendering the body, no code change, no new verdict (PR_PROCESS §7). `neurons/validators/src/services/task/checks/custom_build_orphan_sweep.py:57` STALE-BODY (medium) — The body describes none of this round: the Design notes still say the readiness probe is `ssh_client.run(timeout=min(N, 10))` while the code at docker_service.py:4488 runs `create_process` + `wait(timeout=min(N, 10) + 5… → the override now carries a Round 8 paragraph (the three changes, A23e and the two sweep tests by name, both fail on the round-7 head), the Design-notes probe sentence reads `timeout -k 2 min(N, 10)` on the executor with asyncssh's + 5 s backstop, What changed…

Mechanical notes dismissed: `neurons/validators/src/services/docker_service.py:4114` `--network=host` in `_dind_firewall_helper_names`' docstring names the iptables helper that edits the host DOCKER-USER chain and needs the host network namespace; unchanged by this round (r317b dismissal carries over). · `neurons/validators/src/services/docker_service.py:4747` the remove helper's `docker run --rm --network=host` (host namespace note and the f-string-without-placeholder note): same helper, same purpose; the line is one segment of a concatenated f-string whose neighbours interpolate `remove_helper` / `dind_image`; unchanged since c3ca081. · `neurons/validators/src/services/docker_service.py:4793` `# noqa: BLE001 — best-effort` carries its reason inline; `_teardown_dind_build`'s docstring says failures are logged, never raised; the round only removed the local `settings` import above it. · `neurons/validators/src/services/docker_service.py:4137` `f"exit $rc"` is one segment of a multi-line concatenated f-string in `_bounded_helper_command`; `$rc` is shell; style only, unchanged. · `neurons/validators/src/services/docker_service.py:4293` one segment of the purge script's concatenated f-string whose neighbours interpolate `{dind_ip}` / `{DIND_DNS_RULE_TAG}`; style only, unchanged. · `neurons/validators/src/services/docker_service.py:4294` the line IS an f-string (`f'case "$spec" in *"-s {dind_ip}/32 "*"--comment {DIND_DNS_RULE_TAG} "*) '`); scanner false positive, unchanged. · `neurons/validators/tests/test_custom_dockerfile_build.py:291` `--network=host` in test routers, fixtures and asserts matching the helper's command string (the router key in `_make_dind_ssh` / `_make_process_ssh`); no container runs; applies to all 13 test-file notes (291, 777, 1050, 1057, 1080, 1266, 1273, 1356, 1390, 1400, 1416, 1485, 1493), which are r317b's 13 lines shifted by the round's added fake and A23e. · `neurons/validators/tests/test_custom_dockerfile_build.py:1286` `skipif(shutil.which("timeout") is None)` guards A24b, a subprocess test needing coreutils `timeout(1)`, which the executor and the EC2 runner have (40 passed, 0 skipped at 2c4ffe4); A24c/A24e cover the exit-status path without the binary; unchanged.

### Verified

Gate: PASS (766d624) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1381)

### What changed

- `docker_service.py` `_parse_dind_nameservers`: the DinD's IPv4 nameservers inside a blocked range
- `docker_service.py` `_egress_filter_script(..., dns_servers)`: ACCEPT udp/tcp 53 per resolver above the DROPs, tagged `-m comment --comment lium-dind-dns`; apply and teardown purge every tagged rule for the DinD IP first, whatever resolver it names
- `docker_service.py` `_custom_build_image`: reads the DinD's `/etc/resolv.conf`; `timeout=` on every setup command; the readiness loop keeps N probes plus N seconds of sleep, each bounded by min(N, 10) s and run as `create_process` + `wait` inside `async with`, so a probe that outlives even that bound has its channel closed before the next one opens; both firewall helpers run under the executor's `timeout -k 5`, are named `<dind>-fw-apply` / `<dind>-fw-remove`, any non-zero exit of the apply helper fails the apply (`execute_and_stream_logs(check_exit_status=True)`, so a silent 137 from `timeout -k` never starts the build), and the teardown force-removes both by name before it removes the DinD (its IP)
- `core/config.py`: `CUSTOM_DOCKERFILE_SETUP_STEP_TIMEOUT_SECONDS` (default 180) and `CUSTOM_DOCKERFILE_DIND_READY_TIMEOUT_SECONDS` (default 60), both `gt=0`
- tests A20 to A26 in `tests/test_custom_dockerfile_build.py` (19 functions, 25 cases; 6 fail on main, A21 and A25 on the round-0 head; A21, A21b, A22, A23b, A23c, A25b need the round-2 head; A21c the round-3 head; A24 (new asserts), A24b and A26 the round-6 head; A24c, A24d, A24e the round-7 head; A23e the round-8 head) and 2 in `tests/test_custom_build_orphan_sweep.py` (round 8)

</details>


neurons/validators/src/core/config.py
```diff
@@ -572,6 +572,15 @@ def get_allowed_gpu_hwmodels(self) -> set[str]:
         env="CUSTOM_DOCKERFILE_DIND_READY_TIMEOUT_SECONDS", default=60,
         description="Max seconds to wait for the inner DinD dockerd to become ready.",
     )
+    CUSTOM_DOCKERFILE_SETUP_STEP_TIMEOUT_SECONDS: int = Field(
```
taiberium, line 580: Please require `gt=0`. Zero and negative values make AsyncSSH calls time out immediately, while `execute_and_stream_logs(timeout=0)` disables its timeout.

neurons/validators/src/services/docker_service.py
```diff
@@ -4157,7 +4209,20 @@ def _egress_filter_script(dind_ip: str, cidrs: list[str], apply: bool) -> str:
                     f"$IPT -C DOCKER-USER -s {dind_ip} -d {c} -j DROP 2>/dev/null || "
                     f"$IPT -I DOCKER-USER -s {dind_ip} -d {c} -j DROP"
                 )
+            for ns in dns_servers:
+                for proto in ("udp", "tcp"):
+                    rule = f"-s {dind_ip} -d {ns} -p {proto} --dport 53 -j ACCEPT"
+                    lines.append(
+                        f"$IPT -C DOCKER-USER {rule} 2>/dev/null || "
```
taiberium: `-C` can find a stale ACCEPT below a newly inserted DROP, so DNS still fails. Please remove matching ACCEPTs and insert one at the top.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/core/config.py
surcyf123: Done in `6b7c4f6`: `gt=0` on `CUSTOM_DOCKERFILE_SETUP_STEP_TIMEOUT_SECONDS`, same as `RENTAL_PROBE_SSH_DEADLINE_SECONDS`. A25 checks that 0 and -1 are refused at settings load and 1 is accepted.

surcyf123:
@taiberium Both inline points are in `6b7c4f6` (one commit): `gt=0` on the setup timeout, and the port-53 ACCEPT is deleted-then-inserted instead of `-C` checked. Full validators suite 2578 passed on EC2; A21 and A25 fail on the previous head. Re-requesting your review.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `6b7c4f6`: the ACCEPT is no longer check-then-insert. Every matching copy is deleted first in a `while $IPT -D DOCKER-USER <rule> 2>/dev/null; do :; done` loop, then one is inserted at the top with `-I`; the teardown deletes every copy the same way. The DROPs keep their check: a stale DROP anywhere sits below an ACCEPT inserted at position 1 after it. A21 asserts no check on an ACCEPT and the delete-before-insert order for udp and tcp; it fails on the previous head.

neurons/validators/src/services/docker_service.py
```diff
@@ -4157,7 +4212,19 @@ def _egress_filter_script(dind_ip: str, cidrs: list[str], apply: bool) -> str:
                     f"$IPT -C DOCKER-USER -s {dind_ip} -d {c} -j DROP 2>/dev/null || "
                     f"$IPT -I DOCKER-USER -s {dind_ip} -d {c} -j DROP"
                 )
+            for ns in dns_servers:
+                for proto in ("udp", "tcp"):
+                    rule = f"-s {dind_ip} -d {ns} -p {proto} --dport 53 -j ACCEPT"
+                    # Not `-C || -I`: a stale ACCEPT below the new DROP satisfies
+                    # `-C` and the insert is skipped. Delete every copy, then
+                    # insert one at the top of the chain.
+                    lines.append(f"while $IPT -D DOCKER-USER {rule} 2>/dev/null; do :; done")
```
taiberium: Please tag these rules and purge every DNS exception for `dind_ip` before adding current resolvers. Failed cleanup plus IP reuse otherwise leaves an old resolver allowed.

neurons/validators/src/services/docker_service.py
```diff
@@ -4260,17 +4363,24 @@ async def _custom_build_image(
...
+                        f"/usr/bin/docker exec {shlex.quote(dind_name)} docker info",
+                        check=False,
+                    )
+                    if probe.exit_status == 0:
+                        return True
+                    await asyncio.sleep(1)
+                return False
+
+            try:
+                ready = await asyncio.wait_for(_wait_dind_ready(), timeout=ready_timeout_s)
```
taiberium: Please add `gt=0` to `CUSTOM_DOCKERFILE_DIND_READY_TIMEOUT_SECONDS`; zero or negative values now cancel the only probe, while the previous loop ran one.

taiberium: nit: this caps readiness at 60 s wall time, while the old loop allowed 60 probes plus 60 s of sleep. A per-probe `timeout=` on `ssh_client.run` keeps the old budget and still catches a hung probe; non-blocking.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `7a7f23d`: every DNS ACCEPT now carries `-m comment --comment lium-dind-dns`. Apply and teardown read `iptables -S DOCKER-USER` and delete every rule with the DinD's source IP and the tag, as printed, before the inserts. The old resolver's rule goes whatever address it names, so IP reuse after a failed teardown no longer leaves it allowed. The listing's exit status is checked: on apply a failed listing exits 4 and the build aborts; on teardown it is tolerated so the DROP deletes still run. Test A21b runs the purge line against a fake chain listing and a failing listing.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `7a7f23d`: `CUSTOM_DOCKERFILE_DIND_READY_TIMEOUT_SECONDS` is `gt=0`, and the `max(1, ...)` is gone. Test A25b covers 0 and -1.

surcyf123: Done in `7a7f23d`: the outer `wait_for` is gone. The loop runs N probes a second apart again, and each probe carries `timeout=min(N, 10)`. A hung `docker info` ends the wait within 10 s. The worst case for the stage is N times 11 s, 11 minutes at the default 60. Tests A23b and A23c cover the hung probe and the probe count.

surcyf123:
@taiberium All three rows are done in 7a7f23d. Tagged DNS rules are purged per DinD IP on apply and teardown. The ready timeout is `gt=0`. A per-probe bound replaces the outer `wait_for`. Validator suite 2582 passed on EC2, gate PASS.

neurons/validators/src/services/docker_service.py
```diff
@@ -4138,13 +4193,44 @@ def _egress_filter_script(dind_ip: str, cidrs: list[str], apply: bool) -> str:
...
+        # possibly in place); on teardown it is tolerated so the DROPs below
+        # still go.
+        on_list_failure = (
+            '{ echo "DOCKER-USER listing failed" >&2; exit 4; }' if apply else 'rules=""'
+        )
+        purge_tagged = (
+            f"rules=$($IPT -S DOCKER-USER 2>/dev/null) || {on_list_failure}; "
+            f"printf '%s\\n' \"$rules\" | while read -r _a spec; do "
+            f'case "$spec" in *"-s {dind_ip}/32 "*"--comment {DIND_DNS_RULE_TAG} "*) '
+            f"$IPT -D $spec;; esac; done"
```
taiberium: If a tagged-rule deletion fails, abort the apply to prevent IP reuse from leaving an old resolver allowed.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `ad4778f`: Done in `ad4778f`. On apply, a tagged rule whose delete fails ends the purge loop and the script exits 5 with `tagged DNS rule delete failed` on stderr. The caller returns `build_egress_setup`, so the build never runs with an old resolver's ACCEPT in the chain. The purge now runs before the DROP inserts, so an abort leaves no half-applied DROP behind. The teardown still tolerates a failed delete and goes on to the DROPs. Test A21c runs the whole apply and teardown scripts in `sh -c` against a fake iptables whose delete fails. Apply exits 5 with no insert or check issued. Teardown exits 0 and tries both tagged deletes and all four DROPs. On the previous head the test fails: exit 0, both new ACCEPTs inserted. Validator suite on EC2 run 20260916T135517Z-co1p at this head: 2583 passed.

neurons/validators/src/core/config.py
```diff
@@ -569,8 +569,21 @@ def get_allowed_gpu_hwmodels(self) -> set[str]:
         description="--memory limit for the throwaway DinD build container.",
     )
     CUSTOM_DOCKERFILE_DIND_READY_TIMEOUT_SECONDS: int = Field(
-        env="CUSTOM_DOCKERFILE_DIND_READY_TIMEOUT_SECONDS", default=60,
-        description="Max seconds to wait for the inner DinD dockerd to become ready.",
+        env="CUSTOM_DOCKERFILE_DIND_READY_TIMEOUT_SECONDS", default=60, gt=0,
+        description=(
+            "Readiness budget for the inner DinD dockerd: this many one-second probes, and the "
```
taiberium: nit: one probe waits at most `min(this, DIND_READY_PROBE_MAX_SECONDS)` = 10 s, not this value, so please fix the description. Non-blocking.
```suggestion
            "Readiness budget for the inner DinD dockerd: this many one-second probes; each probe is "
            "bounded by min(this value, 10) s (a hung `docker info` ends the wait). Zero or negative "
```

neurons/validators/src/services/docker_service.py
```diff
@@ -4129,7 +4142,49 @@ def _parse_egress_block_cidrs(raw: str) -> list[str]:
...
+        the egress block, the
+        DROP rules eat every DNS query and `docker build --pull` fails on
+        `FROM` after the resolver timeout (12 to 15 s with one nameserver,
+        23 s with two, measured from `Building custom image` to the failure:
+        18 of 23 production `docker_build` failures in 1 to 15 Sep 2026 had
+        exactly that duration, a constant per node). Those servers get an
+        ACCEPT on port 53 only; a public resolver needs no rule and is left
+        out.
+        """
+        blocked = []
```
taiberium: nit: `cidrs` always comes from `_parse_egress_block_cidrs`, so this `except ValueError` never fires; a typed comprehension is enough, and the production numbers in the docstring belong in the PR. Non-blocking.
```suggestion
        blocked = [ipaddress.ip_network(c) for c in cidrs]
```

taiberium (approved):

neurons/validators/src/core/config.py
surcyf123: Done in `60cc3e8`: Your wording, as suggested.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `60cc3e8`: `blocked = [ipaddress.ip_network(c) for c in cidrs]`, and the docstring says why every entry parses (`_parse_egress_block_cidrs` returns `str(ip_network(...))`). The production numbers left the docstring; they are in the PR body's Problem and Value lines and in the timing table under Details.

neurons/validators/src/services/docker_service.py
```diff
@@ -4260,13 +4400,27 @@ async def _custom_build_image(
...
+                try:
+                    probe = await ssh_client.run(
+                        f"/usr/bin/docker exec {shlex.quote(dind_name)} docker info",
+                        check=False,
+                        timeout=probe_timeout_s,
+                    )
+                except asyncio.TimeoutError:
+                    # A probe that does not return in its bound is a dockerd
+                    # that is not answering: stop here.
+                    break
```
taiberium: A slow probe aborts all remaining retries and can reject a healthy DinD. Treat the timeout as not ready.
```suggestion
                    continue
```

taiberium (changes requested):
See the inline comments.

surcyf123:
@taiberium Pushed at a1c3abc: a slow readiness probe is one not-ready probe, the remaining probes run, and the bound is executor-side so no channel is left open.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `a1c3abc`: Applied, with one change to how the bound is held. The bound is now `timeout -k 2 N` on the executor, as for the DinD start. A slow probe returns exit 124. That is not ready, like an exit 1, and the loop goes on to the next probe. asyncssh's own timeout only stops waiting. With `continue` alone, a dockerd that never answers would leave one open session channel per slow probe, and sshd's MaxSessions is 10. It stays as the backstop at N + 5 s. Test A23d: the first probe hits its 3 s bound, the second answers ready, the build runs. On the previous head it failed at build_dind_unready. A23b and A23c assert the executor-side prefix. Validator suite on EC2 at this head: 2683 passed.

neurons/validators/src/services/docker_service.py
```diff
@@ -4313,7 +4524,8 @@ async def _custom_build_image(
                 command=egress_cmd,
                 log_tag=log_tag,
                 log_text="Applying build egress firewall",
-                log_extra={**default_extra, "dind_ip": dind_ip},
+                log_extra={**default_extra, "dind_ip": dind_ip, "dns_servers": dns_servers},
+                timeout=setup_timeout_s,
```
taiberium: nit: when this apply times out, the remote helper keeps inserting rules, but `egress_applied` stays False and the teardown skips them. Pass `dind_ip` to `_teardown_dind_build` unconditionally, since the remove script is idempotent. Non-blocking, fine as a follow-up.

taiberium (approved):

neurons/validators/src/services/docker_service.py
surcyf123: Done in `8530ffe`: `_teardown_dind_build` gets `dind_ip` whenever the IP is known; `egress_applied` is gone. A13 also checks the remove script runs on the egress-failed path.

neurons/validators/src/services/docker_service.py
```diff
@@ -4313,7 +4523,8 @@ async def _custom_build_image(
...
```
taiberium: Run both firewall helpers under executor-side `timeout -k 5`, name and force-remove each before releasing the DinD IP, and treat exit 124 as failure.

neurons/validators/src/services/docker_service.py
```diff
@@ -4138,26 +4187,81 @@ def _egress_filter_script(dind_ip: str, cidrs: list[str], apply: bool) -> str:
...
+        # is captured first and its exit status checked: a pipe would turn a
+        # failed `-S` (xtables lock) into a silent no-op. On apply that failure
+        # aborts (exit 4, the caller never runs the build with a stale ACCEPT
+        # possibly in place); on teardown it is tolerated so the DROPs below
+        # still go. A tagged rule whose `-D` fails aborts the apply the same
+        # way (exit 5): the loop runs in the pipe's subshell, so `exit 5` ends
+        # the loop and the pipeline's status carries it to the `|| { ...; }`
+        # after `done`; the caller's streamer fails the step on the stderr
+        # line. Without it the old resolver's ACCEPT stays in the chain and
+        # the new ACCEPTs are inserted next to it (taiberium, #1381). On apply
```
taiberium: nit: please drop the review references like "(taiberium, #1381)" here and at the readiness and teardown comments, the PR keeps that history. Non-blocking, fine as a follow-up.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `c3ca081`: The review references are gone from the purge, readiness and teardown comments in `docker_service.py`. The test docstrings in `test_custom_dockerfile_build.py` lost theirs too. The reasoning stays; the history is the PR's.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `c3ca081`: Both helpers now run under the executor's `timeout -k 5 <setup bound>`. The new wrapper prints one stderr line on exit 124, and the streamer fails a step on stderr, so exit 124 fails the apply at build_egress_setup. Before, the streamer's own timeout only stopped reading. The helpers are named after the DinD (suffix -fw-apply and -fw-remove). The teardown force-removes both names before it removes the DinD and releases its IP. The remove helper's exit 124 is logged as a failed teardown. The streamer and asyncssh keep a backstop at the bound + 15 s. Tests: A24 asserts the apply command and its tail. A24b runs the wrapper in sh: 124 with the stderr line, a quick command's own status, a failing command's own stderr. A26 (x2) asserts the remove helper's bound, the order (remove helper, then both helpers removed, then the DinD), and the warning on 124. Validator suite on EC2 at this head: 2587 passed.

neurons/validators/src/services/docker_service.py
```diff
@@ -4072,6 +4081,32 @@ def _custom_build_scratch_dir(pod_id: str) -> str:
...
+        """`command` under executor-side `timeout -k 5 <bound_s>`, exit 124 made loud.
+
+        `execute_and_stream_logs` fails a step on stderr, and `timeout(1)`
+        prints nothing when it kills the command, so a helper that hit its
+        bound would otherwise read as success. The wrapper echoes one stderr
+        line on exit 124 and exits with the command's status either way.
+        """
+        return (
+            f"timeout -k 5 {int(bound_s)} {command}; rc=$?; "
+            f"[ $rc -eq 124 ] && echo {shlex.quote(f'{what} timed out after {int(bound_s)}s')} >&2; "
```
taiberium, line 4136: Please fail the firewall step on every non-zero helper exit, including 137 from `timeout -k`. `execute_and_stream_logs` ignores the exit status, so a silent 137 starts the build without a confirmed egress firewall.

taiberium (changes requested):
See the inline comments.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `2c4ffe4`: The apply step now fails on the helper's exit status, not on a stderr line. execute_and_stream_logs has a new flag, `check_exit_status`. With it, the streamer waits for the channel to close (process.wait_closed), reads process.exit_status, and fails the step on any value other than 0. A missing status (a signal death) fails too. The error carries `Process exited with status <rc>`, and one error line goes to the pod log. Only the egress apply call passes the flag, so no other caller changes. The timeout -k 5 wrapper stays; its 124 line is now only for the log. Tests, in test_custom_dockerfile_build.py: A24c runs _custom_build_image through the real streamer with a fake process that exits 137 or 1 and prints nothing. Both end at `build_egress_setup` with no docker build. A24d does the same with exit 0 and the build runs. A24e covers the streamer alone: 137, 1 and None fail with the flag and raise with raise_exception; without the flag the old stderr-only verdict is unchanged. A24 asserts the flag on the apply call. A24b and A26 still pass. Validator suite on EC2 at this head: 2593 passed; the file: 40 passed.

neurons/validators/src/services/docker_service.py
```diff
@@ -4302,19 +4530,74 @@ async def _custom_build_image(
...
+            # the helper's exit status itself (`check_exit_status`), so a
+            # helper killed without a word (137 from `timeout -k`, an OOM
+            # kill, a `docker run` that never got to iptables) never starts
+            # the build; `_bounded_helper_command` adds the stderr line on 124
+            # so the pod log says why. The streamer's timeout stays as the
+            # backstop, past the executor-side bound.
+            apply_helper, _ = self._dind_firewall_helper_names(dind_name)
+            egress_cmd = self._bounded_helper_command(
                 f"/usr/bin/docker run --rm --network=host --cap-add=NET_ADMIN "
+                f"--name {shlex.quote(apply_helper)} "
```
taiberium, line 4595: `CustomBuildOrphanSweepCheck` treats this helper's suffix as part of the pod ID and can delete it during an active build. Strip the helper suffix before its active-pod check.

neurons/validators/src/services/docker_service.py
```diff
@@ -4260,13 +4455,36 @@ async def _custom_build_image(
...
+                    probe = await ssh_client.run(
+                        f"timeout -k 2 {probe_timeout_s} "
+                        f"/usr/bin/docker exec {shlex.quote(dind_name)} docker info",
+                        check=False,
+                        timeout=probe_timeout_s + 5,
+                    )
+                except asyncio.TimeoutError:
+                    # Backstop for a host where even `timeout(1)` did not
+                    # return: not ready yet, and the probe used its bound.
+                    continue
```
taiberium, line 4495: `run(timeout=...)` leaves its SSH channel open after a timeout, so retries can exhaust MaxSessions. Use `create_process()` and close each timed-out process before retrying.

neurons/validators/src/services/docker_service.py
```diff
@@ -4428,18 +4715,55 @@ async def _teardown_dind_build(
...
+        failure), then `docker rm -f` of both firewall helpers by name, then
+        the DinD itself. The helpers go first because removing the DinD
+        releases its IP: an apply or remove helper that outlived its bound is
+        still editing rules for that IP, and the next build to get the IP
+        would inherit them.
+
         Always called from `_custom_build_image`'s finally. Failures are logged,
         never raised — the rental flow must not break on cleanup.
         """
+        from core.config import settings
```
taiberium: nit: `settings` is already imported at module level (line 135), so this local import adds nothing. Drop the line. non-blocking, fine as a follow-up.

taiberium (changes requested):
**neurons/validators/src/services/docker_service.py:4327**

nit: `docker_service.py:4327` `_custom_build_image` is now ~380 lines; your numbered comments 2 to 4 already mark the seam for `_start_dind_and_wait` and `_apply_build_egress`. non-blocking, fine as a follow-up.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `766d624`: `dind_container_pod_id` strips the helper suffix before the active-pod check. An active build's helper survives the sweep, an orphan's helper still goes; both cases are tested and fail on the round-7 head.

neurons/validators/src/services/docker_service.py
surcyf123: Done in `766d624`: the probe runs as `create_process` + `wait` inside `async with`, as the streamer does. A timed-out probe's channel is closed before the next opens; the success path is unchanged (A23e).

neurons/validators/src/services/docker_service.py
surcyf123: Done in `766d624`: the local `settings` import is gone; the teardown reads the module-level one.

surcyf123:
@taiberium Not changing this in this round because the split moves about 250 lines while three review threads are open on them. The seam is as you say (`_start_dind_and_wait`, `_apply_build_egress`); the split is a follow-up after this lands, measured at 388 lines today against 252 on main.

neurons/validators/src/services/docker_service.py
```diff
@@ -4129,7 +4198,43 @@ def _parse_egress_block_cidrs(raw: str) -> list[str]:
         return cidrs
 
     @staticmethod
-    def _egress_filter_script(dind_ip: str, cidrs: list[str], apply: bool) -> str:
+    def _parse_dind_nameservers(resolv_conf: str, cidrs: list[str]) -> list[str]:
```
taiberium: NIT: The name reads as "parse all nameservers", but it keeps only the ones inside the block. Rename it to `_dind_nameservers_inside_blocked_cidrs`.

neurons/validators/src/services/docker_service.py
```diff
@@ -4260,13 +4455,45 @@ async def _custom_build_image(
                 )
                 return False, "build_dind_start"
 
-            # 3. Wait for the inner dockerd to accept connections.
+            # 3. Wait for the inner dockerd to accept connections: up to
```
taiberium: NIT: This 22-line comment and the 16-line purge comment at line 4268 retell the review history. Cut each one to one or two lines of why.

taiberium (approved):

neurons/validators/src/services/docker_service.py
surcyf123: Renamed in 14e9d79d: `docker_service.py:4201` is `_dind_nameservers_inside_blocked_cidrs`, with the call site and test A20 following. Nothing else named it.

neurons/validators/src/services/docker_service.py
surcyf123: Cut in e740639b: the purge comment (`docker_service.py:4268`) and the probe comment (`docker_service.py:4440`) are three lines of why each, −44 lines in all; the exit-code and MaxSessions history stays in rounds 3, 4 and 8 of the PR body. No statement changed; the 41 custom-build tests and the full validators suite (2598 passed) are unchanged.

surcyf123:
Final head is `e740639`. The 18 `Done in` replies above name `2c4ffe4`, `60cc3e8`, `6b7c4f6`, `766d624`, `7a7f23d`, `8530ffe`, `a1c3abc`, `ad4778f`, `c3ca081`, an earlier push of this round. Every fix named there is on `e740639`.

surcyf123:
Loop lane `cloud-0811` has finished on this PR (hold released 08:15Z).

taiberium:
🚀 temporary dev environment **dah-3521** is starting from `DAH-3521-dind-build-dns-egress` (4h, until 2026-09-23T12:28:27.297Z).

- namespaces `dev-dah-3521-backend`, `dev-dah-3521-frontend`, `dev-dah-3521-subnet`, `dev-dah-3521-miner-portal-backend`, `dev-dah-3521-miner-portal-frontend`, `dev-dah-3521-lium-stats`
- one GPU node, `g6.xlarge` (NVIDIA L4 24 GB) at $0.81/h on demand until the environment is destroyed; it joins the copy's central miner once the deploy is up.

A result comment follows when the deploy ends. Progress: https://github.com/Datura-ai/lium-infra/actions/workflows/temp-dev.yml

taiberium:
❌ temporary dev environment **dah-3521** failed to come up (lium-backend (failure), lium-frontend (not run), stats (not run), miner-portal-backend (not run), miner-portal-frontend (not run))

Requested by @taiberium. [Workflow run](https://github.com/Datura-ai/lium-infra/actions/runs/35837316924)

surcyf123:
Loop lane `cloud-0834` has finished on this PR (hold released 08:40Z).

surcyf123:
Loop lane `cloud-0843` has finished on this PR (hold released 08:48Z).

taiberium:
🚀 temporary dev environment **dah-3521** is starting from `DAH-3521-dind-build-dns-egress` (4h, until 2026-09-23T12:46:31.105Z).

- namespaces `dev-dah-3521-backend`, `dev-dah-3521-frontend`, `dev-dah-3521-subnet`, `dev-dah-3521-miner-portal-backend`, `dev-dah-3521-miner-portal-frontend`, `dev-dah-3521-lium-stats`
- one GPU node, `g6.xlarge` (NVIDIA L4 24 GB) at $0.81/h on demand until the environment is destroyed; it joins the copy's central miner once the deploy is up.

A result comment follows when the deploy ends. Progress: https://github.com/Datura-ai/lium-infra/actions/workflows/temp-dev.yml

taiberium:
✅ temporary dev environment **dah-3521** is up

- https://dah-3521.dev.lium.io (Lium app, its API under /api)
- https://dah-3521-provider.dev.lium.io (provider portal), API at `dah-3521-provider-api.dev.lium.io`
- its own Postgres, TimescaleDB and Redis; stats runs against them
- sign in as `dev-dah-3521@dev.lium.io` — Google login cannot work on a new address, so the deploy seeds a password account
- API token: `python .github/scripts/dev-env/sign_in.py --env dah-3521 --write <path>` in lium-platform (AWS SSO, LiumBackendDev) writes email and token to a new 0600 file; use it as `Authorization: Bearer <token>` on https://dah-3521.dev.lium.io/api
- validator, central miner and a local Bittensor chain in `dev-dah-3521-subnet`
- GPU node `g6.xlarge` (NVIDIA L4 24 GB) at 10.0.1.175, **$0.81/h** on demand until the environment is destroyed; listed, opted in, on the copy's provider portal and central miner; its validator verifies it next

Expires 2026-09-23T12:46:31.105Z. Comment `/temp-dev` again to redeploy and extend, `/temp-dev destroy` to remove it now.

Requested by @taiberium. [Workflow run](https://github.com/Datura-ai/lium-infra/actions/runs/35839041643)

surcyf123:
eng-drain-1200-io: restacking this branch on main (hold until 2026-09-23T13:45:07Z).

surcyf123:
Restacking on main 5754d83: conflicts in `neurons/validators/tests/test_custom_dockerfile_build.py` (imports), `neurons/validators/src/services/docker_service.py`; re-review from 766d624.

This is a rebase (all commits are surcyf123's), so the old head was e740639. Textual conflicts: in 7a7f23d the imports were kept from both sides. In 8530ffe, #1376's `CustomBuildOutcome` return was kept and `egress_applied = True` was dropped, as that commit intends.

There's one non-textual fix, in its own commit b09ca3b. #1376 changed `_custom_build_image` to return a three-field `CustomBuildOutcome`, so the three setup-timeout returns this PR adds (`build_dind_start`, and `build_egress_setup` twice) were still two-tuples and would crash callers with a ValueError. They now return `CustomBuildOutcome` using the sibling steps' renter-facing text, and this PR's tests unpack `ok, step, _tail` the way main's tests do. Validators: 2899 passed, 15 skipped, and 9 failures in `test_port_mapping_dao.py` that also fail on plain main (venv sqlmodel tz). Ran on cloud VM, not EC2.

surcyf123:
**lium-io#1381 read-clean at c4318e6: all 21 taiberium change-request threads are answered in code with loop replies, the restack onto main adapts to CustomBuildOutcome correctly, validators suite green on this head; 1 low.**

Fresh read at `c4318e67c2b74b0d7e6bdb026d041c08ecdc2594` by `subagent-eng-verdict-drain-b` (it wrote none of this PR). 0 high, 0 medium, 1 low.

```json
{
 "sha": "c4318e67c2b74b0d7e6bdb026d041c08ecdc2594",
 "verdict": "clean",
 "checks": 11,
 "change_requests": [
  {
   "reviewer": "taiberium",
   "threads": "4024353049 4024353064 4025299156 4025299164 4025299170 4026333807 4033264144 4033264158 4034039530 4034676873 4044382755 4044382767 4044974929 4045564955 4045564961 4045564967 4059669792 4059669797",
   "answered": "each Done in <sha> with a loop reply; approved at 766d624; restack b09ca3b6 + c4318e67 (drops a stray uv.lock) change no reviewed behaviour"
  }
 ],
 "findings": [
  {
   "class": "CORRECTNESS",
   "severity": "low",
   "file": "neurons/validators/src/services/docker_service.py",
   "line": 5052,
   "text": "_teardown_dind_build runs the remove helper before it force-removes the apply helper, so an apply helper that outlived its bound can insert a tagged DNS ACCEPT after the purge; the next apply on that DinD IP purges tagged rules first (exit 5 on failure), so no build runs with it, but the rule sits in DOCKER-USER until then.",
   "what": "teardown order leaves a window for a late apply helper",
   "fix": "docker rm -f both helpers before the remove script, or accept as-is"
  }
 ],
 "dismissed": [
  {
   "line": 4596,
   "why": "purge loop runs in a pipe subshell; `exit 5` ends the while and the trailing `|| { ...; exit 5; }` aborts the apply, as the reply says"
  },
  {
   "line": 4771,
   "why": "create_process + wait(timeout) inside async with closes the channel on timeout; asyncssh TimeoutError subclasses asyncio.TimeoutError"
  }
 ],
 "evidence": {
  "riskiest": {
   "line": 4596,
   "why": "the tagged-rule purge is shell text matching on `iptables -S` output; a mismatch leaves an old resolver ACCEPT",
   "checked_by": "read the pattern against iptables -S format (-A DOCKER-USER -s <ip>/32 -d <ns>/32 -p udp -m udp --dport 53 -m comment --comment lium-dind-dns -j ACCEPT); `read -r _a spec` drops -A so `$IPT -D $spec` is valid; tests A21/A21c pass"
  },
  "ran": [
   {
    "cmd": "cd neurons/validators && pdm install && pytest tests/ -q (CI env vars)",
    "result": "2908 passed, 15 skipped in 188 s"
   },
   {
    "cmd": "GitHub checks on c4318e6",
    "result": "tests-ok, Validators Tests, e2e gate PASS"
   }
  ],
  "edge_cases": [
   "CUSTOM_DOCKERFILE_*_TIMEOUT_SECONDS=0 or -1 → refused at settings load (gt=0)",
   "hung docker info probe → timeout(1) exit 124 = not ready, next probe runs",
   "helper killed silently (137) → check_exit_status fails the apply, build never runs open",
   "resolver outside blocked CIDRs → no ACCEPT added"
  ],
  "callers": [
   "neurons/validators/src/services/docker_service.py:4838 (execute_and_stream_logs check_exit_status=True, only caller)",
   "neurons/validators/src/services/task/checks/custom_build_orphan_sweep.py:168 (dind_container_pod_id)"
  ],
  "callers_none_why": "",
  "reviewer_page": "checked taiberium threads: all answered"
 },
 "high": 0,
 "medium": 0,
 "low": 1,
}
```

PR #1383, opened 2026-09-16
DAH-3508 - say the SSH key was not accepted instead of "Invalid executor id"

taiberium:
Implements the lium-io half of [DAH-3508](https://app.notion.com/p/3dcb8bfdbde98148836cc315ebb330b6).

**TL;DR** — "Invalid executor id" was the validator's text for two different things. The common one is that the miner returned no executor at all, because the node did not answer the SSH key upload. The id was correct every time. The text sent every reader after a wrong uuid.

**What changed**
- `_ssh_key_not_accepted_text` builds the message from the miner's answer: an empty list says "no executor accepted the SSH key", a mismatch says "the miner returned a different executor id".
- The count of executors the miner returned, and the id it returned, go into `extra`, so they reach Loki and `detail` and stay out of the renter-visible headline.
- The four call sites (WebSocket rent, WebSocket debug key, REST rent, REST debug key) use it.
- `error_code` stays `InvalidExecutorId`. The backend and the provider portal key on that code, not on the text.

**Depends on** the lium-platform PR that adds the new text to `EventsAggregationService.ERROR_PATTERNS`. Without it the analytics category for these events falls back to "Other pod creation error". Merge that one first.

**How to verify**
`pytest neurons/validators/tests/test_ssh_key_not_accepted_text.py` — three cases: empty list, different id, missing list.


surcyf123 (commented):
Would approve — nothing blocks. The loop does not approve, so this is a comment. 1 low thing to look at.

1. **[low]** `neurons/validators/tests/test_ssh_key_not_accepted_text.py:25` — `test_a_missing_executor_list_is_read_as_no_executor` passes `None`, which `AcceptSSHKeyRequest.executors: list[ExecutorSSHInfo]` (`datura/datura/requests/miner_requests.py:63`) cannot carry, so the test names no regression → drop it, or name the caller that can hand over `None`.

<details><summary>Notes</summary>

- minor (not counted): **[info]** `neurons/validators/src/services/miner_service.py:1149` — the four call sites keep `error_code=InvalidExecutorId`. The portal keys on the code (`apps/portal/backend/src/dtos/default_job_failure.py:18`), the backend on the text, so the body's claim holds.
- texts, end to end: headline → `_handle_container_error` → `msg` (headline) and `detail` (`to_full_string()`: headline + ` >>> ` + extra json) → lium-platform `validator_consumer.py:730` and `:796` categorise `msg.detail or msg.msg` → lium-platform#480's patterns `no executor accepted the ssh key` and `returned a different executor id` match as lowercase substrings. The miner filters the list to the requested id (`neurons/miners/src/consumers/validator_consumer.py:179`), so "a different executor id" is the right word for a non-empty mismatch.
- merge order: the "Depends on" holds. Without #480 both texts land in "Other pod creation error" (`events_aggregation.py:105`) and the give-up event quotes that. Nothing else reads the old text: lium-platform's other hits are a doc line and two comments (flagged on #480); `lium-io-backend` is archived.
- tests on EC2 (this head merged onto origin/main, run `20260916T072703Z-be1j`): `tests/test_ssh_key_not_accepted_text.py` 3 passed. CI `Validators Tests` and `e2e-gate` green at `0c06843`.
- ticket: DAH-3508 — Pod deletes/rents fail with "Invalid executor id" (Rustam)

</details>

_review by the loop at `0c06843`; the code owner's read still counts._

surcyf123 (approved):
Approved (loop review). Checked: the new text matches lium-platform#480's category patterns; error_code stays InvalidExecutorId so the portal keeps working on the code; the miner filters to the requested id, so 'a different executor id' is the right word for a non-empty mismatch; EC2 module 3 passed on the head merged onto main. One optional low: test_ssh_key_not_accepted_text.py:25 passes None where the type is list[ExecutorSSHInfo].

PR #1385, opened 2026-09-16
DAH-3540 - [P2] validator one-off: inspect stale executor hosts read-only through…

surcyf123:
Fixes DAH-3540 · reviewer: @taiberium

**Problem** — Rustam 16 Sep 08:23Z: "we need someone from our side to enter the node and check."

**Value** — 39 nodes in two /24 blocks earn 0 idle pay on the old runner image (`actual_score = 0`, "Required executor image is outdated", reports/EXECUTOR_FLEET_STALE_20260910.md §0). The provider's bot measured Docker Hub from a filler container and it answered..

**Goal** — DAH-3540 (P2).

**Fits** — Sits in `lium-io` — `neurons/validators`; touches backend (validator one-off), docs, tests (38).

**Flag:** no flag

**Risks**
- Running it = one verification cycle per node: one-shot key installed, one SSH session of ten reads (≤ 64 KiB per stream), key removed; the provider sees a validator login. Rollback: none, no host state changes.
- Merging changes one prod line: `_remove_ssh_key_via_rest` accepts a remove only on a 200 with an `SSHKeyRemoved` body; its callers ignore the return value, so prod gains one warning log per refused remove and nothing else.
- A remove refused 3× with the same key: the block prints the exact public key to delete from the executor's `authorized_keys`; exit 1.

**How to verify**
- `cd neurons/validators && python -m pytest tests/test_inspect_executor_host.py -q` → 38 passed (test names and what each refuses: Details)
- `echo "<executor uuid> <miner hotkey>" | pdm run python scripts/inspect_executor_host.py --dry-run` prints the plan for that node and opens nothing; `< nodes.txt` with the 43-node list under Details prints all 43

**Verified by the loop:** Gate: PASS e11115b

**Ran it:** `tools/aws_verify.sh lium-io e11115b86 --app io --bundle head.bundle` → EC2 20260917T090643Z-9u4z: validators suite 2666 passed, 15 skipped (+ 2 root-sensitive as… · … +1 under Details

**Self-review:** clean at e11115b (15 checks, 4 low)

**Parts:** backend ✓ (validator one-off) · migration n/a — no schema · frontend n/a — operator script · CLI/SDK n/a — validator-side · docs ✓ module docstring + env block here (no public page) · tests ✓ 38 · dashboards n/a — stdout

**Docs:** none needed because a renter or provider never runs this one-off; the module docstring carries usage, the connect path and where the shell lands, and the env list is `### Env the runner needs` below.

### What changed, in full

- `neurons/validators/scripts/inspect_executor_host.py`: per executor id, `SSHService.generate_ssh_key`, the signed `SSHPubKeySubmitRequest` through `MinerService._make_rest_request` (`_generate_auth_headers`, `_sign_validator_pubkey`, one executor id per request as the express lane does), the miner's `ExecutorSSHInfo` for address/port/user, then `connect_with_phase_timing(host, port, username, client_keys=[pkey], known_hosts=None)` — the call `DockerService.wait_for_port_check_containers` makes — and `_remove_ssh_key_via_rest` in a `finally` that runs after every submit, a refused or timed-out one included (the remove is idempotent; a submit that times out after the miner pushed the key must still be followed by one). The miner's 200 means it accepted the remove request, not that the executor confirmed the removal, and the block says so: `key remove request accepted by miner (not confirmed on executor): True|False`. A remove the miner did not accept sets the node's error (`key remove not accepted by miner after 3 attempts: the key is still installed, remove it by hand (public key in the node block)` — `may still be installed` after a submit the miner did not accept; `NodeReport.public_key_left` carries the key the block prints`) and `exit_code()` returns 1 for any node with an error or with `key_removed is not True`
- Ten reads per node (`COMMANDS`): `docker info --format '{{json .RegistryConfig}}'`, `/etc/docker/daemon.json` (file, `NO_DAEMON_JSON`, or `DAEMON_JSON_IS_DIR`; the executor container bind-mounts the host path, `neurons/executor/docker-compose.app.yml`), `docker images --digests` for the runner and executor images, `docker ps`, one read `ids=$(docker ps -q); docker inspect --format '{{.Name}} {{.Image}}' $ids; echo __inspect_images__; docker image inspect --format '{{.Id}} {{json .RepoDigests}}' $(docker inspect --format '{{.Image}}' $ids)` (every running container and the image id it RUNS, then the RepoDigests of those image ids, both from ONE container-id snapshot: with two snapshots a container that started in between showed only in the digest half and its image read `none running` — review, taiberium 17 Sep), `docker logs executor-watchtower-1 --tail 50` with a `name=watchtower` fallback, `docker logs executor-executor-runner-1 --tail 50` with a `name=executor-runner` fallback (the runner's entrypoint runs `docker compose up --pull always --detach --wait --force-recreate`, so a `--wait` that never came up healthy is in this log, not Watchtower's), `/proc/1/cgroup` + `/proc/self/cgroup`, `/.dockerenv` + `hostname`. Each block says whether the shell ran in the executor container or on the host
- Every read runs as `capped(command)`: `{ { <command>; printf '\n__inspect_rc=%s\n' "$?"; } 2>&1 1>&3 3>&- | head -c 65536 1>&2 3>&-; } 3>&1 | head -c 65536` — POSIX sh (dash included): stdout is parked on fd 3 while stderr goes through the first `head`, then comes back for the second, so the node sends at most 65536 bytes per stream and a command that keeps writing gets SIGPIPE; the exit status rides on the `__inspect_rc=` line, and a stdout that `head` cut has no whole marker, so its exit status is `None` (the table's `daemon_json` then reads `?`, not `yes`). The client (`capped_output`) clips each stream to 65536 characters again: a stream at the cap ends with `[truncated at 65536 bytes]`, a shorter stdout without a marker ends with `[no exit status: stdout cut at 65536 bytes or the shell stopped early]` (a multibyte stream cut by `head` is shorter in characters than in bytes, so it takes this note); `errors="replace"` keeps a split multibyte character from failing the read. The node block and the plan show the bare commands plus one wrapped example
- Summary table per node: mirror list, daemon.json state, `executor_running` (digest of the image the running executor container(s) run; `executor` and `monitor` share it, so `×2`), `executor_vs_hub` against `fetch_executor_image_digest()` — the registry digest of `settings.EXECUTOR_IMAGE_REF`, the value `score_calculator` compares (DAH-2701) — `runner_running`, `runner_pulled` (newest `docker images` row) and `runner_vs_hub` against Docker Hub's `daturaai/compute-subnet-executor-runner:latest` (what Watchtower pulls) so a Watchtower that pulled and failed to restart still reads `STALE`; `running_digests_by_repo` splits the one container read on its `__inspect_images__` line, joins the two halves and matches a container by its image's repository (as `machine_scrape` finds the executor), so no running container on the image reads `none` / `none running`, several containers on one image read `<digest> ×N`, an image that carries two RepoDigests of the repo (pulled by tag and by digest) reads `<digest>+<digest>` with either one counting as current, and containers on different images read `2 images in 2 containers: <name>=<digest>, …` with `MIXED` in the `vs_hub` column — never one picked; where the shell ran, Watchtower's last log line, error
- The node list comes on stdin by default (`< nodes.txt`, `echo "<uuid> <hotkey>" | …`; the repo ships no `nodes.txt`, and the earlier documented `--executor-ids nodes.txt` failed in a clean checkout — review, taiberium 17 Sep); `--executor-ids FILE|id,id` reads the file when it exists, a missing file name reads a piped stdin and says so on stderr, a terminal gets an error naming both forms (`read_ids_argument`, `looks_like_a_path`: a `/` or `.` that no uuid, hotkey or comma list has); a line is `uuid [miner_hotkey]`, `--miner-hotkey` fills bare ids; `--axon HOTKEY=IP:PORT` to skip the metagraph lookup, `--concurrency 4` (at least 1), `--timeout 60` per node with a quarter of it per read (a hung `docker logs` costs its slice; the reads already made are kept and the error names the budget), `--no-hub-digest`, `--dry-run` prints the plan and opens nothing. Each miner's axon is resolved once, before any node task starts (`resolve_axons_once`)
- Tests (`neurons/validators/tests/test_inspect_executor_host.py`, 38): the validator's own fakes — `_FakeSSHClient` from `test_failed_container_diagnostics` records the commands, `_connect_returning` from `test_ssh_connect_timing` stands under `connect_with_phase_timing` and captures the connect kwargs; `test_every_command_is_read_only` allow-lists the program and verb of every shell segment of the wrapped command; `test_capped_wrapper_cuts_each_stream_on_the_node_and_keeps_the_exit_status` runs the wrapper in `sh -c` (small output, 200 kB stdout, 200 kB stderr, a `docker info` style stderr warning next to JSON)

- 17 Sep review (taiberium): `run_commands` returns how many reads the node answered; a node whose login worked and whose every read raised (`ConnectionLost`, channel closed) gets `NO_READ_COMPLETED` as its error and the run exits 1 — before, `report.error` stayed None and the row was clean. `remove_key_with_retries` calls `_remove_ssh_key_via_rest` up to `KEY_REMOVE_ATTEMPTS` (3) times, `KEY_REMOVE_RETRY_DELAY` (2 s) apart, with the public key THIS run submitted — a rerun mints a new key and cannot remove the first one; when every attempt fails, `NodeReport.public_key_left` holds the key and the node block prints it under `PUBLIC KEY LEFT ON THE EXECUTOR` with the `authorized_keys` line to delete; the note no longer says re-run. Tests: `test_a_session_that_drops_after_login_is_a_failed_node_and_exit_1` (every read raises `ConnectionLost` → error, exit 1, key still taken back; one completed read → clean), `test_the_key_remove_is_retried_with_the_same_public_key_and_printed_when_it_never_succeeds` (two refusals then a 200 → three removes carrying the submit's key, clean; three refusals → key printed, no rerun advice, exit 1)

### How to verify, in full

- `cd neurons/validators && python -m pytest tests/test_inspect_executor_host.py -q` (38) — `test_every_command_is_read_only` (allow-list of `cat/echo/head/test/hostname/printf`, fd plumbing and `docker info/images/ps/logs/inspect` on the wrapped command; refuses `docker compose pull`, `docker update`, a redirect, wrapped or not); `test_connects_where_the_miner_said_and_takes_the_key_back` (host/port/user from the miner's answer, `known_hosts=None`, one client key, submit then remove); `test_capped_wrapper_cuts_each_stream_on_the_node_and_keeps_the_exit_status` (four `sh -c` cases); `test_a_stream_at_the_cap_is_cut_and_marked_on_the_client_too`; `test_executor_vs_hub_compares_the_executor_image_the_validator_scores_not_the_runner`; `test_summary_reports_zero_and_several_runner_containers_instead_of_picking_one`; `test_hub_digests_come_from_the_validators_own_executor_lookup`; `test_runner_vs_hub_follows_the_running_image_not_the_newest_pull`; `test_a_hung_read_keeps_the_earlier_reads_and_names_the_budget`; `test_axons_are_resolved_once_per_miner_before_the_nodes_run`; `test_dry_run_prints_the_plan_and_opens_nothing`; `test_summary_says_unknown_not_yes_for_a_read_that_failed`; `test_main_refuses_a_bad_axon_and_zero_concurrency`; `test_a_refused_key_submit_still_takes_the_key_back_and_connects_nowhere`, `test_a_submit_that_times_out_is_still_followed_by_a_remove` and `test_a_remove_the_miner_did_not_accept_is_a_node_error_and_exit_1` (the three key-remove rows of the 11:40Z review); `test_a_200_whose_body_is_not_ssh_key_removed_is_not_a_removal` (three bodies: `FailedRequest`, `{}`, `null`) and `test_runner_log_is_read_next_to_watchtowers` (the 12:53Z review)
- `pdm run python scripts/inspect_executor_host.py --dry-run < nodes.txt` on the validator prints the plan for 43 nodes and opens nothing (the node list is under *For Rustam*; save it as `nodes.txt` first, the repo does not ship it)

### Risks, in full

- Running it does what one verification cycle does to each of the 43 nodes: the miner writes the validator's one-shot public key into the executor's `authorized_keys`, one SSH session runs ten reads (at most 64 KiB per stream each), then the key-remove request takes the key back — the provider sees a validator login in `auth.log`; a block ending `key remove request accepted by miner (not confirmed on executor): False` means the miner did not take the remove request; the node's error column says so, the run exits 1, and on these pre-14-Sep executor images (before DAH-3394's key purge) the key stays until a remove succeeds, so the remove is retried three times with the same key and, if all fail, the block prints that public key for exact manual removal (a rerun mints a new key and cannot remove it). A `True` is the miner's 200 with an `SSHKeyRemoved` body (a 200 carrying `FailedRequest` is not accepted), not a check on the executor — rollback: none, nothing on the host's docker state, config or containers changes.
- Merging changes one prod line, `miner_service.py:1967`: `_remove_ssh_key_via_rest` returns True only on a 200 whose body is `SSHKeyRemoved` (the route answers 200 with a `FailedRequest` body when `deregister_pubkey` raised, so the status alone did not say the key was gone). Its five callers in the verification cycle (`_request_job_to_miner` and the rental paths) ignore the return value, so the deployed validator gains one warning log line per refused remove and nothing else: no migration, no pod restart, no config change, no data rewrite. The script runs only when the validator owner runs it.
- `--concurrency 4`, `--timeout 60` per node plus a 30 s key submit and up to 3 × 10 s (+ 2 × 2 s) key remove: 43 nodes finish in about 25 minutes worst case; a node that hangs costs its budget and an `ERROR` line, not the run.

### Worth it?

39 nodes in two /24 blocks earn 0 idle pay on the old runner image (`actual_score = 0`, "Required executor image is outdated", reports/EXECUTOR_FLEET_STALE_20260910.md §0). The provider's bot measured Docker Hub from a filler container and it answered; the host's docker configuration is the one thing nobody on our side has read. This is about 500 lines of script and about 400 of test for a read that decides the fix for 39 nodes; the one prod change is the `SSHKeyRemoved` body check in `_remove_ssh_key_via_rest` (a stricter return value its callers do not read).

### How the connect path is reused

1. `SSHService.generate_ssh_key(my_key.ss58_address)` — the one-shot key pair a verification cycle mints (`miner_service.py`, `_request_job_to_miner`).
2. `SSHPubKeySubmitRequest(public_key, validator_signature=MinerService._sign_validator_pubkey(...), executor_id=<uuid>, miner_hotkey=...)` posted with `MinerService._make_rest_request` and `MinerService._generate_auth_headers` to `<axon>/api/validator/ssh-pubkey-submit` — the REST route prod uses (`USE_REST_API: "1"` in lium-io-deployment `Pulumi.prod.yaml`). One executor id per request, as the express lane and the rental key-submit do; the miner installs the key on that executor only.
3. The miner's `AcceptSSHKeyRequest` carries the executor's `ExecutorSSHInfo` (`address`, `ssh_port`, `ssh_username`). The validator learns where to connect from the miner, not from the backend; this script does the same.
4. `services.ssh_connect_timing.connect_with_phase_timing(host=executor.address, port=executor.ssh_port, username=executor.ssh_username, client_keys=[pkey], known_hosts=None)` — the call at `docker_service.py` `wait_for_port_check_containers`.
5. `MinerService._remove_ssh_key_via_rest(...)` in a `finally` that runs after every submit, so a failed read, a refused submit or a submit that timed out still gets a remove. `run_commands` takes an `asyncssh.SSHClientConnection`, `capped_output` an `asyncssh.SSHCompletedProcess`.

`make_miner_service()` builds `MinerService(ssh_service=SSHService(), task_service=None, redis_service=None, attestation_service=None)`: the five helpers above read none of the three, and the script schedules no task and issues no attestation nonce. The only Redis use is inside `SubtensorClient` (the provider-portal miner snapshot it reads and refreshes), which the validator process does on every cycle anyway.

### Where the shell lands

The sshd the validator reaches runs inside `executor-executor-1` (`neurons/executor/docker-compose.app.yml`), which mounts the host's `/var/run/docker.sock` and `/etc/docker/daemon.json` and runs with `pid: host`. So `docker info` reports the HOST daemon's `RegistryConfig` (mirrors included) and `cat /etc/docker/daemon.json` reads the HOST file. `/proc/1/cgroup` is the host's init (pid namespace shared); `/proc/self/cgroup` and `/.dockerenv` tell the container apart. The block prints all three so the reader is not left guessing.

### Env the runner needs

Run it where the validator runs (its pod / container), so these are already set: `BITTENSOR_WALLET_NAME`, `BITTENSOR_WALLET_HOTKEY_NAME`, `BITTENSOR_WALLET_DIRECTORY` (the validator hotkey that miners accept), `BITTENSOR_NETWORK` / `BITTENSOR_NETUID`, `MINER_PORTAL_REST_API_URL` and the Redis settings (`SubtensorClient.get_miner` resolves the miner's axon through the provider-portal miner list and its Redis snapshot), plus the `Settings` fields the class requires at import (`SQLALCHEMY_DATABASE_URI`, `ASYNC_SQLALCHEMY_DATABASE_URI`, `VOLUME_MASTER_SECRET` — the validator's `.env` has them). Off the validator, pass `--axon HOTKEY=IP:PORT` for each of the four miners and the metagraph lookup (and Redis) is skipped; the wallet is still needed because the miner checks the signature.

### Inspection run — why the loop did not run it, and the path that works today

The bot's 17 Sep row asks for the output or for access. Measured 17 Sep 07:0xZ from the loop's own credentials: `kubectl auth can-i create pods/exec -n subnet` → `no` (the loop's EKS access entry edits only staging/dev namespaces; view cluster-wide), and the owner's guardrail hook denies `kubectl exec` on the prod context to the loop regardless (STANDING_ORDERS §30/§252; Pixel's ask). The script needs the validator's hotkey to sign the key submit (miners check the signature) and the validator's env, so it runs where the validator runs: `deployment/validator` in namespace `subnet` on `lium-prod-cluster`. The loop does not take that wallet (L-26). So the run is Rustam's, and the access he lacked is in place: [lium-infra#77](https://github.com/Datura-ai/lium-infra/pull/77) (DAH-3541, merged 16 Sep 20:03Z, *Deploy infrastructure* on main green at 22:43Z) gives `sso-LiumBackendDev-namespace-edit` edit on `subnet`. The deployed validator image today is `lium-prod-validator:c032085` (main before this PR).

Run it after the validator deploy that carries this PR — the image built from the merge commit has `/app/scripts/inspect_executor_host.py` (the validator Dockerfile is `COPY . .`) and the `SSHKeyRemoved` body check in `_remove_ssh_key_via_rest`, so no file is copied into the pod:

```
POD=$(kubectl -n subnet get pods -l app=validator -o jsonpath='{.items[0].metadata.name}')
kubectl -n subnet exec -i "$POD" -- pdm run python scripts/inspect_executor_host.py --dry-run < nodes.txt
kubectl -n subnet exec -i "$POD" -- pdm run python scripts/inspect_executor_host.py < nodes.txt | tee inspect-$(date -u +%Y%m%dT%H%MZ).txt
```

Why not copy the script (and `miner_service.py`) into the deployed image `c032085` instead: the script alone would import the image's `_remove_ssh_key_via_rest`, which returns True on any HTTP 200, so a `True` in that run would mean "the miner answered 200", not "the body said removed" (review, taiberium 17 Sep). Copying `miner_service.py` too would fix that read but leaves a validator pod running a module its image tag does not carry until the next restart — a hand-patched prod process nobody can identify from the deployment. Waiting for the deploy costs one release cycle and keeps the pod's code equal to its image. Redact node IPs before the output leaves the pod.

### For Rustam

```
cd neurons/validators && pdm run python scripts/inspect_executor_host.py --dry-run < nodes.txt
cd neurons/validators && pdm run python scripts/inspect_executor_host.py < nodes.txt
```

`nodes.txt` — the 43 executors (uuid, miner hotkey), from the prod `executor` table on 16 Sep 08:33Z:

```
2e2cfbdf-75c2-427c-93b4-f3758d983802 5CCwGES74ZGuyNey5mdLMZYCvGpMPi8Haz8744EyonfNWgZp
3921c29c-90b8-4521-8cfe-24fb5ebaa6f7 5CCwGES74ZGuyNey5mdLMZYCvGpMPi8Haz8744EyonfNWgZp
96c6d868-cc22-4c95-9b46-13c674cdf859 5CCwGES74ZGuyNey5mdLMZYCvGpMPi8Haz8744EyonfNWgZp
dc57f125-2d6d-4c45-980b-6c30f865c68c 5CCwGES74ZGuyNey5mdLMZYCvGpMPi8Haz8744EyonfNWgZp
751312aa-c9ea-49d0-9033-8e2d7b4586df 5CUQK1orC4gBwCwV36AtSvHX4ZDJzqF2ibeWmw69usiG277u
9ba61872-0129-4323-88a7-f3eba83bda3e 5CUQK1orC4gBwCwV36AtSvHX4ZDJzqF2ibeWmw69usiG277u
ac24dc4d-c1d9-4ccd-ae2e-ee7539cc66b7 5CUQK1orC4gBwCwV36AtSvHX4ZDJzqF2ibeWmw69usiG277u
01222847-94e0-49cc-8411-ef4ceebbd90b 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
0b84d64a-b0e0-444b-ac69-8ccfc76ecebc 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
0f24e0e0-e9cc-4e8e-bc0e-bd3bace45e5a 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
1890ae9b-23e0-4ca6-b979-870b6294b959 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
1cad0c8e-207a-4467-ae62-51f0ca204dfa 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
1d319933-1180-4663-8b8b-0c4860bbd199 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
22158230-8f05-4033-b9e5-d0bb580e0dc8 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
27b80abb-65a8-4af4-8cce-98b586d826f5 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
2fa71d5d-865a-4381-bce7-7eb1d46417fc 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
35735736-f2c4-4295-bc00-28240359794f 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
3ef4e528-b1ed-4d70-9e58-f2a9a1bd988c 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
47d56683-908e-4f4f-a15d-001e5f99f93c 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
4d6d7c6c-30c8-4b91-9fa9-92facf285fbc 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
585df07c-2eb3-4b6c-a98a-7a1f4a1c3173 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
5ea152f1-ccbe-4628-b545-c88a92a813de 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
60faf04f-8551-4f09-a836-c4c924ab7ec9 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
6a2f323a-164d-4401-a3c4-563d8b156e32 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
72aab06e-83a3-47c7-adb0-9b2a9ad6f7af 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
735eaf10-f00f-4c1e-a350-664838640ce6 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
840ca476-2fa9-4e57-ad62-5b54e17f2bd7 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
85efff35-1a41-4601-8ac4-0cedddbad00a 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
946807db-a8de-4403-8038-9dfb9670c5e6 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
948d7fd3-57b7-4a5b-9e94-0562420f25ef 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
9a83fc62-e3f7-474c-a2ca-b7c575cd1be8 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
9e8236f3-f3d3-408a-b2b6-b23bb2caa3d6 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
a3d35f57-09e1-4547-a3a1-6e85f023dc0a 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
a4c6dbb0-e07d-46d3-9c7d-ea95817d952e 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
a9062c77-67c0-4473-a64f-853658f9081a 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
ab094bb9-f672-4fec-aa81-fa9637701009 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
b79512a8-ad26-47cd-953b-7cc2079bf2be 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
c6bbbaa6-2aab-4b01-8153-35ae645e8102 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
c86f8985-9894-4aa5-92b9-b5ee3f39dce4 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
dcf67bfd-ea87-4c6c-9193-6ebc64b3e711 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
efef6056-37b2-45f6-84c9-e7c16efb3162 5DhzqmmkD99d8MFbEEFgtQYSUWyYHRLNJKpVjYQFn5d6hJsv
9d8f85f7-992e-45c3-8505-c64159418d68 5HGyf9J3u5iByuRYwNLTjrkwSqrLVEeixhFMZUavVuzihUyj
c90e99c6-6f59-41c6-b912-10fd6d4d09a4 5HGyf9J3u5iByuRYwNLTjrkwSqrLVEeixhFMZUavVuzihUyj
```

Output: one fenced block per node (each command, its stdout, stderr, exit code, `key remove request accepted by miner (not confirmed on executor): True|False`), then the table `node | mirror | daemon_json | executor_running | executor_vs_hub | runner_running | runner_pulled | runner_vs_hub | shell_ran_in | watchtower_last_line | error` under two header lines: the expected executor digest (`fetch_executor_image_digest`) and Docker Hub's runner digest. Exit code 1 when any node has an error line or its key remove was not accepted by the miner.

### Known follow-ups

- `known_hosts=None`, as in `wait_for_port_check_containers`. The create path pins the miner's `ssh_host_key` (`_build_known_hosts`); a follow-up can pass that here when the `ExecutorSSHInfo` carries the key.
- `SubtensorClient.initialize()` starts the validator's metagraph warm-up task, which keeps polling for the run's ~25 minutes; `--axon` avoids it. A follow-up can shut the client down after `resolve_axons_once`.
- Output is printed once every node has finished; an interrupted run prints nothing.
- The metagraph lookup needs the validator's Redis (provider-portal snapshot). `--axon` is the escape hatch; a backend lookup of executor → miner axon would remove it.
- The two expected digests are fetched once per run, anonymously, from where the script runs: `fetch_executor_image_digest()` (the validator's own `EXECUTOR_IMAGE_REF` lookup) and `fetch_registry_digest` for the runner tag.

### Ran it

EC2 warm box (`tools/aws_verify.sh lium-io e11115b86 --app io --bundle head.bundle`, c6i.2xlarge, run `20260917T090643Z-9u4z`, $0.054): `validators pytest tests/: rc=0 in 185 s — 2666 passed, 15 skipped, 2 deselected` + `2 root-sensitive test(s) as loopverify: 2 passed`; the same suite at the round's first cut `cbedb6e` (run `20260917T085609Z-79n6`, $0.019): 2666 passed, 15 skipped. `tests/test_inspect_executor_host.py` collects 38 (36 + the two 17 Sep regressions: `test_running_containers_come_from_one_docker_ps_snapshot` runs the real `running_containers` line under `sh -c` through `capped()` against a fake `docker` on PATH whose `ps -q` answers one more container per call, and asserts one `ps` call, the joined digests and the table row; `test_a_missing_ids_file_reads_a_piped_stdin_and_refuses_a_terminal`). Not run against the prod nodes — see *Inspection run* above for why and for the path that works today.

### Self-review

Verdict `clean` at `e11115b` · 15 checks · by subagent-r276-fresh-io · 2026-09-17 09:19Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | CORRECTNESS | `neurons/validators/scripts/inspect_executor_host.py:136` | Carried from cbedb6e, not addressed: with zero running containers `ids` is empty, `docker inspect --format …` and `docker image inspect --format …` exit 1 ("requires at least 1 argument"), `$?` is 1, `running_digests_by_repo` returns None and the table reads `?` for executor_running/runner_running instead of `none`. Unreachable while the shell lands in the running `executor-executor-1` container, so low. | Guard both halves on the snapshot: `[ -z "$ids" ] // docker inspect …` and `[ -z "$ids" ] // docker image inspect …` (`test`/`[` is on the allow-list), or state in the docstring that `?` is the zero-container reading. |
| low | CODE-QUALITY | `neurons/validators/scripts/inspect_executor_host.py:255` | Carried from cbedb6e, partly addressed (`-` now reads stdin): an extensionless missing file (`--executor-ids nodes`) has no `/`, `.` or `,`, so `looks_like_a_path` is False and it is still returned as an inline id, dying on `nodes: no miner hotkey on the line and no --miner-hotkey` — the message this round set out to remove for that spelling. | When a single comma-free word is neither uuid-shaped (36 chars, 4 dashes) nor ss58-shaped (46–48 base58 chars), append STDIN_HINT to the SystemExit that parse_targets raises, or fold that shape check into looks_like_a_path. |
| low | CODE-QUALITY | `neurons/validators/tests/test_inspect_executor_host.py:705` | New with the fold: after `monkeypatch.chdir(tmp_path)` the 'existing file' case writes `tmp_path / "nodes.txt"`, which is now `./nodes.txt` — the very name the missing-file assertions above rely on being absent. Today every `nodes.txt`-missing assertion runs before the write, so nothing fails; a later assertion appended below the write would read the file and pass or fail for the wrong reason. | Name the present file differently (`tmp_path / "present.txt"`) so the missing-file name stays missing for the whole test. |
| low | CODE-QUALITY | `neurons/validators/scripts/inspect_executor_host.py:71` | Carried from cbedb6e: ruff on the two files still reports I001 (unsorted import block, both files), UP035 (`typing.Awaitable/Callable` → `collections.abc`, script:71) and UP041 (`asyncio.TimeoutError` alias, tests:120) — 4 fixable, report-only in CI. The inline `__import__('os')` is fixed (`import os`). | `ruff check --fix` on the two files in a later pass, or leave as-is. |

Mechanical notes dismissed: `neurons/validators/tests/test_inspect_executor_host.py:182` Fold (a) does what it says. Traced `_segments` + `_is_read_only` on the HEAD source: `` ids=`docker rm x` `` splits into one segment `ids=`docker rm x`` (3 words), the backtick rule fires before the assignment rule → REFUSED raw and wrapped (`{ { ids=`docker rm x`` after brace stripping). `ids=$(docker rm x); docker inspect $ids`: `_lift` takes `docker rm x` as its own segment → `rm` not in READ_ONLY_DOCKER_VERBS → REFUSED raw and wrapped; `docker inspect $ids` alone is accepted, correctly. Also refused: `echo `docker rm x``, `x=`reboot``, `ids=$( (docker rm x) )` (nested parens are not lifted; the 5-word segment fails on words[0]). No regression: all ten COMMANDS, wrapped by `capped()`, still pass (no backtick anywhere in them or in the wrapper). `PATH=/tmp` alone is still accepted — an assignment writes nothing and the validator cannot place a binary on the node, so not a read-only hole. · `neurons/validators/tests/test_inspect_executor_host.py:689` Fold (b): `monkeypatch.chdir(tmp_path)` runs before the first `nodes.txt` lookup, so a `neurons/validators/nodes.txt` saved per the runbook no longer turns the missing-file assertions into a file read; the `existing = tmp_path / "nodes.txt"` case still resolves absolutely and passes. See the low at :705 for the ordering fragility this introduces. · `neurons/validators/tests/test_inspect_executor_host.py:12` Fold (c): `import os` at module top, `os.environ['PATH']` at :672; py_compile clean. · `neurons/validators/scripts/inspect_executor_host.py:266` Fold (d): `if value is None or value == "-": return stdin.read()` sits before `pathlib.Path(value)`, so a file literally named `-` in the cwd is never read and `is_file()` never runs for it; the help text says `default (or `-`): stdin`; the new assertion `read_ids_argument("-", stdin=StringIO("c d\n")) == "c d\n"` fails on cbedb6e (which returned "-"). The terminal case for `-` (`stdin.read()` on a tty blocks for keyboard input) is the same behaviour as the default with no flag, and is the universal `-` convention. · `review-wave/readability/overrides/lium-io-1385.json:16` STALE-BODY from cbedb6e is fixed: Parts (`tests ✓ 38`), the Details Tests bullet (`, 38)`), Fits and How to verify all say 38; 38 collected = 33 functions + 3 + 2 parametrized cases. The 'no recorded run for 38 passed' finding is resolved by the EC2 run at e11115b (2666 passed, 15 skipped, 20260917T090643Z-9u4z); the `ran_it` / `ran_it_details` fields (lines 15, 24) still cite 3690c2e / 36 passed and are expected to be re-rendered with that run before the push. · `neurons/validators/scripts/inspect_executor_host.py:137` Unchanged since cbedb6e and re-checked: the one-snapshot line is POSIX-sh under dash (verified with the test's fake docker in round 1), the `__inspect_images__` marker cannot equal a real output line, `$?` is the last `docker image inspect`, miner_service.py:1967 and its five bare-await callers, the miner route's 200+`FailedRequest`, the Dockerfile `COPY . .` runbook claim and the ten reads all still hold; the delta touches none of them. Stack: base main, merge clean, no overlap with main's new commits on the imported modules.

### Verified

Gate: PASS (e11115b) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1385)

### What changed

- `neurons/validators/scripts/inspect_executor_host.py`: `MinerService` REST key submit → miner's `ExecutorSSHInfo` → `connect_with_phase_timing(known_hosts=None)` → reads → key remove in a `finally` that runs after every submit, retried 3× with the SAME public key (`remove_key_with_retries`); when every attempt fails the node block prints that key for manual removal, the node is an error and the run exits 1. A session that drops after login (no read completed) is a node error too, exit 1
- Ten reads: registry config, `daemon.json` (file / missing / directory), pulled runner and executor images, `docker ps`, every running container's image id and its `RepoDigests` (one read, one `docker ps -q` snapshot feeding both inspects), Watchtower log, the runner's log (where a failed `docker compose up --wait` shows), cgroups; each read runs through `capped()`, a POSIX-sh wrapper that cuts stdout and stderr at 65536 bytes on the node (`head -c`) and carries the exit status on a marker line, and the client clips again and marks a cut stream
- Table per node: mirror, daemon.json, `executor_running` + `executor_vs_hub` (the running executor image against `fetch_executor_image_digest()`, the `EXECUTOR_IMAGE_REF` digest the validator scores), `runner_running` + `runner_pulled` + `runner_vs_hub` (against Docker Hub's runner `:latest`, what Watchtower pulls); containers are matched by image repository, so zero matches read `none` and several containers on different images are listed by name as `MIXED`
- Flags: node list on stdin by default (`< nodes.txt`, `echo "<uuid> <hotkey>" | …`; the repo ships no `nodes.txt`), `--executor-ids FILE|id,id` (a missing file falls back to a piped stdin, a terminal gets an error naming both forms), `--miner-hotkey`, `--axon HOTKEY=IP:PORT`, `--concurrency 4`, `--timeout 60`, `--no-hub-digest`, `--dry-run`; axons resolved once per miner
- Tests (38) on the validator's fakes: `_FakeSSHClient` records the commands, `_connect_returning` captures the connect kwargs; an allow-list test keeps every wrapped command read-only; four `sh -c` cases run the real wrapper; a `ConnectionLost` on every read, a remove refused 2×/3×, a fake `docker` whose `ps -q` grows on every call (one snapshot, run under `sh -c`) and a missing `--executor-ids` file are the 17 Sep regressions

</details>


neurons/validators/scripts/inspect_executor_host.py
```diff
@@ -0,0 +1,505 @@
...
+    client = await SubtensorClient.initialize()
+    neuron = await client.get_miner(hotkey)
+    return neuron.axon_info.ip, int(neuron.axon_info.port)
+
+
+async def run_commands(ssh_client, outputs: dict[str, CommandOutput], per_command_timeout: float) -> None:
+    """Fill ``outputs`` one command at a time, so a node budget that runs out keeps what was read."""
+    for label, command in COMMANDS:
+        try:
+            result = await asyncio.wait_for(ssh_client.run(command), timeout=per_command_timeout)
```
taiberium: `ssh_client.run()` buffers unbounded remote output, so a large daemon file or log line can exhaust the validator. Please cap stdout and stderr per command.

neurons/validators/scripts/inspect_executor_host.py
```diff
@@ -0,0 +1,505 @@
...
+        f"test -d /etc/docker/daemon.json && echo {DAEMON_JSON_IS_DIR}"
+        f" || cat /etc/docker/daemon.json 2>/dev/null || echo {NO_DAEMON_JSON}",
+    ),
+    ("pulled_runner_images", f"docker images --digests {RUNNER_IMAGE}"),
+    (
+        # the image the runner container RUNS, not the newest one pulled: a Watchtower that pulled
+        # and failed to restart leaves both on the host
+        "running_runner_digest",
+        "docker image inspect --format '{{index .RepoDigests 0}}'"
+        f" $(docker inspect --format '{{{{.Image}}}}' $(docker ps -qf name={RUNNER_CONTAINER_FILTER})) 2>&1",
```
taiberium: `name=executor-runner` can match multiple containers, and `summarise()` silently uses the last digest. Please report zero or multiple matches instead of choosing one.

neurons/validators/scripts/inspect_executor_host.py
```diff
@@ -0,0 +1,505 @@
...
+            cols = line.split()
+            if len(cols) >= 3 and cols[2].startswith("sha256:"):
+                pulled_digest = cols[2]
+                break
+
+    running_digest = "?"
+    running = out.get("running_runner_digest")
+    if running and running.exit_status == 0 and "@sha256:" in running.stdout:
+        running_digest = "sha256:" + running.stdout.strip().rsplit("@sha256:", 1)[1]
+    vs_hub = "?"
```
taiberium: `vs_hub` compares the runner, but the validator scores the executor image. Please compare the running executor digest with `fetch_executor_image_digest()`.

taiberium (changes requested):
See the inline comments.

surcyf123:
@taiberium this review says "see the inline comments" but none reached the PR (0 review comments 90 s after the review). If your agent left them as a pending review, submitting it will attach them; otherwise a repost would help — nothing to answer until then.

neurons/validators/scripts/inspect_executor_host.py
surcyf123: Done in `372c4f1`: the name filter is gone; two reads list every running container's image id and its `RepoDigests`, matched by repository. Zero matches read `none`, several containers on one image read `<digest> ×N`, different images are listed by name with `MIXED`.

neurons/validators/scripts/inspect_executor_host.py
surcyf123: Done in `372c4f1`: `executor_vs_hub` compares the running executor image with `fetch_executor_image_digest()` (`EXECUTOR_IMAGE_REF`); the runner comparison stays as `runner_vs_hub` for the Watchtower side.

neurons/validators/scripts/inspect_executor_host.py
surcyf123: Done in `372c4f1`: every read now runs through `capped()`. That POSIX-sh wrapper cuts stdout and stderr at 65536 bytes each on the node with `head -c`. The exit status rides on a marker line. The client clips again and marks a cut stream. Four `sh -c` cases test the wrapper.

neurons/validators/scripts/inspect_executor_host.py
```diff
@@ -0,0 +1,662 @@
...
+            continue
+        lines.append(f"$ {command}")
+        if output.stdout:
+            lines.append(output.stdout)
+        if output.stderr:
+            lines.append(f"[stderr] {output.stderr}")
+        if output.exit_status not in (0, None):
+            lines.append(f"[exit {output.exit_status}]")
+        lines.append("")
+    lines.append(f"key removed from miner: {report.key_removed}")
```
taiberium: nit: the miner answers 200 even when the executor removal failed, so `True` only means the miner accepted the request; please label it that way. Non-blocking.
```suggestion
    lines.append(f"key remove request accepted by miner (not confirmed on executor): {report.key_removed}")
```

neurons/validators/scripts/inspect_executor_host.py
```diff
@@ -0,0 +1,662 @@
...
+                    known_hosts=None,
+                ) as ssh_client:
+                    # a quarter of the node budget per read: one hung `docker logs` costs its slice, not the node
+                    await run_commands(ssh_client, report.outputs, per_command_timeout=max(5.0, node_timeout / 4))
+        except TimeoutError:  # the node budget only; the key submit's own 30 s timeout is reported below
+            report.error = f"node budget of {node_timeout:.0f} s ran out after {len(report.outputs)} of {len(COMMANDS)} reads"
+    except Exception as exc:
+        report.error = f"{type(exc).__name__}: {exc}"
+    finally:
+        if key_accepted:
```
taiberium: If the submit times out after the miner already pushed the key, `key_accepted` stays False and no remove is sent. Could you set the flag before awaiting the submit? Remove is idempotent. Non-blocking.

neurons/validators/scripts/inspect_executor_host.py
```diff
@@ -0,0 +1,662 @@
...
+        reports = await run(
+            targets,
+            concurrency=args.concurrency,
+            node_timeout=args.timeout,
+            resolve_axon=resolve_axon,
+            miner_service=make_miner_service(),
+            my_key=my_key,
+            hub=hub,
+        )
+        return 1 if any(r.error for r in reports) else 0
```
taiberium: Cleanup can return `False` while `report.error` stays empty, so this command exits 0 with a key still installed. Please treat `key_removed is not True` as failure.

neurons/validators/scripts/inspect_executor_host.py
```diff
@@ -0,0 +1,662 @@
...
+async def resolve_axon_from_metagraph(hotkey: str) -> tuple[str, int]:
+    """The validator's own lookup: ``SubtensorClient.get_miner`` (central-miner override included)."""
+    from clients.subtensor_client import SubtensorClient
+
+    client = await SubtensorClient.initialize()
+    neuron = await client.get_miner(hotkey)
+    return neuron.axon_info.ip, int(neuron.axon_info.port)
+
+
+async def run_commands(ssh_client, outputs: dict[str, CommandOutput], per_command_timeout: float) -> None:
```
taiberium: nit: Please add AsyncSSH types to `ssh_client` and `capped_output.result`. This is non-blocking and fine as a follow-up.

taiberium (changes requested):
See the inline comments.

neurons/validators/scripts/inspect_executor_host.py
surcyf123: Done in `fc0b175`: the block now reads `key remove request accepted by miner (not confirmed on executor): True|False`, your wording. The docstring and the body say the same: a 200 is the miner taking the request, not the executor confirming.

neurons/validators/scripts/inspect_executor_host.py
surcyf123: Done in `fc0b175`: the flag is gone and the `finally` sends the remove after every submit, timed out or refused included. New tests: a submit that raises `TimeoutError` is still followed by a remove; a 403 submit sends a remove and opens no SSH session.

neurons/validators/scripts/inspect_executor_host.py
surcyf123: Done in `fc0b175`: `run_commands(ssh_client: asyncssh.SSHClientConnection, ...)` and `capped_output(result: asyncssh.SSHCompletedProcess)`. EC2 20260916T121226Z-cn6a at this head: 30 passed.

neurons/validators/scripts/inspect_executor_host.py
surcyf123: Done in `fc0b175`: a remove the miner did not accept appends a note to the node's error. The note reads `key remove not accepted by miner: the key may still be installed, re-run this node`, so it shows in the table. `exit_code()` returns 1 for any node with an error or with `key_removed is not True`. Test: the remove answers 500, the error is set, exit 1. An axon-lookup failure also exits 1.

neurons/validators/scripts/inspect_executor_host.py
```diff
@@ -0,0 +1,679 @@
...
+                    await run_commands(ssh_client, report.outputs, per_command_timeout=max(5.0, node_timeout / 4))
+        except TimeoutError:  # the node budget only; the key submit's own 30 s timeout is reported below
+            report.error = f"node budget of {node_timeout:.0f} s ran out after {len(report.outputs)} of {len(COMMANDS)} reads"
+    except Exception as exc:
+        report.error = f"{type(exc).__name__}: {exc}"
+    finally:
+        # Every submitted key gets a remove, whatever the submit answered: a submit that timed out
+        # after the miner already pushed the key must not leave it there. The remove is idempotent,
+        # so a refused submit costs one harmless extra request.
+        report.key_removed = await miner_service._remove_ssh_key_via_rest(
```
taiberium: `_remove_ssh_key_via_rest` returns `True` for every HTTP 200, including a `FailedRequest` body. Please require `message_type == "SSHKeyRemoved"` and add the missing regression case.

neurons/validators/scripts/inspect_executor_host.py
```diff
@@ -0,0 +1,679 @@
...
+        "running_containers",
+        "docker inspect --format '{{.Name}} {{.Image}}' $(docker ps -q) 2>&1",
+    ),
+    (
+        # the repository digests of those image ids; `running_digests_by_repo` joins the two reads
+        "running_image_digests",
+        "docker image inspect --format '{{.Id}} {{json .RepoDigests}}'"
+        " $(docker inspect --format '{{.Image}}' $(docker ps -q)) 2>&1",
+    ),
+    (
```
taiberium, line 141: `COMMANDS` reads Watchtower logs, but `docker compose up --wait` failures appear in the runner log. Please collect that log too as a non-blocking follow-up.

taiberium (changes requested):
See the inline comments.

neurons/validators/scripts/inspect_executor_host.py
surcyf123: Done in `67747cf`: Done in `67747cf`. The helper now returns True only for a 200 whose body has `message_type` equal to `SSHKeyRemoved`. A 200 with a FailedRequest body, an empty body or a JSON null is logged with the body's details and returns False. In the script that sets the node's error and exits 1, as before. Regression test with three bodies: FailedRequest, an empty object, null. Each fails on the previous head. The five other callers discard the return value, so they only gain a warning log. Validator suite on EC2 run 20260916T135522Z-4oko at this head: 2662 passed.

neurons/validators/scripts/inspect_executor_host.py
surcyf123: Done in `67747cf`: Done in `67747cf`, in this round since it is one command. A new read, `runner_log`, takes the last 50 lines of `executor-executor-runner-1` with a fallback that finds the container by the service name. The runner's entrypoint runs the compose up with `--wait`, so a start that never came up healthy is in that log. The node block prints it under the Watchtower log. Test: the read is in COMMANDS right after the Watchtower one and passes the read-only allow-list.

taiberium:
Этот тикет не решает задачу для Rustam: он не может запустить скрипт и увидеть вывод из терминала. Сам проведи проверку на машинах и приложи результаты с выводом и выводом о причине либо дай Rustam доступ, чтобы он мог посмотреть сам. До этого PR не имеет практического результата.

taiberium (changes requested):
Request changes before merge:

1. If the SSH connection drops after login, every command can fail while the script exits with code 0. Mark the node as failed when no read completes, and add a test for this case.

2. A rerun creates a new SSH key and cannot remove the key left by the first run. Retry removal with the same public key. If all retries fail, print that public key for exact manual removal and remove the advice to rerun the node.

surcyf123:
@taiberium Done in `3690c2e`: Done in `3690c2e`, both points. (1) `run_commands` now returns how many reads the node answered. Zero completed reads sets the node error `no read completed: the SSH session dropped after login` and the run exits 1, whatever the key remove answered. One completed read is enough to judge the node. Test: `test_a_session_that_drops_after_login_is_a_failed_node_and_exit_1` (every read raises ConnectionLost). (2) The remove is retried three times, two seconds apart, with the public key this run submitted. If every attempt fails, the node block prints that exact key under PUBLIC KEY LEFT ON THE EXECUTOR with the authorized_keys line to delete. The advice to rerun is gone. After a submit the miner refused, the note says the key may still be installed, since it never confirmed it landed. Test: `test_the_key_remove_is_retried_with_the_same_public_key_and_printed_when_it_never_succeeds`. 36 tests pass on EC2 run 20260917T075428Z-8gx8.

surcyf123:
@taiberium You are right that the PR has no result until someone runs it. I could not run it from here. The loop's cluster access has no exec on the subnet namespace (kubectl auth can-i answers no) and the owner's guardrail blocks the loop from exec on prod anyway. The script needs the validator hotkey to sign the key submit, so it has to run where the validator runs. You have that access now: lium-infra#77 (DAH-3541, merged 16 Sep, deployed 22:43Z) gives LiumBackendDev edit on subnet. The body's section *Inspection run* has the five commands: find the validator pod, make `/app/scripts`, copy the script and `nodes.txt` in, dry-run, then the run with tee. Every name the script imports is in the deployed image, so no rebuild is needed. One caveat is written there too: the deployed key-remove helper returns True on any HTTP 200, so in that run True means the miner answered 200. If you prefer that I run it, the loop needs exec on subnet. Say so and I file that as a decision for the owner.

neurons/validators/scripts/inspect_executor_host.py
```diff
@@ -0,0 +1,764 @@
...
+    repo_digests_by_image: dict[str, list[str]] = {}
+    for line in images.stdout.splitlines():
+        image_id, _, rest = line.partition(" ")
+        if not image_id.startswith("sha256:"):
+            continue
+        try:
+            repo_digests_by_image[image_id] = [d for d in (json.loads(rest) or []) if isinstance(d, str)]
+        except (ValueError, TypeError):  # `null` for an image without RepoDigests, or a cut line
+            continue
+    if not set(image_by_container.values()) <= set(repo_digests_by_image):
```
taiberium, line 517: A container that starts between these reads appears only in `running_image_digests`, so this check can report `none running`. Please use one container-ID snapshot.

neurons/validators/scripts/inspect_executor_host.py
```diff
@@ -0,0 +1,764 @@
...
+image Watchtower updates. Running digests come from every running container's image (by the
+image's repository, as ``machine_scrape`` finds the executor), so zero matches read ``none`` and
+several containers on different digests are listed by name instead of one being picked.
+``watchtower_log`` and ``runner_log`` are the last 50 lines of each container: Watchtower says
+what it pulled and restarted; the runner (``executor-executor-runner-1``) says whether its
+``docker compose up --wait`` came up healthy, which is where a stale executor's failed start shows.
+
+Usage (from the validator checkout, with the validator's env loaded):
+
+    cd neurons/validators && pdm run python scripts/inspect_executor_host.py --executor-ids nodes.txt
```
taiberium: The documented command fails in a clean checkout because `nodes.txt` is absent. Please add the file or document stdin input as a non-blocking follow-up.

taiberium (changes requested):
**neurons/validators/src/services/miner_service.py:1967**

nit: the description says no prod code changes, but `miner_service.py:1967` changes, and copying only the script into `c032085` runs the old remove check that accepts any 200. Copy `miner_service.py` too or run after the validator deploy, and fix the description. Non-blocking.

neurons/validators/scripts/inspect_executor_host.py
surcyf123: Done in `e11115b`: the documented command is now `... < nodes.txt` (or `echo "<uuid> <hotkey>" | ...`). The repo ships no `nodes.txt`. `--executor-ids nodes.txt` still reads the file when it exists. When it does not, a piped stdin is read (stderr says so) and a terminal gets an error naming both forms. Test `test_a_missing_ids_file_reads_a_piped_stdin_and_refuses_a_terminal`. The body's How-to-verify and runbook use the stdin form.

neurons/validators/scripts/inspect_executor_host.py
surcyf123: Done in `e11115b`: one container-ID snapshot. The two reads are now one read. `ids=$(docker ps -q)` runs once; the container inspect and the image inspect both use those ids. The two halves are split on a marker line. Ten reads per node now. The new test runs the real line under `sh -c` against a fake docker whose `ps -q` grows on every call, and asserts one `ps` call and the joined digests.

surcyf123:
@taiberium Done in `e11115b`: Done in `e11115b` (body only for this one). The description now names the prod change: `miner_service.py:1967`. `_remove_ssh_key_via_rest` returns True only on a 200 with an `SSHKeyRemoved` body. Its five callers ignore the return value, so prod gains one warning log per refused remove. The runbook now says: run it after the validator deploy that carries this PR. The image built from the merge has the script and the body check, so nothing is copied into the pod. I chose that over copying `miner_service.py` too. A hand-copied module leaves the validator pod running code its image tag does not carry, until the next restart. Gate PASS at e11115b. Validators suite: 2666 passed on EC2 (run 20260917T090643Z-9u4z).

taiberium (changes requested):
Close this PR.

A person does this by hand: one login, ten read commands, half an hour on 43 nodes. We are not keeping a 1655-line script and 38 tests in the validator repo for a job that runs once. This is the sixth round on the same change — the answer has not moved.

Close DAH-3540 as well, and stop opening it again.

One thing is worth keeping: `miner_service.py:1967`, where a key remove counts as done only on an `SSHKeyRemoved` body. Send that line as its own small PR.

surcyf123:
@taiberium Agreed. This PR is closed and DAH-3540 is archived. The one line worth keeping, the `SSHKeyRemoved` check at miner_service.py:1967, will come as its own small PR under a new ticket.

PR #1394, opened 2026-09-17
DAH-3398 - [P2] stale-mount repair counts a hand-removed mountpoint dir as repaired

surcyf123:
Fixes DAH-3398 · reviewer: @taiberium

**Problem** — After a host reboot, the validator's stale-mount recovery (DAH-2306) starts the pod only once `rmdir` of the stale mountpoint dir exits 0. A provider who removed that dir by hand, as our ticket replies ask, leaves nothing to rmdir, so the pod never restarts. ticket-0313: 24 h of `POD_NOT_RUNNING` after the dir was gone.

**Value** — One known case in 14 d (ticket-0313, 7 to 10 Sep): about 67 h of a rented RTX 6000 Ada at $0.80/h (about $53 unearned) and two penalties ($31.87 + $12.67, reverted by hand). In the 7 d the report table holds, 4 encrypted-volume rentals hit `POD_NOT_RUNNING`, all back within 1.2 h. 34 source lines remove the one state that today ends only by a human reboot.

**Goal** — [DAH-3398](https://app.notion.com/p/Stale-mount-recovery-treats-a-hand-removed-mountpoint-dir-as-a-failed-repair-so-a-reboot-stranded-e-3d8b8bfdbde9818b9d58d96c0447c768) · P133 reliability.

**Fits** — Validator `services/docker_service.py`, `repair_stale_vloopback_mountpoint`: one extra helper command after a failed rmdir.

**Flag:** reads `RECOVER_ENCRYPTED_VOLUMES` (lium-io#1190; `"true"` in prod, lium-io-deployment); no new flag

**Risks**
- Merge + next validator image roll: a reboot-stranded pod on a hand-cleaned host is started by the validator within one cycle instead of by a human — provider and renter see the pod come back — rollback: revert, the repair is rmdir-only again.
- Encrypted pods need `RECOVER_ENCRYPTED_VOLUMES` (on in prod) and dockerd's `unless-stopped` retry; open lium-io#1265 (DAH-2780) removes that retry for them, so after it lands encrypted pods do not reach this recovery until DAH-2780 part 3 adds its own signal. Plain-volume pods are unaffected.
- No penalty, scoring or billing math changes.

**How to verify**
- `cd neurons/validators && pytest tests/test_docker_service.py -q` → 206 passed, 1 skipped
- Round-1 tests on main's `docker_service.py` → 3 tests / 4 cases fail; the dir-still-there control passes on both
- `docker run --rm --mount type=bind,src=<dir>,dst=/mnt alpine:3.19 test -e /mnt/<vol>` → 1 missing, 0 present, 125 no source dir or image (*Ran it*)

**Verified by the loop:** 17 Sep 2026 · EC2 run `20260917T104841Z-vu2p` · validators 2661 passed, 15 skipped · Gate: PASS (06656c5) on EC2 sandbox Linux x86_64

**Ran it:** helper commands on a docker host with ticket-0313's layout: missing → `rmdir` 1, `test -e` 1; present empty → `rmdir` 0; non-empty → `rmdir` 1, `test -e` 0… · EC2 `20260917T104833Z-85yj`

**Self-review:** clean at 06656c5 (15 checks, 5 low, 1 fixed in body)

**Parts:** backend ✓ (validator) · migration n/a — no schema · frontend n/a — validator internals · CLI/SDK n/a — not renter-facing · docs n/a — no public page · tests ✓ 3 (5 cases) · dashboards n/a — log event only

**E2E:** covered by unit — no vloopback plugin or host reboot in the e2e stack; `test_recover_pod_starts_when_the_provider_already_removed_the_stale_dir` drives the recovery

**Docs:** none changed in this PR — the repair has no public page.

### Why `--mount type=bind` and not `-v`

`docker run -v <missing-dir>:/mnt` creates the missing host dir. With `_docker_root_dir_or_default` falling back to `/var/lib/docker` on a host whose data-root is elsewhere, `-v` would create an empty `propagated-mount` under the wrong root, rmdir would say "No such file", `test -e` would say gone, and the repair would report ALREADY_ABSENT while the real stale dir is still there (round-1 fresh-review finding). `--mount type=bind` refuses a missing source with exit 125, so both helpers fail the same way they do for a missing image and the repair stays skipped. The rmdir command carried `-v` since the repair was written; the stray dir it could create was harmless then because rmdir failed either way.

### Why `test -e` after `rmdir`, and not before

The happy path (DAH-2306: the stale dir is present and empty) stays one helper run, and the extra command only runs when rmdir has already failed. Checking first would add a `docker run` to every successful repair for the rare case. Checking from the same helper container matters: the propagated-mount dir is root-only under dockerd's data-root, so a host-side `test -e` as the validator's ssh user can answer "absent" for a dir it merely cannot see; the helper bind-mounts the dir as root, the way rmdir does.

Only exit 1 counts as absent. `test -e` answers 0 (there) or 1 (not there); docker's own failures are 125 to 127, and those stay a skipped repair with the rmdir's stderr in the log, so a host where the helper image cannot run is not mistaken for a cleaned one (Ran it, last block).

### What this does not change

- `_repair_stale_mountpoint_of_container_volume` still returns the first volume the repair accepts. A rental container carries one vloopback volume (the external volume is s3fs and `_is_vloopback_driver` refuses it), so "first accepted" is the only one; its comment now says the dir may be gone.
- The create-path retry (`_should_repair_stale_mountpoint`) calls the same function; a create that failed with "file exists" on a dir that is gone by the time we look now retries once instead of surfacing the error. Same outcome as a successful rmdir.
- The recovery gate (`_is_stale_vloopback_mountpoint_error` on the container's `State.Error`) is untouched, so a pod stopped on purpose still never matches.
- DAH-3217 / lium-io#1326 (real docker data-root) is the other half of ticket-0313 and is already on main; this PR sits on top of it. It is not in the prod validator image yet (`dd27dc3` at the time of the ticket), so both land together at the next roll.
- `RECOVER_ENCRYPTED_VOLUMES` (lium-io#1190) gates the encrypted-pod branch of the recovery before the repair is reached; lium-io-deployment sets it `"true"` for the prod validator, so ticket-0313's pod does reach this code. Open lium-io#1265 (DAH-2780 part 2, arhangel66) touches the same two files and removes `unless-stopped` for encrypted pods, so after it lands dockerd no longer produces the `State.Error` this recovery's gate matches for encrypted pods; #1265's body says the replacement recovery signal is not in it, so until DAH-2780's part 3 exists, encrypted pods do not reach this code after #1265. Plain-volume pods do. Whichever merges second rebases; the hunks do not overlap.

### Tests

`test_repair_stale_vloopback_mountpoint_counts_a_hand_removed_target_as_repaired`: ticket-0313's host (data-root `/mnt/lium-xfs/lium-docker`), rmdir answers `No such file or directory`, `test -e` exit 1 → True, `VLOOPBACK_STALE_MOUNTPOINT_ALREADY_ABSENT` carries `target` and `executor_id`, no `REPAIRED` and no `REPAIR_SKIPPED` event, all six commands with the 30 s timeout.

`test_repair_stale_vloopback_mountpoint_refuses_non_empty_target[target_still_present|propagated_mount_dir_missing_helper_did_not_run]`: rmdir fails with `not empty` and `test -e` answers 0, or both helpers answer 125 (`bind source path does not exist`) → False, `REPAIR_SKIPPED` with the target path, the rmdir's exit status and stderr, no `ALREADY_ABSENT`. The pre-existing test grew the sixth command and the parametrisation; its old assertions stand.

`test_recover_pod_starts_when_the_provider_already_removed_the_stale_dir[mountpoint_dir_gone_starts_the_container|mountpoint_dir_still_there_stays_failed]`: `recover_pod_after_stale_vloopback_mount` with the container's real `State.Error` and an ssh stub that answers each host command the way ticket-0313's host does; the dir-gone case reaches `start_existing_container` with `local_volume=volume_pod-1` and logs `POD_STALE_MOUNT_RECOVERED`, the dir-present case stops at `POD_STALE_MOUNT_RECOVERY_REPAIR_FAILED`. The second case is the negative control and passes on main too.

Fail-old: the round-1 tests (same cases, `-v` still in the existing assertions) on main's `docker_service.py`: 3 tests / 4 cases fail (the repair returns False after five commands, so the sixth command and the ALREADY_ABSENT event never happen); EC2 run `20260917T102125Z-fegp`. On verbatim main the final tests would also fail the three `--mount` assertions in the pre-existing tests.

### Measurement (read-only prod replica)

`executor_validation_report` exists since 10 Sep 10:12Z (7-day retention), so the 14-day window collapses to 7 days. Rentals with `POD_NOT_RUNNING` reports while open: 15; with an encrypted volume (`pod.enable_volume_encryption`): 4, all recovered, longest 6 consecutive cycles (1.2 h, RTX 6000 Ada, 11 Sep 08:57Z), one of them the tail of ticket-0313 itself (3 cycles until the hand reboot at 10:37Z). ticket-0313's 7 to 10 Sep outage predates the table. Penalties 14 d: 59 `EXECUTOR_INACTIVE_MID_RENTAL` applied ($3,130), 14 reverted ($504). Sentry logs dataset: no `*STALE_MOUNT*` or `*VLOOPBACK_STALE_MOUNTPOINT*` lines in 14 d (validator warnings do not reach Sentry). Query and rows are in the loop's lane folder for DAH-3398.

### Known follow-ups

- A `POD_STALE_MOUNT_RECOVERY_REPAIR_FAILED` streak alert (the ticket's third suggestion): not in this PR, it is a monitoring change with its own owner.
- Ticket-reply text: once this and lium-io#1326 are in the prod image, the advice can be "do nothing, the validator repairs it within 15 min".
- Round-2 fresh-review lows left as they are: a host whose docker root lookup fell back to `/var/lib/docker` while a copied old data-root still holds `plugins/<id>/propagated-mount` can read ALREADY_ABSENT and fail at `docker start` (`POD_STALE_MOUNT_RECOVERY_START_FAILED`) instead of at the repair; same outcome (pod stays down, one error line per cycle). `_vloopback_repair_helper_cmd(..., argv)` holds the command tail as a string.

### Ran it

EC2 sandbox (AL2023 + docker, c6i.large, run `20260917T104833Z-85yj`, $0.0023). `ROOT` stands in for the host's `<docker-root>/plugins/<id>/propagated-mount`; the commands are the ones `repair_stale_vloopback_mountpoint` issues, with the same image:

```
== hand-removed dir (ticket-0313 after 9 Sep 05:28 UTC)
$ docker run --rm --mount type=bind,src=$ROOT,dst=/mnt docker.io/library/alpine:3.19 rmdir /mnt/volume_gone
rmdir: '/mnt/volume_gone': No such file or directory
exit=1
$ docker run --rm --mount type=bind,src=$ROOT,dst=/mnt docker.io/library/alpine:3.19 test -e /mnt/volume_gone
exit=1
== stale but present, empty dir (the DAH-2306 case: rmdir repairs it)
$ docker run --rm --mount type=bind,src=$ROOT,dst=/mnt docker.io/library/alpine:3.19 rmdir /mnt/volume_present
exit=0
== present, not empty
$ docker run --rm --mount type=bind,src=$ROOT,dst=/mnt docker.io/library/alpine:3.19 rmdir /mnt/volume_nonempty
rmdir: '/mnt/volume_nonempty': Directory not empty
exit=1
$ docker run --rm --mount type=bind,src=$ROOT,dst=/mnt docker.io/library/alpine:3.19 test -e /mnt/volume_nonempty
exit=0
== propagated-mount dir itself missing (a guessed docker root)
$ docker run --rm --mount type=bind,src=/job/work/no-such-root/plugins/plugin123/propagated-mount,dst=/mnt docker.io/library/alpine:3.19 rmdir /mnt/volume_gone
docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: ...
exit=125
$ ... test -e /mnt/volume_gone → exit=125
$ test -e /job/work/no-such-root → exit=1   (nothing was created)
== helper image missing
$ docker run --rm --mount ... docker.io/library/alpine:no-such-tag test -e /mnt/volume_gone → exit=125
```

Suites, EC2 warm box (c6i.2xlarge): `tools/aws_verify.sh lium-io 06656c5 --bundle head.bundle` run `20260917T104841Z-vu2p` ($0.019) → validators `2661 passed, 15 skipped` + 2 root-sensitive passed. Fail-old: this branch's tests on main's `docker_service.py` (a local probe commit, run `20260917T102125Z-fegp`) → `4 failed, 202 passed, 1 skipped` in `test_docker_service.py` (round-1 tests): `counts_a_hand_removed_target_as_repaired`, both `refuses_non_empty_target` cases, `mountpoint_dir_gone_starts_the_container`.

### Self-review

Verdict `clean` at `06656c5` · 15 checks · by subagent-fresh-3398-r2 · 2026-09-17 11:04Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | CORRECTNESS | `neurons/validators/src/services/docker_service.py:1517` | Property: ALREADY_ABSENT must only be reported for the tree the plugin actually uses. When `_docker_root_dir_or_default` has fallen back to /var/lib/docker (docker info failed) and the provider's old, copied data-root still holds `plugins/<id>/propagated-mount` without the volume dir (the usual leftover of a data-root move), `--mount type=bind` succeeds, rmdir says 'No such file', `test -e` answers 1 and the repair returns True; recovery then runs `start_existing_container`, which fails on the same 'file exists' and logs POD_STALE_MOUNT_RECOVERY_START_FAILED at error level (create path: one retry, same exception raised). Outcome for the pod is unchanged, but the 'What changed' bullet 3 ('a dir missing under a guessed docker root cannot read as volume dir gone') overstates what the `--mount` change guarantees: only a missing propagated-mount dir is refused. | Either have `_docker_root_dir_or_default` surface the fallback (return `(root, fell_back)` or `None`) and skip the ALREADY_ABSENT verdict when the root was guessed, or narrow the bullet to 'a propagated-mount dir missing under a guessed docker root' and add the copied-old-root case to Known follow-ups. |
| low | STALE-BODY | `PR body (Details 'Why --mount type=bind') — brief section 4 line 96:96` | Property: every sha in prose must be reachable from the pushed branch. `5d0e16a` is contained only by `remotes/origin/DAH-3398` as it stands today; the push of 06656c5 (whose parent is origin/main) replaces it, so the BODY-FRESHNESS check will see an unreachable sha. | Write it as 'the previous pushed head' or `<eight hex characters>`, or drop the parenthetical. |
| low | PROSE-VS-CODE | `PR body (How to verify bullet 2; Ran it 'Fail-old' sentence) — brief section 4 lines 79 and 161:79` | Property: a fail-old sentence must be reproducible as written. 'Same tests on main's docker_service.py → 3 tests / 4 cases fail' is not what verbatim main gives: `uses_rmdir_helper`, `uses_the_hosts_docker_root` and `falls_back_to_the_default_root` assert the `--mount type=bind,...` spelling, which main's `-v` helper does not emit, so a reviewer checking out main's file sees 6 tests / 7 cases red; the counts also disagree (4 failed + 202 passed = 206 on the probe vs 207 passed + 1 skipped = 208 on the head). | Say what the probe commit was (main's repair logic with the `--mount` spelling kept, or whichever it was) and reconcile the two totals, or report the verbatim-main numbers. |
| low | PROSE-VS-CODE | `PR body (Details 'Why --mount type=bind', last sentence) — brief section 4 line 96:96` | 'The rmdir command carried `-v` since DAH-2306' misattributes it: `git log -S'rmdir /mnt/'` shows the `-v <parent>:/mnt ... rmdir /mnt/<volume>` helper arrived in fd9c02f (29 May 2026, 'fix: repair vloopback propagated mountpoints', the DAH-2133 create-path repair), before the DAH-2306 recovery reused it. | Write 'since the DAH-2133 create-path repair'. |
| low | CODE-QUALITY | `neurons/validators/src/services/docker_service.py:810` | `_vloopback_repair_helper_cmd(propagated_mount_dir: str, argv: str)`: the second parameter holds an already-quoted command tail (`rmdir /mnt/volume_x`), not an argv list; the name says the wrong type. | Rename to `command_tail` (or `helper_command`) — optional, non-blocking. |

Fixed in the body: 1 body-only finding(s) answered by re-rendering the body, no code change, no new verdict (PR_PROCESS §7). `PR body (Risks bullet 2; Details 'What this does not change' bullet 5) — brief section 4 lines 74 and 110:74` PROSE-VS-CODE (medium) — Property: every behaviour the body attributes to another PR must exist on main, on this branch, or in that PR's diff. 'DAH-2780's restart path then calls the same repair' and 'DAH-2780's own restart path takes over and… → Risks bullet 2 and the Details bullet now say the opposite: after lium-io#1265 encrypted pods produce no State.Error, so nothing reaches this recovery for them until DAH-2780 part 3 adds its own signal (per #1265's body); plain-volume pods are unaffected. Re-…

Mechanical notes dismissed: `neurons/validators/src/services/docker_service.py:817` `shlex.quote` on the whole `type=bind,src=<dir>,dst=/mnt` token is correct shell + docker syntax for every value `_docker_root_dir_or_default` returns: an absolute path from `docker info` with the trailing slash stripped, or /var/lib/docker, contain only shlex-safe characters so no quotes are emitted (matches the exact-string test assertions); a root with a space is single-quoted into one argument that docker's CSV mount parser accepts; a root with `,` or `"` makes docker exit 125 and the repair stays skipped (fail-closed), the same class the old `-v` had for `:`. · `neurons/validators/src/services/docker_service.py:1084` (question d) The True-on-absent return is safe on the create path: `vloopback_mount_repair_attempted` is set before the repair, so the retry runs once and a second identical 'file exists' raises; VLOOPBACK_STALE_MOUNTPOINT_RETRY still logs; the SDK twin `_run_rental_docker_create_with_port_retry` (line 1149) has the same guard. That a failed `docker run` may leave a Created-state container whose name collides on the retry is DAH-2133 behaviour, identical for the successful-rmdir case, and not changed here. · `neurons/validators/src/services/docker_service.py:1517` Only exit 1 counts as absent: busybox `test -e` answers 0/1 (2 on usage error), docker's own failures are 125-127, a signal-killed process gives None; every non-1 value keeps REPAIR_SKIPPED with the rmdir's status and stderr. A findmnt→rmdir→test race (a mount or recreate in between) yields 0 → skipped, which is the conservative side. · `neurons/validators/tests/test_docker_service.py:4426` `refuses_non_empty_target` lowers caplog to INFO on purpose: the new negative assertion on VLOOPBACK_STALE_MOUNTPOINT_ALREADY_ABSENT (an info event) would be vacuous at WARNING. Both params exercise the branch they name (exit 0 → still there; 125/125 → helper did not run) and assert the sixth command, the rmdir's exit status and stderr. · `neurons/validators/tests/test_docker_service.py:6475` `_hand_cleaned_host_ssh_client` answers every command the real recovery issues in order (.Destination inspect → PLAIN, volume-name inspect, volume inspect with driver `vloopback:latest`, plugin inspect, `docker info` with the /mnt/lium-xfs root, findmnt 1, rmdir 1, `test -e` param) and raises on anything else; the dir-gone case fails on main (rmdir failure → False → REPAIR_FAILED), the dir-present case is a genuine control that passes on both, as the body says. · `PR body (Flag line):70` `RECOVER_ENCRYPTED_VOLUMES: "true"` confirmed at ~/dev/lium-io-deployment/infrastructure/Pulumi.prod.yaml:65; `_can_remount_encrypted_volume` (docker_service.py:868) reads it before the repair is reached, as the body states. · `PR body (Details bullet 4, `dd27dc3`):109` `dd27dc3` is on origin/main (DAH-3019, #1356), so it stays reachable after the push. · `PR body (Details bullet 5, 'the hunks do not overlap'):110` Checked against `gh pr diff 1265`: #1265 inserts `_assert_no_encryption_downgrade` immediately above `_repair_stale_mountpoint_of_container_volume` and changes `_build_rental_container_run_spec` / `create_container`; this PR's hunks are at lines 810, 1501-1535 and a comment ten lines into `_repair_stale_mountpoint_of_container_volume`; no shared lines, both orders merge clean.

### What changed

- `docker_service.py` `repair_stale_vloopback_mountpoint`: when `rmdir` fails, the same alpine helper runs `test -e` on the target; exit 1 (gone) is `VLOOPBACK_STALE_MOUNTPOINT_ALREADY_ABSENT` and returns True
- Any other answer (0 = still there, 125+ = helper did not run) keeps `VLOOPBACK_STALE_MOUNTPOINT_REPAIR_SKIPPED` and returns False, as before
- `_vloopback_repair_helper_cmd()` builds both helper lines with `--mount type=bind,src=<propagated-mount>,dst=/mnt` (was `-v`): a missing propagated-mount dir (guessed docker root) is refused with exit 125, not created and read as empty
- `test_docker_service.py`: 3 tests, 5 cases: hand-removed target repaired · still-present target and missing propagated-mount dir stay refused · `recover_pod_after_stale_vloopback_mount` through the real repair on ticket-0313's host shape reaches `start_existing_container` (and does not when the dir is still there); the three `-v` assertions in existing tests read `--mount`

### Verified

**Verified by the loop on 17 Sep 2026** at head 06656c5: `tools/aws_verify.sh lium-io 06656c5 --bundle head.bundle` (EC2 warm box c6i.2xlarge, run `20260917T104841Z-vu2p`, $0.019) → validators `2661 passed, 15 skipped` + 2 root-sensitive tests passed as loopverify · ruff format unchanged (report-only in CI). Fail-old probe at the round-1 tests: run `20260917T102125Z-fegp` → 4 failed, 202 passed, 1 skipped. Gate: PASS (06656c5) on EC2 sandbox Linux x86_64 (`tools/pr_quality_gate.py lium-io 1394 --claims --fresh`, 6 claims proven).

</details>


neurons/validators/src/services/docker_service.py
```diff
@@ -807,6 +807,18 @@ def _is_vloopback_driver(driver: str) -> bool:
     return driver == _VLOOPBACK_DRIVER_PREFIX or driver.startswith(f"{_VLOOPBACK_DRIVER_PREFIX}:")
 
 
+def _vloopback_repair_helper_cmd(propagated_mount_dir: str, argv: str) -> str:
```
taiberium: NIT: `argv` holds one command string, not an argv list. Rename it to `helper_command`.

neurons/validators/src/services/docker_service.py
```diff
@@ -1486,13 +1498,30 @@ async def repair_stale_vloopback_mountpoint(
...
+            propagated_mount_dir, f"rmdir /mnt/{shlex.quote(local_volume)}"
         )
         repair_result = await ssh_client.run(helper_cmd, timeout=_VLOOPBACK_REPAIR_COMMAND_TIMEOUT_SEC)
         if getattr(repair_result, "exit_status", 0) != 0:
+            # A directory that is already gone needs no rmdir: the provider removed it by hand, which
+            # is what the ticket replies ask for (DAH-3398 / ticket-0313: the stale dir was gone for
+            # 24 h and every cycle still counted the failed rmdir as a failed repair, so the
+            # container was never started). Only `test -e` exit 1 means absent; a helper that did
+            # not run (125+) stays a skipped repair.
+            absent_result = await ssh_client.run(
```
taiberium: NIT: `absent_result` also holds the "still there" answer. Rename it to `absence_check_result`, the name the tests use.

taiberium (approved):

neurons/validators/src/services/docker_service.py
surcyf123: Renamed in 3b4a2b7f: `docker_service.py:810` takes `helper_command`; both call sites pass the quoted command tail as before. Nothing else named it.

neurons/validators/src/services/docker_service.py
surcyf123: Renamed in 2a2984ea: `docker_service.py:1511` and the exit-status check at 1517 read `absence_check_result`, the parametrize name the tests already use. The 206 docker-service tests and the full validators suite (2663 passed) are unchanged.

surcyf123:
Parked until Rustam returns (after 9 Oct): value 2 in the 23 Sep review plan, so it waits for its reviewer instead of moving to a backup. No action needed before then.

surcyf123:
**lium-io#1394 read-clean at 2a2984e: after a failed rmdir of the stale vloopback mountpoint dir, one more alpine helper runs `test -e` and only exit 1 (absent) counts as repaired, so a provider-removed dir no longer strands the pod; the helper now binds with --mount type=bind so a missing propagated-mount dir fails the helper (125) instead of being created by -v; non-empty or still-mounted dirs stay a skipped repair; taiberium approved 06656c5 and both rename NITs are answered with code and replies; validators 2663 passed, 15 skipped; 1 low.**

Fresh read at `2a2984eaa18e437b92ada5d90a94be0ac8f9025c` by `subagent-eng-verdict-drain-b` (it wrote none of this PR). 0 high, 0 medium, 1 low.

```json
{
 "sha": "2a2984eaa18e437b92ada5d90a94be0ac8f9025c",
 "verdict": "clean",
 "checks": 7,
 "change_requests": [
  {
   "reviewer": "taiberium",
   "thread": 4059722411,
   "item": "NIT: rename argv to helper_command",
   "answer": "code 3b4a2b7f + reply 4060512886"
  },
  {
   "reviewer": "taiberium",
   "thread": 4059722420,
   "item": "NIT: rename absent_result",
   "answer": "code 2a2984ea + reply 4060513039"
  }
 ],
 "findings": [
  {
   "severity": "low",
   "class": "process",
   "file": "PR body",
   "path": "PR body",
   "line": 0,
   "text": "the two renames were pushed after taiberium's APPROVED at 06656c5 (his NITs, no behaviour change) and the Ran it line still says 'head moved to 2a2984e - re-run'",
   "what": "post-approval push, stale Ran it",
   "fix": "ran-at now names this head"
  }
 ],
 "dismissed": [
  {
   "line": 1517,
   "why": "only test -e exit 1 returns True; exit 0 (dir still there: non-empty or busy) and 125+ (helper did not run) fall through to REPAIR_SKIPPED"
  },
  {
   "line": 810,
   "why": "the bind spec and the volume name are shlex-quoted; --mount refuses a missing source, so a wrong docker-root guess cannot read as 'volume dir gone'"
  }
 ],
 "evidence": {
  "riskiest": {
   "line": 1517,
   "why": "a false 'repaired' would start a pod on a stale mountpoint",
   "checked_by": "read the exit-status branches of rmdir and test -e against the helper's --mount behaviour and the tests' parametrized cases"
  },
  "ran": [
   {
    "cmd": "neurons/validators: pytest tests -q (worktree HEAD == 2a2984e, py3.12)",
    "result": "2663 passed, 15 skipped"
   }
  ],
  "edge_cases": [
   "dir removed by hand -> rmdir 1, test -e 1 -> repaired",
   "dir present and non-empty -> test -e 0 -> skipped",
   "propagated-mount dir missing -> helper 125 -> skipped"
  ],
  "callers": [
   "stale-mount recovery (DAH-2306) on pod restart; RECOVER_ENCRYPTED_VOLUMES path"
  ],
  "callers_none_why": "",
  "reviewer_page": "taiberium APPROVED 06656c5 with two rename NITs, both answered"
 },
 "high": 0,
 "medium": 0,
 "low": 1,
}
```

PR #1396, opened 2026-09-17
DAH-3597 - [P2] validator: a first DinD probe miss keeps the known sysbox value (flag off)

surcyf123:
Implements [DAH-3597](https://app.notion.com/p/validator-a-DinD-probe-that-never-reached-its-container-scores-a-healthy-node-0-SYSBOX_REQUIRED_MI-3deb8bfdbde981c89a7bc67a09407e7e) · reviewer: @taiberium

**Problem** — A DinD probe that never reaches its container (`docker run` refused, sshd not up in 30 s, SSH reset) is reported as `sysbox_runtime=False`, and `SysboxRequiredCheck` scores the idle executor 0 for the cycle (`SYSBOX_REQUIRED_MISSING`) though nothing about sysbox was measured. 10 to 17 Sep: 95 such one-cycle zero scores on 50 active executors that passed the cycle before and after.

**Value** — Removes 95 zero-score cycles a week on healthy active executors (10 to 17 Sep prod); cost: 28 cycles a week of one-cycle grace to nodes that stay broken.

**Goal** — DAH-3597 · P133 reliability · WORK_QUEUE row 978 (side finding of DAH-2040).

**Fits** — `neurons/validators` `PortConnectivityCheck`, next to the DAH-2272 rental tolerance; a Redis marker per executor; two settings.

**Flag:** DIND_PROBE_FIRST_MISS_GRACE, off by default, turned on by @taiberium on 1 Oct 2026 (proposed by the loop; change it in one reply and the body follows)

**Risks**
- Flag off: no behaviour change; the check never reads or writes the marker.
- Flag on: a node that stays broken (534 driver-mismatch cycles, Details) is paid one more cycle before its 0; the second miss inside the TTL records the downgrade as today — rollback: flag off, the keys expire.
- A measured no-sysbox verdict (probe reached its container, hello-world failed) is never tolerated: grace only on `dind_ok=False`.
- Redis, flag on: one `SET NX EX` or one `DEL` per executor per cycle under the service lock; keys expire with the TTL.
**How to verify**
- `cd neurons/validators && pytest tests/test_port_connectivity_check.py -q` → 15 passed; the same file on `origin/main` → 6 failed (the settings do not exist there)
- Loki, flag on: `{namespace="subnet",container="validator"} |= "sysbox_downgrade_tolerated_reason"` → one line per tolerated first miss

**Verified by the loop:** Gate: PASS 2c83866

**Ran it:** `pytest tests/test_port_connectivity_check.py` at 2c83866 → 15 passed · full validators suite at the earlier head → 2663 passed · the file on `origin/main` → 6 failed · EC2 20260917T130346Z-ccla · 20260917T125001Z-e1k2 · 20260917T125355Z-82e3

**Self-review:** clean at 2c83866 (21 checks, 4 low)

**Parts:** validator ✓ · config ✓ · migration n/a — no schema · backend n/a — nothing to mirror · frontend n/a — no UI · CLI/SDK n/a — nothing a renter types changes · docs ✓ `.env.template` · tests ✓ 5 · dashboards n/a — no new metric

**E2E:** no user-visible path — a validator scoring gate · covered by unit — `tests/test_port_connectivity_check.py`

**Docs:** `neurons/validators/.env.template` — the two settings with their block; providers and renters neither see nor set them.

### What changed
- `core/config.py` and `.env.template`: `DIND_PROBE_FIRST_MISS_GRACE` (default False) and `DIND_PROBE_FIRST_MISS_GRACE_TTL_SECONDS` (default 3600, `gt=0`).
- `services/redis_service.py`: `record_dind_probe_miss` (`SET NX EX`, True on the first miss) and `clear_dind_probe_miss`.
- `services/task/checks/port_connectivity.py`: with the flag on, a `dind_ok=False` downgrade keeps the last known sysbox value on the first miss; a second miss records it; `dind_ok=True` clears the marker.
- `tests/test_port_connectivity_check.py`: `DummyRedis` learns the marker, `DummyConnectivityService` takes `dind_ok`; 5 tests.

### Measured (prod, read-only, 17 Sep 12:3xZ; `executor_validation_report` joined by `pipeline_id` with the validator's Loki lines, 10 to 17 Sep)

694 `SYSBOX_REQUIRED_MISSING` cycles on 94 executors; 275 on 68 executors that are active and report `sysbox_runtime=true` today. The stage the probe died at:

| stage (validator log line) | all 694 | active + sysbox today (275) |
|---|---|---|
| `docker run` refused: `nvidia-container-cli: nvml error: driver/library version mismatch` | 534 (21 executors) | 178 |
| sshd inside the DinD not up in 30 s (`DinD SSH not ready before deadline`) | 54 | 39 |
| `docker run` refused: host port already bound (`address already in use` / `port is already allocated`) | 85 | 41 |
| SSH to the DinD reset or closed | 10 | 8 |
| `docker run --rm hello-world` under sysbox failed or timed out (a measured verdict) | 7 | 6 |
| container name conflict | 3 | 3 |
| `gpu requires reset` | 1 | 0 |

Cycle shape (`lag`/`lead` over the executor's reports): 149 of 694 are one-cycle misses (passed before and after); on the active population 95 of 275: 37 sshd not up, 37 host port bound, 8 SSH reset, 8 driver mismatch that cleared by the next cycle, 4 hello-world, 1 name conflict. The 180 streak cycles on active nodes are 170 driver mismatch: a node that cannot start any GPU container, mislabelled "install sysbox" (follow-up below, not this PR).

Validator-wide in the same 7 d: 5,474 `DinD creation failed` (3,138 `address already in use`, 1,990 NVIDIA hook, 284 `port is already allocated`, 52 name conflicts), 1,988 `DinD SSH not ready before deadline`, 51 `Sysbox check failed`, 3,370 `Sysbox runtime fallback` (rented nodes, already tolerated), 278 `sysbox_downgrade_tolerated` (renting in progress, DAH-2272).

### Why not lium-io#1381
lium-io#1381 (DAH-3521) changes the custom-Dockerfile build's DinD in `services/docker_service.py` (egress DNS, readiness probes). The probe here is `services/executor_connectivity/dind_probe.py` through `ConnectivityOrchestrator`; #1381 does not touch it, so its "slow probe is one not-ready probe" rule does not reach this path.

### Design notes
- The grace lives in `PortConnectivityCheck`, next to the DAH-2272 block, because `ctx.state.sysbox_runtime` is the value `SysboxRequiredCheck` (rental-security, jam6099) and the scorer read; keeping the last known value there needs no change in the gate or the scoring.
- `dind_ok=False` is the only tolerated shape: the orchestrator sets it when `DindVerifier.verify` returned `success=False`, which is every path that never ran the hello-world (creation refused, sshd not up, SSH exception). `dind_ok=True, sysbox_runtime=False` is the hello-world verdict and is recorded at once.
- `SET NX EX` answers "first miss?" and records it in one round trip; the TTL bounds the window so a miss an hour after a forgotten one is not a "second" miss; a probe that reaches its container deletes the key so the count restarts from a pass.
- The renting-in-progress tolerance runs first and does not record a miss: a rental removing the probe container (DAH-2272) is not a probe miss.

### Not this PR
- The `SYSBOX_REQUIRED_MISSING` event should carry the DinD `docker run` error so the 534 driver-mismatch cycles tell the provider to reload the NVIDIA driver, not to install sysbox; `sysbox_required.py` is rental-security (jam6099), so it is a separate ticket.
- `address already in use` on the DinD port (3,138 validator-wide, 26 executors) points at the port probe's own containers not being gone when the DinD starts, or a rental on that port; a separate measurement.

### Ran it
EC2 warm lium-io box (c6i.2xlarge), the validators venv. The head:

```
20260917T130346Z-ccla  validators pytest tests/test_port_connectivity_check.py: rc=0 in 3 s — 15 passed
```

The earlier head of this branch (same code; the delta to 2c83866 is comments, one docstring and the `.env.template` block):

```
20260917T125001Z-e1k2  validators pytest tests/ (2 root-sensitive deselected, run below as loopverify): rc=0 in 191 s — 2663 passed, 15 skipped, 2 deselected
20260917T125001Z-e1k2  validators pytest 2 root-sensitive tests as loopverify: rc=0 — 2 passed
```

Negative control: this PR's `tests/test_port_connectivity_check.py` committed on `origin/main` (a throwaway commit, not on the branch):

```
20260917T125355Z-82e3  validators pytest tests/test_port_connectivity_check.py: rc=1 in 2 s — 6 failed, 9 passed
FAILED test_first_dind_probe_miss_keeps_the_known_sysbox_value
FAILED test_second_dind_probe_miss_inside_the_window_records_the_downgrade
FAILED test_measured_no_sysbox_verdict_is_never_tolerated_and_clears_the_miss
FAILED test_dind_probe_miss_grace_is_off_by_default
FAILED test_dind_probe_grace_ttl_rejects_zero_and_negative[0]
FAILED test_dind_probe_grace_ttl_rejects_zero_and_negative[-1]
```

### Self-review

Verdict `clean` at `2c83866` · 21 checks · by subagent-dind-r2 · 2026-09-17 13:10Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | BODY | `RECOVERY/workers/bug-dind-probe-zero-score/pr_body.md:21` | The round-1 negative-control fold landed only in part: '3 on the missing grace branch, 3 on the settings that do not exist there' still misattributes the first three failures; on origin/main the settings object has no DIND_PROBE_FIRST_MISS_GRACE attribute, so tests 1-3 fail at monkeypatch.setattr (AttributeError) before the check runs, and all 6 cases fail on the missing settings. | Write '6 failed (the 5 new tests, 6 cases: all on the two settings that do not exist on main)'; if a behavioural negative control is wanted, run the file with only port_connectivity.py reverted and quote that count (3 failed: the two miss tests and the clear test). |
| low | BODY | `RECOVERY/workers/bug-dind-probe-zero-score/pr_body.md:26` | Ran it anchors the full suite to 23f584e, which the force-push of the single-commit branch makes unreachable, cites the negctl commit as a hex literal 07dd517 that is also off-branch (the brief's rule: cite only reachable shas, write others as <eight hex characters>), and How to verify's '15 passed at the head' has no run id in Ran it. | Add the head test-file run (20260917T130346Z-ccla, 15 passed) to Ran it and state that 2c83866 differs from 23f584e only by comments, one docstring and .env.template (no test reads the template), or re-run the suite at the head; write the negctl hash as <eight hex characters>. |
| low | TESTS | `neurons/validators/tests/test_port_connectivity_check.py:384` | test_dind_probe_miss_grace_is_off_by_default asserts the live settings object is False, which a developer .env with DIND_PROBE_FIRST_MISS_GRACE=true flips (Settings loads env_file='.env'); the sibling test_executor_image_check.py:115 checks the shipped default with Settings(_env_file=None) for this reason. | Assert the default with Settings(_env_file=None).DIND_PROBE_FIRST_MISS_GRACE is False and monkeypatch the live flag to False for the behavioural part so the test is deterministic on any box. |
| low | STYLE | `neurons/validators/tests/test_port_connectivity_check.py:364` | The docstring's regression 'a miss an hour apart would have read as the second one' is at the window's edge: the default TTL is 3600 s, so a miss one hour apart finds the key expired with or without the clear; the clear matters for a miss after a pass inside the window. | Say 'a miss after a pass inside the TTL window would have read as the second one'. |

Mechanical notes dismissed: `neurons/validators/.env.template:48` Round-1 medium folded: the two settings sit after the RENTAL_PROBE block under a four-line DAH-3597 comment; the block's true/false wording matches the code (first dind_ok=False miss inside the TTL keeps the value, second records, a probe that reaches its container clears, false = recorded at once); Parts and Docs now say 'docs ✓ .env.template' truthfully and the file count is 5 (+184 -2 matches git diff --stat). · `neurons/validators/src/services/task/checks/port_connectivity.py:85` Flag off: keep_known_sysbox is set only by the renting branch and the elif on the flag is skipped, so the state/specs/extra outputs equal the pre-change code and no record/clear call is made (test_dind_probe_miss_grace_is_off_by_default asserts both lists empty); renting_in_progress is still only awaited when downgraded (short-circuit unchanged). · `neurons/validators/src/services/task/checks/port_connectivity.py:95` Flag on, if/elif enumerated: downgraded+renting keeps and never touches the marker; downgraded+dind_ok=False records (first keeps + reason, second sets dind_probe_miss_repeated and records); dind_ok=True (downgraded or not) clears; state already False + dind_ok=False touches nothing, which stops the grace renewing on a node that stays broken (after the recorded 0 the state is False so sysbox_downgraded is False until a pass). · `neurons/validators/src/services/executor_connectivity/dind_probe.py:109` Measured verdict never tolerated: DindVerifier.verify returns success=True with sysbox=False on a hello-world failure or 30 s timeout (line 109), and success=False only on docker run exit!=0 (line 63) or an exception (line 122: sshd never answered, SSH reset); DindProbe.verify delegates to DindVerifier, so the body's 'DindVerifier.verify returned success=False' and orchestrator.py:78-83 (dind_ok=dind_result.success, sysbox False on failure) match. · `neurons/validators/src/services/redis_service.py:277` redis.asyncio Redis.set(name, value, ex=int, nx=True) returns True when set and None when the key exists, so bool() is first-miss; it runs under self.lock like the neighbouring methods, and clear_dind_probe_miss goes through delete(), which also takes the lock (line 300), so the Risks line 'under the service lock' holds for both. · `neurons/validators/src/core/config.py:145` Round-1 triplication fixed: config carries a two-line pointer without the measured number, the check comment carries the why once, the redis docstring is the return-value contract; Field(gt=0) on pydantic v2 raises ValidationError for 0 and -1 and init kwargs outrank env, so the TTL test is real (its DID-NOT-RAISE on main confirms Settings(_env_file=None) has no required fields). · `neurons/validators/tests/test_port_connectivity_check.py:323` Tests 1-3 each fail on the pre-change check for the reason their docstring names: test 1 on state.sysbox_runtime True (old code recorded), test 2 on the missing dind_probe_miss_repeated key, test 3 on cleared_misses; none asserts only what it set up (test 2's repeated flag and test 1's TTL 1234 come from the code path); DummyRedis mirrors the real signatures and ctx.miner_hotkey/ctx.executor.uuid come from helpers.make_context/default_executor. · `neurons/validators/tests/test_port_connectivity_check.py:301` Round-1 lows folded: the false three-line block comment is now a neutral one-liner and the TTL docstring says Redis rejects EX 0 and negative EX with the same error (true: 'invalid expire time in set').

### Verified

Gate: PASS (2c83866) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1396)

</details>


neurons/validators/src/services/task/checks/port_connectivity.py
```diff
@@ -79,9 +81,31 @@ async def run(self, ctx: Context) -> CheckResult:
         # known value and let the next verification cycle re-measure. Mirrors
         # the rented-executor sysbox fallback in ExecutorConnectivityService.
         sysbox_downgraded = ctx.state.sysbox_runtime and not result.sysbox_runtime
+        keep_known_sysbox = False
         if sysbox_downgraded and await ctx.services.redis.renting_in_progress(
             ctx.miner_hotkey, ctx.executor.uuid
         ):
+            keep_known_sysbox = True
+        elif settings.DIND_PROBE_FIRST_MISS_GRACE:
```
taiberium: NIT: `run()` is now 186 lines. Move this block into `_should_keep_last_known_sysbox(ctx, result, extra_info) -> bool`, and drop the unclear `keep_known_sysbox` flag.

neurons/validators/tests/test_port_connectivity_check.py
```diff
@@ -279,3 +297,109 @@ async def test_port_connectivity_records_sysbox_downgrade_when_not_renting(conte
...
+    assert redis_service.recorded_misses == []
+    assert redis_service.cleared_misses == [(ctx.miner_hotkey, ctx.executor.uuid)]
+
+
+@pytest.mark.asyncio
+async def test_dind_probe_miss_grace_is_off_by_default(context_factory):
+    """Flag off (the default): a probe that never reached its container records the downgrade at
+    once and the miss record is never read or written. Regression: the grace applied with the flag
+    off, which would change scoring before the flag's owner turned it on."""
+    assert settings.DIND_PROBE_FIRST_MISS_GRACE is False
```
taiberium: NIT: a local `.env` with the flag on breaks this test. Assert on `Settings(_env_file=None)` and set the live flag to `False` with `monkeypatch`, like `test_executor_image_check.py` does.

taiberium (approved):

PR #1399, opened 2026-09-17
DAH-3602 - [P1] whitelist the current CVM compose hash and fail CI on the next drift

surcyf123:
Implements [DAH-3602](https://app.notion.com/p/A-new-CVM-built-from-the-current-code-fails-the-validator-whitelist-and-the-guide-does-not-say-whic-3deb8bfdbde98173b4a1c0ed6e96f1cb) · reviewer: @taiberium

**Problem** — Every CVM created from executor-v1.128 to v1.130 or `main` measures compose hash `87d34300…`; the prod whitelist holds only `8224d588…` (v1.127), so the validator does not know the hash (zero score once `ENABLE_ATTESTATION_WHITELIST` is on; off in prod today). lium-io#1339 (14 Sep) edited a measured file and no test compared those files with the whitelist. First provider hit: TEEmo, 17 Sep.

**Value** — Unblocks CVM onboarding on every new TDX host since 14 Sep (3 releases, 0 whitelisted today; the 2 running prod CVMs are unaffected). The test turns the next drift into a red CI check on the PR causing it; the release job puts digest and hash into every release's notes (v1.130 had neither).

**Goal** — [DAH-3602](https://app.notion.com/p/A-new-CVM-built-from-the-current-code-fails-the-validator-whitelist-and-the-guide-does-not-say-whic-3deb8bfdbde98173b4a1c0ed6e96f1cb) · P145 providers first.

**Fits** — the validator whitelist in `services/const.py`, the app-compose builder in `dstacktee/scripts/`, the CI router, the executor prod release workflow.

**Flag:** no flag

**Risks**
- Next validator deploy: version 5 (`87d34300…`) is accepted beside version 4; a provider whose new CVM scored zero sees it accepted next cycle — rollback: revert (version 4 stays).
- Next `executor-v*` tag: job `release-notes` (`contents: write`) writes the "CVM attestation" section into the tag's release (append / replace stale / create) — rollback: delete the job; it runs after `deploy`, never blocks the build.
- `87d34300…` takes version 5; DAH-2780 (#1266) takes 6, and its rebase fails this PR's test until it whitelists its hash. #1264/#1274 touch the same host-setup.md block.

**How to verify**
- `python3 neurons/executor/dstacktee/scripts/compose_hash.py` → `87d3430000bb7046a19eeaa6efe074fd8b5f7856bbfe2906addab238174c542d` (`--release-notes` → the section with it and the digest)
- `cd neurons/validators && pytest tests/test_tdx_compose_hash_whitelist.py` → 15 passed; with the version-5 entry removed from `const.py` the first two fail; eight tests drive `release_notes_update.sh` against a stubbed `gh`

**Verified by the loop:** Gate: PASS 650d200

**Ran it:** `pytest tests/test_tdx_compose_hash_whitelist.py` → 15 passed in 1.27 s · EC2 `20260918T175940Z-ia8l` at 650d200 · `compose_hash.py` → `87d34300…` (Mac)

**Self-review:** clean at 650d200 (15 checks, 2 low)
**Parts:** backend n/a — no platform change · validator ✓ · executor ✓ · CI ✓ router + release job · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — no client · docs ✓ host-setup.md here, public guide lium-platform#532 · tests ✓ 15 new · dashboards n/a — no metric

**Docs:** `dstacktee/docs/host-setup.md` · docs: https://github.com/Datura-ai/lium-platform/pull/532

**E2E:** no user-visible path — the e2e stack runs no CVM

### What changed
- `const.py`: `87d34300…` added as version 5 in `TDX_WHITELIST["COMPOSE_HASH"]["PROD"]`; version 4 kept; the DAH-2780 note now says 6
- `dstack.py`: the app-compose dict and its JSON form are `build_app_compose()` / `app_compose_json()`; `new` writes the same bytes as before (checked: same hash)
- `scripts/compose_hash.py`: rebuilds the hash from the three measured files + `APPROVED_RUNNER_IMAGE_DIGEST`; `--release-notes` prints the section, `--check <app-compose.json>` compares a created CVM (`expected …`/`actual …`/`OK`, exit 1 on a mismatch: what host-setup.md now tells a provider to run after `new`). The section and host-setup.md say what an unknown hash costs under each setting of the validator's `ENABLE_ATTESTATION_WHITELIST` (`config.py`: default `False`; `attestation_service.py:592` is the check): on, the CVM scores zero; off (prod today), the validator accepts the hash and scores the node as before. The section names the runner digest's push date (2026-08-19, `APPROVED_RUNNER_IMAGE_PUSHED`) so nobody reads it as this release's runner
- `tests/test_tdx_compose_hash_whitelist.py` (15 tests: whitelisted, strictly newest version (with a message naming the fix), `dstack.py new` writes the rebuilt bytes, `lium-cvm.sh` flags contract and the paths behind the flags, release-notes section (digest push date, the whitelist flag named), `--check`, `--digest` is 64 hex, and the release script against a stubbed `gh`: append, create, leave when the section equals the generated one, replace a stale section while the hash is quoted elsewhere, replace a section with the current hash and a stale digest, rewrite a doubled section to one, match a heading with trailing whitespace, refuse to create when the release cannot be read) and the CI router: a change under `neurons/executor/dstacktee/` now runs the validator tests
- `executor_cd_prod.yml`: job `release-notes` after `deploy` runs `scripts/release_notes_update.sh <tag>`, which writes the section into the tag's GitHub release (the body's `## CVM attestation` section is compared with the generated one line for line; another hash, a stale digest next to the current hash, or a doubled heading is replaced, never a red job); host-setup.md, README, `.env.example` point at it; host-setup.md names `compose_hash.py --release-notes` as the source while no released tag carries the section and tells the provider to run `compose_hash.py --check` after `new`

### Reproduction (Mikhail's numbers, confirmed)
`RECOVERY/workers/bug-3602-cvm-whitelist/repro_compose_hash.py` (loop repo) rebuilds the hash per git ref with the pre-refactor formula (`json.dump(indent=4)` of the dict `dstack.py new` writes, digest stamped by envsubst, `--local-key-provider`, no logs/sysinfo):

```
executor-v1.127      ba007bf10  8224d58801af6333561f116e2d566b179b399f1d1d700f0e8a5ab9326ae901d9
executor-v1.128      c102a3326  87d3430000bb7046a19eeaa6efe074fd8b5f7856bbfe2906addab238174c542d
executor-v1.129      08789e3e8  87d3430000bb7046a19eeaa6efe074fd8b5f7856bbfe2906addab238174c542d
executor-v1.130      4cf0119b8  87d3430000bb7046a19eeaa6efe074fd8b5f7856bbfe2906addab238174c542d
origin/main          219c4f1c2  87d3430000bb7046a19eeaa6efe074fd8b5f7856bbfe2906addab238174c542d
```

executor-v1.128 is the #1339 squash itself, so the drift starts one release earlier than the ticket says (v1.129). The refactored `DStackManager.setup_instance` was driven with the same inputs on the Mac (a placeholder image path; it only records the name, so `new` runs to the manifest); `sha256sum` of `shared/app-compose.json` → `87d34300…`.

### Why the digest lives in `compose_hash.py`
The validator whitelists a hash, not a digest; the digest is only meaningful together with the files, and the release notes need both. One constant next to the builder, read by the test and the release job, keeps them from drifting apart. Rotating the runner = change the constant, whitelist the new hash, same PR.

### Risks, in full
- Validator: on the next deploy the validator accepts version 5 beside version 4; nothing else in the attestation path changes. Rollback: revert the commit; the old hash stays version 4 either way.
- Release job: `needs: deploy`, `permissions: contents: write`, `github.token`. The step is `scripts/release_notes_update.sh <tag>`. An existing release gets the section appended once (the section = the `## CVM attestation` heading up to the next `## ` heading or the end; CRLF-safe); a missing release is created from the `generate-notes` API body with the section last (one `gh release create --verify-tag`). A release whose section is not the generated one gets it replaced (a moved tag's hash, a stale digest next to the current hash, a heading an edit doubled): the whole section is compared, not the hash, so a body that quotes the new hash elsewhere, or a section with the right hash and a wrong digest, does not pass, and the job never stays red after `deploy` published the images. Text a human wrote after the section survives. Nobody's images are affected.
- Merge order: lium-io#1264 (DAH-2780's other PR) and lium-io#1274 rewrite the §6 upgrade block of `dstacktee/docs/host-setup.md` this PR edits; whichever lands second rebases that block and keeps the `compose_hash.py --check` line after `new`.
- Version numbering: the `const.py` comment had reserved 5 for DAH-2780's `pre_launch_script.sh` rewrite (#1266). `87d34300…` is what three shipped releases measure today, so it takes 5; #1266 takes 6 when it lands.

### Known follow-ups
- The STAGE whitelist (`72c9c91a…`/`feedf7cb…`) does not match the current `docker-compose.staging.yml` with any digest this PR knows; not touched here (the test covers PROD only).
- `dstack.py` keeps `Optional[...]` in the new signature to match the file (ruff UP045 flags the whole file already).
- `compose_hash.py` rebuilds with the flags `lium-cvm.sh` passes today; the flags test pins that contract rather than running the shell script (no QEMU, no image in CI).
- The release-script tests stub `gh` with a bash script (releases are files); the real `gh release`/`gh api generate-notes` calls run for the first time on the next `executor-v*` tag.
- executor-v1.129 and v1.130 have releases without the section and v1.131 has a tag with no release; the job fires only on the next tag push. Their measured files equal `main`'s (`git diff executor-v1.129 main -- neurons/executor/dstacktee/app` is empty for v1.128 through v1.131), so after merge, from a `main` checkout where `python3 neurons/executor/dstacktee/scripts/compose_hash.py` prints `87d34300…`, `GH_TOKEN=… bash neurons/executor/dstacktee/scripts/release_notes_update.sh executor-v1.129` (then `executor-v1.130`, `executor-v1.131`) writes it; the loop does not edit or create releases (SO §31), so this is one for the dev who merges.

### Ran it
- EC2 `20260917T150021Z-ubna` (warm lium-io box i-0de19730713274fa7, c6i.2xlarge), head 89be54d from a bundle: `validators pytest tests/` → `2663 passed, 15 skipped, 2 deselected in 180.79s`; `executor pytest tests/` → `382 passed, 2 skipped in 14.05s`; ruff report-only unchanged. Earlier full runs at 738cd33 (`20260917T142857Z-asaq`, 2661 + 382 passed, before the two extra tests).
- EC2 `20260917T150632Z-y7v1`, head bb47cef: `validators pytest tests/test_tdx_compose_hash_whitelist.py` → `6 passed, 107 warnings in 0.05s`.
- EC2 `20260918T081603Z-fwcm` (same warm box), head f18c66b (Rustam's round: the release script, its six tests, the `--digest` test, the strict newest-version assert): `validators pytest tests/test_tdx_compose_hash_whitelist.py` → `13 passed, 107 warnings in 1.02s`. The round's first draft ran as `20260918T080721Z-f519` (11 passed) before the self-review's two lows (a heading with trailing whitespace; a `gh release view` failure other than not-found must not lead to `release create`) were fixed and tested. Two earlier runs at that round's drafts failed and were fixed before the push: the box's system `python3` is 3.9 (dstack.py needs 3.10 — the test now puts the venv interpreter first in `PATH`, as ubuntu-latest's python3 is 3.12), and `read_text()` translated the CRLF fixture (read back as bytes now).
- Mac (stdlib python3, no venv): `python3 neurons/executor/dstacktee/scripts/compose_hash.py` → `87d3430000bb7046a19eeaa6efe074fd8b5f7856bbfe2906addab238174c542d`; `--env staging` → `17cf477b…` (not in the STAGE whitelist, see Known follow-ups); `--release-notes` → the `## CVM attestation` section: the digest with its 2026-08-19 push date, the expected hash, the `sha256sum` and `--check` lines, and what an unknown hash costs with `ENABLE_ATTESTATION_WHITELIST` on and off.
- Fail-old on the Mac: the same hash against `origin/main`'s `const.py` → `in whitelist: False`; against the branch → `True`, `newest: True`.
- `DStackManager.setup_instance` driven with the resolved compose + the two scripts and a placeholder image path (it runs to the manifest) → `sha256sum` of `shared/app-compose.json` `87d34300…`, identical to the pre-refactor writer; `LC_ALL=C python3 …/compose_hash.py` → the same hash (reads are UTF-8 now).
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/executor_cd_prod.yml'))"` → parses (round 1 had an unquoted `name: Release notes: …`, which PyYAML rejects); `actionlint .github/workflows/executor_cd_prod.yml` → only the pre-existing SC2086 info on the `deploy` step (round 1); at f18c66b the step is one `bash …/release_notes_update.sh "$TAG"` line, `bash -n` on the script → ok, PyYAML parses the workflow.
- EC2 `20260918T175940Z-ia8l` (warm lium-io box i-0d2847e63c80eec7b, c6i.2xlarge), head 650d200 from a bundle (Rustam's round 2: whole-section compare in the release script with two new tests, the whitelist-flag wording, the digest date, `--check` in host-setup.md, the two test nits): `validators pytest tests/test_tdx_compose_hash_whitelist.py` → `15 passed, 107 warnings in 1.27s`, rc 0.
- The `release-notes` job itself has not run: it fires on the next `executor-v*` tag (the loop does not tag or release, SO §31). Not run on macOS beyond the stdlib script.

### Self-review

Verdict `findings` at `650d200` · 15 checks · by subagent-r327b-io · 2026-09-18 18:09Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | CORRECTNESS | `neurons/executor/dstacktee/scripts/release_notes_update.sh:62` | `gh release view --json body --jq .body` prints the literal `null` for a release whose body is null (a release created without notes), so the append path writes `null` as the first line of the new notes; same at line 89 (read-back) and line 78 (generate-notes body). | Use `--jq '.body // ""'` on the three calls so an absent body is an empty body. |
| low | TEST-GAP | `neurons/validators/tests/test_tdx_compose_hash_whitelist.py:159` | `"pushed" not in preview` pins a wording token: any future sentence in the section that contains the word "pushed" fails the test for a reason unrelated to the preview note. | Assert `compose_hash.APPROVED_RUNNER_IMAGE_PUSHED not in preview` (the date is the fact the note must not claim for a preview digest). |

Mechanical notes dismissed: `RECOVERY/workers/responder-r327b/body_1399_draft.md:29` Fixed in the body draft before push: the Parts line now reads `tests ✓ 15 new`, matching the 15 tests in the new file. · `RECOVERY/workers/responder-r327b/body_1399_draft.md:78` Fixed in the body draft before push: the `--release-notes` line now describes the section (digest with its 2026-08-19 push date, expected hash, the `sha256sum` and `--check` lines, the ENABLE_ATTESTATION_WHITELIST on/off sentence), which matches release_notes_section() at HEAD. · `RECOVERY/workers/responder-r327b/body_1399_draft.md:64` Fixed in the body draft before push: line 64 now reads "lium-io#1264 (DAH-2780's other PR)"; #1264 and #1266 are both DAH-2780 PRs per `gh pr view`, so the two numbers were not a contradiction. · `neurons/executor/dstacktee/scripts/compose_hash.py:49` `DIGEST_PLACEHOLDER = "${EXECUTOR_RUNNER_IMAGE_DIGEST}"` is the literal envsubst placeholder lium-cvm.sh:466 substitutes; an f-string would raise NameError. Scanner false positive, unchanged in round 2. · `neurons/validators/tests/test_tdx_compose_hash_whitelist.py:131` `assert "envsubst '${EXECUTOR_RUNNER_IMAGE_DIGEST}'" in script` asserts the literal shell text present at lium-cvm.sh:466; not a format string. · `neurons/validators/tests/test_tdx_compose_hash_whitelist.py:147` Line 147 is `assert f'heading="{compose_hash.RELEASE_NOTES_HEADING}"' in script`, already an f-string and the script has `heading="## CVM attestation"`; scanner false positive. · `neurons/executor/dstacktee/scripts/release_notes_update.sh:51` Checked under bash with set -euo pipefail: `grep -cxE … || true` prints 0 and returns 0 on zero matches, 2 on a doubled heading; `[ 0 -eq 1 ]` fails inside the && list without tripping set -e because section_is_current is only called in `if` / `||` position. · `neurons/executor/dstacktee/scripts/release_notes_update.sh:56` Trailing-newline handling verified: `--release-notes` output (print adds one \n) and the section awk cuts from a body (each line printed with \n, trailing blanks trimmed) compare equal via `cmp -s` on process substitutions for: section last with/without a final newline, section followed by blank lines and a `## Thanks` heading, CRLF stripped; a doubled heading (count 2) and a heading with a trailing space (line printed as `$0`) take the replace path, matching the two new tests and the trailing-whitespace test.

### Verified

Gate: PASS (650d200) on EC2 sandbox Linux x86_64 (aws_run gate-lium-io-1399)

</details>


neurons/validators/tests/test_tdx_compose_hash_whitelist.py
```diff
@@ -0,0 +1,147 @@
...
+        'TDX_WHITELIST["COMPOSE_HASH"]["PROD"]: a CVM created from this tree scores zero. '
+        f"Add it as version {max(PROD_WHITELIST.values()) + 1} in services/const.py."
+    )
+
+
+def test_prod_compose_hash_of_this_checkout_is_the_newest_version():
+    # newest-wins: the checkout is the release providers deploy next, so its hash carries the highest
+    # version, or raising TDX_MINIMUM_COMPOSE_VERSION to retire an old release would retire this one
+    measured = compose_hash.compose_hash("prod")
+    assert PROD_WHITELIST.get(measured) == max(PROD_WHITELIST.values())
```
taiberium: nit: equality lets two hashes share the top version, and then the minimum-version floor cannot retire one without the other. Compare against `max((v for h, v in PROD_WHITELIST.items() if h != measured), default=0)` instead. Non-blocking, fine as a follow-up.

.github/workflows/executor_cd_prod.yml
```diff
@@ -37,3 +37,45 @@ jobs:
...
+
+      - name: Add the approved runner digest and the expected compose hash to the release
+        run: |
+          set -euo pipefail
+          python3 neurons/executor/dstacktee/scripts/compose_hash.py --release-notes > section.md
+          hash="$(python3 neurons/executor/dstacktee/scripts/compose_hash.py)"
+          if gh release view "$TAG" --json body --jq .body | tr -d '\r' > body.md; then
+            # a body edited in the web UI comes back with CRLF; tr above makes the marker match
+            if grep -qx "## CVM attestation" body.md; then
+              if grep -qF "$hash" body.md; then
```
taiberium: Please search for `$hash` only inside the `CVM attestation` section. A stale section currently passes when the new hash appears elsewhere in the release body.

neurons/executor/dstacktee/scripts/compose_hash.py
```diff
@@ -0,0 +1,125 @@
...
+        "--digest", default=APPROVED_RUNNER_IMAGE_DIGEST, help="runner digest, sha256:<64-hex>"
+    )
+    parser.add_argument(
+        "--release-notes", action="store_true", help="print the release-notes section"
+    )
+    parser.add_argument(
+        "--check", metavar="APP_COMPOSE_JSON", help="compare a created CVM's app-compose.json"
+    )
+    args = parser.parse_args(argv)
+    if not args.digest.startswith("sha256:") or len(args.digest) != len("sha256:") + 64:
```
taiberium: nit: this accepts `sha256:` plus any 64 characters, so 64 `z` passes and prints a hash for a digest Docker cannot pull. Use `re.fullmatch(r"sha256:[0-9a-f]{64}", args.digest)` (needs `import re`). Non-blocking, fine as a follow-up.

.github/workflows/executor_cd_prod.yml
```diff
@@ -37,3 +37,45 @@ jobs:
...
+                echo "release $TAG already carries the section with hash $hash"
+                exit 0
+              fi
+              # the tag was moved to a tree with another hash: a stale section is worse than none
+              echo "release $TAG carries a CVM attestation section without hash $hash; fix the release body by hand" >&2
```
taiberium: A moved tag lands here with a stale section, and the prod CD run goes red after `deploy` already published the images, on every re-run. Can this replace the section instead: `sed -i '/^## CVM attestation$/,$d' body.md` then append the new one?

neurons/executor/dstacktee/scripts/compose_hash.py
```diff
@@ -0,0 +1,125 @@
...
+            f"- Approved executor-runner digest (`EXECUTOR_RUNNER_IMAGE_DIGEST` in `.env`): `{digest}`",
+            f"- Expected compose hash of a CVM created from this release: `{compose_hash(env, digest)}`",
+            "",
+            "Check after `sudo ./lium-cvm.sh new <name>` and before `run`:",
+            "",
+            "```bash",
+            "sha256sum run/vms/<name>/shared/app-compose.json",
+            "```",
+            "",
+            "The output must equal the expected compose hash. Any other value means the measured files "
```
taiberium, line 105: nit: this says what a wrong value means, but a matching value still scores zero until the validator release carrying this hash is out. Worth one more sentence saying that, so the notes stay trustworthy. Non-blocking, fine as a follow-up.

neurons/executor/dstacktee/docs/host-setup.md
```diff
@@ -106,15 +106,18 @@ Two containers: `aesmd` (SGX architectural enclaves, host network) and `gramine-
    Required beyond the obvious (`MINER_HOTKEY_SS58_ADDRESS`, ports, `CVM_VCPUS/MEMORY/DISK`, `CVM_GPUS`):
 
    - `ENABLE_TDX_ATTESTATION=true`
-   - `EXECUTOR_RUNNER_IMAGE_DIGEST=sha256:<64-hex>` — copy from the release notes. `lium-cvm.sh new` pins this digest into the measured compose (the attested trust boundary) and refuses to run without it.
+   - `EXECUTOR_RUNNER_IMAGE_DIGEST=sha256:<64-hex>` — copy from the "CVM attestation" section of the release notes of the tag you checked out. `lium-cvm.sh new` pins this digest into the measured compose (the attested trust boundary) and refuses to run without it.
```
taiberium: No released tag has this section yet, so a provider setting up from `main` before the next tag has no source for the digest. Add the fallback you already give for the hash: `python3 scripts/compose_hash.py --release-notes` prints both.

taiberium (changes requested):
See the inline comments.

neurons/validators/tests/test_tdx_compose_hash_whitelist.py
surcyf123: Done in `f18c66b`: `assert PROD_WHITELIST.get(measured, 0) > max((v for h, v in PROD_WHITELIST.items() if h != measured), default=0)`.

neurons/executor/dstacktee/scripts/compose_hash.py
surcyf123: Done in `f18c66b`: `re.fullmatch(r"sha256:[0-9a-f]{64}", args.digest)`; `test_digest_must_be_64_hex` passes 64 `z` and expects argparse's exit 2.

.github/workflows/executor_cd_prod.yml
surcyf123: Done in `f18c66b`: a section without the checkout's hash is replaced; the job no longer exits 1. It is your `sed` bounded by the next `## ` heading, so text written after the section survives. A created release gets the generated notes first and the section last (`gh api …/releases/generate-notes`, then one `gh release create --verify-tag`). Only a `gh release view` failure other than 'release not found' (auth, 5xx) still exits 1, so the job never creates a duplicate release. Six tests cover append, create, leave, replace, a heading with a trailing space, and the view error.

neurons/executor/dstacktee/docs/host-setup.md
surcyf123: Done in `f18c66b`: host-setup.md §5 step 1 and the §6 upgrade block now name the fallback. While the tag's notes have no section, `python3 scripts/compose_hash.py --release-notes` prints the same section for the checkout: digest and expected hash.

neurons/executor/dstacktee/scripts/compose_hash.py
surcyf123: Done in `f18c66b`: the section now ends with one more sentence. A matching value scores only once the validator release that whitelists this hash is deployed; until then a correct CVM scores zero as well.

.github/workflows/executor_cd_prod.yml
surcyf123: Done in `f18c66b`: the job's shell is now `dstacktee/scripts/release_notes_update.sh`. Its section_of() cuts the body from the CVM attestation heading to the next `## ` heading or the end. The hash is grepped only in that slice. A test puts the new hash in a Notes paragraph above a stale section and asserts the section is replaced (stubbed gh, releases are files).

surcyf123:
@taiberium All six inline threads are answered in `f18c66b`. The release job now searches the hash inside the section only and replaces a stale one; it is a tested script. host-setup.md names the `compose_hash.py --release-notes` fallback. The three nits are in. CI is green at that head and 13 tests pass on EC2. Re-requesting your review.

neurons/validators/tests/test_tdx_compose_hash_whitelist.py
```diff
@@ -0,0 +1,308 @@
...
+        "--env-file",
+        "$gpu_args",
+        "$port_args",
+        "$lkp_args",
+        "$logs_arg",
+        "$sysinfo_arg",
+    }, invocation
+    assert 'local lkp_args="--local-key-provider"' in script
+    assert 'local logs_arg=""' in script and 'local sysinfo_arg=""' in script
+    assert "envsubst '${EXECUTOR_RUNNER_IMAGE_DIGEST}'" in script
```
taiberium, line 131: nit: this pins the flags but not the paths behind them, so a changed INIT_SCRIPT or prod compose path moves the real hash with CI still green. non-blocking, fine as a follow-up.

```suggestion
    assert "envsubst '${EXECUTOR_RUNNER_IMAGE_DIGEST}'" in script
    assert 'INIT_SCRIPT="$THIS_DIR/app/init_script.sh"' in script
    assert 'PRE_LAUNCH_SCRIPT="$THIS_DIR/app/pre_launch_script.sh"' in script
    assert 'local compose_file="$THIS_DIR/app/docker-compose.yml"' in script
```

neurons/executor/dstacktee/scripts/compose_hash.py
```diff
@@ -0,0 +1,128 @@
...
+            "or the digest differ from the release; the validator rejects the CVM and it scores zero. "
```
taiberium: Please qualify this outcome by whitelist enforcement. Validators with `ENABLE_ATTESTATION_WHITELIST=false` accept unknown compose hashes.

neurons/executor/dstacktee/scripts/release_notes_update.sh
```diff
@@ -0,0 +1,78 @@
...
+    p { next }
+    /^$/ { blanks++; next }
+    { for (; blanks > 0; blanks--) print ""; print }' "$1"
+}
+has_heading() { grep -qxE "$heading[[:space:]]*" "$1"; }
+
+# a body edited in the web UI comes back with CRLF; tr makes the heading match
+if gh release view "$tag" --json body --jq .body > "$work/raw.md" 2> "$work/view.err"; then
+  tr -d '\r' < "$work/raw.md" > "$work/body.md"
+  if section_of "$work/body.md" | grep -qF "$hash"; then
```
taiberium: Please compare the full generated section before exiting. A current hash with a stale digest or duplicate heading leaves conflicting provider instructions unchanged.

neurons/executor/dstacktee/docs/host-setup.md
```diff
@@ -106,15 +106,18 @@ Two containers: `aesmd` (SGX architectural enclaves, host network) and `gramine-
...
    - `ENABLE_TDX_ATTESTATION=true`
-   - `EXECUTOR_RUNNER_IMAGE_DIGEST=sha256:<64-hex>` — copy from the release notes. `lium-cvm.sh new` pins this digest into the measured compose (the attested trust boundary) and refuses to run without it.
+   - `EXECUTOR_RUNNER_IMAGE_DIGEST=sha256:<64-hex>` — copy from the "CVM attestation" section of the release notes of the tag you checked out. While the notes of your tag do not carry that section yet (no released tag has it before the first release after this change), `python3 scripts/compose_hash.py --release-notes` prints the same section for the checkout: the digest and the expected compose hash. `lium-cvm.sh new` pins this digest into the measured compose (the attested trust boundary) and refuses to run without it.
 
-2. Create and boot (the OS image downloads once per host, then is reused):
+2. Create, check the measurement, boot (the OS image downloads once per host, then is reused):
 
    ```bash
    sudo ./lium-cvm.sh new my-executor
+   sha256sum run/vms/my-executor/shared/app-compose.json
```
taiberium: nit: compose_hash.py --check already does this comparison and exits 1, so the provider does not have to match 64 characters by eye. non-blocking, fine as a follow-up.

```suggestion
   python3 scripts/compose_hash.py --check run/vms/my-executor/shared/app-compose.json
```

neurons/executor/dstacktee/scripts/compose_hash.py
```diff
@@ -0,0 +1,128 @@
...
+    return hashlib.sha256(measured_app_compose(env, digest).encode()).hexdigest()
+
+
+def release_notes_section(env: str = "prod", digest: str = APPROVED_RUNNER_IMAGE_DIGEST) -> str:
+    """The section a provider copies from: the digest for .env and the hash to check after `new`."""
+    return "\n".join(
+        [
+            RELEASE_NOTES_HEADING,
+            "",
+            f"- Approved executor-runner digest (`EXECUTOR_RUNNER_IMAGE_DIGEST` in `.env`): `{digest}`",
```
taiberium: nit: this digest is the 2026-08-19 runner, older than the tag it ships with. Naming the approval date here stops a provider reading it as this release's runner. non-blocking.

neurons/validators/tests/test_tdx_compose_hash_whitelist.py
```diff
@@ -0,0 +1,308 @@
...
+    )
+
+
+def test_prod_compose_hash_of_this_checkout_is_the_newest_version():
+    # newest-wins: the checkout is the release providers deploy next, so its hash carries the highest
+    # version, or raising TDX_MINIMUM_COMPOSE_VERSION to retire an old release would retire this one
+    # strictly above every other hash: two hashes sharing the top version could not be retired apart
+    measured = compose_hash.compose_hash("prod")
+    others = max((v for h, v in PROD_WHITELIST.items() if h != measured), default=0)
+    assert PROD_WHITELIST.get(measured, 0) > others
```
taiberium: nit: give this assert a message. A revert of a measured file lands here as a bare `5 > 6` and the fix is not obvious.

```suggestion
    assert PROD_WHITELIST.get(measured, 0) > others, (
        f"compose hash {measured} of this checkout is not the newest entry in "
        'TDX_WHITELIST["COMPOSE_HASH"]["PROD"]: give it a version above every other hash '
        "(a revert to an older tree lands here)."
    )
```

taiberium (changes requested):
The release-notes job in .github/workflows/executor_cd_prod.yml only fires on the next tag push, so executor-v1.129 and v1.130 keep no section. Can you run the script once against those two tags after merge?

neurons/executor/dstacktee/scripts/release_notes_update.sh
surcyf123: Done in `650d200`: `section_is_current` compares the body's section with the generated one line for line (trailing blank lines dropped on both sides) and requires exactly one heading. A current hash next to a stale digest, or a doubled heading, is replaced; the read-back at the end uses the same compare. Two new tests: `test_release_script_replaces_a_section_with_the_current_hash_but_a_stale_digest` and `test_release_script_rewrites_a_doubled_section_to_one`; both fail on the previous script.

neurons/validators/tests/test_tdx_compose_hash_whitelist.py
surcyf123: Done in `650d200`: applied as suggested: the three path assignments (`INIT_SCRIPT`, `PRE_LAUNCH_SCRIPT`, the prod `compose_file`) are pinned next to the flags.

neurons/validators/tests/test_tdx_compose_hash_whitelist.py
surcyf123: Done in `650d200`: applied as suggested.

neurons/executor/dstacktee/scripts/compose_hash.py
surcyf123: Done in `650d200`: the section now names both settings of `ENABLE_ATTESTATION_WHITELIST`. On: the CVM scores zero. Off (the config default, and production today): the validator accepts the hash and scores the node as before. The module docstring and host-setup.md say the same. The check is `attestation_service.py:592`.

neurons/executor/dstacktee/docs/host-setup.md
surcyf123: Done in `650d200`: applied as suggested in step 2 and in the §6 upgrade block. The text under it says what `--check` prints and that exit 1 means the CVM is not the release's. The release-notes section names `--check` too.

surcyf123:
@taiberium Yes, with one detail: the script is not on those tags, so it runs from a main checkout after the merge and writes main's hash. That is those tags' hash: the measured files under `neurons/executor/dstacktee/app` are identical from executor-v1.128 through v1.131 and main, and `compose_hash.py` prints `87d34300…` there. The commands are `bash neurons/executor/dstacktee/scripts/release_notes_update.sh executor-v1.129`, then the same for executor-v1.130, with a GH_TOKEN that can edit releases. executor-v1.131 has a tag and no release yet, so the same command would create its release with generated notes plus the section. The loop does not edit or create releases; that stays with you, and the body lists it under Known follow-ups for whoever merges. All six inline comments are answered in `650d200`; 15 tests pass on EC2.

neurons/executor/dstacktee/scripts/compose_hash.py
surcyf123: Done in `650d200`: the section now says the runner image was pushed 2026-08-19 and is not rebuilt per executor release. So the digest is older than the release and still the approved one. The date is `APPROVED_RUNNER_IMAGE_PUSHED` next to the digest constant, taken from `const.py:339`. A `--digest` preview says "not the approved runner" instead.

neurons/validators/src/services/const.py
```diff
@@ -341,9 +341,16 @@
...
-            # three files moves the hash: DAH-2780 rewrites pre_launch_script.sh and needs its own
-            # version-5 entry here.
+            # three files moves the hash and needs the next version here: the checkout's hash is
+            # rebuilt by neurons/executor/dstacktee/scripts/compose_hash.py and
+            # tests/test_tdx_compose_hash_whitelist.py fails CI while it is missing. Kept: CVMs
+            # created from executor-v1.127 and earlier still attest with it.
             "8224d58801af6333561f116e2d566b179b399f1d1d700f0e8a5ab9326ae901d9": 4,
+            # DAH-3602 — the same runner digest with the measured files at c102a332 (lium-io#1339,
+            # DAH-2834, three settings added to init_script.sh; executor-v1.128 to v1.130 and main).
+            # Missing here from 2026-09-14 to this entry, so every CVM created from those releases
```
taiberium, line 351: **NIT:** Prod runs with `ENABLE_ATTESTATION_WHITELIST=false`, so these CVMs did not score zero there. Write "would score zero with the whitelist on" here and in the test docstring.

neurons/executor/dstacktee/scripts/compose_hash.py
```diff
@@ -0,0 +1,151 @@
...
+
+def compose_hash(env: str = "prod", digest: str = APPROVED_RUNNER_IMAGE_DIGEST) -> str:
+    return hashlib.sha256(measured_app_compose(env, digest).encode()).hexdigest()
+
+
+def release_notes_section(env: str = "prod", digest: str = APPROVED_RUNNER_IMAGE_DIGEST) -> str:
+    """The section a provider copies from: the digest for .env and the hash to check after `new`."""
+    if digest == APPROVED_RUNNER_IMAGE_DIGEST:
+        digest_note = (
+            f"the runner image pushed {APPROVED_RUNNER_IMAGE_PUSHED}; the runner is not rebuilt per "
```
taiberium: **NIT:** The `deploy` job rebuilds and pushes `executor-runner:latest` on every tag, so "not rebuilt per executor release" is wrong. Write "the approved digest does not change per release".

neurons/executor/dstacktee/scripts/dstack.py
```diff
@@ -58,6 +58,47 @@ def merge_dicts(*dicts):
...
 
+def build_app_compose(
+    compose_content: str,
+    *,
+    local_key_provider: bool,
+    enable_logs: bool,
+    enable_sysinfo: bool,
+    init_script: Optional[str] = None,
+    pre_launch_script: Optional[str] = None,
+) -> dict:
```
taiberium: **NIT:** A bare `dict` hides the measured keys. A `TypedDict` with `total=False` names them and writes the same bytes.

neurons/executor/dstacktee/scripts/compose_hash.py
```diff
@@ -0,0 +1,151 @@
...
+    "local": "docker-compose.local.yml",
+}
+DIGEST_PLACEHOLDER = "${EXECUTOR_RUNNER_IMAGE_DIGEST}"
+RELEASE_NOTES_HEADING = "## CVM attestation"
+
+
+def measured_app_compose(env: str = "prod", digest: str = APPROVED_RUNNER_IMAGE_DIGEST) -> str:
+    """app-compose.json as `lium-cvm.sh new` writes it with default flags (--local-key-provider,
+    no --enable-logs, no --enable-sysinfo) after stamping `digest` into the compose file."""
+    compose = (
```
taiberium: **NIT:** `compose` holds the compose text with the runner digest stamped in. Rename it to `compose_with_runner_digest`.

taiberium (approved):

PR #1405, opened 2026-09-18
DAH-3639 - [P1] jupyter: install the script with docker exec, not docker cp

jam6099:
Implements https://www.notion.so/3dfb8bfdbde981a1bf4ec1206e0067eb · reviewer: @jam6099

**Problem** — Rebooting (or re-renting) a pod that takes the validator's jupyter path fails. With an encrypted local volume (or no local volume) `run_jupyter` copied `run_jupyter.sh` into the container's `/tmp` with `docker cp`, and that copy answers `Could not find the file /tmp in container`, so the create aborts, the edit rolls back to the parked container and the pod is left `REBOOT_FAILED` with `jupyter_installation_status = FAILED`.

**Value** — Reboots work again for every pod on the validator jupyter path. Measured: 15 failed copies and 0 successful copies in prod between 15 and 18 Sep; affected pods include 45a4a16d (staging, 18 Sep), ec240f3f (prod, 16 Sep 20:07Z) and 79de2784 (prod, 17 Sep 17:16-17:20Z, four retries). The staging/prod run of failures started when the default templates moved to the `-lium1` builds (lium-platform#58): `ships_sshd=is_cached` (lium-platform executor.py:1213) became false for the old default tag, so those pods dropped from the image-managed jupyter path onto this one.

**Goal** — [DAH-3639](https://www.notion.so/3dfb8bfdbde981a1bf4ec1206e0067eb) (P1).

**Fits** — `neurons/validators/src/services/docker_service.py`, `run_jupyter`, the encrypted / no-local-volume branch.

**Flag:** no flag

**Risks**
- The script is piped from the executor host into the container with `docker exec -i`. If the host file is missing the step fails on an explicit `test -s` message instead of failing later.
- The install runs through `docker exec`, inside the container's own namespaces, which is already how the mkdir and the jupyter start work.
- The jupyter start step and its stdout handling are unchanged; the plain local-volume branch is untouched.

**How to verify**
- `neurons/validators/tests/test_dah3639_jupyter_tmp.py`: no `docker cp` is issued on this path, the install step is `cat /root/app/run_jupyter.sh | docker exec -i ... 'cat > <target>/run_jupyter.sh && chmod +x ...'` fed by the host's pipe, and the start step still targets `<target>/run_jupyter.sh`.
- The existing `test_docker_service.py::test_run_jupyter_with_encrypted_volume_installs_into_plaintext_mount` covers the same path.
- End to end: one reboot of an encrypted + jupyter pod on staging after the executor image is rebuilt (the 45a4a16d case).

**Ran it:** the unit tests run in the validators CI job; no staging executor image has this commit yet.

**Self-review:** clean, 1 file + 2 tests.

**Parts:** backend n/a · migration n/a · frontend n/a · CLI/SDK n/a · docs n/a · tests ✓ · dashboards n/a

**Docs:** none needed, an internal validator fix.

### What changed

- `run_jupyter`'s encrypted branch: the `docker cp` into `/tmp` plus the in-container `cp` are replaced by one `docker exec -i` step that writes the script straight to the target path and chmods it. The host shell pipes `/root/app/run_jupyter.sh` (the same source `docker cp` used) into that exec, so nothing goes through the daemon and nothing is fed from the executor process.
- Tests: `neurons/validators/tests/test_dah3639_jupyter_tmp.py` (new) and the existing `test_docker_service.py` case for this path, updated.

### Why not `docker cp`

`docker cp` fails on the destination, and not because the directory is missing. On staging the executor logged the guard running first:

```
/usr/bin/docker exec -u 0 pod_45a4a16d... sh -c 'mkdir -p /root /tmp'      (exit 0, no stderr)
/usr/bin/docker cp /root/app/run_jupyter.sh pod_45a4a16d...:/tmp/run_jupyter.sh
  -> Error response from daemon: Could not find the file /tmp in container
```

The directory existed in the container and the daemon still could not resolve it: `docker cp` resolves the destination against the container's rootfs, while a path created by the container's own `docker exec` lives in the container's namespace. `docker exec` is the interface that works in that container, and it is how the repo already installs both other scripts it puts into a pod: the gocryptfs setup script and the ssh bootstrap.

### History of the /tmp destination

`f3b4868e` added `mkdir -p /tmp` before the copy because the DinD container did not guarantee it; `4d158747` moved the destination to `/root` and dropped the guard, and `8fb0a69f` (DAH-2243, encrypted local volumes) moved the encrypted branch's destination back to `/tmp`. This change removes the `/tmp` intermediate entirely.

DAH-3639 — https://www.notion.so/3dfb8bfdbde981a1bf4ec1206e0067eb


jam6099:
Root cause, from the staging executor logs after deploying the first commit (5c3fd7b6):

```
/usr/bin/docker exec -u 0 pod_45a4a16d... sh -c 'mkdir -p /root /tmp'   (exit 0)
/usr/bin/docker cp /root/app/run_jupyter.sh pod_45a4a16d...:/tmp/run_jupyter.sh
  -> Error response from daemon: Could not find the file /tmp in container
```

The mkdir ran and /tmp existed in the container, yet docker cp still could not resolve it: the daemon resolves the destination against the container's rootfs, while the directory created by the container's own exec lives in the container's namespace. So the mkdir was not the bug. The install now streams the script in with `docker exec -i` and never uses `docker cp`. Fixed in 8859140c.

surcyf123 (approved):
Read at head `3c6e3f14` (1 commit on `main` @ `219c4f1c`, 3 files). The five things I checked, all hold:

1. **The script never crosses the SSH channel.** The install step is one host-shell line: `test -s /root/app/run_jupyter.sh || …; cat /root/app/run_jupyter.sh | /usr/bin/docker exec -i -u 0 <c> sh -c 'cat > <target>/run_jupyter.sh && chmod +x …'` (`docker_service.py:3565-3570`). The pipe is between the executor's `cat` and `docker exec -i`, on the executor host; `execute_and_stream_logs` is called with no `stdin_data` (asyncssh stdin stays untouched, `docker_service.py:1774-1776`), so the script's quotes/heredocs never become command-line text. `cat` reads the same file `docker cp` read before (the docker CLI reads its source where it runs), so no new path assumption. `docker exec -i` forwards the pipe's EOF, so the in-container `cat >` terminates. Both tests pin `stdin_data is None` and `-i` on the install step.

2. **`<target>` quoting.** `target_q = shlex.quote(target_path)` is spliced into the inner script, and the whole inner script is `shlex.quote`d again for the host shell (`docker_service.py:3569`). For a path with spaces or `;` the inner script reads `cat > '/root;id;x'/run_jupyter.sh && chmod +x '/root;id;x'/run_jupyter.sh` — one token at both layers. `test_jupyter_command_quoting.py::test_renter_volume_path_stays_one_token_at_both_layers[True]` walks this new command too (its `_inner_script_of` picks the first `-c`, which is the `docker exec … sh -c`) and passed at head.

3. **Plain-volume branch and `image_managed_jupyter`.** The `if local_volume and not encrypted_local_volume:` branch (`docker_service.py:3487-3543`) is not in the diff; `run_jupyter` is only entered from `create_container` when `image_managed_jupyter` is false (`docker_service.py:5613-5631`), unchanged.

4. **Failure handling.** The install step keeps `raise_exception=True`, same as the `docker cp` it replaces; a non-zero `docker exec` (or the `test -s` guard, or `cat >` on a read-only target) writes to stderr, `_stream_process_output` turns that into `status=False`, the Exception propagates out of `run_jupyter` under `current_step = "jupyter_setup"` into the same `except Exception:` → `cleanup_failed_container_creation` → re-raise (`docker_service.py:5687-5703`). Same failure step, same rollback.

5. **Both branches the ticket names.** Encrypted (`target = local_volume_path`) and no-local-volume (`target = "/root"`) share this `else:` block, so one change covers both. I did not re-measure the "15 failed copies" number from prod logs; the fix does not depend on it.

CI at `3c6e3f14`: Validators Tests 2660 passed / 15 skipped (job 105505975221, both new/updated tests and the quoting pair listed as PASSED), e2e-gate PASS, CodeQL clean.

Not blocking, outside this PR: `install_jupyter_server` (`docker_service.py:6884-6893`) calls `run_jupyter` without `encrypted_local_volume`, so a later "install jupyter" on an encrypted pod with a `local_volume` would take the temp-container branch and write the script into the ciphertext directory. Pre-existing; worth its own ticket.

Approving. The loop does not merge.

PR #1377, opened 2026-09-15
DAH-3505 - [P2] validator: keep the Docker SDK's SSH session alive through a custom…

surcyf123:
Implements DAH-3505 · stacked on #1376 · reviewer: @taiberium
**Merge order:** this PR merges INTO #1376's branch, not main. Merge #1376 first; merge this one only to fold it into #1376.

**Problem** — 10 custom-Dockerfile rentals in 14 d built for 9 to 14 min, then failed at `volume_creation` with `'NoneType' object has no attribute 'settimeout'`: the paramiko session under the Docker SDK client had no keepalive and was dead after the build.

**Value** — not measured — the body has no Worth it line (SO §69) and the ticket gives no prod number — a human decides.

**Goal** — DAH-3505 (P2).

**Fits** — Sits in `lium-io` — `neurons/validators`; touches validator, tests (6).

**Flag:** no flag

**Risks**
- **Merge order:** this PR merges INTO #1376's branch, not main. Merge #1376 first; merge this one only to fold it into #1376.
- One SSH global request every 30 s per open Docker SDK client (every `rental_docker_client_factory.connect`, 6 call sites) — negligible; rollback: revert.
- Up to 300 chars more per volume-failure CCF message, and the 78-char fixed hint on any other step whose error is the dead-transport text — the backend ignores the field until lium-platform mirrors it (DAH-3504 part 2).
- Not changed: CCF step names, `msg`, `detail`, docker-py's error path, volume sizing math, the asyncssh keepalive; no config, no migration.

**How to verify**
- `cd neurons/validators && pytest tests/test_rental_docker_sdk.py tests/test_custom_dockerfile_build.py -q` → 76 passed; on the base the build file fails 6 (A20, A20b, A20c, A20d, A21, A21b), the SDK file does not collect
- Loki `{namespace="subnet"} |= "Failed create_container" |= "volume_creation"`, 2 to 13 Sep → every custom-build row carries `'settimeout'`

**Verified by the loop:** Gate: PASS 93f2abc

**Ran it:** validators pytest, build + SDK test files → 76 passed · warm EC2 `20260918T181857Z-5moy` at `93f2abc` · not run on macOS · … +2 under Details

**Self-review:** clean at 93f2abc (14 checks)

[...]

PR #1411, opened 2026-09-19
DAH-3678 - [P2] validator: explain add_public_keys failures on exiting images

surcyf123:
Implements [DAH-3678](https://app.notion.com/p/P2-Validator-every-add_public_keys-failure-on-an-exiting-image-returns-the-DAH-2624-message-6-8-s-3e0b8bfdbde981d4a041f0a16150c474) · reviewer: @taiberium

**Problem** — An image with no long-running command exits at once and the SSH-key exec fails at `add_public_keys`. The backend's DAH-2624 explanation (lium-platform#70) keys on markers in the validator's text; an exec the exit itself kills reads `exit_status=137; stderr=''` — no marker — so 6 of one renter's 8 failures on 19 Sep got the generic `failure_step: add_public_keys` text.

**Value** — 6 of tonight's 10 `rent_failed` were this class (add_public_keys 4 + ssh_key_not_accepted 2); the renter fixed his templates only once told. With the text on every failure the first one tells him.

**Goal** — DAH-3678 (P2): the DAH-2624 / B-58 gap from alert 37358 (R-25a). The loop took the first draft (owner rule, 19 Sep); @taiberium reviews instead of writing.

**Fits** — `neurons/validators` only: the `add_public_keys` step of `DockerService.create_container` plus `RentalDockerSdkClient.inspect_container_state()`; `msg` and `failure_step` unchanged.

**Flag:** no flag

**Risks**
- One extra `docker inspect` per failed `add_public_keys`, before cleanup removes the container — providers see one more API call on a failing create, none on a healthy one — rollback: revert.
- A key exec that fails in a running container that restarted once since `docker run` is now explained as the exiting image — the renter reads the DAH-2624 text instead of the generic one — rollback: revert.
- Beyond the diff nothing moves: no migration, no pod restart, no config or data change — nobody else notices — rollback: revert.

**How to verify**
- `pdm run pytest tests/test_docker_service.py tests/test_rental_docker_sdk.py -k "key_injection or inspect_container_state" -q` → 16 passed
- `pdm run pytest tests/ -q --strict-markers` → 2676 passed, 15 skipped (main: 2660); `ruff check --select F,ASYNC210,ASYNC251 --ignore F541 .` → clean

**Verified by the loop:** Gate: PASS d3de99f

**Ran it:** dockerd 29.1.3, Linux: `SLOW_EXEC=1 pdm run python /tmp/dah3678_probe.py alpine:3.20 sh -c 'sleep 1.2'` → explained 0/6 → 6/6; pytest at d3de99f → 2676 passed / 15 skipped

**Self-review:** clean at d3de99f (15 checks, 3 low)

**Parts:** validator ✓ · backend n/a — lium-platform#70 maps the markers · migration n/a — no schema · frontend n/a — backend text · CLI/SDK n/a — no client change · docs n/a — no page lists these texts · tests ✓ 16 cases · dashboards n/a — label unchanged

**E2E:** covered by unit — the exec-killed-by-exit case needs a real daemon (Ran-it probe); the create-path tests use a fake client

**Docs:** none needed because this is internal error text only — the validator's `detail.error` to the backend and its log; the renter reads the backend's unchanged DAH-2624 mapping (lium-platform#70).

### What changed

- `rental_docker_sdk.py`: `ContainerStateSnapshot` (status, running, restarting, exit_code, restart_count, error; `exited_since_start`, `describe()`) and `RentalDockerSdkClient.inspect_container_state()` — one `inspect_container` in the Docker thread pool, daemon errors wrapped as `RentalDockerOperationError`.
- `docker_service.py`: `_explain_add_public_keys_failure()` — on any exception from the `add_public_keys` step it reads the container state (still present: cleanup runs in the enclosing `except`), and when the container is not running, restarting, or has `RestartCount > 0`, raises `ImageExitedDuringKeyInjection` naming the image, the exit code, the state and the original exec error. A running container that never restarted, or a failed inspect, re-raises the original exception unchanged.
- The step's `try/except` in `create_container` is the only call site; `add_ssh_public_keys_with_rental_docker` itself and the standalone add-key path are untouched.

### The gap (file:line on `main` 61020313)

- `neurons/validators/src/services/rental_docker_sdk.py:226-274` — `exec_in_container` raises with a marker only when `_wait_for_container_exec_ready` sees a terminal/timeout state (`status='exited'` / `status='restarting'`) or the daemon returns the 409 `is restarting`. When readiness passes and the CMD exits during the exec, the result is a plain non-zero `ContainerExecResult`.
- `neurons/validators/src/services/docker_service.py:3190-3212` — that result becomes `Exception("Failed to add SSH public keys: exit_status=137; stderr=; stdout=")`, which `create_container`'s handler (`:5806-5829`) puts in `detail.error` as is.
- lium-platform `apps/lium/backend/apps/server/src/consumers/validator_consumer.py:198-208` — `_image_exited_at_start` matches `("is restarting", "is not running", "status='exited'", "status='restarting'")` in `detail`; none of them is in that text, so the renter got `_headline(msg.msg)`.

### Before / after (the renter-facing text)

- Before, on every exec the exit itself killed: `Failed to add SSH public keys: exit_status=137; stderr=''; stdout=''` → the portal showed `Failed create_container (failure_step: add_public_keys)`.
- After: `Failed to add SSH public keys: image 'alpine:3.20' has no long-running command — its default command exited right after start (exit_code=0) and Docker is restarting it (status='restarting' running=True restarting=True exit_code=0 restart_count=1 error=None) while the SSH keys were being installed; a pod needs a long-running process, for example a start command such as `sleep infinity`. Exec error: …` → the backend's `_image_exited_at_start` matches `is restarting` / `status='restarting'` and the renter reads the DAH-2624 text.

### Design notes

- Message shape: `msg` stays the headline and `failure_step` stays `add_public_keys`, so the `failure_step` dashboards label and the backend's headline logic are untouched; only `detail.error` gains the marker text.
- `RestartCount > 0` is treated as "the image exited": a container that restarted once between `docker run` and the exec landed the exec in the restart gap, which is the same renter-side cause. A running, never-restarted container keeps the original exec error verbatim.
- A failed inspect never hides the exec error: the original exception is re-raised unchanged.

### Not verified here

- The portal/event text for a real prod failure after deploy — the backend mapping is lium-platform#70's, unchanged; the validator now emits its trigger. A prod read of `executor_events.details.error_message` after the next such failure confirms it.
- `ruff format --check` flags both touched source files, as it does on `main` (the format job is report-only and both files were already unformatted).

### Claims

| claim | evidence |
|---|---|
| the exec-killed-by-exit text carries no marker on `main` | Ran it below: `exit_status=137; stderr=''` on all 6 attempts, `marker before=False` |
| the fix explains it | Ran it below: `marker after=True` 6/6; `test_a_key_injection_that_fails_on_an_exiting_image_names_the_image[exited|restarting|running-again-after-a-restart|dead]` |
| a running, never-restarted container keeps the exec error | `test_a_key_injection_that_fails_in_a_running_container_keeps_the_exec_error` asserts `detail.error == str(exec_error)` |
| a failed inspect never hides the exec error | `test_a_key_injection_failure_whose_inspect_fails_keeps_the_exec_error` |
| `failure_step` and `msg` unchanged | asserted in every create-path case (`_create_failing_at_add_public_keys`) |
| the container is inspected before cleanup removes it | same helper: `inspect.await_args` recorded, `cleanup_failed_container_creation` awaited once |
| the instantly-exiting case still works | Ran it: `ubuntu:24.04` → readiness poll text `status='restarting'` 6/6 before and after |
| suite green | 2676 passed, 15 skipped in 174 s (Python 3.11, pdm, CI env vars) |

### Ran it

`/tmp/dah3678_probe.py` (docker 29.1.3, Linux x86_64): `docker run -d --restart unless-stopped <image> [cmd]`, then six real `RentalDockerSdkClient.exec_in_container` calls; each failure is passed through `_explain_add_public_keys_failure` and checked against the backend's four markers.

The gap — CMD alive for 1.2 s, key exec that takes longer (nvidia's entrypoint banner then `bash` is this shape):

```
$ SLOW_EXEC=1 pdm run python /tmp/dah3678_probe.py alpine:3.20 sh -c 'sleep 1.2'
attempt 0: backend marker before=False after=True
  old text : Failed to add SSH public keys: exit_status=137; stderr=''; stdout=''
  new text : Failed to add SSH public keys: image 'alpine:3.20' has no long-running command — its default command exited right after start (exit_code=0) and Docker is restarting it (status='restarting' running=True restarting=True exit_code=0 restart_count=1 error=None) while the SSH keys were being installed; a pod needs a long-running process, for example a start command such as `sleep infinity`. Exec error: …
attempt 1..5: backend marker before=False after=True
final state: status='restarting' running=True restarting=True exit_code=0 restart_count=6 error=None
outcomes:
  6x non-zero exec: Failed to add SSH public keys: exit_status=137; stderr=''; stdout=''
explained with a DAH-2624 marker: 6/6
```

The case that already worked — CMD exits at once, the readiness poll catches it:

```
$ pdm run python /tmp/dah3678_probe.py ubuntu:24.04
attempt 0..5: backend marker before=True after=True
  old text : Docker container did not become ready for exec after 15 seconds: status='restarting' running=True restarting=True paused=False dead=False exit_code=0 error=''
final state: status='restarting' running=True restarting=True exit_code=0 restart_count=11 error=None
explained with a DAH-2624 marker: 6/6
```

Tests at d3de99f: `pdm run pytest tests/ -q --tb=short --strict-markers` → `2676 passed, 15 skipped in 174.09s`. Failing-first: with `src/` stashed the new tests do not collect (`ContainerStateSnapshot` missing); with it, 16 passed.

### Self-review

Verdict `clean` at `d3de99f` · 15 checks · by subagent-eng-box-11-1411 · 2026-09-19 17:54Z (tools/pr_self_review.py)

|---|---|---|---|---|
| low | CROSS-PR | `neurons/validators/src/services/rental_docker_sdk.py:339` | Open loop PRs #1365 (DAH-3467-rm-timeout-inspect, head 52ca475, adds `container_status()` at the same spot after `remove_container` and appends tests) and #1377 (DAH-3504-volume-step, head 93f2abc, appends to test_rental_docker_sdk.py) conflict textually with this PR's appends (`git merge-tree d3de99f <head>` reports CONFLICT in rental_docker_sdk.py / test_rental_docker_sdk.py; both branches merge clean with main on their own); the body has no cross-PR line. | Add a CROSS-PR note to the body (whichever of #1365/#1377 lands second rebases; append-only conflicts) and, when rebasing after #1365, consider having `container_status()` and `inspect_container_state()` share one inspect reader. |
| low | STALE-BODY | `neurons/validators/src/services/miner_service.py:171` | The Value line counts `ssh_key_not_accepted 2` inside "6 of tonight's 10 rent_failed were this class", but ssh_key_not_accepted is a miner_service failure (`_ssh_key_not_accepted_text`) with its own failure_step; the backend's `_image_exited_at_start` requires failure_step == 'add_public_keys' and this diff does not touch that path, so this PR changes 4 of the 10, not 6. The Problem line's "6 of one renter's 8 failures" also does not match the Value line's "add_public_keys 4". | Restate the Value line with the add_public_keys count only (or explain how the two ssh_key_not_accepted failures relate) and reconcile it with the Problem line's 6-of-8. |
| low | PROSE-VS-CODE | `neurons/validators/src/services/docker_service.py:3259` | The Problem line and the Ran-it 'old text' quote the validator's failure as `exit_status=137; stderr=''; stdout=''`, but `add_ssh_public_keys_with_rental_docker` formats it as `exit_status=137; stderr=; stdout=` (no quotes); the test fixture `_EXEC_KILLED_BY_EXIT` (tests/test_docker_service.py:6861) uses the real form, so only the prose is off (the probe script printed its own repr). | Quote the emitted form `stderr=; stdout=` in the body, or note that the probe printed repr(). |

### Verified

Gate: PASS (d3de99f) on Linux aarch64 (lium-loop-runtime)

</details>


neurons/validators/src/services/rental_docker_sdk.py
```diff
@@ -155,6 +155,36 @@ class _ContainerExecReadiness:
...
+
+    @property
+    def exited_since_start(self) -> bool:
+        """The container's main process has ended at least once since `docker run`.
+
+        Not running now (exited/dead/removing), mid-restart, or running again after Docker's
+        restart policy brought it back (RestartCount > 0). A `created` container never started,
+        so it is not an exit.
+        """
+        if self.restarting or self.restart_count > 0:
```
taiberium, line 183: **NIT:** A host kill (OOM, GPU fault) also makes this true, and the renter then reads "add `sleep infinity`". Keep the exec error on `OOMKilled` or `exit_code` 137.

neurons/validators/tests/test_docker_service.py
```diff
@@ -6846,3 +6848,122 @@ async def test_create_failure_before_the_container_exists_is_not_a_host_kill(
...
+        pytest.param(
+            _state(status="dead", running=False, exit_code=137, restart_count=0),
+            "is not running",
+            id="dead",
+        ),
+    ],
+)
+@pytest.mark.asyncio
+async def test_a_key_injection_that_fails_on_an_exiting_image_names_the_image(
+    docker_service, monkeypatch, state, marker
```
taiberium: **NIT:** Rename `marker` to `expected_backend_marker`, so the reader sees which text the backend must find.

neurons/validators/tests/test_rental_docker_sdk.py
```diff
@@ -1312,3 +1312,69 @@ class FakeRejectPolicy:
...
+            _container_state(status="restarting", running=True, restarting=True), True, id="restarting"
+        ),
+        pytest.param(_container_state(status="exited", running=False, exit_code=1), True, id="exited"),
+        pytest.param(_container_state(status="dead", running=False, dead=True), True, id="dead"),
+        pytest.param(_container_state(status="created", running=False), False, id="never-started"),
+        pytest.param(_container_state(status="paused", running=True, paused=True), False, id="paused"),
+    ],
+)
+@pytest.mark.asyncio
+async def test_inspect_container_state_tells_an_exit_since_start(inspect_result, exited):
```
taiberium: **NIT:** Rename `exited` to `expected_exited_since_start`.

taiberium (approved):

PR #1283, opened 2026-09-06
DAH-2977 - [P2] pre-pull the top-N official templates on idle executors (flag

surcyf123:
Implements [DAH-2977](https://app.notion.com/p/3d3b8bfdbde981c995ffe5c85923e6cb) · reviewer: @jam6099

**Problem** — Pod start is p50 23–26 s, p90 67–93 s, p99 3–13 min (read-only prod DB, 5 Sep 2026, 398 rentals).

**Value** — not measured — the body has no Worth it line (SO §69) and the ticket gives no prod number — a human decides.

**Goal** — [DAH-2977](https://app.notion.com/p/3d3b8bfdbde981c995ffe5c85923e6cb) (P2) · serves B-42.

**Fits** — Sits in `lium-io` — `neurons/executor`; touches backend (lium-platform#89), docs (`.env.template`), tests.

**Flag:** `PRE_PULL_TEMPLATES_ENABLED`, off by default, turned on by @taiberium on 2026-09-30 (proposed by the loop; change it in one reply and the body follows)

**Risks**
- Rollout: deploy both halves, then `PRE_PULL_TEMPLATES_TOP_N=2` on the backend (lium-platform#89) and `PRE_PULL_TEMPLATES_ENABLED=true` in executor `.env`s; either alone changes nothing (note under Details)

**How to verify**
- `cd neurons/executor && pdm run pytest -q tests/test_pre_pull_service.py tests/test_cache_prefetch_state.py tests/test_cache_template_digest.py` → green (39 pre-pull tests)

**Verified by the loop:** 14 Sep 2026 · warm EC2 20260914T131912Z-oo2i · executor 335 passed at f334638 · Gate: PASS f334638

**Ran it:** executor `pytest tests/` → 335 passed at f334638 (warm EC2 `20260914T131912Z-oo2i`); flag off, poll params and puller count are today's · no live executor

**Self-review:** clean at f334638 (carried from ed5a6a8: rebase-only, patch unchanged; 19 checks, 2 low)

**Parts:** backend ✓ lium-platform#89 · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — executor neuron · docs ✓ `.env.template` · tests ✓ 39 pre-pull + 31 cache-template · dashboards n/a — one log line per pull

**Docs:** `neurons/executor/.env.template` · public: none changed in this PR — `PRE_PULL_TEMPLATES_ENABLED` is off by default; nothing is provider-visible until an environment enables it

<details><summary>Details</summary>

[...]

PR #1412, opened 2026-09-19
DAH-3604 - [P2] executor: PRE_PULL_TEMPLATES_ENABLED on by default (rollout step 3 of the template pre-pull)

surcyf123:
Implements [DAH-3604](https://www.notion.so/DAH-3604) (rollout step 3 of [DAH-2977](https://app.notion.com/p/3d3b8bfdbde981c995ffe5c85923e6cb)) · reviewer: @taiberium

**Flag:** `PRE_PULL_TEMPLATES_ENABLED` default **true** — merge conditions (all three, in order): (1) lium-platform#577 merged and deployed (backend `PRE_PULL_TEMPLATES_TOP_N=2`), (2) one staging executor soaked ≥ 24 h with the flag on via `.env` (Rustam), (3) ≥ 3 prod nodes soaked 24 h; until then this PR stays a draft. Opt-out per node: `PRE_PULL_TEMPLATES_ENABLED=false` in `.env`.

The loop took the first draft (owner rule 19 Sep); @taiberium reviews and decides the on-date (30 Sep proposed); the staging/prod soak steps are his — this PR is step 3.

**Problem** — Pod creates succeed 97.6 %, but time-to-RUNNING p95 is 278 s against a 240 s target; a rental whose template is not on the node starts in p95 1,147 s, one whose template is cached in p95 57.5 s. lium-io#1283 shipped the fix (idle-time pre-pull of the backend's top-N official templates) behind a flag that defaults to `False`, so no node does it unless its `.env` says so.

**Value** — Worth it: p95 278 s vs 240 s target; uncached 1,147 s vs cached 57.5 s; 316 rentals / 7 d lose ~13 s p50 and the 205 s p95 tail to template pulls that the pre-pull would have done while the node was idle.

**Goal** — [DAH-3604](https://www.notion.so/DAH-3604) (P2) — the default flip is the cut that moves the fleet; steps 1–2 (staging `.env`, a few prod `.env`s) prove it first.

**Fits** — Sits in `lium-io` — `neurons/executor` only: `core/config.py` (the default), `.env.template` (the comment), one docstring line in `cache_template_service.py`, and `tests/test_pre_pull_service.py`. Backend half: lium-platform#577 (`PRE_PULL_TEMPLATES_TOP_N=2`, READY). No new code path: every behaviour below shipped in #1283 and is exercised by its 39 tests.

**Risks**
- Fleet-wide registry load at first enable: every executor that restarts on the release starts asking for the top-N list. Mitigated by the 15-min start jitter (`PRE_PULL_START_JITTER_SECONDS=900`), idle-only pulls (no `pod_*` container, no `dial-stdio` session), one pull per 15-min sweep per node, and the 200 GB disk floor (`PRE_PULL_MIN_FREE_GB`). At N=2 the cost is ≈ 8.6 GB compressed per node, once, refreshed only on digest change.
- Disk pressure on small nodes: the pull is skipped (`outcome=insufficient_disk`) unless 200 GB stay free after it; least-recently-used pre-pulled images are untagged first, and nothing else (default image, rental images) is ever removed.
- A node that must not pre-pull: `PRE_PULL_TEMPLATES_ENABLED=false` in `.env` restores the pre-#1283 loop byte-for-byte (the flag-off test pins that).
- Merge order: if this merges before lium-platform#577 is deployed, executors ask for `include_pre_pull=true` and get an empty extra list — harmless, nothing pulled.

**How to verify**
- `cd neurons/executor && pdm run pytest -q tests/test_pre_pull_service.py` → green, including the new `test_pre_pull_is_on_by_default` (asserts `Settings.model_fields["PRE_PULL_TEMPLATES_ENABLED"].default is True`).
- On a soaked node: `docker logs <executor> | grep 'pre_pull '` shows one `pre_pull image=… digest=… seconds=… outcome=pull_ok|preempted|timeout|pull_failed|insufficient_disk|lock_held` line per attempt, without any `PRE_PULL_*` line in its `.env`.

**Ran it:** executor `pdm run pytest tests/ -v` → 410 passed at ff3ff68f (40 in `test_pre_pull_service.py`: 39 from #1283 + the new default test), Python 3.12 / pdm, sandbox VM · no live executor; CI `Executor Tests` (pdm, py3.11) runs the same suite.

**Metrics:** no counters added. The executor has no Prometheus/metrics registry (`rg "prometheus|Counter\(|Gauge\(" neurons/executor/src` → nothing), and this PR does not invent one; the `pre_pull … outcome=` log line from #1283 stays the observability surface for the soak (`lium_executor_prepull_total{outcome}` and `lium_executor_template_cache_hit_total{hit}` would need a registry first — separate ticket if wanted).

**Parts:** backend ✓ lium-platform#577 (not merged yet — merge condition 1) · migration n/a — no schema · frontend n/a — no UI · CLI/SDK n/a — executor neuron · docs ✓ `.env.template` · tests ✓ 1 new (default pinned) + the 39 pre-pull tests unchanged · dashboards n/a — no metrics registry, one log line per pull

**E2E:** n/a — no new code path; the #1312 subnet-loop stack ran #1283 with the flag on and off (7 Sep, PASS). The soak steps (1)–(3) above are the live verification.

**Docs:** `neurons/executor/.env.template` — the `PRE_PULL_*` block now says "on by default since the executor release carrying DAH-3604 (after 4.0.3); set `PRE_PULL_TEMPLATES_ENABLED=false` to opt this node out". Provider-visible once the release ships: idle nodes pull up to two extra official images (≈ 8.6 GB) and keep 200 GB free; the executor README does not mention the flag and is unchanged here.

## Change
- `neurons/executor/src/core/config.py`: `PRE_PULL_TEMPLATES_ENABLED` `Field(default=False)` → `Field(default=True)`; comment updated. The other three settings (`PRE_PULL_MIN_FREE_GB=200`, `PRE_PULL_TIMEOUT_SECONDS=1800`, `PRE_PULL_START_JITTER_SECONDS=900`) are untouched.
- `neurons/executor/.env.template`: the block header drops "off by default", states the new default and the opt-out; the example line reads `# PRE_PULL_TEMPLATES_ENABLED=true`.
- `neurons/executor/src/services/cache_template_service.py`: the one comment that said "off by default" now says "on by default since DAH-3604".
- `neurons/executor/tests/test_pre_pull_service.py`: module docstring updated; new `test_pre_pull_is_on_by_default` pins the default on the `Settings` class (not the process-wide `settings` instance, so a developer's local `.env` cannot change the result). The existing flag-on/flag-off tests set the flag explicitly via `monkeypatch` and are unchanged.

## Rollout (from #1283's body)
1. One staging executor `.env` → `PRE_PULL_TEMPLATES_ENABLED=true`, 24 h, watch `pre_pull` lines and disk (Rustam).
2. ≥ 3 prod nodes the same way, 24 h.
3. **This PR**: default flip in an executor release; nodes that need to stay out set `false`.

## Not in this PR
- No version bump (`version.txt` 4.0.3 is bumped in its own chore commit at release time, as before).
- No metrics registry / counters (see **Metrics** above).
- No change to `PRE_PULL_MIN_FREE_GB` / timeout / jitter.

</details>

---
Opened by the Lium automation account; commits are anonymous by policy. Ticket: DAH-3604.


surcyf123:
**Verdict: changes requested** at `941e5ace` (2 medium, 1 low). Fresh review. Flipping the default makes #1283's pre-pull path live on every node, so the review covers that path (`pre_pull_service.py`, `cache_template_service.py`), not only the diff.

**Findings**

1. **medium** · `neurons/executor/src/services/pre_pull_service.py:75-87` (`rental_activity`). **Problem:** the idle check sees `pod_*` containers and `dial-stdio`. It does not see the validator's VerifyX bandwidth sample, `verifyx_executor.py`, which runs over a plain SSH shell (`verifyx_validation_service.py:302`) or through local verify. A pre-pull can therefore share the link with that sample. The gate is an EMA of at least 100 Mbps with alpha 0.5 (`neurons/validators/src/services/task/checks/verifyx.py:17`, `checks/network_ema.py:1`), so one contended sample can take a slower node under it (`VERIFY_FAILED_NETWORK_SPEED_TOO_SLOW`), and the node is off the market for that cycle. This has already happened: `verifyx.py:121-125` records a traced host that measured 104 Mbps "whose link the executor's on-boot template pre-pull was sharing". Default-on adds about 10.5 GB per node on the release restart, and slow links are both the most likely to overlap and the most likely to fail. The Risks section does not mention the gate. **Fix:** treat a running `verifyx_executor.py` as busy in `rental_activity`, alongside `dial-stdio` (psutil already sees host PIDs through `pid: host`). A new pull then waits, and a running one is preempted within the 10 s check. Add a test, and add a Risks line.

2. **medium** · `pre_pull_service.py:335-358` (`_make_room`) with the sweep loop at `:262-322`. **Problem:** the eviction victim is any tracked LRU ref, including another *current* top-N entry. On an idle node whose free space is just above the floor, the two images evict each other on every sweep, forever. With today's sizes (6.49 GB and 3.96 GB compressed) the band is about 218–229 GiB free. Reproduced with the PR's test fakes and a stateful disk model at 222 GiB free, taking on-disk size as 2× compressed: 8 sweeps gave 8 pulls, `a, b, a, b, …`. That is about 10 GB of registry and link traffic every 30 min per affected node, and it also feeds finding 1. The body's "then nothing once both are present" does not hold on these nodes. **Fix:** make a victim only a tracked ref that is not in this sweep's `entries`, meaning an image that left the top-N (pass those refs into `lru(skip=…)`). When nothing else can go, log `insufficient_disk` and keep what is present. Add a test with a stateful free-space fake.

3. **low** · `pre_pull_service.py:82`. **Problem:** a renter who stopped a pod (`stop_container`, `neurons/validators/src/services/docker_service.py:6035`) still holds the rental, but the pod's `exited` container counts as idle. A pull, and an eviction, can then run during that rental. The impact is small: the restart goes through `dial-stdio` and preempts the pull within about 10 s, the 200 GiB floor still holds, and docker refuses to remove an image that a stopped container uses. **Fix:** state this in Risks. Alternatively, count an `exited` `pod_*` container as busy until it is removed, since the validator's cleanup removes leftovers.

**Checks**
- Head: `git ls-remote` gave `refs/pull/1412/head` = `941e5ace7f86eef9b3913100b727189fada7c378` before the review and again before posting.
- Tests: `neurons/executor`, Python 3.11.11, pdm 2.29.2, `pdm run pytest tests/ -q` gave **424 passed, 1 skipped**, as claimed.
- CI (Actions runs API, head sha): Tests (Executor Tests, e2e-gate, route, tests-ok, ruff check, ruff format ×3), CodeQL (python, actions) and the PR dynamic run are all success. Miners, Validators and Protocol were skipped (no changes there).
- Disk guard, maths and units: `docker_image_size` is bytes. The live prod response gives 6491891139 and 3961262240. The need is `int(size*3)` bytes and the floor is `PRE_PULL_MIN_FREE_GB*1024**3`, so "200 GiB free after 3× compressed" is right at the byte boundary. `statfs("/")` inside the executor container reports the filesystem that holds the data root (overlay upper). The guard is checked only at pull time; eviction runs only in an idle sweep.
- During a rental: no pull starts while a non-exited `pod_*` container or a `dial-stdio` session exists. A pull that is running re-checks every 10 s and closes the stream, which cancels the pull in the daemon. Overlap with a starting rental is therefore at most about 10 s. The exceptions are findings 1 and 3.
- Never the default image: pre-pull entries whose `(repo, tag)` is a mandatory ref are dropped, `protected` is re-read for every eviction victim, and `_cleanup_old_tags` keeps pre-pull tags.
- Failures never block: the sweep runs as an un-awaited task; `_run_pre_pull_sweep` catches `Exception`; the pull budget ends 90 s before the refresh; the lock is non-blocking; `gpus_utility.py` skips only its own cycle.
- Provider opt-out: `PRE_PULL_TEMPLATES_ENABLED=false` in `.env` reaches the container through compose `env_file` (and the CVM `.env`). The flag-off path is pinned by a test.
- Image provenance: only backend entries with `pre_pull: true` are used. The backend serves official templates only (`find_official_by_recent_rentals`), with a Hub digest and a non-zero size. The default URL is `https://lium.io/api`. The pull is `fromImage=<repo>&tag=<digest>` and the re-tag is to the same repo, so content is checked against the digest. No path lets a response re-tag another repository, and the default ref is filtered out. I found no injection path beyond a compromised backend, which already chooses the default image.

surcyf123:
Pixel is away 24–28 Sep, so this review request moves to Mikhail.

